PR #7, opened 2016-12-31
meson: Use 'libzim' name as a dependency

legoktm:
With <https://gerrit.wikimedia.org/r/#/c/329759/>, the library is now
registered with pkg-config as "libzim".

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,7 +6,7 @@ compiler = meson.get_compiler('cpp')
 
 thread_dep = dependency('threads')
 libicu_dep = dependency('icu-i18n')
-libzim_dep = dependency('zimlib')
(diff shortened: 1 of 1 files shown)


mgautierfr:
But with the meson version https://gerrit.wikimedia.org/r/#/c/328341/ (or the more uptodate version https://github.com/mgautierfr/openzim/tree/meson) it is zimlib.
I don't care about the name, we just have to agree on one :)

By the way, both patches on gerrit are not good. The installed pkg-config file doesn't declare the zimlib dependencies. I've fixed it latter mgautierfr/openzim@f6d7e172bc but forget to submit it on gerrit.

PR #35, opened 2017-03-28
Let's use stem and stop words information (if) present in the database.

mgautierfr:


src/xapianSearcher.cpp
```diff
@@ -65,18 +66,39 @@ std::map<std::string, int> read_valuesmap(const std::string &s) {
...
+           ISO-639-3) */
+        icu::Locale languageLocale(language.c_str());
+
+        /* Configuring language base steemming */
+        try {
+          stemmer = Xapian::Stem(languageLocale.getLanguage());
+        } catch (...) {
+          std::cout << "No steemming for language '" << languageLocale.getLanguage() << "'" << std::endl;
+        }
+        queryParser.set_stemmer(stemmer);
```
kelson42: Would not that be better to put that two line in the try block?

mgautierfr: Totally.
Fixed the last `push -f`.

PR #57, opened 2017-06-16
Build libkiwix on any system using Docker

julianghionoiu:


Dockerfile.build
```diff
@@ -0,0 +1,21 @@
...
+
+FROM ubuntu:16.04
+
+RUN apt-get update
+RUN apt-get -y install sudo build-essential apache2 curl git wget pkg-config
+
+COPY . /usr/kiwix/
+WORKDIR /usr/kiwix/
+
+RUN apt-get install -y
```
julianharty: Does this apt-get install -y do anything productive? From a quick test it seems not be be harmful, however it also doesn't install any packages.

julianharty:
@kelson42 @mgautierfr - @julianghionoiu kindly spent time helping me explore ways to build and deploy the kiwix-apache code so it can be system tested. This pull request is for a base Dockerfile that I hope we can extend to incorporate the extra steps needed for the kiwix-apache module. Would this project be a good home for his contribution? If so, please can you review the code and merge it once you're happy with it. Else please suggest where's a good home for this work. Thank you.

Dockerfile.build
julianghionoiu: I have removed it.

mgautierfr:
1. As the name implies it, `travis/*.sh` scripts are here to be run in travis environment.
Using this kind of script in other environment of travis may work now, but it add unnecessary (and somehow illogical) dependencies between scripts and it will lead to modification in travis script that will break docker.

2. Wouldn't be better to have a docker image that contains the dependencies of kiwix-lib only and let the user go into that image and compile kiwix-lib itself ? Build a full new docker image at each build seems overkill.

3. The main important thing made by `travis/install_deps.sh` is install meson, ninja and get the deps archive generated daily by kiwix-build. Maybe it is better to let kiwix-build generate the docker image at the same time it generate the archive.

4. Just by curiosity, why not include the build of kiwix-apache in kiwix-build (independently of the use of docker or not) ?

kelson42:
@mgautierfr What should we do with this really old PR?

kelson42:
Kiwix build is the solution to build easily libkiwix. So we won't further do anything here.

PR #68, opened 2017-07-19
Fix wording problem

kelson42:


kelson42:
To fix https://github.com/kiwix/kiwix-tools/issues/27

Issue #73, opened 2017-07-24
scripts/compile_resources.py does not have licensing information


legoktm:
It was added by @mgautierfr in cba71b4e75c80fac6a49c3f0373d5fd822fe5394, and they're the only author.

mgautierfr:
fixed by #75

PR #75, opened 2017-08-07
Fix license header #73

kelson42:


mgautierfr:
Exactly what I would have written.
Thanks.

PR #77, opened 2017-08-08
Add man page for kiwix-compile-resources

legoktm:
man pages are required for inclusion in Debian. This file was originally
generated by `help2man` and then cleaned up by hand.

### DIFF
--- a/scripts/kiwix-compile-resources.1
+++ b/scripts/kiwix-compile-resources.1
@@ -1,0 +1,20 @@
+.TH KIWIX-COMPILE-RESOURCES "1" "August 2017" "Kiwix" "User Commands"
+.SH NAME
+kiwix-compile-resources \- helper to compile and generate some Kiwix resources
+.SH SYNOPSIS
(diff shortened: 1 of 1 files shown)


mgautierfr:
I'm not sure that we want a man page for this script.

This script is really internal. Its only purpose is to be used by kiwix-tools at compilation time.
It is not intended to be run by user at all.

legoktm:
Debian policy is that everything in /usr/bin needs a man page, so I had to write one anyways. If you don't think it belongs upstream that's fine, feel free to close this and I'll just keep it in the Debian package repository.

legoktm:
I'm just going to include the man page as part of the Debian package, and not try to submit it upstream for now.[1]

If you could take a look at https://github.com/kiwix/kiwix-lib/pull/76 and https://github.com/kiwix/kiwix-tools/pull/63 I would appreciate it.

[1] https://anonscm.debian.org/cgit/collab-maint/libkiwix.git/commit/?id=31e8436084361c93f4eda5aa44132232da0c4747

kelson42:
@legoktm ok thx.. Not much to say on all of this. Just back from Wikimania and @mgautierfr away for a week.

PR #78, opened 2017-08-09
Multisearch jni

mgautierfr:
WIP ! To be merged only once android team is ready on its side.

NOT BEFORE !!


mhutti1:
@mgautierfr This seems to be failing for an unrelated reason. Can you look into it?

mgautierfr:
@mhutti1, does this API is working for you ?
Is it ok for you if I merge it ?

kelson42:
@mhutti1 @mgautierfr So please don't. But I think the problem fundamentaly is that Kiwix Android 3.0 dev takes place on `master` which is also what we use for the 2.3 version. We should not have to make a choice between 2.3 or 3.0... we should be able to do both at the same time IMO.

kelson42:
After discussion with @mgautierfr is seems clear he needs to go ahead with this. But it's important on the Kiwix Android side to deal with these different versions of the libkiwix properly. For example, I think it might be a good idea to not use it with kiwix for android 2.3?

mgautierfr:
It's time to merge this, we need to advance and add new functionality to the JNI API (as the byte range access).

If kiwix-android team is not ready, they can stay on kiwix-lib 1.0. No need to block here.

If there are problems with this new API (and I'm pretty sure there will be), please open specific issues for them.

PR #88, opened 2017-09-26
Rename compile_resources.py to less generic name

mgautierfr:
This changes has been created by @legoktm 

This is just a rebase on top of master to avoid "complex" git history.


mgautierfr:
I would have preferred to keep the `.py` as it is the source and install the file without the extension.

However, while meson implement this (https://github.com/mesonbuild/meson/issues/1487), let's move on and rename our source file.

Thanks @legoktm for your patch.

PR #98, opened 2017-11-22
Add a API to get only a part of a article content.

mgautierfr:
Add the jni method `getContentPart` to get only a part of the artcicle
content.

The method can be used to get a part of the content or to know the size
of the full content.


mgautierfr:
This PR compile but has not been tested as it needs changes on `kiwix-android` side.
I'm merging it to allow android team to test it.

@mhutti1, can you test with the new API and open issue in case of problem ?

PR #101, opened 2017-12-13
Force usage of meson 0.43.0.

mgautierfr:
Static compilation is broken with meson 0.44.0

### DIFF
--- a/travis/install_deps.sh
+++ b/travis/install_deps.sh
@@ -29,7 +29,7 @@ esac
 
 sudo apt-get update -qq
 sudo apt-get install -qq python3-pip ${PACKAGES}
-sudo pip3 install meson
+sudo pip3 install meson==0.43.0
 
 # Ninja
 cd $HOME


legoktm:
Is there an upstream bug report for this? Debian unstable has meson 0.44.0 (https://tracker.debian.org/pkg/meson).

mgautierfr:
No, mainly because I'm not sure it is a problem in meson or with us.

In all case, it's happen when the libraries are in two directory (`/usr/lib` and `/my/custom/dir/lib`) and pkg-config contains only the custom lib dir `-L/my/custom/dir/lib`.
When linking statically, meson try to find the `.a` itself to use the full path as gcc argument.
(Before 0.44.0, it was simply passing the arguments `-L... -l...`.)

But, if the `/usr/lib` directory is not present in the `-L` argument given by pkg-config, meson cannot find the `.a` and fails.

There is a similar issue mesonbuild/meson#2765 especially with the comment https://github.com/mesonbuild/meson/issues/2765#issuecomment-351132294 explaining what I've said above.

However, writing everything here, I think I will open a issue on meson side right now, even if the problem is potentially on our side.

(Another bug that can directly impact debian, is mesonbuild/meson#2761)

PR #104, opened 2017-12-15
Allow JNI to access search snippets

mhutti1:


mgautierfr:
My bad, I implement the function but don't defined it in the Java part.
I've messed somewhere.

PR #105, opened 2017-12-19
Set the verbosity of the underlying zim::Search.

mgautierfr:


kelson42:
@mgautierfr What should we do with this really old PR and branch.

PR #117, opened 2018-02-20
Meson fixes

nirbheek:
Use fallback deps for libzim and pugixml. libzim has a meson build, and is not packaged on some distros (like Fedora), so use a subproject if necessary. Usage as a subproject requires https://github.com/openzim/libzim/pull/116. Also, pugixml does not have a pkg-config file on Fedora.

The build was also broken when xapian-core isn't found.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -10,8 +10,12 @@ static_deps = get_option('android') or get_option('default_library') == 'static'
 
 thread_dep = dependency('threads')
 libicu_dep = dependency('icu-i18n', static:static_deps)
-libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps)
-pugixml_dep = dependency('pugixml', static:static_deps)
+libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps, fallback:['libzim', 'libzim_dep'])
+pugixml_dep = dependency('pugixml', static:static_deps, required:false)
+if not pugixml_dep.found()
+  # FIXME: add static: kwarg when https://github.com/mesonbuild/meson/issues/328 is fixed
+  pugixml_dep = compiler.find_library('pugixml')
+endif
 
 ctpp2_include_path = ''
 has_ctpp2_dep = false
--- a/subprojects/libzim.wrap
+++ b/subprojects/libzim.wrap
@@ -1,0 +1,5 @@
+[wrap-git]
+directory=libzim
+url=https://github.com/openzim/libzim
+push-url=git@github.com:openzim/libzim.git
+revision=origin/master


src/meson.build
```diff
@@ -2,20 +2,22 @@ kiwix_sources = [
   'library.cpp',
   'manager.cpp',
   'reader.cpp',
-  'searcher.cpp',
```
mgautierfr: searcher.cpp is using xapian only to search on external fulltext index.
If the fulltext is embedded in the zim file, it use libzim to do the search (and not use xapian directly).

It would be better fix searcher.cpp than simply not compiling it. Especially if we install the searcher.h header all the time (and we do).

meson.build
```diff
@@ -10,8 +10,12 @@ static_deps = get_option('android') or get_option('default_library') == 'static'
...
 thread_dep = dependency('threads')
 libicu_dep = dependency('icu-i18n', static:static_deps)
-libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps)
-pugixml_dep = dependency('pugixml', static:static_deps)
+libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps, fallback:['libzim', 'libzim_dep'])
+pugixml_dep = dependency('pugixml', static:static_deps, required:false)
+if not pugixml_dep.found()
+  # FIXME: add static: kwarg when https://github.com/mesonbuild/meson/issues/328 is fixed
+  pugixml_dep = compiler.find_library('pugixml')
+endif
```
mgautierfr: We are building static binaries in https://github.com/kiwix/kiwix-tools (also using meson).
We do this to publish "autonomous" binaries users just have to run without bothering with extra files or libraries.

If pugixml static is not found by meson at `kiwix-lib` compilation, it will probably not be found at `kiwix-tools` compilation neither.
So it seems to me as a unnecessary workaround here.

meson.build
```diff
@@ -10,8 +10,12 @@ static_deps = get_option('android') or get_option('default_library') == 'static'
 
 thread_dep = dependency('threads')
 libicu_dep = dependency('icu-i18n', static:static_deps)
-libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps)
-pugixml_dep = dependency('pugixml', static:static_deps)
+libzim_dep = dependency('libzim', version : '>=3.0.0', static:static_deps, fallback:['libzim', 'libzim_dep'])
```
mgautierfr: I may have wrong here (you are a better meson expert than I am :smile: ), but will meson need the url of `libzim` at a moment ?

mgautierfr:
For your information @nirbheek, we have a wrapper build-system at https://github.com/kiwix/kiwix-build
It handle "everything" as installing packages or download and compile missing subprojects for you.

This doesn't invalidate your PR, but it worth knowing it as we've created it to help people overcome the never ending war against dependencies nigthmare :) Maybe someday, everything will use meson and we will get rid of those extra tools.

src/meson.build
mgautierfr: This fix is a bit more complex than just fixing the `meson.build` file. You may not expected to do such changes. If you don't want to do this change yourself, open a issue and I will do it myself.

Of course, if you want to do it, it would be welcome :)

nirbheek:
So to give a bit of context, I found these issues while building Kiwix for [the 0.44.1 release](https://groups.google.com/d/msg/mesonbuild/nxAki2bkhEo/TfZ_aP0XBAAJ), since your project is on the [Meson Users page](http://mesonbuild.com/Users.html).

I did not know about kiwix-tools; I will try to integrate that as a meson subproject while regression testing for the next release. :)

I noticed kiwix-build, but I thought I'd try to get kiwix-lib building with subprojects since it was almost there. I think having all your libraries and tools building as subprojects would be a neat way to have a trivially-buildable stack which is all-static, builds quickly and with zero dependencies on all platforms, and yields portable binaries.

Would also make it easy for me to test your project for regressions before a new release ;)

meson.build
nirbheek: This change is not about static libraries though; the problem is that libpugixml-dev on Fedora does not ship pkg-config files, so `dependency()` doesn't work, and we have to encode a fallback with `find_library()`. The FIXME is because a missing feature means we can't yet tell meson to force static compilation with find_library().

kelson42:
@mgautierfr What is the status of this PR?

kelvinhammond:
@nirbheek What is the status of this PR?

nirbheek:
Well, I opened this 2 years ago because I found that these changes would make it easier to build this repo for meson release testing. I no longer do that, so I will not be working on this PR.

It seems kiwix is already tested in the meson corpus test that is currently maintained, so this PR is not needed either: https://travis-ci.org/github/jon-turney/meson-corpus-test

kelson42:
@mgautier We mist review this PR!

Issue #70, opened 2017-07-19
Wrong redirection handling


kelson42:
Currently the kiwix-lib, if a redirect is found, search the target and return the target page body. It works so far the targeted page is at the same path level like the redirect. If not all relative URLs will probably fail to load properly.

As a workaround, in kiwix-serve a <base> node as been introduced, but this is not the proper answer to that problem and this fixes only kiwix-serve.

To fix that problem, probably the kiwix-lib should returns a proper code or HTML redirection code.

mgautierfr:
> As a workaround, in kiwix-serve a node as been introduced, but this is not the proper answer to that problem and this fixes only kiwix-serve.

Hooo, that why the `base` node is for ? I was wondering why we add this node while rewriting the url in the document in the same time.

> To fix that problem, probably the kiwix-lib should returns a proper code or HTML redirection code.

I agree with this. Redirections are better handled on the client side.
But it is a bit tricky to do. It is a big API change as all clients assume that they will not get any redirection and directly get the content to display.

Until this issue is fixed, `kiwix-lib` has a method that return the content of the asked `url` but also the `baseUrl` (who is the url of real article used). This could be used to detect a redirection and fix url in the webview.
Sadly, this is not exposed in the JNI API. I can quickly add it (as I made it for the title argument https://github.com/kiwix/kiwix-lib/commit/d3acae1fd2bc56cfc38e005791eac2d9eba64b71)

kelson42:
Yes, we should probably add a new parameter to the existing kiwix-lib method(s) to specificaly avoid to resolve redirects.

okkebal:
The new run of StackExchange is being published to the Android version. The files show up in the "online" tab in "get content". So android users will start downloading and trying these files. Due to this issue the pages do not load correctly. This will be a bad experience for our users. The files should only be published in the Android version if they work in the Android version.

automactic:
I think we just need to add something similar to `std::vector<std::string> getRedirectedURL(const std::string& url) const;`, but I know this might be easier said than done

kelson42:
@automactic What would trigger this call? Would you call it each time you try to open an url?

automactic:
@kelson42 I think so, yes, a call will be made every time opening a URL, since there is no way to tell if a url should redirect and every url could potentially redirect. 
-----
Another way to solve this is to get rid of redirections. Think of urls primary key of resources (html, css, js, etc) in a zim file. We resolve all the redirections during offline time.

[...]

PR #123, opened 2018-03-15
New api

mgautierfr:
Introduction of the new api.

This try to correct problem raised in the issue #70.

The new API introduce the new class `Entry` that act as a proxy to an
article in the zim file.

- Methods of `Reader` now return an `Entry` and the user has to call
`Entry`'s methods to get useful information.
- No redirection is made explicitly.
- If an entry is not found, an exception is raised instead of returning
an invalid `Entry`.

The common pattern to get the content of an entry become :

```
std::string content;
try {
  auto entry = reader.get_entry_from_url(url);
  entry = entry.get_redirect_entry();
  content = entry.content();
} catch (NoEntry& e) {
  ...
}
```

To check if entry is a redirection :

```
try {
   auto entry = reader.get_entry_from_url(url);
   bool is_redirect = entry.is_redirect();
   std::string url = entry.get_long_url();
   std::string redirect_url = entry.get_redirect_entry().get_long_url();
   // url == redirect_url if entry is not a redirect entry.
} catch (NoEnrty& e) {
   ...
}
```

@automactic, @mhutti1 please have a look on this change and tell me what you think about the new API. Do you see any blocker with this API ? Missing methods ?
I haven't create the JNI wrapper for the new API but it will be very close to the C++ API.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -1,0 +1,24 @@
+
+#ifndef _KIWIX_COMMON_H_
+#define _KIWIX_COMMON_H_
+
+#include <zim/zim.h>
+
+#ifdef __GNUC__
+#define DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER)
+#define DEPRECATED __declspec(deprecated)
+#else
+#praga message("WARNING: You need to implement DEPRECATED for this compiler")
+#define DEPRECATED
+#endif
+
+
+namespace kiwix {
+
+typedef zim::size_type size_type;
+typedef zim::offset_type offset_type;
+
+}
+
+#endif //_KIWIX_COMMON_H_
--- a/include/entry.h
+++ b/include/entry.h
@@ -1,0 +1,191 @@
+/*
+ * Copyright 2018 Matthieu Gautier <mgautier@kymeria.fr>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#ifndef KIWIX_ENTRY_H
+#define KIWIX_ENTRY_H
+
+#include <stdio.h>
+#include <zim/article.h>
+#include <exception>
+#include <string>
+#include "common.h"
+
+using namespace std;
+
+namespace kiwix
+{
+
+
+class NoEntry : public std::exception {};
+
+/**
+ * A entry represent an.. entry in a zim file.
+ */
+class Entry
+{
+  public:
+    /**
+     * Default constructor.
+     *
+     * Construct an invalid entry.
+     */
+    Entry() = default;
+
+    /**
+     * Construct an entry making reference to an zim article.
+     *
+     * @param article
+     */
+    Entry(zim::Article article);
+    virtual ~Entry() = default;
+
+    /**
+     * Get the path of the entry.
+     *
+     * The path is the "key" of an entry.
+     *
+     * @return the path of the entry.
+     */
+    std::string getPath() const;
+    
+    /**
+     * Get the title of the entry.
+     *
+     * @return the title of the entry.
+     */
+    std::string getTitle() const;
+        
+    /**
+     * Get the content of the entry.
+     *
+     * The string is a copy of the content.
+     * If you don't want to do a copy, use get_blob.
+     *
+     * @return the content of the entry.
+     */
+    std::string getContent() const;
+        
+    /**
+     * Get the blob of the entry.
+     *
+     * A blob make reference to the content without copying it.
+     *
+     * @param offset The starting offset of the blob.
+     * @return the blob of the entry.
+     */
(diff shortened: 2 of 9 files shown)


automactic:
One question: why are all functions in `Entry` snake cased? Don't we always use camel case? like `getContentByEncodedUrl `?

automactic:
What is the difference between a long url and an absolute url? If long url always have namespace as the first path component, then it is just absolute url, isn't it?

automactic:
Does `entry.get_redirect_entry()` throw `NoEntry` when the entry in question does not have redirection? If yes, please add it in the docs.

Also if yes, I think in this case it would be better to throw `NoRedirectedEntry`. NoEntry reminds me too much of the traffic sign. :D

automactic:
Does redirection also happen on non html resources?

automactic:
What does `link_target` do?
Under what situation will an Entry be deleted?
Also why name the class Entry, would Resource be a better name?

kelson42:
@automactic Yes, you can have redirect on any type of content.

mhutti1:
This seems fine from an Android point of view it should integrate fine with our current setup.

mgautierfr:
> One question: why are all functions in Entry snake cased? Don't we always use camel case? like getContentByEncodedUrl?

Well, I use a lot of snake case coding style (and I prefer it). I've automatically switch to it.
But whatever my personal preferences are, we should stick to the same coding style. I will change it.

> What is the difference between a long url and an absolute url? If long url always have namespace as the first path component, then it is just absolute url, isn't it?

No real difference, it is just that long url is the name used in libzim.
Maybe that `long_url` is "N/url" where `abs_url` should be "/N/url".

But we have some plan to remove namespace in the future (https://github.com/openzim/libzim/issues/15)
In this context the long url, absolute url and url would be the same.
Maybe we should use the names `url` and `short_url` (url without namespace).
Most usage always use the full url (so `url`) and in the future `short_url` will deprecated so we should probably directly use the right name (`url`) to avoid changes in the future ?
What do you think, `url`, `namespace` and `short_url` is good for you ?

> if entry.is_redirect() returns true, what would it's long url be like?

The (long) url of the entry itself.
There is no following of redirection made implicitly.
If you what the "final" entry, you should use `entry.get_redirect_entry()`

> Does entry.get_redirect_entry() throw NoEntry when the entry in question does not have redirection? If yes, please add it in the docs.
>
> Also if yes, I think in this case it would be better to throw NoRedirectedEntry. NoEntry reminds me too much of the traffic sign. :D

`entry.get_redirect_entry()` may throw a `NoEntry` exception but only if the entry redirect to a non existing entry. (Saying that it should throw a `zim::ZimFileFormatError`, but `get_redirect_entry() is a kiwix function and `ZimFileFormatError` is a libzim exception. I'm not sure we should do that).

If the entry is not a redirect_entry, `entry.get_redirect_entry` is the entry itself.
Saying that, the method should be called `get_final_entry` or something like that.

`NoEntry` can be thrown by any function returning a entry. So it make no sense to rename it to `NoRedirectEntry` and I'm not sure it worth it to create a new exception for this specific case.

> Does redirection also happen on non-html resources? like image, css, etc.?

As said kelson, yes.
It is totally independent of kiwix/libzim. It depends of how the zim file has been constructed.

> What is a link target entry?
Under what situation will an Entry be deleted?

As far I know, this is not used. And I don't really know what they are.
@kelson42 may help here.

> Also why name the class Entry, would Resource be a better name?

I don't know. Why `Resource` would be a better name ?
I just know that I don't want to call it `Article` because it can represent a lot more things than a article (html pages, layout, css, image, js, metadata, xapian database, whatever ...)

`Entry` definition is "a separate piece of information that is recorded in a book, computer, etc" and that is what it is : Something in a zim file.

But maybe `Resource` is better, I really don't know :)

mgautierfr:
@automactic I've you any comments on the new version of this PR or on my comments ?

Do you think I can merge it or there is still modification to do ?

automactic:
@mgautierfr, I haven't got the time to try it in the iOS app yet. But I think it should work and you can merge, as long as the functions are in camel case.

kelson42:
kiwix-serve should be patched to rely on this.

automactic:
@mgautierfr, I have 2 questions:

1. does `Entry::getUrl()` return absolute url? If so, the returned url should always start with `'/'`
2. `Entry::isRedirect()`, is this API meant to test if the entry is a redirection from another entry or if the entry has redirection?

automactic:
I agree that it should be called path, which I actually do in code of kiwix for iOS. But in the context of an Entry, I think `getUrl` should always return a path start with `/`, so that user of this library know the url is related to the root of the zim file. On the other hand, if the paths does not start with `/`, what is it related to inside the zim file?

I guess what I really want and need is for APIs in libkiwix to be consistent. We should decide if path returned by APIs all start with `/` or not.

automactic:
I fixed some typos in the doc string. Xcode keeps emits warnings wanting me to fix them.

mgautierfr:
@automactic Thanks

I've just push a new version of the API. The main changes are :
- Rename all new method with the name "path" instead of "url".
- As we want in the future to remove the namespaces in zim files, I get a head start and remove the notion of namespace from the new API. Path are always "full" (with the namespace included). It should not be a problem as I don't see a real use case where you (kiwix-android, kiwix-apple, ...) want to know the namespace of a Entry (and it can still be infer from the first char of the path if needed).

@automactic, @mhutti1 If there is no objection to this changes, I merge the PR quickly.

automactic:
@mgautierfr, just to be clear, the paths you mentioned, do they always start with `/` or never start with `/`?

automactic:
@mgautierfr, OK. I am good as long it is consistent. (Please update Searcher, etc. to also reflect this change.)

But does it mean old zim files will still have `/` as a prefix in path? Since it sounds like kiwix-lib just return whatever is stored in the zim file.

mgautierfr:
> But does it mean old zim files (or some certain zim files) will still have / as a prefix in path? Since it sounds like kiwix-lib just return whatever is stored in the zim file.

Yes. But I would consider those zim as invalid zim files. (And as far as I know, old zim files doesn't have a `/` prefix in paths)

PR #129, opened 2018-03-27
[WIP] Opds

mgautierfr:
Add support to opds :

- Add a new class `OPDSDumper` to generate a opds feed from a library.
- Add a new method `readOpds` to `Manager` to load a opds feed content into a library.
- Add a new class `Download` to download file with aria2.

[WIP] This depends on libaria2. So waiting `kiwix-build` to generate correct base dependencies archive.


legoktm:
Just to confirm, is the libaria2 provided by https://tracker.debian.org/pkg/libaria the correct dependency? (cc @kelson42 since this is the code I think you mentioned to me on IRC the other day)

kelson42:
@legoktm I checked, it's now. We talk about the lib behind the package "aria2c", see http://aria2.sourceforge.net/

legoktm:
Got it - it appears to be https://tracker.debian.org/pkg/aria2 then. I'll reach out to the maintainers about creating a package for the library itself. :)

kelson42:
@mgautierfr The README should be updated regarding new aria2lib dep

PR #141, opened 2018-05-20
Decode reserved characters in URLs

mhutti1:
Fixes https://github.com/kiwix/kiwix-android/issues/633

The URLs in the ZIM file are formated ..../Book%3ACancer.... and the %3A is not decoded correctly at the moment.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -464,7 +464,7 @@ Entry Reader::getEntryFromPath(const std::string& path) const
 
 Entry Reader::getEntryFromEncodedPath(const std::string& path) const
 {
-  return getEntryFromPath(urlDecode(path));
+  return getEntryFromPath(urlDecode(path, true));
 }
 
 Entry Reader::getEntryFromTitle(const std::string& title) const


mgautierfr:
I've difficulties to understand the `component` argument of `decodeUrl` and why it has been introduced.
@kelson42 It seems that it is you that introduce it (in 28b0588df4903f2bdc62ef2abf3ed858d84bcbd3), can you explain it ?

mgautierfr:
I mainly disagree with this PR.
The commit 28b0588 mainly add the `component` argument to avoid decode special characters.
This commit was to fix bugs where we were "wrongly"[\*] trying to decode url with `%`. (kiwix/kiwix-tool#56, kiwix/kiwix-android#50, kiwix/kiwix-tools#96)

If we set `component` to `true` we try again to decode special characters, so the opposite of the commit 28b0588. Can you explain a bit more how this change will not break the previous fix ?

\* For kiwix/kiwix-android#50 at least it seems to me that the problem was not in `kiwix-lib` but in `kiwix-android` (and it is probably the same problem in other clients). The API used take a url encoded url. It seems normal to me that `kiwix-lib` tries to decode it and fails if the url given is not url encoded. The client (`kiwix-android`, ...) should encode the url before passing it to `kiwix-lib` or use a API that take a "not url encoded" url (`kiwix-lib` has to be extended if this API doesn't exist but we should not change the behavior a good API wrongly used)

mhutti1:
I don't quite understand the part about clients. The URL in the ZIM file is encoded with : being encoded as %3A. When we pass this URL to kiwix-lib just as we found it to a method that is supposed to decode this URL for us it doesn't and as such can't find the file it points to. This is surely a problem with kiwix-lib not kiwix-android.

mgautierfr:
Yes, but it is this PR https://github.com/kiwix/kiwix-lib/pull/97 that change the behavior of kiwix-lib to not decode special character. And it was because `kiwix-android` (and other) were trying to open url with a `%` not escaped inside (so not url encoded).

This PR may fix your bug, but it is probable that it will break the fix of #97.
I would like to discuss this further to be sure that nothing is broken and that we correctly fix all those issues.

kelson42:
@mgautierfr @mhutti1 From what I remember and read from the comment kiwix/kiwix-tools#96 fondamental problem was that the random url returned by the kiwix-serve was wrongly encoded. This is what I have fixed at that time by copying the whole encoding/decoding logic from an other project. By doing so, it looks like I have added a regression by forgetting to properly decoded all the characters of the entry path, in particular here the slash. I have tested again the random feature with nightly kiwix-serve and it seems to work well, ie. that the url returned by the random feature is properly encoded.

mgautierfr:
If the problem was at encoding steps, maybe we could simply remove the `component`  argument of `urlDecode` ?
I don't see a use case where we would like to have two different behaviors for url decoding.
I'm really not fan to keep extra arguments we don't use and cannot explain.

kelson42:
@mgautierfr Yes, this might be something to consider... but we might also need these options if we give full URL with all the reserved characters. It is not the use case here, but might change.

PR #156, opened 2018-06-30

cyrillemoureaux:
This will prevent https://github.com/kiwix/kiwix-tools/issues/207, although it still doesn't address the fact that carrying on in kiwix-serve with a redirect entry ends up in a crash. It also doesn't explain how the writer created a file with a self-redirect in the first place.


kelson42:
@cyrillemoureaux Thx, can you just please rebase your branch on git master HEAD?

mgautierfr:
Superseded by #157 and #214

PR #157, opened 2018-06-30

cyrillemoureaux:
Take 2 at preventing https://github.com/kiwix/kiwix-tools/issues/207, previous pull request (https://github.com/kiwix/kiwix-lib/pull/156) ended in a weird state and my git-fu isn't good enough to sort it out.

### DIFF
--- a/src/entry.cpp
+++ b/src/entry.cpp
@@ -122,7 +122,6 @@ Entry Entry::getFinalEntry() const
   if (final_article.good()) {
     return final_article;
   }
-
   int loopCounter = 42;
   final_article = article;
   while (final_article.isRedirect() && loopCounter--) {
@@ -131,7 +130,10 @@ Entry Entry::getFinalEntry() const
       throw NoEntry();
     }
   }
-
+  // Prevent infinite loops.
+  if (final_article.isRedirect()) {
+    throw NoEntry();
+  }
   return final_article;
 }
 


src/entry.cpp
```diff
@@ -122,7 +122,11 @@ Entry Entry::getFinalEntry() const
   if (final_article.good()) {
     return final_article;
   }
-
+  // Prevent infinite loops.
+  if (article.isRedirect() &&
+      article.getIndex() == article.getRedirectIndex()) {
+    throw NoEntry();
+  }
```
mgautierfr: I think that this test should be in (or after) the `while`.

Here, it is only catching if the first redirection is redirecting to itself. But the loop could be later on, add a bit more complex (A->B->C->A).

The simpler would be to check that `loopCounter>0` after the `while`. If `loopCounter==0`, it means that we end the `while` because of too many steps. It is probably because of a infinite loop and we should throw `NoEntry`.
This way, we should any kind of infinite loop.

cyrillemoureaux:
I modified the change to have a more general approach to detecting cycles within the limit of attempts. My very original fix for the issue was to throw if at the end final_article was still a redirect, but then I had doubts that there could be some clients which depended on the previous state of returning "final" articles which were still redirect, so talking with kelson42 we came to the middle ground of just detecting self-redirects but leaving the "going past the counter" behaviour intact.

This said, I agree detecting broader cycles than just immediate ones is a good idea so that's what I tried to do with this new version.

mgautierfr:
I think it would be far simple to just check that the final article is not a redirect before returning it (and throw NoEntry if it is).
It would avoid the creation of a set each time we want to get an entry.
There is no real use case of having a redirection chain longer than 42. If we have a too long redirection chain, this is an error.

(Please, rewrite your commit instead of creating another one. Once we merge your PR, we don't need to have the intermediate (wrong) commits)

src/entry.cpp
cyrillemoureaux: I dropped the second commit I'd attempted and went back to pretty much the very first version of the change I had, namely just checking that final_article isn't a redirect, which should address the issue.

mgautierfr:
I have merged your change with #214.
Thanks @cyrillemoureaux.

Issue #189, opened 2018-12-17
ctpp2 is no longer available


hashworks:
The website (http://ctpp.havoc.ru/) referenced in the README file is no longer available.

kelson42:
Thank you for the feebdack. We are working anyway to replace ctpp2 by a better template engine. See #21. I will close this ticket as it basically is a duplicate of it.

PR #191, opened 2019-01-08
Move the templating system to mustache instead of ctpp2.

### LINKED ISSUE #21: Replacing ctpp2 by ctemplate HTML template engine
ctpp2 seems to not been really maintained anymore (no release in 5 years) and is really not easy to use (complicated compilation scripts and needs to be patched).

It might be smartter to replace it. I have found a library which might be able to do the job:
https://github.com/olafvdspek/ctemplate

It's maintained and already packaged in Debian/Ubuntu

mgautierfr:
Mustache templating system is a bit simpler than ctpp2 and ctpp2 is no
more maintained (see #189).
We are moving to the kainjow's Mustache project
(https://github.com/kainjow/Mustache).

It simplify a lot our system has it is header only and we don't have to
precompile the template.

Fix #21


legoktm:
Catching up on kiwix things...THANK YOU! ctpp2 was such a pain to deal with. I'm packaging the kainjow/Mustache library for Debian now and it is so pleasant to use.

PR #212, opened 2019-04-10
Removed Rpath from compiled binaries

xvitaly:
Rpath is not needed to run installed binaries. Most of GNU/Linux distributions strictly forbid it.

See also:
 * https://lintian.debian.org/tags/binary-or-shlib-defines-rpath.html
 * https://wiki.debian.org/RpathIssue
 * https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -44,5 +44,4 @@ kiwixlib = library('kiwix',
                    dependencies : all_deps,
                    version: meson.project_version(),
                    install: true,
-                   install_dir: install_dir,
-                   install_rpath: '$ORIGIN')
+                   install_dir: install_dir)


kelson42:
@xvitaly Thank you very much. We definitly want to get rid of the rpath (we have a couple of tickets open about that). But @mgautierfr was just about to remove it... so please don't take it personally if your PR is just cancelled.

src/meson.build
```diff
@@ -43,6 +43,4 @@ kiwixlib = library('kiwix',
                    include_directories : inc,
                    dependencies : all_deps,
                    version: meson.project_version(),
-                   install: true,
-                   install_dir: install_dir,
-                   install_rpath: '$ORIGIN')
+                   install: true)
```
mgautierfr: You also remove the `install_dir` option.

This is needed for android installation.

xvitaly: Yes. Restored.

xvitaly:
Restored `install_dir` option.

mgautierfr:
Can you just rebase on new origin/master please ?

xvitaly:
Rebased to fresh master branch.

mgautierfr:
Thanks.

PR #214, opened 2019-04-15

mgautierfr:
This will prevent kiwix/kiwix-tools#207.


mgautierfr:
This PR is the same than #157. I had just rebased it to master.

PR #218, opened 2019-04-29
Add pause and unpause functions for aria2

jetownfeve21:
This functions enable to stop and resume download with aria2.

The Downloader's constructor now checks the paused downloads with the
function "tellWaiting()" to get them at the start of kiwix-desktop.

### DIFF
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -233,4 +233,11 @@ void Aria2::unpause(const std::string& gid)
     doRequest(methodCall);
 }
 
+void Aria2::remove(const std::string& gid)
+{
+    MethodCall methodCall("aria2.remove", m_secret);
+    methodCall.newParamValue().set(gid);
+    doRequest(methodCall);
+}
+
 } // end namespace kiwix
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -171,7 +171,6 @@ std::vector<std::string> Aria2::tellActive()
   MethodCall methodCall("aria2.tellActive", m_secret);
   auto statusArray = methodCall.newParamValue().getArray();
   statusArray.addValue().set(std::string("gid"));
-  statusArray.addValue().set(std::string("following"));
   auto responseContent = doRequest(methodCall);
   MethodResponse response(responseContent);
   std::vector<std::string> activeGID;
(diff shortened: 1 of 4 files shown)


src/aria2.cpp
```diff
@@ -186,6 +186,28 @@ std::vector<std::string> Aria2::tellActive()
...
+{
+  MethodCall methodCall("aria2.tellWaiting", m_secret);
+  methodCall.newParamValue().set(0);
+  methodCall.newParamValue().set(99); // max number of downloads to be returned, don't know how to set this properly assumed that there will not be more than 99 paused downloads.
+  auto statusArray = methodCall.newParamValue().getArray();
+  statusArray.addValue().set(std::string("gid"));
+  statusArray.addValue().set(std::string("following"));
+  auto responseContent = doRequest(methodCall);
+  MethodResponse response(responseContent);
+  std::vector<std::string> activeGID;
```
mgautierfr: This should be `waitingGID` :)

src/aria2.cpp
```diff
@@ -186,6 +186,28 @@ std::vector<std::string> Aria2::tellActive()
...
 }
 
+std::vector<std::string> Aria2::tellWaiting()
+{
+  MethodCall methodCall("aria2.tellWaiting", m_secret);
+  methodCall.newParamValue().set(0);
+  methodCall.newParamValue().set(99); // max number of downloads to be returned, don't know how to set this properly assumed that there will not be more than 99 paused downloads.
+  auto statusArray = methodCall.newParamValue().getArray();
+  statusArray.addValue().set(std::string("gid"));
+  statusArray.addValue().set(std::string("following"));
```
mgautierfr: It seems that we don't use the "following" key here.
This is the same in `getActive` method, we could remove this in both methods.

src/downloader.cpp
```diff
@@ -93,6 +93,22 @@ void Download::updateStatus(bool follow)
   }
 }
 
+void Download::resumeDownload()
+{
```
mgautierfr: I'm not sure but we may need to update the status here.
Either before, to be sure that the `m_followedBy` is correctly set.
Or after, to be sure that the status is updated after the call.

src/downloader.cpp
```diff
@@ -111,6 +111,14 @@ void Download::pauseDownload()
     updateStatus(true);
 }
 
+void Download::cancelDownload()
+{
+    if (!m_followedBy.empty())
+        mp_aria->remove(m_followedBy);
+    else
+        mp_aria->remove(m_did);
```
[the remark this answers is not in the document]
jetownfeve21: I did but it crashed because aria2 removed the download and when we try to update the status, it is not available anymore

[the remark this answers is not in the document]
mgautierfr: Forcing the status to `Download::K_REMOVED` as you did is good.

src/downloader.cpp
```diff
@@ -139,6 +139,7 @@ Download* Downloader::startDownload(const std::string& uri)
 Download* Downloader::getDownload(const std::string& did)
 {
   try {
+    m_knownDownloads[did]->updateStatus(true);
```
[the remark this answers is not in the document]
mgautierfr: This is still to change.

src/aria2.cpp
```diff
@@ -199,5 +219,25 @@ void Aria2::shutdown()
   doRequest(methodCall);
 }
 
+void Aria2::forcePause(const std::string& gid)
+{
+    MethodCall methodCall("aria2.forcePause", m_secret);
```
mgautierfr: I would prefer to use the "pause" method instead of "forcePause" (same for "remove"/"forceRemove").
The normal behavior should be "pause" to allow aria2 to correctly close the connection. We should use "force*" only if really necessary.

mgautierfr:
Good ! Time to merge this.

PR #220, opened 2019-05-06
descending sort book

jetownfeve21:

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -190,6 +190,7 @@ class Library
   std::vector<std::string> listBooksIds(
     int supportedListMode = ALL,
     supportedListSortBy sortBy = UNSORTED,
+    const bool sortOrderAsc = true,
     const std::string& search = "",
     const std::string& language = "",
     const std::string& creator = "",
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -247,6 +247,7 @@ std::string Comparator<PUBLISHER>::get_keys(const std::string& id)
 std::vector<std::string> Library::listBooksIds(
     int mode,
     supportedListSortBy sortBy,
+    const bool sortOrderAsc,
     const std::string& search,
     const std::string& language,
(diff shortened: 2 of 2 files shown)


kelson42:
@jetownfeve21 @mgautierfr This branch is quite old and should be reviewed ASAP and if OK merged. I just have rebased it.

mgautierfr:
This is pretty complex and with a lot of duplication.
There are two things here :
- The key we are using to sort the books
- The order we are using.

This is two different things and it should (and could) be handle separatly.
The Comparator could be extended with another template parameter for the order. And the template specitalization being used for implementing the order.

Something like that (not tested) :
```c++
template<supportedListSortBy sort, bool asc_order>
struct Comparator {
  Library* lib;
  Comparator(Library* lib) : lib(lib) {}

  bool operator() (const std::string& id1, const std::string& id2) {
    return get_keys(id1) < get_keys(id2);
  }

  template<typename KEYTYPE>
  bool compare(const KEYTYPE& a, const KEYTYPE& b);
  std::string get_keys(const std::string& id);
  unsigned int get_keyi(const std::string& id);
};
```

I always thing that the api of the listBookIds is pretty complex (too many arguments).
We may rewrite it later but it is not for now. To avoid to add a extra arguments to the functions, we could indeed extend the `supportedListSortBy` enum, but the implementation doesn't have to follow (and use) the same enum.

mgautierfr:
This is an API change. And kiwix-tools (at least) use this api. Please provide a PR on kiwix-tools side to adapt to the new api (this is already done on kiwix-desktop side with kiwix/kiwix-desktop#142)

mgautierfr:
I've just created a PR for a better API for library filtering and sorting here (https://github.com/kiwix/kiwix-lib/pull/234). I let you review this PR but I think it would be better to use the new API than extends the existing one.

PR #221, opened 2019-05-10
kiwix-serve integration in kiwix-lib

jetownfeve21:


jetownfeve21:
@mgautierfr 
I moved the subprocess.h into the "include" directory to compile but I don't know if it's the right way to do it

kelson42:
It is really important attachProcessOpt works fine

jetownfeve21:
Right now I delete the process with the kill() method of the "SubProcess" class instead

kelson42:
AFAIK all the problems to do so have already been fixed in https://github.com/kiwix/kiwix-xulrunner/blob/master/src/components/serverManager/serverManager.cpp

jetownfeve21:
Just add ' ' around the pid string solved the problem

include/meson.build
```diff
@@ -9,7 +9,9 @@ headers = [
   'downloader.h',
   'reader.h',
   'entry.h',
-  'searcher.h'
+  'searcher.h',
+  'kiwixserve.h',
+  'subprocess.h'
```
mgautierfr: This is the list of header files to install.
Header we don't want to install (not part of the public API) don't need (and must not) to be installed.
So we don't need `subprocess.h` here.

include/kiwixserve.h
```diff
@@ -0,0 +1,32 @@
+#ifndef KIWIXLIB_KIWIX_SERVE_H_
+#define KIWIXLIB_KIWIX_SERVE_H_
+
+#ifdef _WIN32
+// winsock2.h need to be included before windows.h (included by curl.h)
+# include <winsock2.h>
+#endif
+
+#include "subprocess.h"
```
mgautierfr: We must not include subprocess here. This is a private header.
You can simply declare the `Subprocess` class and include the `subprocess.h` header in the cpp file.

PR #223, opened 2019-05-14
add tags in opds stream

jetownfeve21:
https://github.com/kiwix/kiwix-tools/issues/252


kelson42:
@mgautierfr said this is OK

PR #229, opened 2019-06-06
other category selects books that don't have categorized tags

jetownfeve21:
fix https://github.com/kiwix/kiwix-desktop/issues/119

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -194,6 +194,7 @@ class Library
     const std::string& language = "",
     const std::string& creator = "",
     const std::string& publisher = "",
+    bool excludeTags = false,
     const std::vector<std::string>& tags = {},
     size_t maxSize = 0);
 
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -251,6 +251,7 @@ std::vector<std::string> Library::listBooksIds(
     const std::string& language,
     const std::string& creator,
     const std::string& publisher,
+    bool excludeTags,
     const std::vector<std::string>& tags,
     size_t maxSize) {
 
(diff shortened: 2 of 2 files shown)


kelson42:
@jetownfeve21 After last weekly meeting, it looks like this PR does not behave link intend. "Other" should list all entries not available in other categories. What you seem to have done is to gather any entry without tags.

include/library.h
```diff
@@ -194,6 +194,7 @@ class Library
     const std::string& language = "",
     const std::string& creator = "",
     const std::string& publisher = "",
+    bool other = false,
```
mgautierfr: Please use another name than "other".
"Other" make sense for the client of the library (kiwix-desktop) where we are looking for all other categories.
Here we are filtering book depending if the book has or has not the provided tags.
Something like "excludeTags" makes more sense (even if this is far than perfect, but all the api is crap)

src/library.cpp
```diff
@@ -277,11 +278,13 @@ std::vector<std::string> Library::listBooksIds(
...
       for (auto& t: tags) {
-        if (sBookTags.find(t) == sBookTags.end()) {
+        if ((!other && sBookTags.find(t) == sBookTags.end()) || (other && sBookTags.find(t) != sBookTags.end())) {
           // A "filter" tag is not in the book tag.
           // No need to loop for all "filter" tags.
           ok = false;
-          break;
+          if (!other) {
+            break;
+          }
```
mgautierfr: We don't need to check `other` here.
If `ok` is set to false, it can never be made true again. So we don't need to continue the loop. Whatever the value of `other` is.

mgautierfr:
The new api of filter (#234) has an option to exclude tags. We should use it.

PR #231, opened 2019-06-20
Create library and set up publishing task

macgills:
Fixes: https://github.com/kiwix/kiwix-android/issues/489 and https://github.com/kiwix/kiwix-build/issues/257
This is an adaptation of the library supplied to me by @mhutti1 
The actually important file here is publish.gradle specifically
  
`    libraryVersion = '1.0.12'`

and

```
user = System.getenv('BINTRAY_USER') 
key = System.getenv('BINTRAY_API_KEY')
```

once these are provided and the task `gradlew bintrayUpload` is run then we will have a new published artifact on bintray


kelson42:
@mgautierfr @mhutti1 I think the library version - as stored in bintray - should be dynamically related to the lib version or the lib code (git tag?). Otherwise, I already see problem happening that `publish.gradle` has been forgotten to be updated.

macgills:
@kelson42 There is an `overwrite` flag on the bintray plugin that defaults to false so by default it will not allow you to upload the same version. I was waiting for @mgautierfr response on how he wants to provide the version, I made the assumption of environment variables for the username/apikey just because I wasn't going to upload those into version control

macgills:
@mgautierfr those files are there so local testing can be done of the upload task, I think they still have value in being present if only for this but if in light of this information you still think they should be gone I will remove them.
What proposal do you have the `libraryVersion`? Should I just make it a env var and you can set it to whatever you want?

mgautierfr:
> those files are there so local testing can be done of the upload task, I think they still have value in being present if only for this but if in light of this information you still think they should be gone I will remove them.

Testing is good but if you want to merge, it is not about testing anymore :)
And it may make the integration in kiwix-build more error prone. If I forget a file while creating the directory structure before running the upload task, it may not break because the "test" files will be present.

> What proposal do you have the libraryVersion? Should I just make it a env var and you can set it to whatever you want?

I would keep it as it is. I already have to change the version in meson.build (https://github.com/kiwix/kiwix-lib/blob/master/meson.build#L2) and in the Changelog. It will be simpler to change it in the project (and after all, this is the advantage to have the upload code in the same repository) than having some another cross configuration between kiwix-build and kiwix-lib.

macgills:
@mgautierfr there will be further small development of this project as I will be adding potentially a snapshot publisher as well, I am happy to delete the java JNI and .so files but the gradle makes me a bit hesitant. Can you live with the gradle there or is it a "must go"?

kelson42:
@macgills I hope I understand the discussion properly but I would support @mgautierfr on this: none of the compiled/generated files should be in the git. Only the sources should be there. Not event the build dependences (like gradle). This is a general rule. For testing the system, people need to compile and if they can't, then we might have a problem -> a ticket should be open in that repo.

mgautierfr:
> Can you live with the gradle there or is it a "must go"?

I confirm what @kelson42 said. And yes, gradle itself (gradle-wrapper.jar, graddlew, graddlew.bat, ...) should be removed. The gradle "configuration" files (build.gradle, publish.gradle, ...), should be keep.
The "golden rule" is that "If something can be found elsewhere or generated, it should not be versioned"

---
I haven't look your code itself, for now I'm just telling that some files should not be there.

macgills:
@kelson42 @mgautierfr  I completely agree, source is source, however gradle is not compiled or generated at run time, it is a default, necessary thing for the compilation of the project. Every android project in existence will have these files here. 
You specify which version of gradle you want in your properties, this is a guarantee about which version of gradle your project is compatible with, the gradlew command will look in your cache and if it does not find this version will go and download it. Gradle and Android build tools have a very close relationship and mixing versions willy nilly is a recipe for disaster

mgautierfr:
Let me tell how I disagree with this :smiley: 
Build process should not download things. This global rule is used almost everywhere in linux environment. Even more it is enforced by meta build tool as rpmbuild, flatpack-builder, ...
There is a step to setup source (when the build tool itself will dl your source) and then a build step most of the time run in a container without network access.

There are other ways to enforce the version of the build tools :
- meson has an option meson_version in the project function (https://mesonbuild.com/Reference-manual.html#project)
- cmake has https://cmake.org/cmake/help/v3.0/command/cmake_minimum_required.html
- automake has https://autotools.io/automake/options.html, autoconf has AC_PREREQ (https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Versioning.html)
- ...

If all android projects embed gradlew it is mainly because it is easier than asking developers to correctly install gradle.

This subproject (android-kiwix-lib-publisher) is here only because it is easier to use gradle tool than create another one in python (or anything else) using the public rest api of bintray. It probably resume itself in two or three files with just a "upload" command.
It is intended to be launch by kiwix-build only (because it needs credential only known by kiwix-build) and kiwix-build already download and install gradle (https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/dependencies/gradle.py) and build android without using gradlew (https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/dependencies/base.py#L505-L527)

macgills:
> There is a step to setup source (when the build tool itself will dl your source) and then a build step most of the time run in a container without network access.

This would be absolutely untenable, builds rely on dependency resolvers and download all their dependencies from maven/jcenter/google's repository(on android)/a company's own nexus. We do not package our dependencies in source so you would have to - before starting this networkless container - parse the build.gradle downloading all these dependencies yourself and placing them manually into the cache. 
The question of testing arises as well, a build should not be successful without all the tests passing, a framework on android has caused me no end of trouble because when you tell it to target an SDK level (this is jvm tests but it pretends to be android) download it hits off to sonatype before the test will start if it has not cached the SDK and this was hitting a firewall in a previous company. And even though I am generally against a test hitting the network, if a specific requirement demanded it I would not prohibit it.

Gradlew is packaged because it guarantess that all developers and all builds will use the same version, the projects version.

```
gradle-5.2-bin.zip'
```
^^ from gradle.py, this will have to be constantly maintained and could easily break and lead to confusing errors. I would not be surprised if this has already happened with kiwix-android.
Come to think of it you are building kiwix-android with its gradle files all there?

mgautierfr:
> ^^ from gradle.py, this will have to be constantly maintained and could easily break and lead to confusing errors.

This is the same here, you will have to do PR on kiwix-lib to change the version you want. Even more, from this doc (https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper) it seems that the recommended way it to regenerate a wrapper when you want to update the gradle version.

We could discuss a lot about how android developers manage their build system but it will not help a lot :)
Lets back move to our case.

There are two visions here.
- The first one (the one I had at beginning) is that this project is not a "real" android project. It is just a upload script to bintray. It doesn't build anything. It will always be launch by kiwix-build who have already a gradle installed. In this case it means that we are adding 13 new files just to upload files to a rest service. It seems so much compared to what we want to do (in the end, this is just a PUT http request).
There is half a dozen of python modules to upload to bintray (https://pypi.org/search/?q=bintray) that resume themselves to few tens of lines that could easily be directly integrated in kiwix-build.
- The second one, we can consider that the project is a "full" project in charge of building something and publishing it. So the project is in charge to build everything (even the jni, the kiwix libraries, ...). I am pretty sure that gradle can handle the build of native app/lib.

It seems that the archive uploaded to bintray (https://bintray.com/kiwix/kiwix/download_file?file_path=org%2Fkiwix%2Fkiwixlib%2Fkiwixlib%2F1.0.12%2Fkiwixlib-1.0.12.aar) is a bit more that what is just build with meson. But a lot of file/directory are useless/empty/easy to create. There is at least the `classes.jar` that have to be generated but it is probably a jar of all jni classes (in the java directory generated by meson) and could be easily created by meson.

If we are in the first vision (a simple upload script), I really don't want to add a full, somehow complex build system just to upload a file.
If we are in the second vision, let's move completely. We create a real gradle project and the kiwix-lib will be build for android using this project (instead of meson). Android developer will be able to integrate the project in their IDE/build system and do whatever they need to do. And in this case, I agree to follow the android convention and integrate a gradle wrapper. (This also means that you, or someone else, become fully responsible for the android build system. I'm not speaking of the jni wrapper itself)

---

This is not particularly against gradle. I would also refuse a full autoconf/autotools/makefile project just to create an archive and a publish it. A small makefile why not. But not a full build system.

macgills:
Seeing as how the finished product is an Android library file `.aar` I would lean towards the second option. I already know that I will have to make at least 1 further update to this project and if I delete the gradle right now I will just have to re-aad it when the time for that change comes.

I have never set up an NDK project before but it seems I can just point it at a makefile? https://expertise.jetruby.com/android-ndk-using-c-c-native-libraries-to-write-android-apps-21550cdd86a I am very much in unchartered territory with the  C ecosystem, probably last touched it in college a decade ago.
Is your proposal a full NDK project that compiles the `.so`s but the JNI java files get copied over by kiwixlib? ` I'm not speaking of the jni wrapper itself` a bit confused by this line

mgautierfr:
I've just opened a PR kiwix/kiwix-build#343 to build the `kiwix-lib.aar` and publish it on bintray.
We discuss a bit with @macgills on slack. I will try to resume the discussion here.

- The build of the library itself (kiwixlib.so) is pretty complex and already handle by kiwix-build. We should not change this.
- The build of the `.aar` is made with this gradle subproject.
- The publication of the `.aar` is a bit more complex.

The build of "things" on kiwix-build CI is made inside a docker container. All the build tools are installed inside it. The publication is made outside of the docker container. (mainly because docker container doesn't have access to secrets).
The publication have to be made on the CI host but we have "nothing" installed here (no gradle, no kiwix-lib repository, ...). So, if we use the gradle subproject to publish the `.aar` it means that we have to clone kiwix-lib, install gradle (with or without gradlew, that not the point) just to publish an archive.
I've decided that a python script is far simpler. The PR add this script (https://github.com/kiwix/kiwix-build/pull/343/files#diff-ba57ef7b85dae48083a6ba7323a6005d).
It is pretty simple (but I have not tested it for now).
If we agree to go this way, it means that the publication task here is useless.

Please comment on the main idea (here) or specific implementation details (on the PR).

---
The question of gradlew is still open but it is a implementation detail, we will see later.

kelson42:
@macgills @mgautierfr Great to see you working together to find the best solution.

macgills:
@mgautierfr are there any changes I need to make or can you merge this so the project is available to kiwix-build?

mgautierfr:
It would be better to remove the publication of the content as we are now making it in kiwix-build with PR kiwix/kiwix-build#343

And please, if you know how to do this, squash your commit all together. Else we will add all the java file to git history while we are never using them.

macgills:
@mgautierfr squashed them, rebasing already pushed commits is pretty treacherous to a repo's history though so I felt bad doing it. Your repo your rules though

mgautierfr:
> @mgautierfr squashed them,

Thanks

> rebasing already pushed commits is pretty treacherous to a repo's history though so I felt bad doing it

Agree on master branch.
However, on PR I consider this as a good practice to rewrite it. PRs are here to discuss code and create the "best way" to do it. History of the (intermediate) work can be found on the PR issue. The git history should reveal the intermediate steps to achieve a change, not the tries of the author.

kelson42:
@mhutti1 @macgills If everybody agrees here, we should carry on with merging.

mgautierfr:
It seems that integration with travis is broken. The status of the build is not returned. However the build is ok (https://travis-ci.org/kiwix/kiwix-lib/builds/562493330)

I'm merging this PR.

PR #236, opened 2019-06-26
kiwix local server port can be set at the constructor

jetownfeve21:


mgautierfr:
I would prefer to add the port as argument (at creation).
This way, we make the kiwixServe configurable and we will have nothing to change on kiwix-lib side when we will have the port configurable in kiwix-desktop.

jetownfeve21:
at creation it means the constructor or the run() method ?

mgautierfr:
In the constructor.

mgautierfr:
I prefer to not change the return value of `run`. The port is already known (it is passed at creation) and I prefer to keep the return value to some potential future more useful information.

mgautierfr:
And add a default value for the port in the constructor. This way we don't break the API.

PR #240, opened 2019-07-25
#239 ReLinker should be used to load library

### LINKED ISSUE #239: ReLinker should be used to load library
See [kiwix-android issue](https://github.com/kiwix/kiwix-android/issues/1259)

We should hopefully be able to reduce our crash rate by using [Relinker](https://github.com/KeepSafe/ReLinker)

macgills:
Fixes #239 

there will be a required update in kiwix-android as well as JniKiwix now accepts a parameter to its constuctor

### DIFF
--- a/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
+++ b/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
@@ -15,3 +15,7 @@ android {
         }
     }
 }
+
+dependencies {
+    implementation 'com.getkeepsafe.relinker:relinker:1.3.1'
+}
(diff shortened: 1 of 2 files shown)


macgills:
@mgautierfr What is the best way to resolve this? If the java files were in the android "publishing" project (actually I should rename that, it is misleading) then it would be able to find the packages that are causing the build error.
I know next to nothing about this build process so your expertise would be greatly appreciated

mgautierfr:
In PR #241, I've move the `setIcuDataDirectory`, so we don't have to "compile" the java code that use relinker when generating the JNI header files with meson.
So this PR is no more needed.

PR #241, opened 2019-07-29
#239 ReLinker should be used to load library

mgautierfr:
Try to fix #240

The thing is that we need to compile the java file to generate the JNI headers and compile kiwix-lib.so.
However, even if we are using javac, we are not in a "real" android environment and we don't have
the Relinker java module present.

But the constructor of `JNIKiwix` is not native and so, we don't need it when generating the JNI headers. By moving the native method `setIcuDataDirectory` in another class, we don't have to "compile" a plain java code.
We still install all java source file, so gradle will be able to "compile" the `JNIKiwix.java` correctly.

I've not tested with android application, but `.aar`

### DIFF
--- a/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
+++ b/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
@@ -15,3 +15,7 @@ android {
         }
     }
 }
+
+dependencies {
+    implementation 'com.getkeepsafe.relinker:relinker:1.3.1'
+}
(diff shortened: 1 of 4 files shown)


kelson42:
@mgautierfr @macgills At least the CI pass... but does that mean anything? What comes next?

macgills:
`.aar` will be fine I imagine, approving

mgautierfr:
Let's merge it, we will see if generated `.aar` is good (I'm pretty confident)

PR #251, opened 2019-08-12
Jni fix

mgautierfr:


src/android/kiwixreader.cpp
```diff
@@ -234,21 +234,24 @@ JNIEXPORT jbyteArray JNICALL Java_org_kiwix_kiwixlib_JNIKiwixReader_getContent(
...
     setIntObjValue(cSize, sizeObj, env);
-
-    data = env->NewByteArray(cSize);
-    env->SetByteArrayRegion(
-        data, 0, cSize, reinterpret_cast<const jbyte*>(entry.getBlob().data()));
-
     setStringObjValue(entry.getMimetype(), mimeTypeObj, env);
     setStringObjValue(entry.getTitle(), titleObj, env);
+    if (isRedirect) {
+      setStringObjValue(entry.getPath(), url, env);
```
mhutti1: Shouldn't this be along the lines of `getFinalPath()`? It doesn't seem to change the URL for me at the moment

mhutti1: Ok it seem that entry.getPath() is correct. Still not getting the right output in android though.

mhutti1: My bad it seems intellij wasn't updating the variable.

PR #278, opened 2019-09-17
Fix bookmarks

jetownfeve21:
better handling of bookmarks when deleting a zim

fix https://github.com/kiwix/kiwix-desktop/issues/252

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -73,10 +73,16 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
   return false;
 }
 
-
+void Library::removeAllBookmarksByZimID(const std::string& zimId)
+{
+  auto removeItem = std::remove_if(m_bookmarks.begin(),m_bookmarks.end(),
+    [zimId](kiwix::Bookmark &bookmark) { return (bookmark.getBookId() == zimId) ? true : false; });
+  m_bookmarks.erase(removeItem, m_bookmarks.end());
+}
 
 bool Library::removeBookById(const std::string& id)
 {
+  removeAllBookmarksByZimID(id);
   return m_books.erase(id) == 1;
 }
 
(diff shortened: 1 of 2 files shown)


src/library.cpp
```diff
@@ -59,7 +59,12 @@ bool Library::addBook(const Book& book)
 
 void Library::addBookmark(const Bookmark& bookmark)
 {
-  m_bookmarks.push_back(bookmark);
+  auto booksId = getBooksIds();
+  for (auto &bookId:booksId) {
+    if (bookId == bookmark.getBookId()) {
+      m_bookmarks.push_back(bookmark);
+    }
+  }
```
mgautierfr: A `if (booksId.find(bookmark.getBookId()) != bookIsd.end())` is better.

src/library.cpp
```diff
@@ -73,10 +78,21 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+{
+  auto it=m_bookmarks.begin();
+  while (it!=m_bookmarks.end()) {
+    if (it->getBookId() == zimId) {
+      m_bookmarks.erase(it);
+    } else {
+      it++;
+    }
+  } 
+}
```
mgautierfr: I somehow (always) dislike this way of doing to remove a this from a vector.

I've just discover this idiom : https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

We should use it :)

kelson42:
@jetownfeve21 I thought a bit about this problem since Tuesday. I don't think this is a good idea to delete bookmarks if the file is deleted because:
* If someone upgrade a file, it will delete the old one and the bookmark should be kept
* To do things properly, we should inform the user that he is not only about to delete a file, but also to delete its bookmarks... which complexify again the process
* I see a use case where for some reason a ZIM file might be deleted but the bookmarks kept

As a consequence, I don't think this problem should be handle that way. Bookmarks for a ZIM file which is not available should be kept. They can be disabled/greyed or simply not displayed. The other one should work as normal.

PR #280, opened 2019-09-19
[API Break] Fix pathTools (and a bit stringTools).

mgautierfr:
Api changes :
 - removeLastPathElement do not takes extra arguments
   `removePreSeparator` and `removePostSeparator`.
   This is not needed as path do not need special tailing separator.
 - Only one function `split`. Arguments can be implicitly convert to
   string. No need for overloading functions to explicitly cast them.
 - `split` function takes another argument `trimEmpty`. If true, empty
   element are removed.

Path manipulation now almost pass trough a vector<string> to store each
path's part.

Most of the complex works is now made in the normalizeParts function.


mgautierfr:
Probably fix https://github.com/kiwix/kiwix-desktop/issues/259

PR #290, opened 2019-10-04
add setPort() method

jetownfeve21:


kelson42:
@jetownfeve21 Any news here?

jetownfeve21:
@kelson42 I forgot to handle invalid port number now it's fine.
I don't understand which variable should be const.

PR #299, opened 2019-12-06
Opensearch

kelson42:
Provides Open Search endpoints to search into content and retrieve results thanks to an Atom feed.

Fixes https://github.com/kiwix/kiwix-tools/issues/52

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+kiwix-lib 8.3.0
+===============
+
+ * Introduce OpenSearch backend
+
 kiwix-lib 8.2.1
 ===============
 
--- a/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
+++ b/android-kiwix-lib-publisher/kiwixLibAndroid/build.gradle
@@ -26,7 +26,7 @@ task writePom {
     project {
         groupId 'org.kiwix.kiwixlib'
         artifactId 'kiwixlib'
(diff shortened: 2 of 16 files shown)


kelson42:
Few remarks I have:
* It is not automatically tested at all
* We need a minimal documentation somewhere (probably in the wiki https://wiki.kiwix.org)
* The OPDSdumper class seems here to be abused (content opensearch has nothing to do with the OPDS)

mgautierfr:
What is the context of this PR ?
It seems that :
- It add a atom feed. (Why we need it)
- Is make this able to search for zim file based on the (full text) content of the zim files.

Ie, searching for "apple" will give us all zim files were "apple" appears in the content.
I'm not sure we need that. We may want to search on the name/description/... of the zim file but not on the real content.
If we want to search on the content, we directly want the articles, not the list of zim file.

kelson42:
@mgautierfr I have added the corresponding ticket https://github.com/kiwix/kiwix-tools/issues/52 to put a bit of context. This is needed to integrate `kiwix-serve` capability in a broader search feature (in Kiwix Hotspot for example). Look to the opensearch API specs to get more details. You last remark seems perfectly pertinent to me, the results provided should point to the real articles, not only the the ZIM (homepage?).

mgautierfr:
I agree with the need to have a API to do fulltext search on the content. But OPDS is not about the content, it is a listing of the content (aka, the catalog).
(And there is already a opensearch functionnality on the catalog, even if not perfect)

PR #302, opened 2020-01-07
update book even if the members aren't empty

jetownfeve21:
remove the conditions to always update the book

fix https://github.com/kiwix/kiwix-desktop/issues/300

### DIFF
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -45,28 +45,14 @@ bool Book::update(const kiwix::Book& other)
     return false;
 
   m_readOnly = other.m_readOnly;
+  m_path = other.m_path;
+  m_pathValid = other.m_pathValid;
+  m_url = other.m_url;
+  m_tags = other.m_tags;
+  m_name = other.m_name;
+  m_favicon = other.m_favicon;
+  m_faviconMimeType = other.m_faviconMimeType;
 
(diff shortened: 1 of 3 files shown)


kelson42:
@jetownfeve21 I don't see any testing to avoid erasing by updating if a path is already there AND zim loaded properly. Normal?

kelson42:
@jetownfeve21 I guess this should be reviewed by @mgautierfr!?

mgautierfr:
From what I see in kiwix-desktop#300, we try to update the book (referenced by the bookId) but, as the path already exists it is not updated.

As far as I know, the update function was used only when moving from a library version to another. And when with the `add` action of kiwix-manage.

@kelson42 Your question is legit. And more globally it is what happens if we have two zim files (same id, two different paths) ?
I don't think there is a unique solution, it depends of the context and it should be handled on the caller side.

It seems to me that, on the library side, if a caller want to update a book in the library, we update it with the values given. I don't see a real reason to refuse a value update.
But I'm open to counter-arguments.

kelson42:
@mgautierfr Agree. Different scenarios should be handled at a higher level. PR looks good to me.

kelson42:
@mgautierfr Please review

PR #310, opened 2020-01-22
Add parameter options to start a download with aria2

jetownfeve21:
Downloader::startDownload has a new parameter option which is a vector of pair that represents the options that can be set for adding a uri with aria2 with the function Aria2::addUri.

Aria2::addUri uses this parameter to set the struct of parameters for the aria2 command

### DIFF
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -92,7 +92,7 @@ class Downloader
 
   void close();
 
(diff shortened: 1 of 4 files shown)


mgautierfr:
I'm a bit surprised it works. Default arguments should be defined on function declaration, not definition.
I'm even more surprised it compiles.

PR #326, opened 2020-02-19
[JAVA] Use a long to store the offset of a article in the zim file.

mgautierfr:
Fixes kiwix/kiwix-android#1769


mgautierfr:
Testing this with @macgills

macgills:
It has worked well for me, though the issue this originally came from is outstanding that is due to another unrelated problem

mgautierfr:
Good, this is a fix of a real bug anyway. I'm merging it.

PR #330, opened 2020-03-26
Add external links blocking in serve

rgaudin:
In many use cases, it is not wanted to have user accidentaly click on external links
and leave the served ZIM content.
This could be because the result is unpredictible (reader not implementing this properly)
or because the serve user knows there's no backup internet connexion or because there is
an induced cost behind external links that doesn't affect served content.

using a new flag (`blockExternalLinks`) on `Response`/`setTaskBar`, a piece of JS code
is injected into the taskbar code.
This code adds a JS handler on all link click events and verifies the destination.
If the destination appears to be an external link (1), the link target is changed to
a specific URL:

```
/external?source=<original_uri>
```

(1) external is a link that's not on the same origin and starts with either `http:` `https:` or `//`.

Server implements a new handler on `/external` that displays a new page (`captured_external.html`)
which returns a generic message explaining the situation and offering to click on the link
again should the user really want to.
This is done by specifically asking `set_taskbar` to not block external requests on that page.

This approach allows integrators using a reverse proxy to handle that endpoint differently (rebrand it)

1. `Server` now has an `m_blockExternalLinks` defaulting to `false`
1. `Server.setTaskbar` is extended to support an additional bool to set the variable.
1. `Response` now has an `m_blockExternalLinks`
1. `Response` constr expects an additional bool for `blockExternalLinks`.
1. `Response.set_taskbar` is extended to support an additional bool to set the variable.
1. JNI/Java Wrapper reflects the extensions.
1. New resource file `templates/block_external.js` (included in head_part). Should it be in skin?
1. New resource file `templates/captured_external.html` for `handle_captured_external()`
1. Added a comment on `head_part.html` to help with JS insertion at the right place
1. `introduce_taskbar()` conditionnaly inserts the JS inside the taskbar


rgaudin:
@mgautierfr, let us know what you think of this and I'll look into updating the tests.

include/server.h
```diff
@@ -56,7 +56,9 @@ namespace kiwix
        void setNbThreads(int threads) { m_nbThreads = threads; }
        void setVerbose(bool verbose) { m_verbose = verbose; }
        void setTaskbar(bool withTaskbar, bool withLibraryButton)
-         { m_withTaskbar = withTaskbar; m_withLibraryButton = withLibraryButton; }
+        { setTaskbar(withTaskbar, withLibraryButton, m_blockExternalLinks); }
```
kelson42: Small indentation problem.

src/server.cpp
```diff
@@ -711,6 +720,26 @@ Response InternalServer::handle_random(const RequestContext& request)
...
+  try {
+    source = kiwix::urlDecode(request.get_argument("source"));
+  } catch (const std::out_of_range& e) {}
+
+  if (source.empty())
+    return build_404(request, "");
+
+  auto data = get_default_data();
+  data.set("source", source);
+  auto response = get_default_response();
```
mgautierfr: Instead of `get_default_response`, we could simply directly create the `Response` with `blockExternalLinks`==false.

This way, we would not have to change the signature of `set_taskbar`

include/server.h
```diff
@@ -56,7 +56,9 @@ namespace kiwix
...
+       void setTaskbar(bool withTaskbar, bool withLibraryButton, bool blockExternalLinks)
```
mgautierfr: I dislike the idea of changing the signature of `setTaskbar`, especially to add an argument not related to taskbar.

It would be simpler to add a new method `setBlockExternalLink`.

src/server/response.cpp
```diff
@@ -124,6 +125,14 @@ void Response::introduce_taskbar()
...
     "<body[^>]*>",
     taskbar_part);
+
+  if ( m_blockExternalLinks ) {
+    const std::string capture_external_part = getResource("templates/block_external.js");
+    m_content = appendToFirstOccurence(
+      m_content,
+      "block external links\n",
+      capture_external_part);
+  }
```
mgautierfr: I would move this code in another function than introduce taskbar.

- This is not related to taskbar.
- Now, how it is, if taskbar is deactivated, you don't include `block_external.js`. So user cannot block external links without a taskbar.

static/templates/block_external.js
```diff
@@ -0,0 +1,17 @@
...
+        if (href.substr(0, 2) == "//")
+          return capture(e);
+        if (href.substr(0, 5) == "http:")
+          return capture(e);
+        if (href.substr(0, 6) == "https:")
+          return capture(e);
+        return;
+      }
+    }});
+  });
```
mgautierfr: I wonder if the js could check if the current url start with "external", if this case, it would not capture links.

So, we would have to set `blockExternalLink` at response level. It would be only a server configuration (introduce this js or not).

It would be pretty useful if we do this later (https://github.com/kiwix/kiwix-lib/issues/322#issuecomment-584225014). As the inserted js would be always present, it would not be possible to set a boolean to avoid the block of the links. And the boolean in the response would be useless.

static/templates/head_part.html
```diff
@@ -5,6 +5,7 @@
 <script type="text/javascript" src="{{root}}/skin/jquery-ui/jquery-ui.min.js"></script>
 <script>
   var jk = jQuery.noConflict();
+  // block external links
```
mgautierfr: This file is included only if we have a taskbar. As I already said you should not include the catpure function inside the taskbar.

This line will be probably removed in the next version of this PR anyway, but I just want to mention that I would prefer a comment like `//<BLOCK_EXTERNAL> <-- block external function will be inserted here`.
If you read this file without the context of this PR, it seems it is just a comment out of context and someone will probably remove it next time he changes this file.

static/templates/captured_external.html
```diff
@@ -0,0 +1,23 @@
...
+  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.min.css" rel="Stylesheet" />
+  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.theme.min.css" rel="Stylesheet" />
+  <script type="text/javascript" src="{{root}}/skin/taskbar.js" async></script>
+</head>
+<body class="kiwix">
+
+<h1>External link blocked</h1>
+<p>This instance of Kiwix protects you from accidentaly going to external (out-of ZIM) links.</p>
+<p>If you intend to go to such locations, please click the link bellow.</p>
+<p><a href="{{ source }}">Go to {{ source }}</a></p>
```
mgautierfr: I wonder how it is working with url with space or urlencoded.

rgaudin:
Thanks for the feedback ; Agrees I tied it to the taskbar while it's not actually mandatory and that changes a lot.

Will come back with a revised version

mgautierfr:
Another comment, to be future proof we may use another url than `/external`. We may need to have the same mechanism for other thing that catching external url. It is a long future but I think about zim with extensions (and so url to non existent page) or link between zim files.

Maybe changing the url to something like `/catch/external?source=<original_uri>` or `/catch?raison=external&source=<original_uri>` is be bit better (and it is not a complex thing, we can handle it even if it is finally for nothing)

include/server.h
rgaudin: done

static/templates/captured_external.html
rgaudin: Should not be a problem as we are receiving it urlencoded always (it's in the request). Tried with many unexpected inputs and it seems to be working as expected.

rgaudin:
OK, here are the main changes from the last iteration

* As suggested, introduced a `setBlockExternalLinks()` on the server and moved away from the taskbar
* Added JNI binding for those
* ~~As suggested, using `Response` in handler to disable URL blocking~~ (see below)
* Rewrote the capture code to use vanilla JS as jquery is part of the taskbar
* Consequence is that it's larger but it's stored in a separate (cached then?) file so less verboseness in <head> of each page
* changed URL to `/catch/external?source=<source>`
* Eventually changed JS code to detect external page and not apply blocking there so it's not set in handler code

src/server.cpp
rgaudin: Did that then came back to default without taskbar call as not required anymore (page detected in JS)

src/server.cpp
```diff
@@ -383,7 +392,7 @@ Response InternalServer::build_500(const std::string& msg)
 {
   kainjow::mustache::data data;
   data.set("error", msg);
-  Response response(m_root, true, false, false);
+  Response response(m_root, true, false, false, m_blockExternalLinks);
```
mgautierfr: I would not block external here. 500 means internal server error.
If we have this, we probably don't want to insert so js that redirect to us.

(I'm not sure we have links in `_500_html`, but be future proof, we may want to add a link to a bugtracker or something else)

static/templates/captured_external.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <meta charset="UTF-8" />
+  <title>Welcome to Kiwix Server</title>
+  <script type="text/javascript" src="{{root}}/skin/jquery-ui/external/jquery/jquery.js"></script>
+  <script type="text/javascript" src="{{root}}/skin/jquery-ui/jquery-ui.min.js"></script>
+  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.min.css" rel="Stylesheet" />
+  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.theme.min.css" rel="Stylesheet" />
+  <script type="text/javascript" src="{{root}}/skin/taskbar.js" async></script>
```
mgautierfr: Do we need all thoses jquery includes ?

rgaudin: No, you're right. I tried to include the taskbar at first but that revealed quite complex and useless. Will clean that up.

src/server.cpp
rgaudin: Right.

Issue #264, opened 2019-08-27
Incompatibility with Mustache 4


kelson42:
It seems the solution we have implemented to manage `urlencoded` does not work with Moustache4 (it takes the whole line up to the last carriage return) instead of only the content between
`{{#urlencoded}}` and `{{/urlencoded}}`. It looks like a bug in Mustache as this seems work fine with Mustache 3.2

hashworks:
Related: https://github.com/kiwix/kiwix-tools/issues/321

mgautierfr:
kainjow/Mustache 4 is still in beta. It may be a bug in mustache.

kelson42:
@mgautierfr Could you please make a short bug report then? I don't have seen anything open,

kelson42:
@mgautierfr Mustache 4 has been officially released https://github.com/kainjow/Mustache/commit/b1136d822e1d04dd111fb7bda4ca3adeb788ad8d. I tested it. The bug is still there unfortunately.

mgautierfr:
Issue has already been opened there : https://github.com/kainjow/Mustache/issues/38

I'm afraid there is nothing we can do here (except fixing mustache itself)

kelson42:
@mgautierfr No workarounds?

mgautierfr:
I haven't found one. A workaround would be to simply move the `{{#urlencoded}}stuff{{/urlencoded}}` in its own line. But, in our case, as we use it inside a string, it would insert `\n` in the string and break the url.

A easiest solution is to stay on mustache 3

kelson42:
Bug has been fixed now upstream. I don't see big value to keep this ticket open.

PR #331, opened 2020-03-27
add required info to README

rgaudin:
Mustache 4 was release in late october 2019 and is not compatible with kiwix-lib as is.
Replacement in taskbar is messed-up and breaks suggestions.
Until this is fixed/adapted, inform users to use version 3.

@mgautierfr not sure you're aware of this bug…


mgautierfr:
This is reported in #264
As there is no solution for now, let's merge this.

PR #338, opened 2020-04-07
Updated: meson.build to generate pkg-config file

kelvinhammond:


mgautierfr:
Thanks for this PR.

May I ask why this change ?
I agree that using the pkgconfig module may be better than using a configuration file.
But they are not generating the same output. Maybe the new output is better but it would be nice to explain it.

For exemple, the "old" kiwix.pc on native_dyn is :
``

kelvinhammond:
@mgautierfr The main issue I had with the original included file was that the lib64 directory was hard coded and broke when I tried to build using that generated pkg-config file. The one generated by this PR should work with different system configurations. 

I don't know if `Requires.private` is required or not, pkg-config generates it based on the library build settings, we can add it to `Requires`.

`Requires` is what will be output by pkg-config when you run `pkg-config --libs kiwix` so if a library is required for linking when building something using kiwix-lib it should be in `Requires`. Chances are `libmicrohttpd` and `-pthread` should be in `Requires`.

kelvinhammond:
@mgautierfr any updates?

mgautierfr:
I have to check that it doesn't break all our builds as it change pkg-config file output.

About `Requires.private` and `Requires`. If kiwix-lib use a library but doesn't not expose it through its public api, it should go in `Requires.private` (so libmicrohttpd should go in `Requires.private`). If the dependency is expose it should go in `Requires` (this is the case for `pugixml` and `icu4c`)

kelson42:
@mgautierfr What is the status here wiht the PR of @kelvinhammond ?

kelvinhammond:
@kelson42 Oops, missed this. You'll need to check with @mgautierfr I thought it was ready to go.

PR #341, opened 2020-04-13
Fix open external zim

jetownfeve21:
Check if the parameter `pathToSave` is empty before use it otherwise the
book path is empty too, which causes crash on opening external zim files

fix https://github.com/kiwix/kiwix-desktop/issues/403


kelson42:
@jetownfeve21 There is no way to unit test this properly?

PR #346, opened 2020-04-20
addBookFromPathAndGetId unit test

jetownfeve21:


test/manager.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+TEST(ManagerTest, addBookFromPathAndGetIdTest)
+{
+    kiwix::Library lib;
+    kiwix::Manager manager = kiwix::Manager(&lib);
+
+    auto bookId = manager.addBookFromPathAndGetId("./test/example.zim");
+    EXPECT_NE(bookId, "");
+    kiwix::Book book = lib.getBookById(bookId);
+    EXPECT_NE(book.getPath(), "");
```
mgautierfr, line 16: Wouldn't be better to check that `book.getPath()` equals "./test/example.zim" ?

test/manager.cpp
```diff
@@ -0,0 +1,27 @@
...
+    
+    const std::string pathToSave = "./pathToSave";
+    const std::string url = "url";
+    bookId = manager.addBookFromPathAndGetId("./test/example.zim", pathToSave, url, true);
+    book = lib.getBookById(bookId);
+    auto savedPath = isRelativePath(pathToSave) ?
```
mgautierfr, line 22: You already know that `pathToSave` is relative.
And by the way, you should not use the same code that the code you want to test to compute the right result.

legoktm:
@jetownfeve21 @mgautierfr: What is the source/copyright on the `test/data/example.zim` file that was added?

kelson42:
@jetownfeve21 Except for the logo of wikibooks, this ZIM files contains two empty pages.

jetownfeve21:
@legoktm @kelson42  this zim file only exists for the purpose of the unit tests

PR #353, opened 2020-05-17
Add Ubuntu 20.04 Focal to CI

kelson42:


.github/workflows/ci.yml
```diff
@@ -104,7 +112,8 @@ jobs:
     - name: Install deps
       shell: bash
       run: |
-        ARCHIVE_NAME=deps2_${OS_NAME}_${{matrix.target}}_kiwix-lib.tar.xz
+        TARGET_RADICAL=$(echo ${{matrix.target}} | sed -E 's/_([^_]*)$//')
```
veloman-yunkan, line 115: I guess the intent of this change is to convert `native_static_focal` and `native_static_xenial` to `native_static`, and  `native_dyn_focal` and `native_dyn_xenial` to `native_dyn`. However it will also convert `win32_static` and `win32_dyn` to `win32`,  and `android_arm` and `android_arm64` to `android`, which, I believe, was not intended.

kelson42: Indeed. This is one of a few other problems. Not so straight to setup CI on focal as well unfortunately.

PR #360, opened 2020-05-22
Fixed the support for byte range requests

veloman-yunkan:
Will close kiwix/kiwix-tools#148

This PR enables handling of partial content requests with a single byte-range. Requests for two or more byte ranges (even if they effectively constitute a single continuous range) are rejected with a 416 (Range Not Satisfiable) error response. Such behaviour complies with somewhat liberal interpretation of the [spec](https://tools.ietf.org/html/rfc7233#section-4.4)):

> The 416 (Range Not Satisfiable) status code indicates that none of the ranges in the request's Range header field (Section 3.1) overlap the current extent of the selected resource or that the set of ranges requested has been rejected due to invalid ranges or **an excessive request** of small or overlapping ranges.

Support for [`If-Range`](https://tools.ietf.org/html/rfc7233#section-3.2) conditional requests is not implemented.

Partial requests work only for content (entries) from the served ZIM file(s). If both `Range:` and `Accept-Encoding: deflate` headers are present in the request, the former has precedence and uncompressed partial content is sent in a 206 response (even if a 200 response to a respective non-range request would have its content compressed).

The changes have been validated with new test cases in the server unit-test (as well as some manual experiments with kiwix-serve and curl).

This PR should be easy to review in its final state though some issues with the state of the current code in master are more obvious when looking at the first several commits in the PR history. I didn't spend any time cleaning up the development history, since I believe that this PR can be merged as a single squashed commit.

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -41,10 +41,10 @@ bool is_compressible_mime_type(const std::string& mimeType)
 
 int get_range_len(const kiwix::Entry& entry, RequestContext::ByteRange range)
 {
-  const int entrySize = entry.getSize();
-  return range.second == -1
-       ? entrySize - range.first
-       : std::min(range.second + 1, entrySize) - range.first;
+  const int64_t entrySize = entry.getSize();
+  return range.last() == -1
+       ? entrySize - range.first()
+       : std::min(range.last() + 1, entrySize) - range.first();
 }
 
 } // unnamed namespace
@@ -330,7 +330,7 @@ void Response::set_entry(const Entry& entry, const RequestContext& request) {
     set_compress(true);
   } else {
     const int range_len = get_range_len(entry, request.get_range());
-    set_range_first(request.get_range().first);
+    set_range_first(request.get_range().first());
     set_range_len(range_len);
   }
 }
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -173,7 +173,7 @@ MHD_Response*
 Response::create_error_response(const RequestContext& request) const
 {
   MHD_Response* response = MHD_create_response_from_buffer(0, NULL, MHD_RESPMEM_PERSISTENT);
-  if ( m_returnCode == MHD_HTTP_RANGE_NOT_SATISFIABLE ) {
+  if ( m_returnCode == 416 ) {
     std::ostringstream oss;
     oss << "bytes */" << m_byteRange.length();
 
@@ -343,7 +343,7 @@ void Response::set_entry(const Entry& entry, const RequestContext& request) {
     set_content(content);
     set_compress(true);
   } else if ( m_byteRange.kind() == ByteRange::INVALID ) {
-    set_code(MHD_HTTP_RANGE_NOT_SATISFIABLE);
+    set_code(416);
     set_content("");
     m_mode = ResponseMode::ERROR;
   }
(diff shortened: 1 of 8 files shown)


src/server/response.cpp
```diff
@@ -322,14 +314,16 @@ void Response::set_entry(const Entry& entry, const RequestContext& request) {
   set_mimeType(mimeType);
   set_cacheable();
 
+  // XXX: Which of the Accept-Encoding and Range headers has precedence if both
+  // XXX: are present? Can partial content be compressed?
```
mgautierfr: Searching (a bit) the internet, I've found nothing about this.
I would say that yes, partial content can be compressed.

I think we should resolve the byteRange first and then compress the content if needed.
However, our use case here is that we need byteRange for media (video) and they are not compressible.
I don't know if we should handle both case in the same time (but we must be sure to answer with the correct headers anyway)

mgautierfr: Hum... I'm not sure what I've reviewed. I've just comment this lines and github say me this is outdated...

veloman-yunkan: > Searching (a bit) the internet, I've found nothing about this.

Same for me. I guess I better have mentioned that in the PR description.

> I think we should resolve the byteRange first and then compress the content if needed.

The interaction of the `Accept-Encoding` and `Range` is a little tricky (and it's strange that there is no formal specification of what the correct interpretation should be). Suppose that the client issues a regular (non-range) request for a large compressible resource (e.g. a huge JSON or XML document). The resource is compressed and streamed to the client but the connection is interrupted at 95%. The client issues a range request to fetch the remaining 5% of the *compressed representation of the resource*, with the purpose of recombining it with whatever was sent in the first session and being able to decompress it. In such an scenario (which is different from your point of view of how `Range` should play with `Accept-Encoding`) the load on the server can be excessive - entire resource must be compressed with the purpose of sending only a small part of it.

Note that for partial/interrupted download of a compressed resource the client has no chance to decompress the received part, find out the byte-range for the uncompressed version of the resource and issue a range request for the remaining part.

That's why the safest bet is to *not* combine compression with partial responses.

veloman-yunkan:
@mgautierfr I re-requested your review so that you can provide your feedback or approve the latest version.

src/server/response.cpp
mgautierfr: > Note that for partial/interrupted download of a compressed resource the client has no chance to decompress the received part, find out the byte-range for the uncompressed version of the resource and issue a range request for the remaining part

Why ? That's what I was thinking that what it has to be done.

For me the range is about the uncompressed content. If a connection is lost at 95%, client must uncompress the data it received and ask for what is left to have (in uncompressed range). This range will be compressed again in a new stream. But it cannot ask for the last 5% of the compressed stream.

mgautierfr:
I've reviewed the full PR. The code is good. We still have this discussion about `Range` and `Accept-Encoding` but it is not a blocker.

src/server/response.cpp
veloman-yunkan: My assumption was that in general it is impossible to decompress a partial archive. That assumption is probably wrong for the compression algorithms/formats used in HTTP. Still, I doubt that if we implement that in kiwix-serve the handling on the client side will be like we expect. Therefore I think that we should not spend any time on this until we know of authoritative formal specification of how it must work.

PR #364, opened 2020-06-02
Fix for the failing assertion in the ByteRange constructor

### LINKED ISSUE #363: Error in byte-range mgmt
I see this time to time in the log:
```
======================
Requesting : 
full_url  : /meta
method    : GET (0)
version   : HTTP/1.1
request#  : 247
headers   :
 - Accept : 'image/webp,image/apng,image/*,*/*;q=0.8'
 - Accept-Encoding : 'gzip, deflate'
 - Connection : 'close'
 - Host : 'library.kiwix.org'
 - User-Agent : 'Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.6 Chrome/69.0.3497.128 Safari/537.36'
 - X-Forwarded-For : '111.187.34.191'
 - X-Forwarded-Port : '80'
 - X-Forwarded-Proto : 'http'
 - X-Forwarded-Ssl : 'off'
 - X-Real-IP : '111.187.34.191'
arguments :
 - content : 'wikipedia_zh_medicine_maxi_2020-05'
 - name : 'favicon'
Parsed : 
full_url: /meta
url   : /meta
acceptEncodingDeflate : 1
has_range : 0
is_valid_url : 1
.............
Response :
httpResponseCode : 200
headers :
 - ETag: '"1591098928403549187/c"'
kiwix-serve: ../../SOURCE/kiwix-lib_release/src/server/byte_range.cpp:68: kiwix::ByteRange::ByteRange(kiwix::ByteRange::Kind, int64_t, int64_t): Assertion `last >= first' failed.
--
acceptEncodingDeflate : 1
has_range : 0
is_valid_url : 1
.............
** running handle_content
Response :
httpResponseCode : 404
headers :
 - Cache-Control: 'no-cache, no-store, must-revalidate'
 - Access-Control-Allow-Origin: '*'
 - Content-Type: 'text/html'
 - Content-Encoding: 'deflate'
 - Vary: 'Accept-Encoding'
Request time : 0.000577s
```

veloman-yunkan:
Will close #363 .

The assertion in the `ByteRange` constructor was written under the assumption that the content must have non-zero size. Now it allows that corner case. The rest of the `ByteRange` code should handle that case normally (the only arguable scenario being that a 416 (Not Satisfiable) response is sent to any range request of such a 0-sized resource).

### DIFF
--- a/src/server/byte_range.cpp
+++ b/src/server/byte_range.cpp
@@ -65,7 +65,7 @@ ByteRange::ByteRange(Kind kind, int64_t first, int64_t last)
 {
   assert(kind != NONE);
   assert(first >= 0);
-  assert(last >= first);
+  assert(last >= first || (first == 0 && last == -1));
 }
 
 ByteRange::ByteRange(int64_t suffix_length)


kelson42:
@veloman-yunkan No unit test?

veloman-yunkan:
@kelson42 I need to update the ZIM file used in the server unit-test by adding an empty entry to it. Yesterday I tried to do that via `zimdump` and `zimwriterfs`, but the resulting ZIM file changed a lot. Can you suggest a way to achieve what I need?

kelson42:
@veloman-yunkan You shoul recreate one, there is not solution to modify one.

veloman-yunkan:
> Your approach with `zimdump` and `zimwriterfs` is the proper one.

@kelson42 It's not a proper one, because the structure of the new ZIM file is very different from the structure of the original ZIM file. Some issues with that flow:

1. zimdump loses redirection info. Instead it creates empty articles.
2. zimdump saves the articles using their title rather than the name (the trailing part of the URL).
2. zimwriterfs was designed to take as input an original HTML structure and reorganizes it into a set of namespaces. When zimwriterfs receives the output of zimdump it messes up the structure by adding another level of namespaces.

veloman-yunkan:
I propose to merge this PR before the unit test is ready. I verified that it fixes the issue with empty articles on a ZIM file created using the `zimdump`+`zimwriterfs` flow (which, however, is not suitable for the unit-test).

mgautierfr:
> Can you suggest a way to achieve what I need?

I would go with a patched version on `zimrecreate`. As you said, `zimdump` and `zimwriterfs` are not the inverse of the other.

veloman-yunkan:
I chose to add the unit test for this case via a separate ZIM file. See #365.

PR #365, opened 2020-06-03
Enhanced the server unit-test with corner cases

veloman-yunkan:
The server unit-test now exercises requests for empty (0-sized) resources (related issue is #363).

The corner case data is located in a new ZIM file at `test/data/corner_cases.zim`. That ZIM file was created from source data at ` test/data/corner_cases` using the script `test/data/create_corner_cases_zim_file`. When a new corner case is added the script must be re-ran and the updated ZIM file must be committed along with the updated source data. When running the `create_corner_cases_zim_file` script `zimwriterfs` must be in the PATH.


mgautierfr:
Shouldn't we test ByteRange with `corner_cases.zim` ?

veloman-yunkan:
> Shouldn't we test ByteRange with `corner_cases.zim` ?

Now we do

PR #368, opened 2020-06-12
Add non-minified version of jquery-ui.js

legoktm:
Debian wants to have the source files for minified scripts. Also the
jqueryui.com downloader is broken, so if we ever needed to modify/fix the
library, it would be good to have a non-minified version on hand.


mgautierfr:
I have no strong opinion if we should use mimified on non-mimified scripts. (I tend to think we should not use jquery at all to have a drop-down list for completion).

But it would be good to have the source of the file in the commit message (especially if the same message says that jquery downloader is broken).
And, as I will not review jquery, it is even more necessary to have to source somewhere.

kelson42:
Actually Jquery should not be part of our code base https://github.com/kiwix/kiwix-lib/issues/369

legoktm:
> But it would be good to have the source of the file in the commit message (especially if the same message says that jquery downloader is broken).

I spent 2 hours trying to check out an ancient version of jquery-ui, failed at hacking the build scripts and eventually looked around the internet and found https://github.com/components/jqueryui/blob/1.11.0/jquery-ui.js - it looks to be the same given the header and my skim of a deminified version of the current minified one.

PR #374, opened 2020-07-01
Add new thread safe suggestion API.

mgautierfr:
Previous API were using an internal vector to store the suggestions search
results.

The new API takes a vector as out argument. So user can call the functions
without having to protect the search.

We should change the android API to reflect the change but it is a bit
more complex to do at JNI level. As android do not call it multithreaded
we are safe for now. And we need the new API asap for kiwix-desktop.

So we keep the same API on android for now, the new api will be made
in next version.

### DIFF
--- a/include/reader.h
+++ b/include/reader.h
@@ -43,6 +43,8 @@ namespace kiwix
  * The Reader class is the class who allow to get an entry content from a zim
  * file.
  */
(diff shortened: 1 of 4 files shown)


kelson42:
@mgautierfr Should we merge?

PR #378, opened 2020-07-09
Increase test timeout to 160s

### LINKED ISSUE #375: test timeout on slow architectures
Hello, looks like 30 seconds for server testing is really not enough to make the testsuite pass on emulated hardware such as qemu-riscv64
https://launchpadlibrarian.net/487000588/buildlog_ubuntu-groovy-riscv64.libkiwix_9.2.2+dfsg-3_BUILDING.txt.gz

Can you please increase a little bit the timeouts?

kelson42:
Fixes #375

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -42,6 +42,6 @@ if gtest_dep.found() and not meson.is_cross_build()
                               link_args: extra_link_args,
                               dependencies : all_deps + [gtest_dep],
                              build_rpath : '$ORIGIN')
-        test(test_name, test_exe)
+        test(test_name, test_exe, timeout : 160)
     endforeach
 endif


LocutusOfBorg:
thanks!

mgautierfr:
It would be better to change the timeout only on the `server` test. We don't need to increase the timeout for others. This add a timeout of 2m40s for each test. If somehow there is a infinite loop in kiwixlib, as there are 10 tests the job will run 24m before exiting with error.

kelson42:
@mgautierfr To me this is clearly a priority that tests work out of the box for everybody if possible. This has clearly priority on the edge case of the infinite loop which is going to happen to a C++ developer anyway (so he know what his is doing) who knows when.

mgautierfr:
If you want to have test work everywhere, we should simply remove the timeout.
Our tests are working on our CI without problem. A user come with a specific need and there is a way to solve its problem at runtime with a simple command line option and you prefer to impact all our build system to just avoid a user to pass an option.
There will be always a new architecture running slow, or without enough memory or .. On the opposite side we want to add more tests. This is infinite.
A user compiling on a slow architecture *who knows when* can simply add a `-t 10` option to `meson test`.

kelson42:
> If you want to have test work everywhere, we should simply remove the timeout.

This is how it is currently and it seems to fail! See the bug report.

legoktm:
I did some more testing yesterday and it's rather straightforward for us to bump the timeout using e.g `meson test -t 10` in the Debian/Ubuntu packaging, so I think we're fine either way whether we bump the timeout in kiwix-lib or leave it to packagers to do.

PR #379, opened 2020-07-10
Automatically build Debian packages and publish to PPA

legoktm:
We can currently only build for Ubuntu versions because we're not yet
publishing libzim for Debian.

Development builds (on commits to master) will build against master libzim
while release builds (on tag pushes) will build against the most recent
release of libzim.


debian/changelog
```diff
@@ -0,0 +1,5 @@
+libkiwix (0.0.0) unstable; urgency=medium
```
mgautierfr, line 1: Souldn't be the version of kiwix-lib ?

mgautierfr:
> Development builds (on commits to master) will build against master libzim
while release builds (on tag pushes) will build against the most recent
release of libzim.

I don't see anything for this on the PR. Is it a behavior inferred from how we build libzim ?

debian/changelog
legoktm: No, this is just a placeholder. The real version is either taken from the tag being pushed or extracted out of meson.build (see https://github.com/legoktm/gh-action-auto-dch/blob/master/auto_dch.py#L29). In theory we could put the real version here, but then it would be another thing for y'all to bump and update when making a new release, so I tried to avoid that.

legoktm:
> > Development builds (on commits to master) will build against master libzim
> > while release builds (on tag pushes) will build against the most recent
> > release of libzim.
> 
> I don't see anything for this on the PR. Is it a behavior inferred from how we build libzim ?

Yes. There are two PPAs, [kiwix](https://launchpad.net/~kiwixteam/+archive/ubuntu/kiwix) (releases) and [kiwix-dev](https://launchpad.net/~kiwixteam/+archive/ubuntu/kiwix-dev) (master). On commit pushes to master we build a dev package (e.g. ` 6.1.7~git202007081458.b82fcd0~focal`) and upload it to the kiwix-dev PPA. On tag pushes we build a release package (`6.1.7~focal`) and upload it to the kiwix PPA. So the logic is implemented that if we are building a dev package, it will use the packages from kiwix-dev as the dependencies, and release builds will depend upon the kiwix PPA.

Kelson asked me to document all this on the wiki, which I'm planning to do this weekend. I did want to make sure that the workflow of building with a dependency from the PPA worked as I expected it to first though.

legoktm:
Note that per discussion with Kelson, I just renamed the PPAs so its now [release](https://launchpad.net/~kiwixteam/+archive/ubuntu/release) and [dev](https://launchpad.net/~kiwixteam/+archive/ubuntu/dev). I expect the package builds to fail until https://github.com/openzim/libzim/pull/367 is merged and the PPA has new libzim builds.

legoktm:
This is ready for reviewing now.

mgautierfr:
> So the logic is implemented that if we are building a dev package, it will use the packages from kiwix-dev as the dependencies, and release builds will depend upon the kiwix PPA.

I missed that libzim package is published on kiwix repository. So you use the dependency packages in there.

It leads to the question of if we should publish libzim package on kiwix ? They are two different projects.. but having a second ppa for libzim is maybe a bit too complex.
I've no real opinion on this. I raising the question here to be sure that everyone agree on this.

Beside this, nothing to say on this PR.

kelson42:
> > So the logic is implemented that if we are building a dev package, it will use the packages from kiwix-dev as the dependencies, and release builds will depend upon the kiwix PPA.
> 
> I missed that libzim package is published on kiwix repository. So you use the dependency packages in there.
> 
> It leads to the question of if we should publish libzim package on kiwix ? They are two different projects.. but having a second ppa for libzim is maybe a bit too complex.
> I've no real opinion on this. I raising the question here to be sure that everyone agree on this.

I share the same thoughs about this topic. I tend to think for the moment this is good enough. Next weeks we will start to think about having our own PPR infrastructure to be able to provide packages for Debian as well. Depending how this effort works, we might consider to create two PPAs.

legoktm:
> It leads to the question of if we should publish libzim package on kiwix ? They are two different projects.. but having a second ppa for libzim is maybe a bit too complex.
> I've no real opinion on this. I raising the question here to be sure that everyone agree on this.

The main downside to having a second openZIM PPA is that users would need to add a second PPA just to be able to install stuff. We'd be asking people to do `sudo add-apt-repository ppa:openzim/release && sudo add-apt-repository ppa:kiwixteam/release` so they could get the correct libzim version that the kiwix package they want depends on.

We can always adjust the human-readable names/descriptions to say "Kiwix & openZIM".

PR #385, opened 2020-07-21
Better kiwix.pc creation

kelson42:
Follower of  #338


mgautierfr:
The same code will leads to the same comments. You will not fool me we another PR :)

I agree that using the pkgconfig module may be better than using a configuration file.
But they are not generating the same output. Maybe the new output is better but it would be nice to explain it.

For example, the "old" kiwix.pc on native_dyn is :

```
prefix=/home/mgautier/Project/KIWIX/WORKING_DIR/BUILD_native_dyn/INSTALL
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: libkiwix
Description: A library that contains a lot of things used by used by other kiwix programs
Version: 9.0.1
Requires: libzim icu-i18n pugixml libcurl
Libs: -L${libdir} -lkiwix 
Cflags: -I${includedir}/ 
```

and the new one is :

```
prefix=/home/mgautier/Project/KIWIX/WORKING_DIR/BUILD_native_dyn/INSTALL
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: kiwix
Description: A library that contains a lot of things used by used by other kiwix programs
URL: https://github.com/kiwix/kiwix-lib
Version: 9.0.1
Requires: libzim >= 5.0.0, icu-i18n, pugixml, libcurl
Requires.private: libmicrohttpd
Libs: -L${prefix}/lib64  -lkiwix
Libs.private: -pthread
Cflags: -I${includedir} 
```

It is good because libmicrohttpd is added to Requires.private but should we also move icu-i18n, pugixml and libcurl to Requires.private ?

On win32_static the old pc is :

```
prefix=/home/mgautier/Project/KIWIX/WORKING_DIR/BUILD_win32_static/INSTALL
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: libkiwix
Description: A library that contains a lot of things used by used by other kiwix programs
Version: 9.0.0
Requires: libzim icu-i18n pugixml libcurl
Libs: -L${libdir} -lkiwix 
Cflags: -I${includedir}/ -DCURL_STATICLIB
```

and the new one is :

```
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: kiwix
Description: A library that contains a lot of things used by used by other kiwix programs
URL: https://github.com/kiwix/kiwix-lib
Version: 9.0.1
Requires: icu-i18n, libzim >= 5.0.0, pugixml, libcurl, libmicrohttpd
Libs: -L${prefix}/lib64  -lkiwix -pthread
Cflags: -I${includedir} -pthread -DCURL_STATICLIB
```

There are two -pthread. We should probably have only one.

This PR is globally a good idea, but I think it need more works. If we use pkgconfig module's implicit dependencies resolution, we may not need the pkg_requires variable.

kelson42:
@mgautierfr I was assuming you or me will have to work on it and as @kelvinhammond was not reacting, I thought we better have it as a branch in our repo. No intention to fool you of course ;). Your comment is a bit more detailed and I believe will help to prepare a better version. Thx.

kelvinhammond:
@mgautierfr The [pkg-config docs](https://people.freedesktop.org/~dbn/pkg-config-guide.html) state that:
> **Requires.private:** A list of private packages required by this package but not exposed to applications.

And
> Requires and Requires.private define other modules needed by the library. It is usually preferred to use the private variant of Requires to avoid exposing unnecessary libraries to the program that is linking with your library.

The `-pthread` being defined twice _**should**_ be OK. `pthread` is both a library and a compiler flag.

> If we use pkgconfig module's implicit dependencies resolution, we may not need the pkg_requires variable.
[Reference](https://mesonbuild.com/Pkgconfig-module.html#implicit-dependencies)
Correct, that may be where `libmicrohttpd` is coming from too. What happens when you try it?

PR #388, opened 2020-07-24
Request header case is ignored

### LINKED ISSUE #387: Range header check should be case-insensitive
The HTTP Range header parsing appears to only support an exact match for header `Range`, looking at:
https://github.com/kiwix/kiwix-lib/blob/e743e04b94a501bb9dba67b61f0f6bb3518a8b33/src/server/request_context.cpp#L88

HTTP headers are generally case-insensitive, and it would be good to also at least support `range`, if not full case-insensitivity.

In particular, the browser `Headers` object treats headers as case-insensitive, and may result in a lower-case header if loading a request via `fetch()`

veloman-yunkan:
Fixes #387 

Originally reported against case sensitivity of the `Range` header, this fix applies to all request headers (since according to RFC 7230 all header fields are case-insensitive, see https://tools.ietf.org/html/rfc7230#section-3.2). However, a corresponding unit-test was added only for the Range header.

### DIFF
--- a/src/server/request_context.cpp
+++ b/src/server/request_context.cpp
@@ -26,6 +26,8 @@
 #include <cstdio>
 #include <atomic>
 
+#include "tools/stringTools.h"
+
 namespace kiwix {
 
 static std::atomic_ullong s_requestIndex(0);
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
@mgautierfr I think this PR has missed your attention

Issue #393, opened 2017-11-18
Feature Request - Status code for no search results


darkenvy:
When searching, everything is returned as a 200 OK. 
In the event that no results are shown (such as reaching past the last page), we should return a 204 No Content.

**Why 204?** 
The server performed it's task correctly, there is simply nothing to show. 404 would work as well, but may indicate a more severe error when there isn't one.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

**Why is this helpful?**
* Scripts that test the integrity of zims can benefit from status codes (in general).
* We are taking advantage of the fact that zims are indexed. We are using this to scrape articles names on search terms for future zim creation. Having a status code helps stop the script when it reaches a 204 status code.

**Priority**: Low

![https://puu.sh/yoL2g/950f20c666.png](https://puu.sh/yoL2g/950f20c666.png)

MananJethwani:
can I work on this?

kelson42:
@MananJethwani yes

MananJethwani:
basically what I am doing is just checking before returning that if response content is empty then change the status code......is this fine or should some other changes must be included?

kelson42:
no, you should check if the search returns result and change the status code if not.

MananJethwani:
should I make changes in test/server.cpp as well? as it would fail if 204 is returned when it expects 200 status code

kelson42:
@MananJethwani Please do a PR which secures that what is asked in the ticket is implemented. We would then review it and ask for changes if needed.

kelson42:
FYI, we might have to rollback the fix of this ticket because it creates new problem, see #466

PR #396, opened 2020-07-29
added 204 code for empty return of search.

MananJethwani:
added status code 204 for empty search return in src/server.cpp as asked in issue #393 ....... I would like to suggest that some changes are also required in test/service.cpp as when empty search is returned, earlier response code was 200 but this time it might show 204 at its place causing fail in some tests.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -740,6 +740,13 @@ Response InternalServer::handle_search(const RequestContext& request)
   } catch (const std::exception& e) {
     std::cerr << e.what() << std::endl;
   }
+
+  //changing status code if no result obtained
+  if(searcher.getEstimatedResultCount() == 0)
+  {
+    response.set_code(MHD_HTTP_NO_CONTENT);
+  }
+
   return response;
 }
 
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -155,7 +155,7 @@ const ResourceCollection resources200Compressible{
   { NO_ETAG,   "/catalog/searchdescription.xml" },
   { NO_ETAG,   "/catalog/search" },
 
-  { NO_ETAG,   "/search?content=zimfile&pattern=abcd" },
+  { NO_ETAG,   "/search?content=zimfile&pattern=a" },
 
   { NO_ETAG,   "/suggest?content=zimfile&term=ray" },
 
@@ -195,6 +195,15 @@ TEST_F(ServerTest, 200)
 {
   for ( const Resource& res : all200Resources() )
     EXPECT_EQ(200, zfs1_->GET(res.url)->status) << "res.url: " << res.url;
+}
+
+// seperate test for 204 code
+
+TEST_F(ServerTest, EmptySearchReturnsA204StatusCode)
+{
+  const char* url="/search?content=zimfile&pattern=abcd";
+  auto res=zfs1_->GET(url);
+  EXPECT_EQ(204, res->status) << "res.url: " << url;
 }
 
 TEST_F(ServerTest, CompressibleContentIsCompressedIfAcceptable)


veloman-yunkan:
>  I would like to suggest that some changes are also required in test/service.cpp as when empty search is returned, earlier response code was 200 but this time it might show 204 at its place causing fail in some tests.

@MananJethwani The proposed change must be included in the PR. PRs must not break the unit-tests.

MananJethwani:
ok sorry I will make those changes

test/server.cpp
```diff
@@ -194,14 +194,14 @@ ResourceCollection all200Resources()
 TEST_F(ServerTest, 200)
 {
   for ( const Resource& res : all200Resources() )
-    EXPECT_EQ(200, zfs1_->GET(res.url)->status) << "res.url: " << res.url;
+    EXPECT_GE(204, zfs1_->GET(res.url)->status) << "res.url: " << res.url;
```
veloman-yunkan, line 197: That's not a good check. It allows for any response code <= 204, and doesn't validate which status is returned for which request.

A good test must possess the following properties:

1. It must pass if the code is correct
2. It must fail if the code is not correct

It is a typical mistake to forget about item 2, since people usually feel happy when they observe passing tests. As a result sometimes tests are created that have no chance of failing, or gradually degrade toward that state. In this particular case the test has degraded - the server may return status code 123 but the test will not catch that.   

The same comment applies to all your other changes in `tests/server.cpp`

MananJethwani: ok I didn't notice that one I would try to make changes around that, thanks if you could give me any idea of sort of to help please feel free to tell

veloman-yunkan: To the best of my understanding, the root cause for the server tests failing because of your changes is that the tested URLs contain at least one search query that returns an empty result. You can modify those URLs so that the search results are non-empty. And then you can add an additional (separate) test case for your enhancement.

MananJethwani: Ok I was thinking about removing EXPECT_EQ macro for 200 status code and implementing it with 2 OR statements in condition but if that is an ill practice I will go by your idea thanks.

src/server.cpp
```diff
@@ -740,6 +740,15 @@ Response InternalServer::handle_search(const RequestContext& request)
...
     std::cerr << e.what() << std::endl;
   }
+
+  // changing status code if no results obtained
+  searcher.restart_search();
+  
+  if(!searcher.getNextResult())
+  {
+    response.set_code(MHD_HTTP_NO_CONTENT);
+  }
```
mgautierfr, line 750: With the `restart_search` you are doing two searches.
One to generate the http content, one to test if there is results.

You can use `getEstimatedResultCount();` to check if there are result or not.

MananJethwani: ok i would include that changes as well in the next commit thanks

MananJethwani: ok I will include that change as well in next commi.....btw i thought of that earlier but didn't understood  meaning of estimate is that a thing related to upper bound?

MananJethwani:
please check the changes I have made and let me know if i have done something wrong......thank you

test/server.cpp
```diff
@@ -523,3 +537,17 @@ TEST_F(ServerTest, RangeHasPrecedenceOverCompression)
...
+  const auto r0 = zfs1_->GET(url, { {"Range", "bytes=100-200"} } );
+
+  const char* header_variations[] = { "RANGE", "range", "rAnGe", "RaNgE" };
+  for ( const char* header : header_variations ) {
+    const auto r = zfs1_->GET(url, { {header, "bytes=100-200"} } );
+    EXPECT_EQ(206, r->status);
+    EXPECT_EQ("bytes 100-200/20077", r->get_header_value("Content-Range"));
+    EXPECT_EQ(r0->body, r->body);
+  }
+}
```
veloman-yunkan, line 553: You picked up this change from a different commit accidentally. It must not show up in this PR.

test/server.cpp
```diff
@@ -155,7 +161,7 @@ const ResourceCollection resources200Compressible{
   { NO_ETAG,   "/catalog/searchdescription.xml" },
   { NO_ETAG,   "/catalog/search" },
 
-  { NO_ETAG,   "/search?content=zimfile&pattern=abcd" },
+{ NO_ETAG,   "/search?content=zimfile&pattern=a" },
```
veloman-yunkan, line 164: Please restore the whitespace.

test/server.cpp
```diff
@@ -197,6 +203,14 @@ TEST_F(ServerTest, 200)
...
 }
 
+// seperate test for 204 code
+
+TEST_F(ServerTest, 204)
+{
+  for ( const Resource& res : resources_204 )
+    EXPECT_EQ(204, zfs1_->GET(res.url)->status) << "res.url: " << res.url;
+}
+
```
veloman-yunkan, line 213: It doesn't make much sense to have multiple search URLs returning empty results (the reason for multiple URLs  in the other cases is to provide maximal coverage). This test can exercise a single URL similar to `ServerTest.SuccessfulSearchForAnArticleTitleRedirectsToTheArticle` (and you can rename it to `EmptySearchReturnsA204StatusCode`)

test/server.cpp
```diff
@@ -139,6 +139,12 @@ std::ostream& operator<<(std::ostream& out, const Resource& r)
 
 typedef std::vector<Resource> ResourceCollection;
 
+
+const ResourceCollection resources_204{
+    { NO_ETAG,   "/search?content=zimfile&pattern=abcd" },
+    { NO_ETAG,   "/search?content=zimfile&pattern=xyzw" },
+};
+
 const ResourceCollection resources200Compressible{
```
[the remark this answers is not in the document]
MananJethwani: sorry I thought I removed them. just a second i will remove that

kelson42:
@MananJethwani Please rebase your branch on origin/master and squash everything in one commit. I will then merge it.

MananJethwani:
@kelson42 done as asked

Issue #321, opened 2020-02-07
Better handling of search pagination parameters.


mgautierfr:
For now, the pagination is handled using two parameters `start` and `end`. `end-start` being the `nb_element_per_page`.
If you go to the last page using the button added at end of the search page, `start` is `last_page_index * nb_element_per_page` and `end` is `number_of_result`. In this situation the `end-start` will be smaller that the original `nb_element_per_page`.
If we navigate again in the search pages, the page lengh is now smaller than original. (In worst case, only one element per page).

Instead of a `start`, `end` parameters, we should use a `start`, `pagelength` parameter. This way the `pagelength` is always the same. We must only change the `start` parameter to change page.

MananJethwani:
I want to work on this

MananJethwani:
can you elaborate a bit on what I should do.....as I can see I have 2 options either I can just change the request part in server.cpp to get the pagelength instead of end and calculate end accordingly(start+pagelength and some exceptional cases) or I could change request as well as searcher and search_rendered to take pagelength as an argument everywhere and make changes accordingle

kelson42:
@mgautierfr @MananJethwani Should we notcloe this ticket?

mgautierfr:
Yes.

PR #399, opened 2020-08-05
added pageLength for search_pagination for issue

MananJethwani:
added a function to set  result count per page from server.cpp for issue #321

### DIFF
--- a/include/search_renderer.h
+++ b/include/search_renderer.h
@@ -61,6 +61,13 @@ class SearchRenderer
   void setSearchProtocolPrefix(const std::string& prefix);
 
   /**
+   * set result count per page
+   */
+  void setPageLength(unsigned int pageLength){
+    this->pageLength  = pageLength;
+  }
+
+  /**
    * Generate the html page with the resutls of the search.
    */
   std::string getHtml();
@@ -73,10 +80,9 @@ class SearchRenderer
   std::string searchPattern;
   std::string protocolPrefix;
   std::string searchProtocolPrefix;
-  unsigned int resultCountPerPage;
+  unsigned int pageLength;
   unsigned int estimatedResultCount;
   unsigned int resultStart;
-  unsigned int resultEnd;
 };
 
 
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -93,30 +93,30 @@ std::string SearchRenderer::getHtml()
   kainjow::mustache::data pages{kainjow::mustache::data::type::list};
 
   auto resultStart = mp_searcher->getResultStart();
-  auto resultEnd = mp_searcher->getResultEnd();
-  auto resultCountPerPage = resultEnd - resultStart;
+  auto resultEnd = 0U;
   auto estimatedResultCount = mp_searcher->getEstimatedResultCount();
   auto currentPage = 0U;
   auto pageStart = 0U;
   auto pageEnd = 0U;
   auto lastPageStart = 0U;
-  if (resultCountPerPage) {
-    currentPage = resultStart/resultCountPerPage;
+  if (pageLength) {
+    currentPage = resultStart/pageLength;
     pageStart = currentPage > 4 ? currentPage-4 : 0;
     pageEnd = currentPage + 5;
-    if (pageEnd > estimatedResultCount / resultCountPerPage) {
-      pageEnd = estimatedResultCount / resultCountPerPage;
+    if (pageEnd > estimatedResultCount / pageLength) {
+      pageEnd = (estimatedResultCount + pageLength - 1) / pageLength;
     }
-    if (estimatedResultCount > resultCountPerPage) {
-      lastPageStart = static_cast<int>(round(estimatedResultCount/resultCountPerPage)) * resultCountPerPage;
+    if (estimatedResultCount > pageLength) {
+      lastPageStart = ((estimatedResultCount-1)/pageLength)*pageLength;
     }
   }
+
+  resultEnd = resultStart+pageLength; //setting result end
 
   for (unsigned int i = pageStart; i < pageEnd; i++) {
     kainjow::mustache::data page;
     page.set("label", to_string(i + 1));
-    page.set("start", to_string(i * resultCountPerPage));
-    page.set("end", to_string((i + 1) * resultCountPerPage));
+    page.set("start", to_string(i * pageLength));
 
     if (i == currentPage) {
       page.set("selected", true);
@@ -137,9 +137,8 @@ std::string SearchRenderer::getHtml()
   allData.set("searchPatternEncoded", urlEncode(this->searchPattern));
   allData.set("resultStart", to_string(resultStart + 1));
   allData.set("resultEnd", to_string(min(resultEnd, estimatedResultCount)));
-  allData.set("resultRange", to_string(resultCountPerPage));
+  allData.set("pageLength", to_string(pageLength));
   allData.set("resultLastPageStart", to_string(lastPageStart));
-  allData.set("lastResult", to_string(estimatedResultCount));
   allData.set("protocolPrefix", this->protocolPrefix);
   allData.set("searchProtocolPrefix", this->searchProtocolPrefix);
   allData.set("contentId", this->searchContent);
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -708,33 +708,36 @@ Response InternalServer::handle_search(const RequestContext& request)
   try {
     start = request.get_argument<unsigned int>("start");
   } catch (const std::exception&) {}
-  auto end = 25;
-  try {
-    end = request.get_argument<unsigned int>("end");
-  } catch (const std::exception&) {}
-  if (start>end) {
-    auto tmp = start;
-    start = end;
-    end = tmp;
+  auto pageLength=25;
+  try{
+    pageLength=request.get_argument<unsigned int>("pageLength");
+  }catch(const std::exception&){}
+  if (pageLength > MAX_SEARCH_LEN) {
+    pageLength = MAX_SEARCH_LEN;
   }
-  if (end > start + MAX_SEARCH_LEN) {
-    end = start + MAX_SEARCH_LEN;
+
+  if(pageLength==0)
+  {
+    pageLength=25;
   }
+
+  auto end=start+pageLength;
 
   /* Get the results */
   try {
     if (patternString.empty()) {
       searcher.geo_search(latitude, longitude, distance,
-                           start, end, m_verbose.load());
+                           start, end , m_verbose.load());
     } else {
       searcher.search(patternString,
-                       start, end, m_verbose.load());
+                       start, end , m_verbose.load());
     }
     SearchRenderer renderer(&searcher, mp_nameMapper);
     renderer.setSearchPattern(patternString);
     renderer.setSearchContent(bookName);
     renderer.setProtocolPrefix(m_root + "/");
     renderer.setSearchProtocolPrefix(m_root + "/search?");
+    renderer.setPageLength(pageLength);
     response.set_content(renderer.getHtml());
   } catch (const std::exception& e) {
     std::cerr << e.what() << std::endl;
--- a/static/templates/search_result.html
+++ b/static/templates/search_result.html
@@ -133,7 +133,7 @@
         <ul>
           {{#resultLastPageStart}}
             <li>
-              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start=0&end={{resultRange}}">
+              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start=0&pageLength={{pageLength}}">
                 ◀
               </a>
             </li>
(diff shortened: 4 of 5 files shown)


src/server.cpp
```diff
@@ -713,6 +713,10 @@ Response InternalServer::handle_search(const RequestContext& request)
   try {
     end = request.get_argument<unsigned int>("end");
   } catch (const std::exception&) {}
+  auto pageLength=25;
+  try{
+    pageLength=request.get_argument<unsigned int>("pageLenth");
```
veloman-yunkan, line 718: Typo: pageLenth -> pageLength

MananJethwani:
@veloman-yunkan do i need to remove end parameter from search and search_renderer completely? i thought just the handling pagination part must be done using start and pagelength and it is being done just like that. resultEnd is not used in pagination part it is just being assigned for other things I guess.

veloman-yunkan:
@MananJethwani My interpretation is that the `end` parameter must be removed from query (i.e. from `InternalServer::handle_search()`)

MananJethwani:
@veloman-yunkan I removed end from server,search and search_renderer. I also changed starting point to be according to page no.

src/server.cpp
```diff
@@ -709,33 +709,31 @@ Response InternalServer::handle_search(const RequestContext& request)
...
-    end = start + MAX_SEARCH_LEN;
+  auto pageLength=25;
+  try{
+    pageLength=request.get_argument<unsigned int>("pageLength");
+  }catch(const std::exception&){}
+  if (pageLength > MAX_SEARCH_LEN) {
+    pageLength = MAX_SEARCH_LEN;
   }
+  if(pageLength==0)pageLength=25; // pageLength can't be zero
+  start=(start/pageLength)*pageLength; // changed start point according to page no.
```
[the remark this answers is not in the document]
MananJethwani: @veloman-yunkan no anything bad won't happen but I just thought it would be better if result start is the first result of the page before entering pagination as earlier it was a bit tough to understand setting range but it would be better now.

MananJethwani:
> Please test your changes by building kiwix/kiwix-tools and running kiwix-serve (asking for a unit-test would be too much for this PR).

@veloman-yunkan can you help me with kiwix/kiwix-tools build as I don't understand how i will link my local kiwix-lib repo. with kiwix-tools and whenever i try to run meson .  build it shows kiwix not found

veloman-yunkan:
@MananJethwani First make a clean build of kiwix-tools via kiwix-build. Then in `SOURCES/kiwix-lib` switch to your branch, and afterwards run `kiwix-build` with the `--skip-source-prepare` option.

MananJethwani:
@veloman-yunkan can i get your irc name to contact you I am having some problems regarding kiwix-build .would it be fine if I ask you there?

veloman-yunkan:
@MananJethwani I just joined as veloman

MananJethwani:
@veloman-yunkan I made the required changes to apply search_pagination properly. and tested it by running kiwix-serve.

static/templates/search_result.html
```diff
@@ -133,22 +133,22 @@
         <ul>
           {{#resultLastPageStart}}
             <li>
-              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start=0&end={{resultRange}}">
+              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start=0&pageLength={{resultRange}}">
                 ◀
```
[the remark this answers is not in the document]
MananJethwani: ok

src/search_renderer.cpp
```diff
@@ -93,36 +93,54 @@ std::string SearchRenderer::getHtml()
...
+    kainjow::mustache::data page;
+    page.set("label",to_string(pageEnd+1));
+    page.set("start",to_string(pageEnd*pageLength));
+
+    if(pageEnd==currentPage)
+    {
+      page.set("selected",true);
+    }
+    pages.push_back(page);
+  }
```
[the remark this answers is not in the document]
MananJethwani: @veloman-yunkan yes you are right just a min i will change that.

src/search_renderer.cpp
```diff
@@ -93,37 +93,42 @@ std::string SearchRenderer::getHtml()
...
-    if (estimatedResultCount > resultCountPerPage) {
-      lastPageStart = static_cast<int>(round(estimatedResultCount/resultCountPerPage)) * resultCountPerPage;
+    if (estimatedResultCount > pageLength) {
+      lastPageStart = static_cast<int>(round(estimatedResultCount/pageLength)) * pageLength;
     }
+    if(pageEnd*pageLength==estimatedResultCount && pageEnd>0)pageEnd--;
   }
 
-  for (unsigned int i = pageStart; i < pageEnd; i++) {
+  for (unsigned int i = pageStart; i <= pageEnd; i++) {
```
veloman-yunkan, line 115: Did this change stay after attempts to find a working solution?

src/search_renderer.cpp
```diff
@@ -93,37 +93,42 @@ std::string SearchRenderer::getHtml()
...
-      pageEnd = estimatedResultCount / resultCountPerPage;
+    if (pageEnd > estimatedResultCount / pageLength) {
+      pageEnd = estimatedResultCount / pageLength;
     }
-    if (estimatedResultCount > resultCountPerPage) {
-      lastPageStart = static_cast<int>(round(estimatedResultCount/resultCountPerPage)) * resultCountPerPage;
+    if (estimatedResultCount > pageLength) {
+      lastPageStart = static_cast<int>(round(estimatedResultCount/pageLength)) * pageLength;
     }
+    if(pageEnd*pageLength==estimatedResultCount && pageEnd>0)pageEnd--;
```
veloman-yunkan, line 112: I feel that the `pageEnd` management logic can be simplified even further by taking advantage of the implicit relation between `pageEnd` and `lastPageStart`

src/search_renderer.cpp
MananJethwani: yes as now I am not adding last page individually it should be added using the loop only

src/search_renderer.cpp
MananJethwani: @veloman-yunkan  ok let me take a look I have some more ideas to apply to make code shorter and more efficient, I will make a final review attempt tomorrow after applying all the changes by then if you have any more suggestions please let me now :)

MananJethwani: @veloman-yunkan I guess pageEnd and lastPageStart are not always related because pageEnd is the page which is at max  +5 of current page but lastPageStart is start of the last page of whole result.

src/search_renderer.cpp
veloman-yunkan: @MananJethwani I cannot agree here. If `pageEnd` stays equal to `currentPage+5` then this change includes an extra page in the page list. I think that the problem you were dealing with was about the case when `pageEnd` had to be adjusted, so instead it just needs to be adjusted properly.

src/search_renderer.cpp
```diff
@@ -93,30 +93,38 @@ std::string SearchRenderer::getHtml()
...
     pageEnd = currentPage + 5;
-    if (pageEnd > estimatedResultCount / resultCountPerPage) {
-      pageEnd = estimatedResultCount / resultCountPerPage;
+    if (pageEnd > estimatedResultCount / pageLength) {
+      pageEnd = estimatedResultCount / pageLength;
+      //to include last page in pagination if page is not empty
+      if(pageEnd*pageLength!=estimatedResultCount)
+      {
+        pageEnd++;
+      }
```
veloman-yunkan, line 112: So you want to find the upward rounded result of `estimatedResultCount / pageLength` (i.e. the smallest value of `pageEnd` such that `pageEnd*pageLength>=estimatedResultCount`). This can be computed with a single expression as follows:
```
pageEnd = (estimatedResultCount + pageLength - 1) / pageLength;
```

src/search_renderer.cpp
```diff
@@ -93,30 +93,38 @@ std::string SearchRenderer::getHtml()
...
     }
-    if (estimatedResultCount > resultCountPerPage) {
-      lastPageStart = static_cast<int>(round(estimatedResultCount/resultCountPerPage)) * resultCountPerPage;
+    if (estimatedResultCount > pageLength) {
+      lastPageStart = static_cast<int>(round(estimatedResultCount/pageLength)) * pageLength;
+      //changing last page if it is not empty
+      if(lastPageStart==estimatedResultCount && lastPageStart>pageLength)
+      {
+        lastPageStart-=pageLength;
+      }
```
veloman-yunkan, line 120: I think this can be simplified to

```
lastPageStart = ((estimatedResultCount-1)/pageLength)) * pageLength;
```

since `(estimatedResultCount-1)/pageLength` is the index of the page containing the last result.

src/search_renderer.cpp
```diff
@@ -136,10 +134,9 @@ std::string SearchRenderer::getHtml()
   allData.set("searchPattern", kiwix::encodeDiples(this->searchPattern));
   allData.set("searchPatternEncoded", urlEncode(this->searchPattern));
   allData.set("resultStart", to_string(resultStart + 1));
-  allData.set("resultEnd", to_string(min(resultEnd, estimatedResultCount)));
-  allData.set("resultRange", to_string(resultCountPerPage));
+  allData.set("resultEnd", to_string(min(resultStart + pageLength, estimatedResultCount)));
```
mgautierfr, line 137: `resultStart` and `resultEnd` are only used in the template to know how many result is displayed (on the total number of result).
It may be better to pass this number using a specific variable name.
("pageLength" could be a good name, but it must not be mixed with the requested page length.
The current page length can be different than the request page length if there is less result than requested or if we are at the last page and it contains less result that the requested length).

mgautierfr:
When you update your branch with master, I prefer than you rebase your branch on master instead of merging master into your branch.

src/search_renderer.cpp
```diff
@@ -93,30 +93,28 @@ std::string SearchRenderer::getHtml()
   kainjow::mustache::data pages{kainjow::mustache::data::type::list};
 
   auto resultStart = mp_searcher->getResultStart();
-  auto resultEnd = mp_searcher->getResultEnd();
-  auto resultCountPerPage = resultEnd - resultStart;
+  auto pageLength = this->pageLength;
```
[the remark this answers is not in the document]
MananJethwani: so I should use a getPageLength method in searcher to get pageLength right?

src/search_renderer.cpp
MananJethwani: @mgautierfr so there should be a `resultEnd` variable for this instead of `resultStart + pageLength`

src/search_renderer.cpp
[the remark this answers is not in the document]
MananJethwani: @mgautierfr so I should not copy it and use it directly right?

src/server.cpp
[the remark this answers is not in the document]
MananJethwani: @mgautierfr so which of these lines should i remove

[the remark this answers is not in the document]
MananJethwani: so I will remove these two then

src/search_renderer.cpp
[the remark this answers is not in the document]
mgautierfr: Yes, as you have done it now.

src/search_renderer.cpp
mgautierfr: No, it was not my comment.
My comment was that the two tokens `resultEnd` and `resultStart` was only used in the template to calculate the different between two to get the number of result displayed in the page. And so we could calculate this is cpp and pass it directly to the token.

But I was wrong. The token doesn't calculate the different but display the both values `{{resultStart}}-{{resultEnd}}`. I was a bit quick on this point.

So there was nothing to change here. However, the change you've made to reintroduce `resultEnd` variable is ok, even if I was not asking for that. So we are good :)

mgautierfr:
You have removed the `if(pageLength==0)pageLength=25;` It is a good think the keep it, it make no sens to ask for 0 results.

On a more global way, I think we shouldn't change the search api on the cpp side.
The initial issue was only about the http request, and the fact that with the `end` argument and how we render the template, we may change the induce page length.
We don't have this problem on the cpp side.
- `pagelength` name doesn't make sens on the cpp side, it only make sens when rendering page, so in the searchRenderer/server request.
- And changing this introduces a API break (difficult to catch as the type and the number of arguments don't change).

MananJethwani:
@mgautierfr so I should keep the line `if(pageLength==0)pageLength=25;` , and are there any other changes? I don't understand.

mgautierfr:
Yes, we should keep the line `if(pageLength==0)pageLength=25;` (but maybe put it in different line, with `{}`)

One "big" change to do is to revert the change on the search api (c++ side). This introduce a API break we don't want. We should keep only the change on the server/search_renderer side.
(It will simplify this PR)

MananJethwani:
@mgautierfr should I retain the end argument in request ? I have reverted the changes in src/searcher.cpp and include/searcher.h please check them as well .

include/search_renderer.h
```diff
@@ -60,6 +60,13 @@ class SearchRenderer
    */
   void setSearchProtocolPrefix(const std::string& prefix);
 
+  /**
+   * set result count per page
+   */
+  void setResultCountPerPage(unsigned int pageLength){
```
mgautierfr, line 66: Maybe rename this `setPageLength`.

mgautierfr:
> should I retain the end argument in request ? 

No, we want to remove the "end" argument and replace it with "pageLength".

mgautierfr:
We are good. Thanks for this PR.

PR #400, opened 2020-08-05
Set fixed size for favicon in home page listing

rgaudin:
While [spec](https://wiki.openzim.org/wiki/Metadata#Favicon) says that the favicon
should be a 48x48 image, ZIM creators might not respect it.

If a ZIM contains a larger favicon, the UI is broken. This fixes it and ensures all favicon have equal sizes, removing the unpleasing lack of harmony that we can see sometimes.

Note that ZIM with smaller size favicon would get blurry as those would be upscaled.


rgaudin:
rebased

PR #403, opened 2020-08-06
Prevent taskbar and blocker at article level

### LINKED ISSUE #391: Prevent taskbar at article-level
In the context of zimit we need to be able to query HTML article and retrieve them as is, without the taskbar.
In current zimit, we have the JS replayer query articles and displaying them itself.

In the current libkiwix, the taskbar can only be disabled globally at the library caller level (kiwix-serve in this case) and if enabled, automatically applies to all `text/html` articles.

It used to be fine with other scraper as the typical use case is to retrieve JSON or JS data from other URLs but in this case, the replayer requests full HTML content and so the taskbar is injected where it doesn't belong.

This is currently circumvented by storing those HTML articles with a different MIME-type.

**Question**: what is the wanted behavior for this? Should we be able to request raw content? Is the mime-trick the way to go? Is there anything in this field in libzim2?

@mgautierfr @kelson42 @ikreymer

rgaudin:
Some HTML articles are meant to be displayed through a viewer. In this case,
we know we don't want the server to inject the taskbar nor the link blocker
because the content is not a user-ready web page but a partial element of it.

Such articles still need to be `text/html` to be parsed properly by browsers.

This changes the way we decide to display the tasbar or not.
Previously, we were adding it to every article with a MIME __starting with__ `text/html`.
Now, we're additionally preventing it on `text/html` MIME if there is a `;raw=true` string inside.

This leaves articles with MIME `text/html;raw=true` (warc2zim convention) outside
of the taskbar target.

For similar reasons, the external-link blocker is set to apply to the same set of articles.
Previously, it was applied to all articles which was an (unoticable) mistake.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -963,7 +963,7 @@ Response InternalServer::handle_content(const RequestContext& request)
     printf("mimeType: %s\n", response.get_mimeType().c_str());
   }
 
-  if (response.get_mimeType().find("text/html") != string::npos)
+  if ( startsWith(response.get_mimeType(), "text/html") )
     response.set_taskbar(bookName, reader->getTitle());
 
   return response;
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -127,8 +127,8 @@ std::string render_template(const std::string& template_str, kainjow::mustache::
 
 void Response::introduce_taskbar()
 {
-  if (! m_withTaskbar)
-    // Taskbar is globally disabled.
+  if ( !m_withTaskbar || !contentDecorationAllowed() )
+    // Taskbar either globally disabled or not allowed for this type
     return;
   kainjow::mustache::data data;
   data.set("root", m_root);
@@ -152,6 +152,8 @@ void Response::introduce_taskbar()
 
 void Response::inject_externallinks_blocker()
 {
+  if ( !contentDecorationAllowed() )
+    return;
   kainjow::mustache::data data;
   data.set("root", m_root);
   auto script_tag = render_template(RESOURCE::templates::external_blocker_part_html, data);
@@ -167,6 +169,12 @@ Response::can_compress(const RequestContext& request) const
   return request.can_compress()
       && is_compressible_mime_type(m_mimeType)
       && (m_content.size() > KIWIX_MIN_CONTENT_SIZE_TO_DEFLATE);
+}
+
+bool
+Response::contentDecorationAllowed() const
+{
+    return m_mimeType.find(";raw=true") == std::string::npos;
 }
 
 MHD_Response*
@@ -189,7 +197,7 @@ Response::create_raw_content_mhd_response(const RequestContext& request)
   if (m_addTaskbar) {
     introduce_taskbar();
   }
-  if ( m_blockExternalLinks ) {
+  if ( m_blockExternalLinks && startsWith(m_mimeType, "text/html") ) {
     inject_externallinks_blocker();
   }
 
--- a/src/server/response.h
+++ b/src/server/response.h
@@ -71,6 +71,7 @@ class Response {
     void inject_externallinks_blocker();
 
     bool can_compress(const RequestContext& request) const;
+    bool contentDecorationAllowed() const;
 
   private: // functions
     MHD_Response* create_mhd_response(const RequestContext& request);


rgaudin:
> @rgaudin This can not work that way because we can have a mime-type with other parameters, for example `text/html;charset=utf-8`. We need to use a specific parameter, I propose `src=true`. `partial=true` is good as well.

OK, so you want to apply the taskbar to all but those that include this parameter, right ?

In this case, I don't think `src` is clearer than `partial`. Maybe `raw=true` ?

kelson42:
> > @rgaudin This can not work that way because we can have a mime-type with other parameters, for example `text/html;charset=utf-8`. We need to use a specific parameter, I propose `src=true`. `partial=true` is good as well.
> 
> OK, so you want to apply the taskbar to all but those that include this parameter, right ?

yes, we have to.

> In this case, I don't think `src` is clearer than `partial`. Maybe `raw=true` ?

`src` is meant here because the page in loaded via an iframe `src` attribute. In all other cases this is a `href` attribute... but I agree with all thre proposals.

rgaudin:
> `src` is meant here because the page in loaded via an iframe `src` attribute. In all other cases this is a `href` attribute... but I agree with all thre proposals.

OK, I think `src` is too tied to the current iframe-based use case. I've chosen `raw=true` and updated the code.

Please let me know if this is the right location for this function ; I'm not familiar with the code base.

kelson42:
@rgaudin Better in the principle, but:
* I would rename the function to `isRawHtmlMime` (keep the naming coherent)
* I would make the function `static`

rgaudin:
> @rgaudin Better in the principle, but:
>  I would rename the function to `isRawHtmlMime` (keep the naming coherent)

OK, not sure what it wasn't coherent with but OK. Requires two tests now

```cpp
// before
if (isRegularHtmlMime(response.get_mimeType()))
// now
if ( kiwix::startsWith(response.get_mimeType(), "text/html") && !isRawHtmlMime(response.get_mimeType()))
```

>  I would make the function `static`

I don't get that. this is a bare function in the `kiwix` namespace. Do you mean to skip the namespace prefix?

src/tools/stringTools.cpp
```diff
@@ -391,3 +391,9 @@ bool kiwix::startsWith(const std::string& base, const std::string& start)
         && std::equal(start.begin(), start.end(), base.begin());
 }
 
+
+bool kiwix::isRawHtmlMime(const std::string& mime)
+{
+  return startsWith(mime, "text/html")
+        && mime.find(";raw=true") != std::string::npos;
+}
```
veloman-yunkan, line 399: `stringTools.h/cpp` is not a good home for such a narrow check. I would rather make this a member function of the `Response` class and decouple it from HTML-ness, since the `raw=true` attribute may later be applied to other content types as well. Thus the member function (with reversed logic) will be

```c++
bool Response::contentDecorationAllowed() const
{
    return m_mimeType.find(";raw=true") == std::string::npos;
}
```

rgaudin: Yes that's better. thanks.

mgautierfr:
I'm a bit disturb by this PR, both with the solution proposed and the implementation.

### Solution proposed

`text/html;raw=true` seems interesting in first instance as we use optional mimetype parameter and we don't change a lot. But in fact it create a different mimetype than `text/html`.

I dislike this as we mix the type of the content with the a specific use case of only one application, needed for only one kind of zimfile.

- The full mimetype will be send to the client (browser and native applications). We should not as the `raw=true` is entirely use for one use case (kiwix-serve). How a browser will interpreter the `raw=true` parameter ? 
- Both mimetype will be stored in the `Counter` metadata. It will break at least [getArticleCount](https://github.com/kiwix/kiwix-lib/blob/master/src/reader.cpp#L132-L148) (And any other usage of this public metadata)

There is maybe better solution that changing the contenttype store in the zim file for this specific case.
For example, as the request is made from the viewer, the raw request can be set as a request parameter. Thus, we would not change the content of zimfile. (Not sure if it is a better idea, but we should think about the alternative before implementing thing quickly).

### Implementation

I dislike the idea of a check on the mimetype both on the server side (if it is html, add taskbar) and the response side (if it raw, don't add taskbar). Tests should be made on the same place. (Same with the inject_externallink_blocker, where there is several tests both before calling the method and inside the method).

And finally, it would have been nice to merge https://github.com/kiwix/kiwix-lib/pull/397 before. It would have avoid the merge conflict (At least it would have been simple to rebase this change on #397 instead of rebasing #397 on this change)

veloman-yunkan:
> And finally, it would have been nice to merge #397 before. It would have avoid the merge conflict (At least it would have been simple to rebase this change on #397 instead of rebasing #397 on this change)

@mgautierfr You can merge `master` into the `refactor_response` branch before merging the latter back into `master`. I don't think that there is a strong need for rebasing each PR before merging it.

PR #409, opened 2020-08-12
Run the CI on pull request.

### LINKED ISSUE #410: CI should be run on pull_request event
It was the case before and as been changed by this commit : https://github.com/kiwix/kiwix-lib/commit/286fbba9f6b6921a23305bc9dd370334e42d930e

The `CI` workflow is only used to check that the code is valid (it compiles it and test it), nothing is publish or keep. (And so, it should be called `Check`)
It "publish" only code-coverage reports and comment on the PR (who make sens only in a PR context).

The workflow is that any changes must pass by a PR.
We want to check that the PR is valid. If the PR is on a branch in a fork repository, there is not push on our side and the CI is not launch.

Running the Check on master is useless as it produces nothing and if there is something wrong in the code, what is wrong is that the code should never has been push in first instance. It add no benefice.

mgautierfr:
This is needed to have the CI run on pull request created from of fork
repository.

Fixes #410

### DIFF
--- a/.github/workflows/check_pr.yml
+++ b/.github/workflows/check_pr.yml
@@ -1,6 +1,6 @@
-name: CI
+name: Check PR
 
-on: [push]
+on: [pull_request]
 
 jobs:
   Macos:


kelson42:
I would like to have a ticket so we can discuss the problem properly. I’m not happy with this PR because I want to secure that master head passes properly the CI.

legoktm:
Why not do `[push, pull_request]` like the packages.yml workflow does?

rgaudin:
Shouldn't it be updated to checkout@v2?

mgautierfr:
This is (too) quick change. It probably should be a revert of https://github.com/kiwix/kiwix-lib/commit/286fbba9f6b6921a23305bc9dd370334e42d930e.

> Why not do [push, pull_request] like the packages.yml workflow does?

Because it make the CI run twice on "local" pull request and this is a wast of resources.

mgautierfr:
Issue discussing about this is #410

kelson42:
Superseeded by https://github.com/kiwix/libkiwix/issues/410

PR #411, opened 2020-08-25
Fix segmentation fault around Aria2 RPC.

mgautierfr:
Fix wrong curl request structure and avoid segmentation fault.

Also fix destruction of subprocess and associated monitor thread.

### DIFF
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -26,8 +26,7 @@ Aria2::Aria2():
   mp_aria(nullptr),
   m_port(42042),
   m_secret("kiwixariarpc"),
-  mp_curl(nullptr),
-  m_lock(PTHREAD_MUTEX_INITIALIZER)
+  mp_curl(nullptr)
 {
   m_downloadDir = getDataDirectory();
   makeDirectory(m_downloadDir);
@@ -115,6 +114,7 @@ Aria2::Aria2():
 
 Aria2::~Aria2()
 {
+  std::unique_lock<std::mutex> lock(m_lock);
   curl_easy_cleanup(mp_curl);
 }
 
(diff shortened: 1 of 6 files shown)


src/aria2.cpp
```diff
@@ -84,24 +84,24 @@ Aria2::Aria2():
...
-      fprintf(stderr, "\nlibcurl: (%d) ", res);
-      if (errbuf[0] != 0) {
-        std::cerr << errbuf << std::endl;
+      std::cerr <<"ERROR : curl_easy_perform() failed." << std::endl;
+      std::cerr << "libcurl: (" << res << ") " << std::endl;
+      if (m_curlErrorBuffer[0] != 0) {
+        std::cerr << m_curlErrorBuffer.get() << std::endl;
       } else {
         std::cerr << curl_easy_strerror(res) << std::endl;
       }
```
veloman-yunkan, line 107: I would deduplicate this piece of code with an identical piece of code in `Aria2::doRequest()`

veloman-yunkan (changes requested):
The PR title doesn't correctly summarize the contents.

src/aria2.cpp
mgautierfr: Done with a macro.

mgautierfr:
> The PR title doesn't correctly summarize the contents.

Yes, the PR was one commit only and github uses the commit's title and description for the PR.
I've update the PR's title and description.

veloman-yunkan (approved):
I wonder why you don't like helper functions, but I won't block because of that.

mgautierfr:
> I wonder why you don't like helper functions, but I won't block because of that.

I tend to use functions for what they provides : They encapsulate a functionality implemented using several instruction and provided to user code through a defined interface. They identify a feature and a set of instructions.

Using function to do code deduplication has several cons :
- Function has a runtime cost (stack allocation, argument passing..). Even if compiler may optimize that I agree.
- Function can complicate the code. It is a unit to manage (should we unit test it ?). It may be move in another file. Modifying the function must be done with care (who are the users of the function, will I break the code ?)

I'm not against helper functions. But I against helper functions **only** to avoid code deduplication.
Macro are better adapted for that. They are a way to insert/replace/construct text. And this is about code deduplication is do not write several time the same text.

On top of that, I do not do code deduplication. I do code detruplication. Most of the time, things only written twice are not important enough to make me think about them and give them a "identity" (at least, I don't think about them *only* because they are written twice).

(The same way, I think about class hierarchy in terms of what classes are and how they specialize and not how they share feature and code (but I don't forget that sharing features can be a clue that classes may be related))

PR #412, opened 2020-08-25
Resetting CURLOPT_ERRORBUFFER before calling Aria2::doRequest

ZaWertun:
Fix for https://github.com/kiwix/kiwix-desktop/issues/513 (Kiwix always crashing after exit on Fedora).
Crash was caused by using of this gcc flags: `-fstack-protector-strong` or/and `-fstack-clash-protection`.


mgautierfr:
This is superseded by #411 

Thanks for patch @ZaWertun anyway. We fix the issue thanks to you.

PR #418, opened 2020-10-28
Fix counter parsing

mgautierfr:


mgautierfr:
This issue is probably common to all parser reading `M/Counter`. They probably have to adapt themselves.

@Jaifroid, you may be interested by this.

veloman-yunkan (commented):
IMHO, this implementation of parsing is too low-level and difficult to understand. I took the liberty of providing a more readable implementation as a new commit 0f8fe1f63.

mgautierfr:
I'm not sure it is more readable. Probably because I've write the first version.
But it is not so important, tests are passing with both implementations.
I'm merging.

PR #419, opened 2020-10-29
debian: Have libkiwix-dev depend on libmicrohttpd-dev

legoktm:
As of 9c925f6778, it's now required in the pkg-config file.


kelson42:
Looks good but surprisingly the compilation does not pass on macOS, adding @mgautierfr to the reviewers.

kelson42:
This PR seems to fail because of the `subprojects` directory. If we remove it it works fine.

kelson42:
Rebased

kelson42 (approved):

PR #428, opened 2020-11-24
Adapt kiwx-lib to the new api of libzim (libzim_next).

mgautierfr:
Need openzim/libzim#454 first.


src/reader.cpp
```diff
@@ -438,56 +365,47 @@ Entry Reader::getEntryFromEncodedPath(const std::string& path) const
...
 {
-  if (!zimFileHandler || zimFileHandler->is_multiPart() )
+  if (!zimArchive)
   {
     return false;
   }
 
-  return ( pathExists("Z//fulltextIndex/xapian")
-        || pathExists("X/fulltext/xapian"));
+  for(auto path: {"Z//fulltextIndex/xapian", "X/fulltext/xapian"}) {
```
veloman-yunkan, line 397: Is the double slash in the path fine?

veloman-yunkan (commented):
Just one question

src/reader.cpp
mgautierfr: Yes, it is a bug in the old zim files. Recent files uses `X/fulltext/xapian`.

veloman-yunkan (approved):

PR #434, opened 2020-12-15
Build kiwix-lib only if meson is compiled with xapian.

mgautierfr:
Depends of openzim/libzim#465

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -25,7 +25,6 @@ endif
 
 thread_dep = dependency('threads')
 libicu_dep = dependency('icu-i18n', static:static_deps)
-libzim_dep = dependency('libzim', version : '>=6.3.0', static:static_deps)
 pugixml_dep = dependency('pugixml', static:static_deps)
 libcurl_dep = dependency('libcurl', static:static_deps)
 microhttpd_dep = dependency('libmicrohttpd', static:static_deps)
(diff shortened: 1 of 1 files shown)


kelson42:
@mgautierfr Do we really want to do that now? Not convinced this is a good idea as there is no real need to do this now.

meson.build
```diff
@@ -39,6 +38,12 @@ else
   error('Cannot found header mustache.hpp')
 endif
 
+libzim_dep = dependency('libzim', version : '>=6.3.0', static:static_deps)
+if not compiler.has_header_symbol('zim/zim.h', 'LIBZIM_WITH_XAPIAN')
```
veloman-yunkan, line 42: Shouldn't `zim/zim_config.h` be checked instead?

kelson42:
@mgautierfr Any feedback on this?

meson.build
kelson42: @mgautierfr ?

kelson42 (commented):
Would be good to get that merged.

meson.build
mgautierfr: `zim.h` includes `zim_config.h`.
Both header file would work but I prefer to keep using `zim.h` as I tend to consider `zim_config.h` as a bit more internal.

This is really a matter of opinion here. There is no real separation (and documentation) of what can be directly including or not.

kelson42:
@veloman-yunkan `zim.h` being the top header to be include, it sounds logicial to me to rely on it. Would that be ok to merge for you?

veloman-yunkan (approved):

PR #437, opened 2021-01-01
More robust getMediaCount()

### LINKED ISSUE #436: Webp images are not counted as images
So the number given by Kiwix-Serve are wrong, see for example:
image

kelson42:
Fixes #436

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -120,9 +120,9 @@ unsigned int Reader::getMediaCount() const
       = this->parseCounterMetadata();
   unsigned int counter = 0;
 
-  for (auto it = counterMap.begin(); it != counterMap.end(); ++it) {
-    if (it->first.rfind("image/", 0) == 0) {
-      counter += it->second;
+  for (auto &pair:counterMap) {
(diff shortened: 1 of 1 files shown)


src/reader.cpp
```diff
@@ -120,20 +120,12 @@ unsigned int Reader::getMediaCount() const
...
-  auto it = counterMap.find("image/jpeg");
-  if (it != counterMap.end()) {
-    counter += it->second;
-  }
-
-  it = counterMap.find("image/gif");
-  if (it != counterMap.end()) {
-    counter += it->second;
+  for (auto it = counterMap.begin(); it != counterMap.end(); ++it) {
+    if (it->first.rfind("image/", 0) == 0) {
```
veloman-yunkan, line 124: Better use `kiwix::startsWith()` (declared in `include/tools/stringTools.h`)

veloman-yunkan (changes requested):

src/reader.cpp
kelson42: done

veloman-yunkan (approved):

PR #439, opened 2021-01-07
Support 'video/*' * 'audio/*' mimetypes in getMediaCount()

### LINKED ISSUE #438: Handle video/audio in getMediaCount
Following #437, we should also count video, sound and maybe other in the getMediaCount.

kelson42:
Fixes #438


veloman-yunkan (approved):
I overcame the temptation to turn the review process of this simple change into an exhausting marathon. [Life is too short](http://paulgraham.com/vb.html) for that.

PR #440, opened 2021-01-11
Fix taskbar insertion in case of '<head>' attributes

kelson42:
Fixes https://github.com/kiwix/kiwix-tools/issues/412


kelson42:
Too trivial, just merging.

mgautierfr:
Not so trivial, it will break `<head foo="<oups>">`

PR #442, opened 2021-01-18
Better taskbar insertion

kelson42:
Fixes https://github.com/kiwix/kiwix-lib/issues/322

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -1,5 +1,21 @@
-
-
+/*
+ * Copyright 2019 Matthieu Gautier <mgautier@kymeria.fr>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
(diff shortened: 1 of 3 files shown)


src/server/response.cpp
```diff
@@ -203,9 +203,9 @@ void ContentResponse::introduce_taskbar()
   data.set("title", m_bookTitle);
   data.set("withlibrarybutton", m_withLibraryButton);
   auto head_content = render_template(RESOURCE::templates::head_part_html, data);
-  m_content = appendToFirstOccurence(
+  m_content = prependToFirstOccurence(
     m_content,
-    "<head[^>]*>",
+    "</head[^>]*>",
```
veloman-yunkan, line 208: For a closing tag we don't need the **`[^>]*`** in the regular expression (or, at most, we can allow some whitespace there).

veloman-yunkan: Also I think that for this particular case the strings don't need to go through the UTF-8 conversion in both directions, i.e. the regular STL functions would work perfectly fine, but that comment probably belongs to the premature optimization category.

veloman-yunkan (changes requested):

src/server/response.cpp
kelson42: @veloman-yunkan I have kept `[^>]*` as I thought this is just the best approach to handle comments/tabs/spaces. I still think so? Regarding the second part, I have really not much of a clue. Because of that, I would tend to keep it. But I can remove it if you want.

rgaudin (approved):
Not competent to discuss the code. I see you are inserting the taskbar at the end of head instead of at the beginning which was what was discussed in the ticket.

src/server/response.cpp
veloman-yunkan: > I have kept `[^>]*` as I thought this is just the best approach to handle tabs/spaces/.....

That's definitely not the best approach to handle whitespace. That patter will also match `</header>`, `</headache>`. Better explicitly use a whitespace designator. And probably you would also want to allow whitespace *before* `head` too?

kelson42: @veloman-yunkan Hopefully better now?

veloman-yunkan (approved):

PR #445, opened 2021-01-20
Use C++11 std::thread instead of pthread.

mgautierfr:
This is necessary as libzim now does the same and we don't install wrapper pthread library on windows CI.


src/subprocess_unix.cpp
```diff
@@ -11,47 +11,30 @@
...
-  m_mutex(PTHREAD_MUTEX_INITIALIZER),
-  m_waitingThread()
+  m_running(false)
 {
 }
 
 UnixImpl::~UnixImpl()
 {
   kill();
 // Android has no pthread_cancel :(
```
veloman-yunkan, line 21: This comment doesn't make sense anymore.

src/subprocess_unix.cpp
```diff
@@ -11,47 +11,30 @@
...
-  sigaction(SIGUSR1, &actions, NULL);
-#endif
-
   UnixImpl* self = static_cast<UnixImpl*>(_self);
-  waitpid(self->m_pid, NULL, 0);
+  while (true) {
+    if (!waitpid(self->m_pid, NULL, WNOHANG)) {
+      break;
+    }
+    if (self->m_shouldQuit) {
```
veloman-yunkan, line 32: I couldn't find where `m_shouldQuit` is set

src/subprocess_unix.cpp
```diff
@@ -11,47 +11,30 @@
 
 UnixImpl::UnixImpl():
   m_pid(0),
-  m_running(false),
-  m_mutex(PTHREAD_MUTEX_INITIALIZER),
-  m_waitingThread()
+  m_running(false)
 {
 }
```
veloman-yunkan, line 16: `m_shouldQuit` is left uninitialized

veloman-yunkan (changes requested):

src/subprocess_unix.cpp
mgautierfr: Removed.

src/subprocess_unix.cpp
mgautierfr: Added in the destructor.

src/subprocess_unix.cpp
mgautierfr: Done

veloman-yunkan (approved):

Issue #404, opened 2020-08-06
Random should be based off suggestions' pool


rgaudin:
At the moment, the Random feature (`Reader::getRandomPage()`) is the result of a random article id inside the `A` namespace.

https://github.com/kiwix/kiwix-lib/blob/98a10d2ca1b404fca4386d624e6e7f5e342ea416/src/reader.cpp#L211-L232

in the libzim2 branch, @mgautierfr added a check to ensure the retrieved article has a `text/html` mime-type as the random article is taken from the whole articles list, since this branch dropped namespaces.

https://github.com/kiwix/kiwix-lib/blob/86fb0cc4a6a1eac7c72aca46e6560bbe686ce0dd/src/reader.cpp#L168-L187

We believe this is not the appropriate behavior as non-`text/html` content could be welcomed here.

There is no concept of “*this should be included in random*” but we do have a concept of “*this should be navigable to*” and that is the suggestion system which rely on `get_title()` being true which make some semantic sense: if you set a title on an article, you're exposing it.

The random feature should thus offer a random article that matches the suggestion criteria. Whether it's possible to reuse the suggestion methods or not is unknown to me though.

kelson42:
@veloman-yunkan Any opinion on this? Would you know how to make a random feature based on the title index?

veloman-yunkan:
@kelson42 I will look into it

veloman-yunkan:
@kelson42

Unfortunately, currently in the title index the articles without a title participate with their path (url) used instead of the title. As a result articles with real titles are not guaranteed to form a continuous range. Because of this an effective and effiecient approach for randomly picking up a titled article is not possible (without creating an additional index when the ZIM file is opened). The current algorithm can be trivially modified to only return articles with a title, however it may fail to do so within the set limit on the number of attempts (which is now equal to 42).

@mgautierfr Do you plan to change in libzim2 the way the title index is built (so that articles without titles are excluded from it)?

kelson42:
We are again impacted by https://github.com/openzim/libzim/issues/208 here. We need a way to go through the title index without any side effect of the url index.

kelson42:
Actuallyimplementing should be trivial, just computes the number of entries base on the offset of urlpointerlst and titlepointerkst divided by 4. Then find a number in that range, read the 4 bytes entry.

veloman-yunkan:
@kelson42 

> Only articles with titles given during the ZIM writting process will be returned and this is what we want.

[...]

PR #446, opened 2021-01-20
Use the new libzim's getRandomEntry instead of implementing it ourselves.

mgautierfr:
Need openzim/libzim#476 first.


kelson42:
@mgautierfr Does this will fix #404?

veloman-yunkan (approved):

Issue #318, opened 2020-02-01
Meson does not (always) check for mustache.hpp at the right place


kelson42:
On Ubuntu 20.20:

```
# apt-file search mustache.hpp
libkainjow-mustache-dev: /usr/include/kainjow/mustache.hpp
# apt install libkainjow-mustache-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  libkainjow-mustache-dev
0 upgraded, 1 newly installed, 0 to remove and 21 not upgraded.
Need to get 9088 B of archives.
After this operation, 53.2 kB of additional disk space will be used.
Get:1 http://archive.ubuntu.com/ubuntu focal/universe amd64 libkainjow-mustache-dev amd64 4.0+ds-1 [9088 B]
Fetched 9088 B in 0s (167 kB/s)                  
debconf: delaying package configuration, since apt-utils is not installed
Selecting previously unselected package libkainjow-mustache-dev:amd64.
(Reading database ... 18628 files and directories currently installed.)
Preparing to unpack .../libkainjow-mustache-dev_4.0+ds-1_amd64.deb ...
Unpacking libkainjow-mustache-dev:amd64 (4.0+ds-1) ...
Setting up libkainjow-mustache-dev:amd64 (4.0+ds-1) ...
root@461d1d888597:~/kiwix-lib# rm -rf build && meson . build && ninja -C build && ninja -C build install
The Meson build system
Version: 0.52.0
Source dir: /root/kiwix-lib
Build dir: /root/kiwix-lib/build
Build type: native build
Project name: kiwix-lib
Project version: 8.2.2
C++ compiler for the host machine: c++ (gcc 9.2.1 "c++ (Ubuntu 9.2.1-25ubuntu1) 9.2.1 20200123")
C++ linker for the host machine: GNU ld.bfd 2.33.90.20200122
Host machine cpu family: x86_64
Host machine cpu: x86_64
Run-time dependency threads found: YES 
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Run-time dependency icu-i18n found: YES 63.2
Run-time dependency libzim found: YES 6.0.2
Run-time dependency pugixml found: YES 1.9
Run-time dependency libcurl found: YES 7.66.0
Run-time dependency libmicrohttpd found: YES 0.9.66
Has header "mustache.hpp" : NO 

meson.build:29:2: ERROR: Problem encountered: Cannot found header mustache.hpp
```

[...]

PR #459, opened 2021-03-05
Support for book categories in OPDS feed

veloman-yunkan:
Fixes kiwix/kiwix-tools#318

Also contains a couple of other minor fixes:
- fixed month in the `<feed>/<updated>` element in the response to a /catalog/search query
- fixed the date format (removed a trailing colon) in the `<feed>/<entry>/<updated>` elements in responses to catalog requests

Added several unit-tests covering (non-exhaustively) the catalog related functionality of the kiwix server.


src/opds_dumper.cpp
```diff
@@ -77,6 +77,7 @@ pugi::xml_node OPDSDumper::handleBook(Book book, pugi::xml_node root_node) {
   ADD_TEXT_ENTRY(entry_node, "updated", gen_date_from_yyyy_mm_dd(book.getDate()));
   ADD_TEXT_ENTRY(entry_node, "name", book.getName());
   ADD_TEXT_ENTRY(entry_node, "flavour", book.getFlavour());
+  ADD_TEXT_ENTRY(entry_node, "category", book.getCategory());
```
mgautierfr, line 80: Adding the category node in the opds feed is not enough.
We have to read the node in `Book::updateFromOpds` also.
(And we need to do the same with `the library.xml` format (`Book::updateFromXml` and in `libxml_dumper.cpp`)

src/book.cpp
```diff
@@ -220,4 +220,16 @@ bool Book::getTagBool(const std::string& tagName) const {
...
 
+std::string Book::getCategory() const
+{
+  try
+  {
+    return getTagStr("category");
+  }
+  catch ( const std::out_of_range& )
+  {
+    return "unknown";
```
mgautierfr, line 231: Do we really want to have "unknown" ?
Maybe we can use "other", or a empty string, or skip the node ?

src/server/internalServer.cpp
```diff
@@ -692,28 +709,24 @@ std::unique_ptr<Response> InternalServer::handle_catalog(const RequestContext& r
       startIndex = extractFromString<unsigned long>(request.get_argument("start"));
     } catch (...) {}
     try {
-      filter.acceptTags(kiwix::split(request.get_argument("tag"), ";"));
+      tags = kiwix::split(request.get_argument("tag"), ";");
+    } catch (...) {}
+    try {
+      tags.push_back("_category:" + request.get_argument("category"));
```
mgautierfr, line 715: If the book now has a `getCategory`, we should use it (as a new method in `Filter`) instead of reusing the `acceptTags`.

mgautierfr (changes requested):

src/opds_dumper.cpp
mgautierfr: I've just seen this comment : https://github.com/kiwix/kiwix-tools/issues/318#issuecomment-790691534

I'm not particularly agree with it, but let's not implement it (library.xml change) for now (until we agree in the issue discussion).

veloman-yunkan: That is contrary to @kelson42's [reply](https://github.com/kiwix/kiwix-tools/issues/318#issuecomment-790691534) to [my question](https://github.com/kiwix/kiwix-tools/issues/318#issuecomment-790680130)

mgautierfr: We need at least to read the node from opds.
If we don't this is totally useless to add the category in the opds feed.

mgautierfr (changes requested):
The last two commits are good but we need to read the category from the opds node. Even if we don't modify the library.xml.

mgautierfr:
The opds stream is not just a output of the library. It is also a input of our library system.
Our in-memory library can be dump as opdsStream (to exchange through internet) or as a library.xml file (to store on fs).
And we can initialize (load) it with a opdsStream or a library.xml.
If we add a information in one of our serialization format, we MUST extend the loading part to use this information.
(I personally think we should extend both serialization formats in the same time, but it is the discussion we have in #318)

veloman-yunkan:
> If we add a information in one of our serialization format, we MUST extend the loading part to use this information.

Agree. But currently the category information dumped in the `<category>` node simply duplicates the information present in the `<tags>` node. If we assume that we read only what we dump, we don't lose anything. Otherwise we must define how any conflict between `<category>` and `<tags>` must be handled.

mgautierfr:
We want to be able to set a specific category independently of what is zim files.
This way we can recategorize the zim file without recreate it. So `<category>` takes precedence over `<tags>`.
When reading a `library.xml` we will use tags to get the category (until we change `library.xml` format).
When reading opds stream, always use the category node.

veloman-yunkan:
> When reading opds stream, always use the category node.

If category node is missing from the OPDS stream, but the `_category` tag is present then I use the tag value. This is done in order to handle the case of an OPDS stream served by a version of libkiwix not containing this enhancement.

veloman-yunkan:
I have also added support for the `category` attribute in library.xml in a manner similar to OPDS stream - if the new attribute is not present then the value of the `_category` tag (if any) is used.

mgautierfr (changes requested):
Nothing to say on the existing code.

But we need to update the book's category in the `update` methods also (from another book or a reader).

mgautierfr:
We are good. Can you rebase-fixup on master please.

veloman-yunkan:
> We are good. Can you rebase-fixup on master please.

Done

PR #461, opened 2021-03-12
moved autocomplete from head_part.html to taskbar.js

### LINKED ISSUE #444: Put all taskbar code in external resource
Currently with still insert the following code on each page:
```
  var jk = jQuery.noConflict();
  jk(function() {
  jk( "#kiwixsearchbox" ).autocomplete({

  source: "/suggest?content=wikisource_eu_all_nopic_2020-12",
  dataType: "json",
  cache: false,

  select: function(event, ui) {
  jk( "#kiwixsearchbox" ).val(ui.item.value);
  jk( "#kiwixsearchform" ).submit();
  },

  });
  });

  /* cybook hack */
  if (navigator.userAgent.indexOf("bookeen/cybook") != -1) {
  	jk("html").addClass("cybook");
  }
```

We should put everything in `/skin/taskbar.js` or similar.

MananJethwani:
@kelson42 this resolves #444

### DIFF
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -1,5 +1,37 @@
+function htmlDecode(input) {
+    var doc = new DOMParser().parseFromString(input, "text/html");
+    return doc.documentElement.textContent;
+}
 
 (function ($) {
+    const root = $( `link[type='root']` ).attr("href");
+
+    $( "#kiwixsearchbox" ).autocomplete({
+
+        source: `${root}/suggest?content=${window.location.pathname.split(`${root}/`)[1].split('/')[0]}`,
+        dataType: "json",
+        cache: false,
+
+        response: function( event, ui ) {
+
+              for(const item of ui.content) {
+                  item.label = htmlDecode(item.label);
+                  item.value = htmlDecode(item.value);
+              }
+        },
+
+        select: function(event, ui) {
+        $( "#kiwixsearchbox" ).val(ui.item.value);
+        $( "#kiwixsearchform" ).submit();
+        },
+
+    });
+
+    /* cybook hack */
+    if (navigator.userAgent.indexOf("bookeen/cybook") != -1) {
+        $("html").addClass("cybook");
+    }
+
     if ($(window).width() < 520) {
         var didScroll;
         var lastScrollTop = 0;
--- a/static/templates/head_part.html
+++ b/static/templates/head_part.html
@@ -1,39 +1,7 @@
 <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.min.css" rel="Stylesheet" />
 <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.theme.min.css" rel="Stylesheet" />
 <link type="text/css" href="{{root}}/skin/taskbar.css" rel="Stylesheet" />
+<link type="root" href="{{root}}">
 <script type="text/javascript" src="{{root}}/skin/jquery-ui/external/jquery/jquery.js"></script>
 <script type="text/javascript" src="{{root}}/skin/jquery-ui/jquery-ui.min.js"></script>
-<script>
-  function htmlDecode(input) {
-        var doc = new DOMParser().parseFromString(input, "text/html");
-        return doc.documentElement.textContent;
-  }
-  var jk = jQuery.noConflict();
-  jk(function() {
-  jk( "#kiwixsearchbox" ).autocomplete({
-
-  source: "{{root}}/suggest?content={{#urlencoded}}{{{content}}}{{/urlencoded}}",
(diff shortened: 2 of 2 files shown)


static/skin/taskbar.js
```diff
@@ -1,5 +1,35 @@
+function htmlDecode(input) {
+    var doc = new DOMParser().parseFromString(input, "text/html");
+    return doc.documentElement.textContent;
+}
 
 (function ($) {
+    $( "#kiwixsearchbox" ).autocomplete({
+
+        source: `/suggest?content=${window.location.pathname.split('/')[1] || ''}`,
```
veloman-yunkan, line 9: `window.location.pathname.split('/')[1]` assumes an empty root location under which the library is served (search for [`m_root` in src](https://github.com/kiwix/kiwix-lib/search?q=m_root) and for `{{root}}` in `static`).

veloman-yunkan (changes requested):

static/skin/taskbar.js
veloman-yunkan: Also taking the root location into account the suggestion `source` URL should start with `/${root}/suggest?...`

MananJethwani: @veloman-yunkan I get the problem but what solution do you have in mind is it possible to somehow get the value of {{root}} in JS

veloman-yunkan: One hack is to look for the `<link>` elements in the `<head>` section and deduce the root location. A somewhat cleaner solution is to inject a dedicated link pointing to the the root explicitly (something like: `<link type="root" href="rooturl">`.

MananJethwani: @veloman-yunkan is this fine ?
```
source: `${root}/suggest?content=${window.location.pathname.split(`${root}/`)[1] || ''}`
```

veloman-yunkan: Assuming that `root` is not empty, `window.location.pathname.split(`${root}/`)[1]` won't produce a correct value for the book filename (for `http://localhost/myzimlibrary/zimfile/A/welcome.html` it will evaluate to `zimfile/A/welcome.html`)

veloman-yunkan: Test your changes by starting `kiwix-serve` with a non-empty `-r` option, for example:

```
kiwix-serve -p 8080 -r myzimlibrary ...
```

MananJethwani: Sorry for that😅
I have 2 ways in mind rn to get it working correctly

1) keep the script tag we deleted with just this code
```
<script>const content={{#urlencoded}}{{{content}}}{{/urlencoded}}</script>
```
And use it wherever needed.
But, would this destroy the whole reason of moving content to `taskbar.js`?
As we don't want js code in headhtml.

2) change the already applied method to work properly

```
source:  `${root}/suggest?content=${(root ? window.location.pathname.split(`${root}/`)[1].split('/')[0] : window.location.pathname.split(`/`)[1])  || ''}`
```
In this case is it confirm that content won't have any `/` and will be just simple SIM file names?
I will also add encodeURI to this code if you suggest that 
Please let me know what I should do.

veloman-yunkan: I prefer the second option. Yet, doesn't

```
window.location.pathname.split(`${root}/`)[1].split('/')[0]
```
always work (including the case of an empty `root`)?

> In this case is it confirm that content won't have any `/` and will be just simple zim file names?

Looks like so:

https://github.com/kiwix/kiwix-lib/blob/533541cf4592db041c7759a6a8384a054f4a13c0/src/book.cpp#L178-L195

(complemented by the fact that `/` cannot be used in filenames under Windows).

MananJethwani: ok then will use the second method.....I misunderstood your earlier comment 😅

> Assuming that `root` is not empty, `window.location.pathname.split(`${root}/`)[1]` won't produce a correct value for the book filename (for `http://localhost/myzimlibrary/zimfile/A/welcome.html` it will evaluate to `zimfile/A/welcome.html`)

and used ternary operator for root.
will update the PR.
thanks

veloman-yunkan (changes requested):
Please squash all your commits into one and request a review from @mgautierfr too.

MananJethwani:
@veloman-yunkan I don't have permissions to add a reviewer can you please do it?....I will squash all the commits meanwhile.

mgautierfr (approved):
Agree with the code.

I like the idea of adding the `root` link in the page.
I wonder if we could add a js variable for content also. (It is a open idea, we are good with how the pr is).

Just to mention also that content can be empty (if we are in the main page).
It is working as it is so no problem. Just to mention.

veloman-yunkan (approved):

PR #463, opened 2021-03-14
Change Mustache-3.0 to Mustache-4.1

soumyankar:
`ninja -C build` fails on strictly using Mustache-3.0/3.1/3.2 as mentioned. Using Mustache-4.1 resolves the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ libraries need to be available:
 * [Pugixml](https://pugixml.org/) (package `libpugixml-dev` on Ubuntu)
 * [Mustache](https://github.com/kainjow/Mustache) (Just copy the
 header `mustache.hpp` somewhere it can be found by the compiler and/or
-set CPPFLAGS with correct `-I` option). Use Mustache version 3 only.
+set CPPFLAGS with correct `-I` option). Use Mustache version 4.1 or above.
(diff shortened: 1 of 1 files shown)


kelson42:
@soumyankar meson should be able to check this properly

soumyankar:
Meson always replies with `Has header "mustache.hpp" : YES` on using Mustache-3.0/3.1/3.2/4.1
However on running `ninja -C build` the build fails on multiple errors. It only succeeds on using Mustache-4.1.

kelson42:
@soumyankar I know, this is the description of the problem. Meson should be able to check the version of the library and complain properly if too old.

soumyankar:
I suppose so - I am a beginner to software development and I hope I followed the steps correctly. To fulfil the dependency requirements I downloaded [Mustache-3.0](https://github.com/kainjow/Mustache/releases/tag/v3.0) and copied the `mustache.hpp` file into `/usr/local/include`. Are my steps alright or should I have done something differently?

mgautierfr:
The PR itself is ok. I'm merging it. Thanks @soumyankar 

> Meson should be able to check the version of the library and complain properly if too old.

It is not easy to check the mustache version. I've opened a issue on mustache for that (https://github.com/kainjow/Mustache/issues/54)


> Are my steps alright or should I have done something differently?

Your steps ok. Although I would advice to not modify you system include directory.
You may want to use kiwix/kiwix-build to build kiwix-lib (and other project). It will install all dependencies (and compile) for you.

PR #464, opened 2021-03-17
adding kind and path attributes to suggest response object and using it in autocomplete

MananJethwani:
@kelson42 @mgautierfr this resolves kiwix/kiwix-tools#205

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -416,6 +416,8 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
       MustacheData result;
       result.set("label", suggestion[0]);
       result.set("value", suggestion[0]);
+      result.set("kind", "path");
+      result.set("path", suggestion[1]);
       result.set("first", first);
       first = false;
       results.push_back(result);
@@ -428,6 +430,7 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
     MustacheData result;
     result.set("label", "containing '" + term + "'...");
     result.set("value", term + " ");
+    result.set("kind", "pattern");
     result.set("first", first);
     results.push_back(result);
   }
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -5,26 +5,30 @@ function htmlDecode(input) {
 
 (function ($) {
     const root = $( `link[type='root']` ).attr("href");
+    const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
 
     $( "#kiwixsearchbox" ).autocomplete({
 
-        source: `${root}/suggest?content=${window.location.pathname.split(`${root}/`)[1].split('/')[0]}`,
+        source: `${root}/suggest?content=${bookName}`,
         dataType: "json",
         cache: false,
 
         response: function( event, ui ) {
-
               for(const item of ui.content) {
                   item.label = htmlDecode(item.label);
                   item.value = htmlDecode(item.value);
+                  if (item.path) item.path = htmlDecode(item.path);
               }
         },
 
         select: function(event, ui) {
-        $( "#kiwixsearchbox" ).val(ui.item.value);
-        $( "#kiwixsearchform" ).submit();
+            if (ui.item.kind === 'path') {
+                window.location.href = `${root}/${bookName}/${encodeURI(ui.item.path)}`;
+            } else {
+                $( "#kiwixsearchbox" ).val(ui.item.value);
+                $( "#kiwixsearchform" ).submit();
+            }
         },
-
     });
 
     /* cybook hack */
--- a/static/templates/suggestion.json
+++ b/static/templates/suggestion.json
@@ -2,6 +2,10 @@
   {{#suggestions}}{{^first}},{{/first}}
   {
     "value" : "{{value}}",
-    "label" : "{{label}}"
+    "label" : "{{label}}",
+    "kind" : "{{kind}}"
+    {{#path}}
+      , "path" : "{{path}}"
+    {{/path}}
   }{{/suggestions}}
 ]


static/templates/suggestion.json
```diff
@@ -1,7 +1,7 @@
 [
   {{#suggestions}}{{^first}},{{/first}}
   {
-    "value" : "{{value}}",
+    "value" : "{{url}}",
```
veloman-yunkan, line 4: `url` will be HTML escaped/encoded here (in particular `&` will be replaced with `&amp;`). That may seem wrong, but the current design of the full (front-end + back-end) system [assumes](https://github.com/kiwix/kiwix-lib/issues/398#issuecomment-789169478) that `value` is HTML-decoded in the front-end. Please put a comment here making that explicit.

veloman-yunkan (changes requested):
This will break projects depending on kiwix-serve suggestion endpoint. How many of them do we know about? Don't we have to preserve backward compatibility (for example by adding a new endpoint, say 'suggesturl')? I am just raising those questions to make sure that we didn't overlook anything.

mgautierfr:
I've took the wrong direction with this PR.
The workflow with the current code is:
- Do a suggestion search (`http://server.com/suggest?content=<zim>&term=<query>`).
- This return a (json) list of title/value(title) item.
- When user clicks on a suggestion, (or any validation kind of validation of the search), we do a search (`http://server.com/search?content=<zim>&pattern=<query>`)
- The search in the backend start to look if there is a article with the exact title <query> (https://github.com/kiwix/kiwix-lib/blob/master/src/server/internalServer.cpp#L497-L519). If yes it redirect to the" final" url (`http://server.com/<zim>/<article_path>`), if not it return a search page result.

So when the user click on a suggestion, it is redirected to the article.
The problem is that we do extra steps and it is not efficient.

The workflow should be :
- Do a suggestion search (`http://server.com/suggest?content=<zim>&term=<query>`).
- This return a (json) list of title/path item.
- When user clicks on a suggestion, directly go to `http://server.com/<zim>/<path>`
- If user do a real search, do it (`http://server.com/search?content=<zim>&pattern=<query>`)
- The search in the backend do not try to find a exact match and directly generate the search page.

> This will break projects depending on kiwix-serve suggestion endpoint. How many of them do we know about? Don't we have to preserve backward compatibility (for example by adding a new endpoint, say 'suggesturl')? I am just raising those questions to make sure that we didn't overlook anything.

I'm not aware of any projects using it. But @kelson42 or @rgaudin may know some.
If there is no known project I would say that we can change it as we which as it is mainly designed to be internal.
If there is some it would be nice to see with them to not use it, and then start to design a real restfull api projects can use.

src/server/internalServer.cpp
```diff
@@ -427,7 +427,7 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
   if (reader->hasFulltextIndex()) {
     MustacheData result;
     result.set("label", "containing '" + term + "'...");
-    result.set("value", term + " ");
+    result.set("url", "search?content=" + bookName + "&pattern=" + term + "+");
```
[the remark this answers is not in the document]
MananJethwani: @veloman-yunkan right will do that......I should use kiwix::urlEncode for that purpose right?

static/skin/taskbar.js
```diff
@@ -20,9 +20,15 @@ function htmlDecode(input) {
               }
         },
 
+        focus: function(event, ui) {
+            $( "#kiwixsearchbox" ).val(ui.item.label);
+            return false;
```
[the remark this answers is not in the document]
MananJethwani: by default select and focus  try to place `ui.item.value` in the searchbox but we need to place `ui.item.label` as our value now contains a URL, even if I write the line `$( "#kiwixsearchbox" ).val(ui.item.label);` the autocomplete widget will continue with propagating the event ahead and will again change the value to `ui.item.value`.

src/server/internalServer.cpp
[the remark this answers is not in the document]
veloman-yunkan: If `kiwix::parseMimetypeCounter()` cannot be utilized for that purpose and/or you can make `kiwix::urlEncode()` work with minimal tweaks, then yes, go for it.

P.S. :wink:

static/templates/suggestion.json
MananJethwani: what you mean here is I should write `&amp;` in place of `&` in L430 InternalServer.cpp, other than that what comment should be added?
something like `url received should be html encoded`

veloman-yunkan: See http://mustache.github.io/mustache.5.html. The `{{variable}}` syntax applies HTML escaping. For a URL that may seem wrong. The comment must say that this is intentional, since the receiving side in taskbar.js un-escapes this field.

veloman-yunkan (changes requested):

veloman-yunkan:
> You've took the wrong direction with this PR.
> The workflow with the current code is:
> 
>     * Do a suggestion search (`http://server.com/suggest?content=<zim>&term=<query>`).
> 
>     * This return a (json) list of title/value(title) item.
> 
>     * When user clicks on a suggestion, (or any validation kind of validation of the search), we do a search (`http://server.com/search?content=<zim>&pattern=<query>`)
> 
>     * The search in the backend start to look if there is a article with the exact title  (https://github.com/kiwix/kiwix-lib/blob/master/src/server/internalServer.cpp#L497-L519). If yes it redirect to the" final" url (`http://server.com/<zim>/<article_path>`), if not it return a search page result.
> 
> 
> So when the user click on a suggestion, it is redirected to the article.
> The problem is that we do extra steps and it is not efficient.
> 
> The workflow should be :
> 
>     * Do a suggestion search (`http://server.com/suggest?content=<zim>&term=<query>`).
> 
>     * This return a (json) list of title/path item.
> 
>     * When user clicks on a suggestion, directly go to `http://server.com/<zim>/<path>`
> 
>     * If user do a real search, do it (`http://server.com/search?content=<zim>&pattern=<query>`)
> 
>     * The search in the backend do not try to find a exact match and directly generate the search page.
> 

@mgautierfr What is in particular wrong about this PR? As far as I understand it is along the lines outlined in your proposed flow - now selecting a suggestion avoids sending a search request and directly goes to the suggested page.

veloman-yunkan (approved):

veloman-yunkan:
@MananJethwani Please rebase your branch and squash all commits into one.

MananJethwani:
@veloman-yunkan done

kelson42:
> This will break projects depending on kiwix-serve suggestion endpoint. How many of them do we know about? Don't we have to preserve backward compatibility (for example by adding a new endpoint, say 'suggesturl')? I am just raising those questions to make sure that we didn't overlook anything.

@veloman-yunkan It is true that this will break compatibility. We know about a few of third party software dealing with this API. Considering that this should be easy to adapt their code and that basically how it was working before was buggy, I'm ready to break the backward compatibility.

kelson42:
@mgautierfr Thx for the cristal clear explanation about how it works and should work!

MananJethwani:
@mgautierfr so should I just change the response back to `label, value` but this time keep the value as the article name and generate the URL in the frontend would that be fine?

static/skin/taskbar.js
```diff
@@ -5,24 +5,38 @@ function htmlDecode(input) {
...
+        },
+
         select: function(event, ui) {
-        $( "#kiwixsearchbox" ).val(ui.item.value);
-        $( "#kiwixsearchform" ).submit();
+            window.location.href = ui.item.value;
+            $( "#kiwixsearchbox" ).val(ui.item.label);
+            return false;
```
mgautierfr, line 39: We already have the search path "computed" by the form (https://github.com/kiwix/kiwix-lib/blob/master/static/templates/taskbar_part.html#L5-L9).
It would be good to avoid compute it in another place (for now it is compute in `response` function, but as said in another commit, it should be handled here).

If the item is a existing entry path, we must directly change the `window.location.href` and if this is a real search, let's use the form to do it.
(If we need to change the search url, or add a query parameter, we will have to do it in only one place).

static/skin/taskbar.js
```diff
@@ -5,24 +5,38 @@ function htmlDecode(input) {
...
 
         response: function( event, ui ) {
 
               for(const item of ui.content) {
                   item.label = htmlDecode(item.label);
-                  item.value = htmlDecode(item.value);
+                  item.value = encodeURI(htmlDecode(item.value));
+
+                  if (item.label === `containing '${item.value}'...`) {
+                      item.value = `${root}/search?content=${bookName}&pattern=${item.value}+`;
```
mgautierfr, line 23: (This is a open questioning, not a asked change)

The `+`(space) at the end of the request is to force a real search (avoid a redirection in the backend if the title exact matches https://github.com/kiwix/kiwix-lib/blob/master/src/server/internalServer.cpp#L514-L519).

If we remove the redirection in the backend, we don't need to a space at the end of the query.
As for suggestion, it may break some third party user (I don't know about any).
@kelson42 may have a opinion about this but I think we can remove this redirection in the backend and make the `search` endpoint always do a search.

mgautierfr (changes requested):
> @mgautierfr so should I just change the response back to label, value but this time keep the value as the article name and generate the URL in the frontend would that be fine? as that would make the PR follow the separation of concern you talk about.

This is the global idea yes. But see my comments in my review.

static/skin/taskbar.js
```diff
@@ -5,24 +5,38 @@ function htmlDecode(input) {
...
         cache: false,
 
         response: function( event, ui ) {
 
               for(const item of ui.content) {
                   item.label = htmlDecode(item.label);
-                  item.value = htmlDecode(item.value);
+                  item.value = encodeURI(htmlDecode(item.value));
+
+                  if (item.label === `containing '${item.value}'...`) {
```
MananJethwani, line 22: @veloman-yunkan would it be better if I don't place this check for all `item.label` and just check the label of last one if it is of this form then change it's URL?

static/skin/taskbar.js
```diff
@@ -5,18 +5,26 @@ function htmlDecode(input) {
...
```
veloman-yunkan, line 22: Even if this condition holds true for the original values of `item.label` and `item.value`, it may fail to be so for their transformed values.

veloman-yunkan (changes requested):
The title of this PR must be updated

static/skin/taskbar.js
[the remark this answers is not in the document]
MananJethwani: @mgautierfr I am adding another entery named `"isKindPath"` and keeping it true for  one case and false for another, as I can't compare `kind == path` in mustache as it is a logic-less template and I am adding path, pattern as well.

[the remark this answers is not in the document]
mgautierfr: > as I can't compare kind == path in mustache

Why do you need to compare `kind == path` in mustache ?
You need to check the value of `kind` in the js.
On the backend, the kind, path, pattern, value,... is set in the cpp code (https://github.com/kiwix/kiwix-lib/blob/master/src/server/internalServer.cpp#L415-L422)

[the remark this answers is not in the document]
MananJethwani: I am going with the second option as having a 3rd entry in json will stop jqueryUI from rendering any further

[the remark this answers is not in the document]
MananJethwani: got it sorry

static/skin/taskbar.js
MananJethwani: @mgautierfr if we are submitting the from then what is the need of creating another response element named `pattern(or term)` as for that case `ui.item.value` stores the exact same thing

MananJethwani: we should then just create `path` and `kind`

src/server/internalServer.cpp
```diff
@@ -428,6 +430,8 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
     MustacheData result;
     result.set("label", "containing '" + term + "'...");
     result.set("value", term + " ");
+    result.set("kind", "pattern");
+    result.set("path", "");
```
MananJethwani, line 434: needed to add path here as iif path was undefined for some part of suggestion mustache was giving a rendering error
another way was to remove path here and add
```
{{^path}}
      "path" : ""
{{/path}}
```
in `suggestion.json`
other than that If I tried to remove the path from this line and not add above-mentioned part in `suggestion.json` it won't render and errors won't appear

MananJethwani: I tried finding if this is how conditioning works in mustache but couldn't find much, this works but might be a bit wrong 
@mgautierfr @veloman-yunkan can you help me if this seems wrong

static/templates/suggestion.json
```diff
@@ -2,6 +2,10 @@
   {{#suggestions}}{{^first}},{{/first}}
   {
     "value" : "{{value}}",
-    "label" : "{{label}}"
+    "label" : "{{label}}",
+    "kind" : "{{kind}}",
+    {{#path}}
+      "path" : "{{path}}"
+    {{/path}}
```
veloman-yunkan, line 9: For an empty `path` this will result in an invalid JSON (a comma followed by a closing brace, which confuses the strict parsers).

static/skin/taskbar.js
```diff
@@ -5,26 +5,29 @@ function htmlDecode(input) {
...
-        source: `${root}/suggest?content=${window.location.pathname.split(`${root}/`)[1].split('/')[0]}`,
+        source: `${root}/suggest?content=${bookName}`,
         dataType: "json",
         cache: false,
 
         response: function( event, ui ) {
-
               for(const item of ui.content) {
                   item.label = htmlDecode(item.label);
                   item.value = htmlDecode(item.value);
```
veloman-yunkan, line 19: Since `path` is HTML-encoded in `suggestion.json` (that is important for producing a valid JSON in case if `path` contains a double quote symbol), it must be `htmlDecode`-ed here, too.

veloman-yunkan (changes requested):

static/templates/suggestion.json
MananJethwani: so what should I do about it?....I confirm for empty path my current solution works fine

veloman-yunkan: Move the comma

src/server/internalServer.cpp
mgautierfr: The mustache format is documented here : https://mustache.github.io/mustache.5.html

static/templates/suggestion.json
MananJethwani: thanks I understand what you meant now

veloman-yunkan (approved):

mgautierfr:
I approve this PR.
Can you rebase on last `master` before I do the final approval and merge it ?

MananJethwani:
sure

MananJethwani:
@mgautierfr done, should I rebase it as well?

mgautierfr:
> @mgautierfr done, should I rebase it as well?

humm... You've just done it no ? What do you want to rebase more ?

MananJethwani:
sorry I meant squash

mgautierfr:
Yes, you may.
There is no real reason to split this change in different commits.

mgautierfr (approved):

stardiviner:
Is this PR in Docker image now?

kelson42:
@stardiviner No, we will need a new release of the libkiwix and a new release/build of the kiwix-tools (kiwix-serve) first. This will take a few additional weeks.

stardiviner:
Ok, Thanks for answering. @kelson42

Issue #430, opened 2020-12-02
Drop libzim wrapper


mgautierfr:
With the last version of libzim, a lot of things in kiwix-lib become a simple wrapper around the libzim objects.

`kiwix::Reader` is a wrapper of `libzim::Archive`.
`kiwix::Entry` is a wrapper of `libzim::Entry` and `libzim::Item`
`kiwix::Searcher` is a wrapper of `libzim::Search`

They provide almost nothing more than an intermediate step and for the few real differences, they should probably be moved to libzim itself or are relics due to api compatibility.

It would be better to drop all those wrapping and let user code directly use libzim when needed.
We would have a clear distinction between libzim and kiwix-lib :
- libzim is about reading a zim archive, getting its content or searching in it.
- kiwix-lib is about handling zim archive themselves (download them, organise them in a library, ...) and other things (server, bookmark, ...)

Most of the cpp application could be simply being adapted (even without waiting for the wrappers being dropped).
The main question are about the wrapper in another language of kiwix-lib that need to be adapted.
- macos/ios : The wrapping is managed by @automactic, I suppose it would not be to complicated to change.
- android : The wrapping is made in kiwix-lib itself. We should probably move it to a separated repository. Then we may have to wrapper for kiwix-lib and libzim or keep one wrapper wrapping the both library. @macgills probably has his word to say here.

What do you think about this ?

(I was pretty sure to already have opened a issue about this but I cannot find it. If you do, feel free to link both issues).

kelson42:
Sounds a good idea but of course it has to work fine in reality. If good for @mgautierfr and @veloman-yunkan.

macgills:
On the Android side a separate repository to wrap probably makes sense. It was done as it is as a build convenience I think? Having both wrapped is convenient for how the app is currently structured but if it makes more sense to separate them it could be good for project health.
Sidenote: Bookmarks are managed through kiwix-lib? That is not the case in android

kelson42:
@mgautierfr @veloman-yunkan @maneeshpm I believe this is important we do that in libkiwix10. But the strategy to do it and who will do what is unclear to me. Any hint?

mgautierfr:
The plan I see is the following:
1. Stop using the wrappers (`Reader`, `Entry`, `Searcher`, ....) in libkiwix itself. This step does not change the libkiwix API. It make libkiwix use libzim structures internally and provide the wrappers as a API only.
[...]

PR #465, opened 2021-03-17
Add fullsearch link to 404 page if article is not found.

soumyankar:
Fixes [Issue #430](https://github.com/kiwix/kiwix-tools/issues/430) on kiwix-tools.

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -742,6 +742,15 @@ std::string get_book_name(const RequestContext& request)
   }
 }
 
+std::string searchSuggestionHTML(const std::string& searchURL, const std::string& pattern)
+{
+  kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{{searchURL}}}\">{{pattern}}</a>");
+  MustacheData data;
+  data.set("pattern", pattern);
+  data.set("searchURL", searchURL);
+  return (tmpl.render(data));
+}
+
 } // unnamed namespace
 
 std::shared_ptr<Reader>
@@ -765,6 +774,8 @@ InternalServer::build_redirect(const std::string& bookName, const kiwix::Entry& 
 
 std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& request)
 {
+  const std::string url = request.get_url();
+  const std::string pattern = url.substr((url.find_last_of('/'))+1);
   if (m_verbose.load()) {
     printf("** running handle_content\n");
   }
@@ -775,7 +786,10 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?pattern="+pattern; // Make a full search on the entire library.
+    const std::string details = searchSuggestionHTML(searchURL, kiwix::urlDecode(pattern));
+
+    return Response::build_404(*this, request, bookName, details); 
   }
 
   auto urlStr = request.get_url().substr(bookName.size()+1);
@@ -805,7 +819,10 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+pattern; // Make a search on this specific book only.
+    const std::string details = searchSuggestionHTML(searchURL, kiwix::urlDecode(pattern));
+
+    return Response::build_404(*this, request, bookName, details);
   }
 }
 
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -93,10 +93,11 @@ std::unique_ptr<Response> Response::build_304(const InternalServer& server, cons
   return response;
 }
 
-std::unique_ptr<Response> Response::build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName)
+std::unique_ptr<Response> Response::build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName, const std::string& details)
 {
   MustacheData results;
   results.set("url", request.get_full_url());
+  results.set("details", details);
 
   auto response = ContentResponse::build(server, RESOURCE::templates::_404_html, results, "text/html");
   response->set_code(MHD_HTTP_NOT_FOUND);
--- a/src/server/response.h
+++ b/src/server/response.h
@@ -47,7 +47,7 @@ class Response {
 
     static std::unique_ptr<Response> build(const InternalServer& server);
     static std::unique_ptr<Response> build_304(const InternalServer& server, const ETag& etag);
-    static std::unique_ptr<Response> build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName);
+    static std::unique_ptr<Response> build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName, const std::string& details="");
     static std::unique_ptr<Response> build_416(const InternalServer& server, size_t resourceLength);
     static std::unique_ptr<Response> build_500(const InternalServer& server, const std::string& msg);
     static std::unique_ptr<Response> build_redirect(const InternalServer& server, const std::string& redirectUrl);
--- a/static/templates/404.html
+++ b/static/templates/404.html
@@ -9,5 +9,10 @@
     <p>
       The requested URL "{{url}}" was not found on this server.
     </p>
+    {{#details}}
+    <p>
+      {{{details}}}
+    </p>
+    {{/details}}
   </body>
 </html>


src/server/response.cpp
```diff
@@ -97,11 +97,13 @@ std::unique_ptr<Response> Response::build_404(const InternalServer& server, cons
 {
   MustacheData results;
   results.set("url", request.get_full_url());
-
+  // Get URL slug and Name of the book and pass to results
+  std::string url_slug=request.get_full_url().substr((request.get_full_url().find_last_of('/'))+1);
```
mgautierfr, line 101: request already have method to get a part of a url.

mgautierfr (changes requested):
I somehow disagree with the idea of `build_404` trying to be smart.
This method is used in numerous place where it make no sens at all to have a link. (random page, catalog handling, ...).

I would suggest that the `404.html` template contains a extra `{{details}}` token and add a argument to `build_404` to allow the method callers to put the details they want.
This way, `handle_content` could add a link to a search (or main library if it is the bookName which is wrong).
Other methods may add the details they want (if they want).

src/server/response.cpp
soumyankar: are you speaking of `std::string RequestContext::get_url_part(int number)` ? 
I did not use this method because the url path might extend to multiple `'/'s` and I require the slug only. I felt using this simple `find_last_of('/')` was just as easy.

soumyankar:
Understood. It seems as a much better way of doing it. I'll change my code, and also try to incorporate the changes asked in [this conversation](https://github.com/kiwix/kiwix-tools/issues/430).

mgautierfr:
>  I'll change my code, and also try to incorporate the changes asked in this conversation.

Please, do two separated PR.

soumyankar:
> I would suggest that the 404.html template contains a extra {{details}} token and add a argument to build_404 to allow the method callers to put the details they want.

I am not quite sure what you mean by this. If I add another argument to `build_404` I would need to change the `response.h` file, correct? Like so:
`static std::unique_ptr<Response> build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName, const std::string& details);`

And if I do this wouldn't this break a lot of method calls in `internalServer.cpp`? Apologies if I'm wrong.

kelson42:
Adding @veloman-yunkan as a reviewer as @mgautierfr is going to be away for a few days.

static/templates/404.html
```diff
@@ -9,5 +9,8 @@ <h1>Not Found</h1>
     <p>
       The requested URL "{{url}}" was not found on this server.
     </p>
+    <p>
+      Do you want to search for <a href="{{root}}/search?content={{bookName}}&pattern={{url_slug}}">{{url_slug}}</a> in this zim file?
+    </p>
```
veloman-yunkan, line 14: This must now be passed as `{{details}}`. Also, rephrase the sentence like [requested by @kelson42](https://github.com/kiwix/kiwix-tools/issues/430#issuecomment-802651937)

src/server/response.h
```diff
@@ -47,7 +47,7 @@ class Response {
 
     static std::unique_ptr<Response> build(const InternalServer& server);
     static std::unique_ptr<Response> build_304(const InternalServer& server, const ETag& etag);
-    static std::unique_ptr<Response> build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName);
+    static std::unique_ptr<Response> build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName, const std::string& details);
```
veloman-yunkan, line 50: You can specify a default value of an empty string for `details`. Then you won't have to update all call sites of `build_404()`.

veloman-yunkan (changes requested):

src/server/response.h
soumyankar: oh yes, of course! I'll do that.

soumyankar:
Hi @veloman-yunkan, would you please take a look at the changes I've made? Thanks.

static/templates/404.html
```diff
@@ -9,5 +9,10 @@ <h1>Not Found</h1>
     <p>
       The requested URL "{{url}}" was not found on this server.
     </p>
+    {{#details}}
+    <p>
+      Make a full text search for "<a href="{{root}}/{{searchPattern}}">{{details}}</a>"
+    </p>
+    {{/details}}
```
veloman-yunkan, line 16: My understanding of [@mgautierfr's comment](https://github.com/kiwix/kiwix-lib/pull/465#pullrequestreview-614317023) was that all this extra stuff must be passed as `{{details}}`.

veloman-yunkan (changes requested):

static/templates/404.html
soumyankar: @veloman-yunkan Understood. I'll do it as was asked.

veloman-yunkan (approved):

veloman-yunkan:
Before/when merging it makes sense to squash all commits into one.

soumyankar:
So should I close this PR and create a fresh one? with these exact changes?

kelson42:
> So should I close this PR and create a fresh one? with these exact changes?

No, you rebase on `upstream/master` https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa and then you squash all commits together and the you force push them. Make a backup of your repository before.

soumyankar:
Did I do this correctly? These have been my steps:
```bash
git remote add upstream https://github.com/kiwix/kiwix-lib.git
git fetch upstream
git rebase upstream/master
git rebase -i HEAD~4
[pick oldest commit, squash 3 following commits]
[reword the commit message]
git push -f origin master
```
Sorry I haven't done this before so I just wanted to double-check.

veloman-yunkan:
@soumyankar Yes, everything is correct

soumyankar:
@kelson42 would you take a last look at everything so I can continue working on other issues?

kelson42:
@soumyankar Looks good but I made a merge meanwhie (my bad). Would you be able please to rebase a last time? And please give your sole commit a meaningfull comment so we can understand what it does without refering to github.

veloman-yunkan:
@soumyankar You did something wrong while rebasing. As a result your PR has picked up commits from other PR that were merged before it.

soumyankar:
Yes, I feel so too. I'll retry. Somehow the commit message didnt change either.

soumyankar:
Apologies for pushing so many commits. I hope everything is in order this time.

soumyankar:
@kelson42 would you mind taking a look? Thanks.

kelson42:
@veloman-yunkan @mgautierfr Could you please review again?

src/server/internalServer.cpp
```diff
@@ -801,8 +804,9 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
   } catch(kiwix::NoEntry& e) {
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
-
-    return Response::build_404(*this, request, bookName);
+    searchPattern = urlStr.substr((urlStr.find_last_of('/'))+1); //Using an extra variable to not overcomplicate the line below.
+    details = "Make a full text search for <a href=\""+m_root+"/search?content="+bookName+"&pattern="+searchPattern+"\">"+searchPattern+"</a>";
```
veloman-yunkan, line 808: This skipped my attention during a previous review. The second `searchPattern` must be URL-decoded and HTML-escaped.

src/server/internalServer.cpp
```diff
@@ -772,7 +774,8 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    details = "Make a full text search for <a href=\""+m_root+"/search?pattern="+bookName+"\">"+bookName+"</a>";
```
veloman-yunkan, line 777: This skipped my attention during a previous review. The second `bookName` must be URL-decoded and HTML-escaped.

veloman-yunkan (changes requested):

src/server/internalServer.cpp
soumyankar: @veloman-yunkan I know that I can use `kiwix::urlDecode` for URL-decode but I cannot find a relevant method on `stringTools.h` for HTML-escape. What should I do there?

veloman-yunkan: Adding a new function for HTML-escaping should be easy. However, I propose to create a new template for the **"Make a full text search for <a ...> ... \</a>"**, string and delegate the HTML-escaping job to mustache. Then if we need to make that piece of HTML more fancy we will have to edit that template rather than touch the C++ code in two places.

soumyankar:
I just want this little bit brought to attention:
image
The taskbar is set through this line: `response->set_taskbar(bookName, "");`
However, if bookName happens to `""` then the taskbar disappears. Which is what is happening in the screenshot.
In my code, when the bookName doesn't find a valid Reader
```
  if (reader == nullptr) {
    std::string url = request.get_url();
    details = url.substr((url.find_last_of('/'))+1);
    std::string bookName="";
    . . .
```
I make `bookName=""` so that the search link `search?content={{bookName}}&pattern={{details}}` is not rendered useless.
Having said all this, I believe that the taskbar not rendering should be fixed in another issue and is not exactly a bug in this case.

soumyankar:
@veloman-yunkan mind taking a look? Thanks.

static/templates/search_link_404.html
```diff
@@ -0,0 +1,6 @@
+<!-- Printing the search link for the article that wasnt found / a full search for the article. -->
+{{#details}}
+<p>
+	Make a full text search for <a href="{{root}}/search?content={{bookName}}&pattern={{details}}">{{details}}</a>
+</p>
+{{/details}}
```
veloman-yunkan, line 6: Variable names should reflect their meaning. Please rename `details` to `pattern` or `searchPattern`.

static/templates/search_link_404.html
```diff
@@ -0,0 +1,6 @@
+<!-- Printing the search link for the article that wasnt found / a full search for the article. -->
+{{#details}}
+<p>
+	Make a full text search for <a href="{{root}}/search?content={{bookName}}&pattern={{details}}">{{details}}</a>
```
veloman-yunkan, line 4: Variables expanded in the URL must *not* be HTML escaped.

src/server/response.h
```diff
@@ -103,6 +104,7 @@ class ContentResponse : public Response {
     bool m_blockExternalLinks;
     std::string m_bookName;
     std::string m_bookTitle;
+    std::string m_details;
```
veloman-yunkan, line 107: No need to add new data members to `ContentResponse`.

src/server/response.cpp
```diff
@@ -93,15 +93,16 @@ std::unique_ptr<Response> Response::build_304(const InternalServer& server, cons
...
-
+  results.set("details", details);
+  results.set("bookName", bookName);
   auto response = ContentResponse::build(server, RESOURCE::templates::_404_html, results, "text/html");
   response->set_code(MHD_HTTP_NOT_FOUND);
   response->set_taskbar(bookName, "");
-
+  response->set_search_link(bookName, details);
   return std::move(response);
 }
```
veloman-yunkan, line 107: Render the details of a 404 response outside this function and pass `details` as a fragment of HTML that can be embedded in the response as is.

src/server/internalServer.cpp
```diff
@@ -801,8 +806,8 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
   } catch(kiwix::NoEntry& e) {
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
-
-    return Response::build_404(*this, request, bookName);
+    std::string details = urlStr.substr((urlStr.find_last_of('/'))+1); //Using an extra variable to not overcomplicate the line below.
```
veloman-yunkan, line 809: This variable name is confusing

src/server/internalServer.cpp
```diff
@@ -772,7 +774,10 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string url = request.get_url();
+    details = url.substr((url.find_last_of('/'))+1);
```
veloman-yunkan, line 778: Shouldn't this be `searchPattern` instead?

veloman-yunkan (changes requested):

src/server/response.cpp
soumyankar: @veloman-yunkan I did this in a different way, but I think it's similar to what you advised

static/templates/search_link_404.html
```diff
@@ -0,0 +1,6 @@
+<!-- Printing the search link for the article that wasnt found / a full search for the article. -->
+{{#pattern}}
+<p>
+	Make a full text search for <a href="{{root}}/search?content={{{bookName}}}&pattern={{{pattern}}}">{{pattern}}</a>
```
veloman-yunkan, line 4: I think it makes sense to pass the entire search URL as a separate parameter. Then you can avoid specifying the book name when it is not needed.

src/server/response.cpp
```diff
@@ -139,28 +139,28 @@ std::unique_ptr<Response> Response::build_redirect(const InternalServer& server,
...
 };
 
 static ssize_t callback_reader_from_item(void* cls,
-                                  uint64_t pos,
-                                  char* buf,
-                                  size_t max)
+  uint64_t pos,
+  char* buf,
+  size_t max)
 {
```
veloman-yunkan, line 164: Please undo all whitespace changes like these.

veloman-yunkan (changes requested):
> @veloman-yunkan I did this in a different way, but I think it's similar to what you advised 

Please recall why the `Response::build_404()` function got a new `details` parameter (see @mgautierfr's [original review comment](https://github.com/kiwix/kiwix-lib/pull/465#pullrequestreview-614317023)). Generation of 404 page must be able to add arbitrary additional information, not just a suggested search link.

In fact, in my last but one review I expected you to modify your commit 434fd961e2 only in how the contents of `details` variable was produced - instead of building that piece of HTML via C++ string concatenation operations you should have utilized a mustache template.

src/server/response.cpp
```diff
@@ -93,15 +93,15 @@ std::unique_ptr<Response> Response::build_304(const InternalServer& server, cons
...
   MustacheData results;
   results.set("url", request.get_full_url());
-
+  results.set("bookName", bookName);
   auto response = ContentResponse::build(server, RESOURCE::templates::_404_html, results, "text/html");
   response->set_code(MHD_HTTP_NOT_FOUND);
   response->set_taskbar(bookName, "");
-
+  response->set_search_link(bookName, pattern);
```
mgautierfr, line 104: @veloman-yunkan already told you to not modify the `ContentResponse` object but I continue :)

You should simply modify the `_404_html` template to add a details "token/variable" and add the "details" variable to `results`.

mgautierfr (changes requested):
I haven't made a full review but 

First, I agree with @veloman-yunkan review, he somehow already told you how to change this PR
Second, here a small explanation to help you understand the global picture and why it must be done this way : 

The 404 error page is a "global" error page when some url is not found. It can appear at a lot of different places.
Some of them (but not exhaustive) are (see the use of `build_404` in `src/server/internalServer.cpp`):
- When a particular article (path) has not be found in a zim archive (what we want to handle here).
- When a zim archive is not found when user want to get the content of a article (what you try to handle also).
- When a zim archive is not found when user want to .. get a random article, get meta information, do suggestion search.
- Wrong use (call) of capture external.
- Wrong request on the catalog
- Any wrong query 

For now the 404 page is pretty generic (wrong url, thanks, bye).
It is a nice idea to add a search url for the first case, but it make no sens for all other use cases.
But we may in the future add some information for some other cases (what went wrong, here a link to home page, ....)
So we want to build a generic way to add information to 404 page. As those information have nothing in common we have to stick by the most basic way to add information : add text in the html content.

This text (detail) MUST be generated on the user side (side of `build_404`, so in our code anyway) and simply pass to the 404 which will "add" it to the 404 page.

So the search url must be generated in `handle_content` (only if the article has not be found) and the text (a html content, with `a` tag and such) must be passed to the `build_404` method. All other calls to `build_404` must not pass any value for the detail (and `build_404` must not add detail by itself)

soumyankar:
>  instead of building that piece of HTML via C++ string concatenation operations you should have utilized a _mustache template_.

@veloman-yunkan I did not understand that you spoke of a mustache template the first time around. I probably thought too far ahead than was needed. I apologise for dragging out this PR for so long.

soumyankar:
@mgautierfr Thank you for the explanation, I'll keep all of this in mind for the future :)

src/server/internalServer.cpp
```diff
@@ -802,8 +810,15 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("bookName", bookName);
+    data.set("pattern", pattern);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{{root}}}/search?content={{{bookName}}}&pattern={{{pattern}}}\">{{pattern}}</a>");
+    std::string searchPattern = tmpl.render(data);
```
veloman-yunkan, line 818: Now this variable should be named `details`

src/server/response.cpp
```diff
@@ -93,10 +93,11 @@ std::unique_ptr<Response> Response::build_304(const InternalServer& server, cons
   return response;
 }
 
-std::unique_ptr<Response> Response::build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName)
+std::unique_ptr<Response> Response::build_404(const InternalServer& server, const RequestContext& request, const std::string& bookName, const std::string& pattern)
```
veloman-yunkan, line 96: `pattern` → `details`

static/templates/404.html
```diff
@@ -1,13 +1,19 @@
...
+    The requested URL "{{url}}" was not found on this server.
+  </p>
+  <!-- Printing the search link for the article that wasnt found / a full search for the article. -->
+  {{#pattern}}
+  <p>
+    {{{pattern}}}
+  </p>
+  {{/pattern}}
+</body>
 </html>
```
veloman-yunkan, line 19: Please undo the whitespace change

static/templates/404.html
```diff
@@ -1,13 +1,19 @@
...
+  <h1>Not Found</h1>
+  <p>
+    The requested URL "{{url}}" was not found on this server.
+  </p>
+  <!-- Printing the search link for the article that wasnt found / a full search for the article. -->
+  {{#pattern}}
+  <p>
+    {{{pattern}}}
+  </p>
+  {{/pattern}}
```
veloman-yunkan, line 17: `pattern` → `details`

src/server/internalServer.cpp
```diff
@@ -772,7 +774,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("pattern", pattern);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{{root}}}/search?pattern={{{pattern}}}\">{{pattern}}</a>");
+    std::string searchPattern = tmpl.render(data);
```
veloman-yunkan, line 781: Now this variable should be named `details`

src/server/internalServer.cpp
```diff
@@ -802,8 +810,15 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("bookName", bookName);
+    data.set("pattern", pattern);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{{root}}}/search?content={{{bookName}}}&pattern={{{pattern}}}\">{{pattern}}</a>");
```
veloman-yunkan, line 817: At this point `pattern` is URL-encoded (since it was extracted as a fragment of a URL). When generating the HTML, pattern must stay URL-encoded in the search URL, but must be URL-decoded as text to be presented to the user (and we let mustache HTML-encode it).

The best solution may be to compose the search URL via C++ and pass it as a single parameter to mustache. Then you can create and share a mustache template with the other fragment of related code.

src/server/internalServer.cpp
```diff
@@ -772,7 +774,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("pattern", pattern);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{{root}}}/search?pattern={{{pattern}}}\">{{pattern}}</a>");
```
veloman-yunkan, line 780: At this point `pattern` is URL-encoded (since it was extracted as a fragment of a URL). When generating the HTML, pattern must stay URL-encoded in the search URL, but must be URL-decoded as text to be presented to the user (and we let mustache HTML-encode it).

The best solution may be to compose the search URL via C++ and pass it as a single parameter to mustache. Then you can create and share a mustache template with the other fragment of related code.

veloman-yunkan (changes requested):

soumyankar:
@veloman-yunkan I let mustache URL encode and I used `kiwix::urlDecode` to print the text for the user. And I used C++ to build the string as you said.

src/server/internalServer.cpp
```diff
@@ -772,7 +774,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("pattern", pattern);
+    std::string searchURL = m_root+"/search?pattern="+pattern;
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\""+searchURL+"\">"+kiwix::urlDecode(pattern)+"</a>");
+    std::string details = tmpl.render(data);
```
veloman-yunkan, line 782: Such usage of mustache doesn't make sense - is `details` different from the contents of `tmpl`¹? `tmpl` must stay a meaningful mustache template with two parameters - `url` and `pattern` (the former must be expanded as is, while the latter must be HTML-encoded).

¹ Actually it can be, if either `searchURL` or the URL decoded pattern contain `{{m_root}}` or `{{pattern}}`, but I bet this is not why you are using mustache here.

src/server/internalServer.cpp
```diff
@@ -802,8 +811,16 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    MustacheData data;
+    data.set("root", m_root);
+    data.set("bookName", bookName);
+    data.set("pattern", pattern);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+pattern;
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\""+searchURL+"\">"+kiwix::urlDecode(pattern)+"</a>");
+    std::string details = tmpl.render(data);
```
veloman-yunkan, line 820: Same here

veloman-yunkan (changes requested):
Your editor removes the newline symbol on the last line. Please restore it in all modified files.

soumyankar:
@veloman-yunkan Is it correct this time around?

src/server/internalServer.cpp
```diff
@@ -775,7 +777,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?pattern="+pattern;
+    MustacheData data;
+    data.set("pattern", pattern);
```
veloman-yunkan, line 782: URL decode the `pattern` here

src/server/internalServer.cpp
```diff
@@ -775,7 +777,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+    data.set("searchURL", searchURL);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{searchURL}}\">{{{pattern}}}</a>");
```
veloman-yunkan, line 784: The other way around - `searchURL` must be substituted as is, and `pattern` must be HTML encoded.

src/server/internalServer.cpp
```diff
@@ -805,7 +814,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+pattern;
+    MustacheData data;
+    data.set("pattern", pattern);
```
veloman-yunkan, line 819: URL decode the `pattern` here

src/server/internalServer.cpp
```diff
@@ -805,7 +814,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+    data.set("searchURL", searchURL);
+    kainjow::mustache::mustache tmpl("Make a full text search for <a href=\"{{searchURL}}\">{{{pattern}}}</a>");
```
veloman-yunkan, line 821: This template must be shared with the one above (remove the literal duplication).

veloman-yunkan (changes requested):

soumyankar:
@veloman-yunkan Made the changes.

src/server/internalServer.cpp
```diff
@@ -775,7 +778,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?pattern="+kiwix::urlDecode(pattern);
```
veloman-yunkan, line 781: In the URL the pattern must stay URL-encoded.

src/server/internalServer.cpp
veloman-yunkan: URL-decode the pattern on this line:

```c++
data.set("pattern", pattern);
```

src/server/internalServer.cpp
veloman-yunkan: URL-decode the pattern on this line:

```c++
data.set("pattern", pattern);
```

src/server/internalServer.cpp
```diff
@@ -805,7 +814,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+kiwix::urlDecode(pattern);
```
veloman-yunkan, line 817: In the URL the pattern must stay URL-encoded.

src/server/internalServer.cpp
```diff
@@ -775,7 +778,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?pattern="+kiwix::urlDecode(pattern);
+    MustacheData data;
+    data.set("pattern", pattern);
+    data.set("searchURL", searchURL);
+    std::string details = tmpl.render(data);
+
+    return Response::build_404(*this, request, bookName, details); // Make a full search on the entire library.
```
veloman-yunkan, line 787: Remove the comment or move it to a location where it applies better

src/server/internalServer.cpp
```diff
@@ -805,7 +814,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+kiwix::urlDecode(pattern);
+    MustacheData data;
+    data.set("pattern", pattern);
+    data.set("searchURL", searchURL);
+    std::string details = tmpl.render(data);
+    
+    return Response::build_404(*this, request, bookName, details); // Make a search on this specific book only.
```
veloman-yunkan, line 823: Remove the comment or move it to a location where it applies better

veloman-yunkan (changes requested):
Please read the review comments more attentively and don't address them mechanically. Expect one more iteration after completing this one (I omitted some feedback in order to avoid more confusion).

src/server/internalServer.cpp
```diff
@@ -775,7 +778,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   const std::shared_ptr<Reader> reader = get_reader(bookName);
   if (reader == nullptr) {
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?pattern="+pattern; // Make a full search on the entire library.
+    MustacheData data;
+    data.set("pattern", kiwix::urlDecode(pattern));
+    data.set("searchURL", searchURL);
+    std::string details = tmpl.render(data);
```
veloman-yunkan, line 785: Convert this piece of code to a small helper function, so that it is replaced with a single line:

```c++
const std::string details = searchSuggestionHTML(searchURL, kiwix::urlDecode(pattern));
```

Declare/define the `searchSuggestionHTML()` function in an unnamed namespace. If you can think of a better name for it, go ahead and do it.

src/server/internalServer.cpp
```diff
@@ -805,7 +814,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    return Response::build_404(*this, request, bookName);
+    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+pattern; // Make a search on this specific book only.
+    MustacheData data;
+    data.set("pattern", kiwix::urlDecode(pattern));
+    data.set("searchURL", searchURL);
+    std::string details = tmpl.render(data);
```
veloman-yunkan, line 821: Reuse the `searchSuggestionHTML()` (or whatever you name it) function here.

veloman-yunkan (changes requested):
This pretty much completes my feedback. A suggestion for a small change may follow, but the light in the end of the tunnel is so bright and warm that it may evaporate before I decide to publish it. :smiley:

soumyankar:
@veloman-yunkan I really hope that I did it correctly this time!

veloman-yunkan (approved):
Thanks for bearing with me this long. I suggest that you review the entire history of this PR and distill those things that can safely be carried to other programming tasks.

soumyankar:
@veloman-yunkan I should be the one thanking you for being so patient with me. I have absorbed a lot of knowledge from this PR. Thank you

mgautierfr:
Thanks to both of you @soumyankar and @veloman-yunkan for this long review.

@soumyankar, can you rebase on master ? I will merge it then.

mgautierfr (approved):

PR #472, opened 2021-03-22
added root functionality for block external link feature

### LINKED ISSUE #407: Block external link feature doesn't work with root prefix.
The "block external link" feature doesn't take into account the root prefix at all.

### How to reproduce

Run `kiwix-serve -p 8000 foo.zim -b -r ROOT` then go to `http://localhost:8000/ROOT/foo` and click on a external link.

MananJethwani:
@kelson42 this resolves #407 
I have one more point currently `<link type="root" href="{{root}}">` is placed in head_part.html which is rendered only if `m_withTaskbar` is true,
it should be moved to `index.html` according to me so should I do that as well in this PR?

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -221,9 +221,9 @@ void ContentResponse::inject_externallinks_blocker()
   kainjow::mustache::data data;
   data.set("root", m_root);
   auto script_tag = render_template(RESOURCE::templates::external_blocker_part_html, data);
-  m_content = appendToFirstOccurence(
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
> currently `<link type="root" href="{{root}}">` is placed in head_part.html which is rendered only if `m_withTaskbar` is true,
> it should be moved to `index.html` according to me so should I do that as well in this PR?

`index.html` is used only for the homepage. Let's pursue that fix in a separate PR.

veloman-yunkan (approved):

Issue #322, opened 2020-02-07
taskbar should be inserted after charset definition


rgaudin:
The kiwix-serve *taskbar* adds a bunch of content to the beginning of the `<head>` element rendering it **difficult** to place the `<meta charset />` element within the first 1024 characters of the page.

According to [the spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta), the charset element must be **inside the `<head />`** and **within the first 1024 bytes** of the content.

Given the following, **custom-crafted** HTML content

``` html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>My title</title>
  </head>
```

The library with the taskbar enabled would serve:

``` html
<!DOCTYPE html>
<html>
  <head><link type="text/css" href="/skin/jquery-ui/jquery-ui.min.css" rel="Stylesheet" />
<link type="text/css" href="/skin/jquery-ui/jquery-ui.theme.min.css" rel="Stylesheet" />
<link type="text/css" href="/skin/taskbar.css" rel="Stylesheet" />
<script type="text/javascript" src="/skin/jquery-ui/external/jquery/jquery.js"></script>
<script type="text/javascript" src="/skin/jquery-ui/jquery-ui.min.js"></script>
<script>
  var jk = jQuery.noConflict();
  jk(function() {
  jk( "#kiwixsearchbox" ).autocomplete({

  source: "/suggest?content=la_chaine_de_maths_et_tiques_fr_all_2020-02",
  dataType: "json",
  cache: false,

  select: function(event, ui) {
  jk( "#kiwixsearchbox" ).val(ui.item.value);
  jk( "#kiwixsearchform" ).submit();
  },

  });
  });

  /* cybook hack */
  if (navigator.userAgent.indexOf("bookeen/cybook") != -1) {
    jk("html").addClass("cybook");
  }
</script>
<script type="text/javascript" src="/skin/taskbar.js" async></script>
    <meta charset="utf-8">
    <title>My title</title>
  </head>
```
The taskbar payload is 984 bytes alone, leaving almost nothing and variable according to the ZIM name.

It's OK on this optimized html example but in many cases, the charset would not make it to the first 1024b, forcing the browser to restart rendering after eventually encountering a charset info. This is visible in the logs with Firefox

``` 
[...]

PR #474, opened 2021-03-23
injecting root link directly and renamed head_part to head_taskbar

### LINKED ISSUE #473: <link type="root" ...> must be injected even if the taskbar is disabled
`<link type="root" href="{{root}}">` is placed in head_part.html which is rendered only if `m_withTaskbar` is true,
it should be moved to `index.html`.

MananJethwani:
@kelson42 @veloman-yunkan this resolves #473

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -202,7 +202,7 @@ void ContentResponse::introduce_taskbar()
   data.set("hascontent", !m_bookName.empty());
   data.set("title", m_bookTitle);
   data.set("withlibrarybutton", m_withLibraryButton);
-  auto head_content = render_template(RESOURCE::templates::head_part_html, data);
+  auto head_content = render_template(RESOURCE::templates::head_taskbar_html, data);
   m_content = prependToFirstOccurence(
     m_content,
     "</head[ \\t]*>",
@@ -225,6 +225,13 @@ void ContentResponse::inject_externallinks_blocker()
     m_content,
     "</head[ \\t]*>",
     script_tag);
+}
+
+void ContentResponse::inject_root_link(){
+  m_content = prependToFirstOccurence(
+    m_content,
+    "<head[ \\t]*>",
+    "<link type=\"root\" href=\"" + m_root + "\">");
 }
 
 bool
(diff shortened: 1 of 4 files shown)


src/server/response.cpp
```diff
@@ -227,6 +227,13 @@ void ContentResponse::inject_externallinks_blocker()
     script_tag);
 }
 
+void ContentResponse::inject_root_link(){
+  m_content = prependToFirstOccurence(
```
veloman-yunkan, line 231: Shouldn't this be `appendToFirstOccurence()`?

veloman-yunkan (changes requested):
Please split the single commit into two

src/server/response.cpp
MananJethwani: Prepend is actually adding it just after the head tag so that any module can use it and we don't have to move it up again if some other module needs it which is above this link.

kelson42: @MananJethwani "append" is add after and "prepend" is add before. If it does not work like this, this is a bug. Here you add a `<link>` node before the `<head>` node, which is wrong. I wonder if you have tested accurately the result! Actually this seems double wrong, it should be prepend to the `</head>` tag, for the same reason as for #322.

kelson42 (changes requested):

kelson42:
@MananJethwani @veloman-yunkan I wrongly merged this PR.
@MananJethwani Would you be able to ASAP create a new PR fixing the problems reported?
@veloman-yunkan You had other problems to report in the PR?

src/server/response.cpp
MananJethwani: I confirm that it was working fine, it was adding link to start of head(just after head tag before all other content of head tag) I was as well a bit confused but thought that's how it is supposed to work as a similar function is present in [jQuery(named prepend)] (https://api.jquery.com/prepend/#:~:text=prepend(%20function%20)&text=A%20function%20that%20returns%20an,of%20the%20element%20as%20arguments.)

That works exactly the same as this one is working.

veloman-yunkan: `prependToFirstOccurence()` works differently from the jQuery `prepend()`. `appendToFirstOccurence()`/`prependToFirstOccurence()` know nothing about HTML and don't operate on HTML nodes/elements. They (should) insert the specified string relative to the first occurrence of the specified regex.

PR #483, opened 2021-04-09
fixed suggestion system

MananJethwani:
@kelson42 this fixes https://github.com/kiwix/kiwix-tools/issues/446 but will revert back #322 so please draft it for discussion for now so that better suggestions could be discussed

### DIFF
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -84,5 +84,5 @@ jq(document).ready(() => {
                 $('label[for="kiwix_button_show_toggle"], .kiwix_button_cont').removeClass('searching');
             }
         });
-    })(jq);
+    })(jq.fn.jquery === '1.10.2' ? jq : $);
 })
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -3,83 +3,86 @@ function htmlDecode(input) {
     return doc.documentElement.textContent;
 }
 
-(function ($) {
-    const root = $( `link[type='root']` ).attr("href");
-    const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
-
-    $( "#kiwixsearchbox" ).autocomplete({
-
-        source: `${root}/suggest?content=${bookName}`,
-        dataType: "json",
-        cache: false,
-
-        response: function( event, ui ) {
-              for(const item of ui.content) {
-                  item.label = htmlDecode(item.label);
-                  item.value = htmlDecode(item.value);
-                  if (item.path) item.path = htmlDecode(item.path);
-              }
-        },
-
-        select: function(event, ui) {
-            if (ui.item.kind === 'path') {
-                window.location.href = `${root}/${bookName}/${encodeURI(ui.item.path)}`;
-            } else {
-                $( "#kiwixsearchbox" ).val(ui.item.value);
-                $( "#kiwixsearchform" ).submit();
-            }
-        },
-    });
-
-    /* cybook hack */
-    if (navigator.userAgent.indexOf("bookeen/cybook") != -1) {
-        $("html").addClass("cybook");
-    }
-
-    if ($(window).width() < 520) {
-        var didScroll;
-        var lastScrollTop = 0;
-        var delta = 5;
-        // on scroll, let the interval function know the user has scrolled
-        $(window).scroll(function (event) {
-            didScroll = true;
+const jq = jQuery.noConflict();
+jq(document).ready(() => {
+    (function ($) {
+        const root = $( `link[type='root']` ).attr("href");
+        const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+    
+        $( "#kiwixsearchbox" ).autocomplete({
+    
+            source: `${root}/suggest?content=${bookName}`,
+            dataType: "json",
+            cache: false,
+    
+            response: function( event, ui ) {
+                  for(const item of ui.content) {
+                      item.label = htmlDecode(item.label);
+                      item.value = htmlDecode(item.value);
+                      if (item.path) item.path = htmlDecode(item.path);
+                  }
+            },
+    
+            select: function(event, ui) {
+                if (ui.item.kind === 'path') {
+                    window.location.href = `${root}/${bookName}/${encodeURI(ui.item.path)}`;
+                } else {
+                    $( "#kiwixsearchbox" ).val(ui.item.value);
+                    $( "#kiwixsearchform" ).submit();
+                }
+            },
         });
-        // run hasScrolled() and reset didScroll status
-        setInterval(function () {
-            if (didScroll) {
-                hasScrolled();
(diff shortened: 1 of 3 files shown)


kelson42:
@MananJethwani Why all the javscript is changed? Have you simply made a revert?

MananJethwani:
@kelson42 no I haven't made a revert, actually it is changed as I have added a `$(document).ready()` and moved most JS part inside it as if we load it early, autocomplete failed sometime because searchbox was not loaded that early.

MananJethwani:
@kelson42 somehow `inserted_style.css` from mwoffliner static files interfere with jquery.ui so we need to keep jquery.ui above that particular file(I couldn't figure what was wrong exactly in inserted_style.css at the time but removing that file made suggest work properly I checked by removing that file while scrapping using mwoffliner). But, by moving these files to the line right after the `<head>` tag the autocomplete setup couldn't find the `#kiwixsearchbox` element as it wasn't loaded at that point.

my solution to the problem was to move it right after opening the `<head>` tag and adding a `jquery(document).ready()` function which is used to check if all the required components inside the ready function are loaded or not also even if there exist another solution and it is not necessary to add  `jquery(document).ready()` to solve this particular case,  it is then also considered safe to call any jquery related function inside `jquery(document).ready()` as mentioned [here](https://learn.jquery.com/using-jquery-core/document-ready/).

as you have suggested we should use `<meta>` tag and append these scripts right after it and if not present append right after `<head>` tag is opened.
so should I proceed or you have a better solution in mind?

mgautierfr:
The issue discribing the problem is pretty small.
Do you have a zim/article for which the suggestion system is broken ?

If somehow the content css (`inserted_style.css` is a content from the libzim/kiwix-lib pov) is conflicting whit what we insert, we must understand why before doing fixes.
(This makes me think to https://github.com/kiwix/kiwix-lib/issues/394)

> we need to keep jquery.ui above that particular file

What do you mean by "keep above" ? 

> so please draft it for discussion for now so that better suggestions could be discussed

I'me moving the PR as draft.

MananJethwani:
@mgautierfr by
>  we need to keep jquery.ui above that particular file

I mean we should call the script tags from kiwix-lib(including jquery-ui one) before the link tags for css files added by mwoffliner

about how inseted_style.css in conflicting with jquery I will have a proper look and let you know.

mgautierfr:
> I mean we should call the script tags from kiwix-lib(including jquery-ui one) before the link tags for css files(added by mwoffliner).
>
> also about how inseted_style.css is conflicting with jquery I will have a proper look and let you know, sorry for incomplete analysis.

It is surprising that css file conflict with js code.
The `inserted_style.css` may "change" the style of some completion items (popup, ...) and so break the ui.
But it is surprising that the order change that. Either they all time conflict or not.

MananJethwani:
sorry I was the wrong problem seems to be something else at this point I believe autocomplete is being called even before it is loaded but I guess the problem could be something else because even after waiting for 3-4 mins on a page autocomplete is still not loaded

MananJethwani:
Ok so I finally figured out what the exact problem is sometimes the jQuery added by us is being loaded first or sometimes jQuery added by the zim file is being loaded first using `jQuery.noConflict()` therefore assigns sometimes `kiwix-lib` jQuery to `jq` otherwise the one loaded by the zim file. and jQuery-ui is attached to the one which is loaded first.

so @mgautierfr what should be done?

static/skin/taskbar.js
```diff
@@ -3,83 +3,86 @@ function htmlDecode(input) {
...
+            focus: function () {
+                $('.kiwix_searchform').addClass('full_width');
+                $('label[for="kiwix_button_show_toggle"], .kiwix_button_cont').addClass('searching');
+            },
+            blur: function () {
+                $('.kiwix_searchform').removeClass('full_width');
+                $('label[for="kiwix_button_show_toggle"], .kiwix_button_cont').removeClass('searching');
+            }
+        });
+    })(jq.fn.jquery === '1.10.2' ? jq : $);
```
MananJethwani, line 87: @mgautierfr @kelson42 this fixes the issue but we would then need to keep updating the line if we update the jQuery version we are using to check to which variable kiwix-lib's `jQuery` and `jQuery-ui` are attached.

static/skin/taskbar.js
```diff
@@ -3,83 +3,86 @@ function htmlDecode(input) {
...
-    }
-
-    if ($(window).width() < 520) {
-        var didScroll;
-        var lastScrollTop = 0;
-        var delta = 5;
-        // on scroll, let the interval function know the user has scrolled
-        $(window).scroll(function (event) {
-            didScroll = true;
+const jq = jQuery.noConflict();
```
MananJethwani, line 6: @mgautierfr I have already used `jQuery.noConflict` to remove conflict between 2 jquery but the problem is `jQuery.noConflict` captures the jquery that is loaded first so if the jQuery we add is loaded first it assigns that to `jq` but, if jQuery they use(zim file) is loaded first then `jq` has their version of jQuery stored in it.

I suppose using `iFrame` would be the best course of action.

kelson42:
@MananJethwani This PR is still a draft? I still find suspicious that so many lines have been changed for such a small thing to fix.

kelson42:
@MananJethwani AFAIK, this PR does not secure which version of JQuery is loaded first, as a consequence we don't secure that the content is loaded with its version of JQuery right?

MananJethwani:
@kelson42 yes, this PR couldn't figure out which version of jQuery was being loaded first and we needed `jQuery.noConflict` to remove conflict as 2 jquery versions are being loaded but as above mentioned we don't know which one is loaded first and is assigned by `jQuery.noConflict` to `jq` but, we know the version of `jQuery` we have added so I am using it to determine which version of `jQuery` is loaded in `jq` and assigning correct `jQuery` to function so that autocomplete can be loaded.

kelson42:
what about forcing the taskbar j's loading order with https://www.w3schools.com/tags/att_script_defer.asp ?

MananJethwani:
@kelson42 from your last comment I was able to figure out the actual problem there was an async attribute added to the `taskbar.js` script tag which made it load earlier than other jquery files added by kiwix-lib sometimes and caused the problem we were facing.
I have removed that attribute and kept jQuery.noConflict as still the jquery we load should not conflict with jquery added by the zimfile.
also whole `taskbar.js` is still under a `jq(document).ready()` as all jquery operation should always occur after all elements required by jQuery are completly loaded.

mgautierfr:
On solution would be to stop using jquery to do the completion. https://tarekraafat.github.io/autoComplete.js seems to be a viable completion library which not uses jquery.

MananJethwani:
@mgautierfr I agree jQuery is a legacy module that is no longer used mostly and everything it does can be achieved by vanilla js.
for now my other PR fixes the issue I will try to remove jquery part in my GSOC project if I get selected(we can open a separate ticket for that purpose).

PR #485, opened 2021-04-09
Library::removeBookById() updates the reader list and the search DB

### LINKED ISSUE #478: kiwix::Library::removeBookById() doesn't update m_readers and m_bookDB
https://github.com/kiwix/kiwix-lib/blob/e214efe/src/library.cpp#L96-L99

veloman-yunkan:
Fixes #478 

I added a new test-suite `XmlLibraryTest` so that I could make use of the `Library::getReaderById()` method (a library constructed from an OPDS stream doesn't contain local books).

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -106,6 +106,7 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
 
 bool Library::removeBookById(const std::string& id)
 {
+  m_bookDB->delete_document("Q" + id);
   m_readers.erase(id);
   return m_books.erase(id) == 1;
 }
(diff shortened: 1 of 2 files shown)


src/library.cpp
```diff
@@ -331,7 +331,7 @@ Library::BookIdCollection Library::filter(const Filter& filter)
 {
   BookIdCollection result;
   for(auto id : getBooksByTitleOrDescription(filter)) {
-    if(filter.acceptByNonQueryCriteria(m_books[id])) {
```
veloman-yunkan, line 333: This bug was introduced in #460 because `Library::filter()` [is not declared `const`](https://github.com/kiwix/kiwix-lib/issues/487).

mgautierfr (approved):

PR #488, opened 2021-04-11
Fully xapian powered catalog search

### LINKED ISSUE #484: [library] Use Xapian for library search on all fields
#460 provided a fix for #106, targeting (as requested) only the title/description fields.

Now it's time to use Xapian for search over all fields of the library entries.

veloman-yunkan:
Fixes #484 with the exception of the `local`, `remote`, `valid` and `maxSize` filtering criteria. Those can be processed via Xapian too, but is it worth the effort (though we will gain some uniformity in how a query is fulfilled - no postprocessing through `Filter::accept()` will be needed)? 

Also
- adds diacritics insensitivity to catalog filtering by title/description
- the unit-test of `Library::filter()` is greatly enhanced

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -596,11 +596,6 @@ bool Filter::hasQuery() const
   return ACTIVE(QUERY);
 }
 
-bool Filter::accept(const Book& book) const
-{
-  return acceptByNonQueryCriteria(book) && acceptByQueryOnly(book);
-}
-
 bool Filter::acceptByNonQueryCriteria(const Book& book) const
 {
   auto local = !book.getPath().empty();
@@ -645,17 +640,6 @@ bool Filter::acceptByNonQueryCriteria(const Book& book) const
     }
   }
   return true;
-}
-
-bool Filter::acceptByQueryOnly(const Book& book) const
-{
-  if ( ACTIVE(QUERY)
-    && !(matchRegex(book.getTitle(), "\\Q" + _query + "\\E")
-        || matchRegex(book.getDescription(), "\\Q" + _query + "\\E")))
-    return false;
-
-  return true;
-
 }
 
 }
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -279,14 +279,17 @@ void Library::updateBookDB(const Book& book)
   const std::string title = normalizeText(book.getTitle(), lang);
   const std::string desc = normalizeText(book.getDescription(), lang);
   const std::string name = book.getName(); // this is supposed to be normalized
+  const std::string category = book.getCategory(); // this is supposed to be normalized
   doc.add_value(0, title);
   doc.add_value(1, desc);
   doc.add_value(2, name);
+  doc.add_value(3, category);
   doc.set_data(book.getId());
 
   indexer.index_text(title, 1, "S");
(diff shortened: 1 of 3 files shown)


veloman-yunkan:
Incorporated @maneeshpm's fix of openzim/libzim#534 into the commits titled "Handling of non-words in publisher query" and "Catalog filtering by creator works via Xapian" (respectively commits 7bfd050 and 109e54d in this revision of the PR).

kelson42:
@mgautierfr It is important this PR is reviewed so @veloman-yunkan can move on.

test/library.cpp
```diff
@@ -486,6 +486,14 @@ TEST_F(LibraryTest, filterByName)
   EXPECT_FILTER_RESULTS(kiwix::Filter().name("wikibooks_de"),
     "An example ZIM archive"
   );
+
+  EXPECT_FILTER_RESULTS(kiwix::Filter().query("name:wikibooks_de"),
+    "An example ZIM archive"
+  );
```
mgautierfr: How is it working with `kiwix::Filter().name("foo").query("name:bar")` ?

src/library.cpp
```diff
@@ -276,14 +276,33 @@ void Library::updateBookDB(const Book& book)
...
+  const std::string creator = normalizeText(book.getCreator());
+  const std::string tags = normalizeText(book.getTags());
   doc.add_value(0, title);
   doc.add_value(1, desc);
+  doc.add_value(2, name);
+  doc.add_value(3, category);
+  doc.add_value(4, lang);
+  doc.add_value(5, publisher);
+  doc.add_value(6, creator);
+  doc.add_value(7, tags);
```
mgautierfr, line 293: Do we need to store those values ?
It seems we only need the bookId stored in the data.

mgautierfr (commented):
Not a lot to say, code is good.
Just two questions.

test/library.cpp
veloman-yunkan: I don't think we have to define the behaviour for this case (rather we can define the result to be unspecified).

src/library.cpp
veloman-yunkan: Now we don't need them.

veloman-yunkan:
> Not a lot to say, code is good.
> Just two questions.

And what about the question raised in the description of the PR?

test/library.cpp
mgautierfr: Ok, we will see what to do if somehow we have this case. But as you, I doubt we have to.

src/library.cpp
mgautierfr: Can you remove them so ?

veloman-yunkan: Done

mgautierfr (approved):

mgautierfr:
Please rebase on top of master (and merge yourself if I took too much time to come back on this PR)

veloman-yunkan:
> @veloman-yunkan Please rebase on top of master (and merge yourself if I took too much time to come back on this PR)

@mgautierfr I don't have enough privileges in this repository

mgautierfr:
> I don't have enough privileges in this repository

Wow, I will see with to change that.

Issue #209, opened 2019-03-20
Use OPDS Feed with partial entries


mgautierfr:
For now, we have only one OPDS feed : A "Acquisition Feeds" (http://library.kiwix.org/catalog/root.xml)
This is what we could call the "Complete Acquisition Feed" (See https://specs.opds.io/opds-1.1.html#Complete_Acquisition_Feeds)
The search API (http://library.kiwix.org/catalog/search) is "just" a way to filter the main feed.

The main feed is about 1.5Mb for now.
This is less that ("library_zim.xml) (7.5Mb) because we do not include the icons but link to the icon url.
However, this is still a bit big especially for people with small internet bandwidth.

OPDS Spec allow to reduce this size, at the price of a bit more complexity.
It is possible to have a feed with partial entries instead of complete entries (https://specs.opds.io/opds-1.1.html#Partial_and_Complete_Catalog_Entries)

The feed would list only the entries' id and a link to download the rest of the metadata. The catalog would have a new API endpoints (something like "http://library.kiwix.org/catalog/entry/<entry_id>") to get other metadata (title, description, ...)
The whole partial feed (all entries, but partial content) would be around 427Ko.

A client would have to do more request to get the whole content, but the size would be greatly reduced.

427Ko is  for all entries. A partial catalog for English zim file only would be 60Ko only (instead of 232Ko). For french it would be 11Ko (instead of 40Ko)

automactic:
> The feed would list only the entries' id and a link to download the rest of the metadata.

Oh, please don't. A list API that only returns the IDs would be virtually useless, as no user facing interface would show user these UUIDs. The essentially info for helping user decide which ones are of interest and are worthy to dive into should be included in the list API. Otherwise, the consumer of the API would essentially need to do 1+N API call.

automactic:
We can create a API that returns all IDs in the online library, so that apps that decide to keep a local snapshot of the online library can quickly figure out the diff. But I think we should still have an API that gives the consumer all book and their full meta data (it can be paginated).

kelson42:
@mgautierfr Can you please:
1 - Update ticket and ideally close it
2 - Split it in precise topics, this ticket seems far too generix
3 - Answer to @automactic

mgautierfr:
> 1- Update ticket and ideally close it.

There is nothing to update nor close.

> 2 - Split it in precise topics, this ticket seems far too generix

[...]

PR #492, opened 2021-04-15
Opds categories feed

### LINKED ISSUE #518: OPDS endpoint for categories
In order to improve ZIM files filtering on readers, we need to maintain a centralized list of *Categories* to which ZIM files are attached to. This has been mentionned in [kiwix-lib#130](https://github.com/kiwix/kiwix-lib/issues/130) and [kiwix-desktop#193](https://github.com/kiwix/kiwix-desktop/issues/193).

The Categories list will start with the followings and should support internationalizations (`en` translations bellow).

* `gutenberg`: `Gutenberg`
* `phet`: `PhET`
* `psiram`: `Psiram`
* `stack_exchange`: `Stack Exchange`
* `ted`: `TED`
* `vikidia`: `Vikidia`
* `wikibooks`: `Wikibooks`
* `wikinews`: `Wikinews`
* `wikipedia`: `Wikipedia`
* `wikiquote`: `Wikiquote`
* `wikisource`: `Wikisource`
* `wikispecies`: `Wikispecies`
* `wikiversity`: `Wikiversity`
* `wikivoyage`: `Wikivoyage`
* `wiktionary`: `Wiktionary`

veloman-yunkan:
Will fix #518

This PR introduces a new OPDS API at `/catalog/v2` (as well as fixes a few noticed issues unrelated to that enhancement). 
Currently available endpoints are:

- `/catalog/v2/root.xml`: top level navigation feed, pointing to the full acquisition feed and to the list of categories
- `/catalog/v2/entries`: an equivalent of both `/catalog/root.xml` and `/catalog/search` of the old OPDS API.
- `/vatalog/v2/categories`: a navigation feed containing an entry for each category with a link to the corresponding acquisition feed filtered by that category.

Notes:

- facets are not used (contrary to the suggestion in https://github.com/kiwix/libkiwix/issues/518#issuecomment-792627977)
- `OPDSDumper` is not utilized. `mustache` was used instead.


veloman-yunkan:
We will not have have clean CI until #522 is merged and #523 is fixed. If possible, please review without waiting for the CI to pass.

src/server/internalServer.cpp
```diff
@@ -678,30 +697,171 @@ InternalServer::search_catalog(const RequestContext& request,
...
+std::unique_ptr<Response> InternalServer::handle_catalog_v2_entries(const RequestContext& request)
+{
+  const std::string root_url = normalizeRootUrl(m_root);
+
+  const auto now = gen_date_str();
+  kainjow::mustache::list bookData;
+  const auto filter = get_search_filter(request);
+  const auto allMatchingEntries = mp_library->filter(filter);
+  const size_t count = request.get_optional_param("count", 10UL);
+  const size_t start = request.get_optional_param("start", 0UL);
```
mgautierfr, line 793: This seems pretty close to `search_catalog` method. We should reuse it.

src/server/internalServer.cpp
```diff
@@ -678,30 +697,171 @@ InternalServer::search_catalog(const RequestContext& request,
...
+             RESOURCE::catalog_v2_entries_xml,
+             kainjow::mustache::object{
+               {"date", now},
+               {"endpoint_root", root_url + "/catalog/v2"},
+               {"feed_id", gen_uuid(m_library_id + "/entries?"+request.get_query())},
+               {"filter", query},
+               {"books", bookData }
+             },
+             "application/atom+xml;profile=opds-catalog;kind=acquisition"
+  );
```
mgautierfr, line 834: Even if this is not mandatory by opds spec, it would be better to also
add openSearch element (https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#elements)
(As OPDSDumper does)

---

If you need to search for opensearch spec, it seems that now opensearch.org is use for the opensearch software (:facepalm: :angry: :cry:). So all links you can find on the web are wrong, especially the ones in opds specs. The opensearch spec is now at https://github.com/dewitt/opensearch

mgautierfr (changes requested):
> OPDSDumper is not utilized. mustache was used instead.

I dislike the idea. Now we have two ways to generate opds feed.
I'm not against using mustache, but it would be better to port OPDSDumper to mustache and use OPDSDumper everywhere.
(And having a specific OPDSDumper class would allow us to test it without creating a instance of the server)

> /catalog/v2/entries: an equivalent of both /catalog/root.xml and /catalog/search of the old OPDS API.

With issue https://github.com/kiwix/libkiwix/issues/209 the plan is to move to a partial feed for the entries.
In this PR, the entries feed is a complete acquisition feed.
It is not a blocker for this PR, it is probably better to do it in two PR.
But we must care to not do a release with a complete acquisition feed.

Also, now we have only one url for listing all entries and search for them. I don't know if it is good thing or not, but I prefer to mention it explicitly.

kelson42:
If we change the URL paths/schemas we need to be careful. We have Kiwix Desktop and Kiwix iOS relying on them!

mgautierfr:
> If we change the URL paths/schemas we need to be careful. We have Kiwix Desktop and Kiwix iOS relying on them!

We don't change the paths/schemas, we add new ones (`catalog/v2/*`)

src/server/internalServer.cpp
```diff
@@ -76,6 +76,23 @@ extern "C" {
...
+{
+  return to_string(zim::Uuid::generate(s));
+}
+
+inline std::string normalizeRootUrl(const std::string& rootUrl)
+{
+  return (rootUrl.empty() || rootUrl[0] == '/')
+       ? rootUrl
+       : "/" + rootUrl;
+}
```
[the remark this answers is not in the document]
veloman-yunkan: Fully agree. I simply wanted to avoid any impact on the old code, but it seems that the chance of breaking something with this is negligible.

src/server/internalServer.cpp
veloman-yunkan: Done

src/server/internalServer.h
```diff
@@ -73,6 +73,10 @@ class InternalServer {
     std::unique_ptr<Response> build_homepage(const RequestContext& request);
     std::unique_ptr<Response> handle_skin(const RequestContext& request);
     std::unique_ptr<Response> handle_catalog(const RequestContext& request);
+    std::unique_ptr<Response> handle_catalog_v2(const RequestContext& request);
+    std::unique_ptr<Response> handle_catalog_v2_root(const RequestContext& request);
+    std::unique_ptr<Response> handle_catalog_v2_entries(const RequestContext& request);
+    std::unique_ptr<Response> handle_catalog_v2_categories(const RequestContext& request);
```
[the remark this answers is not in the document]
veloman-yunkan: Done

veloman-yunkan:
> > OPDSDumper is not utilized. mustache was used instead.
> 
> I dislike the idea. Now we have two ways to generate opds feed.
> I'm not against using mustache, but it would be better to port OPDSDumper to mustache and use OPDSDumper everywhere.
> (And having a specific OPDSDumper class would allow us to test it without creating a instance of the server)

I moved most of the new code into `OPDSDumper` but didn't port pre-existing code to mustache. That can be done in another PR.

veloman-yunkan:
New commits start from [/catalog/v2/entries going through OPDSDumper](60b7e667e571ee87d245590b83e8e773ec505d28)

mgautierfr:
The ideas is not to "have a OPDSDumper" but to avoid having two different implementations to do the same thing (transform a list of books, with some metadata, to a opds stream).
Moving mustache implementation inside OPDSDumper but in another method and keep the previous implementation is useless.
If mustache is a better implementation, then we should replace the previous implementation.

veloman-yunkan:
> The ideas is not to "have a OPDSDumper" but to avoid having two different implementations to do the same thing (transform a list of books, with some metadata, to a opds stream).
> Moving mustache implementation inside OPDSDumper but in another method and keep the previous implementation is useless.
> If mustache is a better implementation, then we should replace the previous implementation.

Done

src/opds_dumper.cpp
```diff
@@ -168,6 +79,32 @@ string OPDSDumper::dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const
...
+  const kainjow::mustache::object template_data{
+     {"date", gen_date_str()},
+     {"root", rootLocation},
+     {"feed_id", gen_uuid(libraryId + "/catalog/search?"+query)},
+     {"filter", query.empty() ? MustacheData(false) : MustacheData(query)},
+     {"totalResults", to_string(m_totalResults)},
+     {"startIndex", to_string(m_startIndex)},
+     {"itemsPerPage", to_string(m_count)},
+     {"books", bookData }
+  };
```
mgautierfr, line 99: It seems that this data is pretty closed from the ones used in `dumpOPDSFeedV2`.
Could be have "only one method" using two different templates ?
(Or maybe even better, one template with just the endpoint changing ? If possible)

kelson42:
@veloman-yunkan @mgautierfr This PR is key and need review and rebase.

mgautierfr (changes requested):
Two small commenst on the code to fix and we are good. This is far better with all the code in the rendered and using mustache everywhere.

---

On a more general comment, #209 implies that the root/search feed is partial and then we get the full information doing another request (and so another endpoint).
We will definitively not do it in this PR. But I would like to mention that the v2 API may change in close future and client should not start to use it right now.
(And to have a more stable API, we may move the endpoint `/catalog/v2/entries` to `/catalog/v2/search` (and so keep `entry`/`entries` for the endpoint to get all information about entries)

src/opds_dumper.cpp
veloman-yunkan: Let's consider that after the new API is finalized. I don't want the old API to be a drag on the new API, and a premature attempt to share code between them will do just that.

veloman-yunkan:
@mgautierfr  @kelson42 Only the last two commits are new since the last review (ignoring a a trivial conflict in one of the earlier commits that was resolved while rebasing).

mgautierfr:
We are good. Please rebase-fixup and we can merge.

mgautierfr (approved):

veloman-yunkan:
> We are good. Please rebase-fixup and we can merge.

Done

PR #497, opened 2021-04-19
removed redirect to articles in search

### LINKED ISSUE #481: Remove redirect to article in search
Following #464, we now have the redirect code in search request (https://github.com/kiwix/kiwix-lib/blob/master/src/server/internalServer.cpp#L500-L522) which somehow become useless.

It may broke a API (if someone use the search api with the article title to directly go to the article (in case of bookmark)), but I think this is something we can accept.

MananJethwani:
@mgautierfr @kelson42 this resolves #481


mgautierfr:
@MananJethwani Can you merge the two commits together ?
It is not needed to have two separated step here.

Else, we are good.

MananJethwani:
@mgautierfr done

mgautierfr:
Thanks

PR #498, opened 2021-04-21
fixed suggestion system

MananJethwani:
@mgautierfr @kelson42 this resolves kiwix/kiwix-tools#446
and this moves #483 to a branch in kiwix/kiwiwx-lib

problem - suggestion system worked sometimes and failed otherwise as there was an async attribute added to `taskbar.js` script tag which made it load and executed earlier than jquery modules causing it to pick jquery which was available without following the order of loading modules.

fix - I have removed that async attribute and added defer to the scripts we load so that we can be sure of the order of loading of scripts, I have also kept `jQuery.noConflict` as the jquery we load should not conflict with jquery added by the zimfile.
also, the whole `taskbar.js` is still under a `jq(document).ready()` as all jquery operations should always occur after all elements required by jQuery are completely loaded.

### DIFF
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -3,83 +3,86 @@ function htmlDecode(input) {
     return doc.documentElement.textContent;
 }
 
-(function ($) {
-    const root = $( `link[type='root']` ).attr("href");
-    const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
-
-    $( "#kiwixsearchbox" ).autocomplete({
-
-        source: `${root}/suggest?content=${bookName}`,
-        dataType: "json",
-        cache: false,
-
-        response: function( event, ui ) {
-              for(const item of ui.content) {
-                  item.label = htmlDecode(item.label);
-                  item.value = htmlDecode(item.value);
-                  if (item.path) item.path = htmlDecode(item.path);
-              }
-        },
-
-        select: function(event, ui) {
-            if (ui.item.kind === 'path') {
-                window.location.href = `${root}/${bookName}/${encodeURI(ui.item.path)}`;
-            } else {
-                $( "#kiwixsearchbox" ).val(ui.item.value);
-                $( "#kiwixsearchform" ).submit();
-            }
-        },
-    });
-
-    /* cybook hack */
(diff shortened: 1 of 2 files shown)


static/skin/taskbar.js
```diff
@@ -3,83 +3,86 @@ function htmlDecode(input) {
...
-    if ($(window).width() < 520) {
-        var didScroll;
-        var lastScrollTop = 0;
-        var delta = 5;
-        // on scroll, let the interval function know the user has scrolled
-        $(window).scroll(function (event) {
-            didScroll = true;
+const jq = jQuery.noConflict();
+jq(document).ready(() => {
+    (function ($) {
```
mgautierfr, line 8: Your code is 
```
jq(document).ready(() => {
  (function ($) { ... })(jq);
})
```
But documentation propose 
```
jq(document).ready(function( $ ) {
   ...
});
```

The second seems simpler. What is the difference ?

static/skin/taskbar.js
```diff
@@ -3,83 +3,86 @@ function htmlDecode(input) {
...
-    }
-
-    if ($(window).width() < 520) {
-        var didScroll;
-        var lastScrollTop = 0;
-        var delta = 5;
-        // on scroll, let the interval function know the user has scrolled
-        $(window).scroll(function (event) {
-            didScroll = true;
+const jq = jQuery.noConflict();
```
mgautierfr, line 6: From https://api.jquery.com/jQuery.noConflict/ it seems that we should pass a `true` argument to remove referent to jQuery itself.

mgautierfr (commented):
I'm not a jquery expert so I may be wrong.
But few comments that need to be answered before merging.

static/skin/taskbar.js
MananJethwani: @mgautierfr the second code specifies to use the global variable $ which has the zim file variant of jquery associated with it, in the first type we are mentioning to use `jq` as a $.
P.S. - this code was not added by me and I just tried to keep it similar to older code.

kelson42 (approved):
principle + code looks good to me. I have tested it works. @MananJethwani Thank you very much.

kelson42:
@mgautierfr I guess we can merge this now?

PR #499, opened 2021-04-28
const-correct kiwix::Library

### LINKED ISSUE #487: Shouldn't some of the methods of kiwix::Library be const?
The following public methods of `kiwix::Library` are conceptually `const` but are not declared such:

```c++
  std::shared_ptr<Reader> getReaderById(const std::string& id);
  bool writeToFile(const std::string& path);
  bool writeBookmarksToFile(const std::string& path);
  unsigned int getBookCount(const bool localBooks, const bool remoteBooks);
  std::vector<std::string> getBooksLanguages();
  std::vector<std::string> getBooksCreators();
  std::vector<std::string> getBooksPublishers();
  const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = true);
  BookIdCollection getBooksIds();
  DEPRECATED BookIdCollection filter(const std::string& search);
  BookIdCollection filter(const Filter& filter);
  void sort(BookIdCollection& bookIds, supportedListSortBy sortBy, bool ascending);
  DEPRECATED BookIdCollection listBooksIds(...);
```

Besides, it makes sense to have `const` versions of the following methods:

```c++
  Book& getBookById(const std::string& id);
  Book& getBookByPath(const std::string& path);
```

Is there anything that is preventing us from doing this?

veloman-yunkan:
Fixes #487 

- Made most methods of kiwix::Library const.
- Also added const versions of getBookById() and getBookByPath()
  methods.

Should we bump the libkiwix version because of this change?

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -193,7 +193,9 @@ class Library
    */
   bool removeBookmark(const std::string& zimId, const std::string& url);
 
+  const Book& getBookById(const std::string& id) const;
   Book& getBookById(const std::string& id);
+  const Book& getBookByPath(const std::string& path) const;
   Book& getBookByPath(const std::string& path);
   std::shared_ptr<Reader> getReaderById(const std::string& id);
 
@@ -211,7 +213,7 @@ class Library
    * @param path the path of the file to write to.
    * @return True if the library has been correctly saved.
    */
-  bool writeToFile(const std::string& path);
+  bool writeToFile(const std::string& path) const;
 
   /**
    * Write the library bookmarks to a file.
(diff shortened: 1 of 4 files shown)


mgautierfr:
> Should we bump the libkiwix version because of this change?

Yes, this is a ABI break and we don't have a separated versioning of ABI/API.

kelson42:
> > Should we bump the libkiwix version because of this change?
> 
> Yes, this is a ABI break and we don't have a separated versioning of ABI/API.

OK, but we have already bumped to 10.0.0, so we are good?

mgautierfr:
Indeed. We can merge it.

PR #505, opened 2021-05-08
Use title snippets if available

### LINKED ISSUE #82: Snippets for suggestions should return highlighted title, not content.
The suggestions method of the searcher is searching in the title, not the content.
And if we are searching for suggestion, we will display the title to the user, not the content.

So in this case, snippets should return highlighted title.

maneeshpm:
Fixes #82 

With openzim/libzim#545 we now support snippet generation for titles. It can be used as the display label on the ui for highlighted titles via the "label" field. The old version used plain-title which is still available in the value field.

Changes included in this pr:
- Sets `label` to title snippet if available
- Set up proper rendering of snippet using `_renderItem` property of autocomplete ui

### DIFF
--- a/include/reader.h
+++ b/include/reader.h
@@ -77,7 +77,7 @@ class SuggestionItem
  * file.
  */
 
-using SuggestionsList_t = std::vector<std::vector<std::string>>;
+using SuggestionsList_t = std::vector<SuggestionItem>;
 class Reader
 {
  public:
--- a/include/reader.h
+++ b/include/reader.h
@@ -38,6 +38,41 @@ namespace kiwix
 {
 
 /**
+ * The SuggestionItem is a helper class that contains the info about a single
+ * suggestion item.
+ */
+
+class SuggestionItem
+{
+  // Functions
+  private:
+    // Create a sugggestion item.
+    explicit SuggestionItem(std::string title, std::string normalizedTitle,
+                            std::string path, std::string snippet = "") :
+      title(title),
+      normalizedTitle(normalizedTitle),
+      path(path),
+      snippet(snippet) {}
+
+  public:
+    const std::string getTitle()           {return title;}
+    const std::string getNormalizedTitle() {return normalizedTitle;}
+    const std::string getPath()            {return path;}
+    const std::string getSnippet()         {return snippet;}
+
+    const bool hasSnippet()                {return !snippet.empty();}
+
+  // Data
+  private:
+    std::string title;
+    std::string normalizedTitle;
+    std::string path;
+    std::string snippet;
+
+  friend class Reader;
+};
+
+/**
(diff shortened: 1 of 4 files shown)


kelson42:
@MananJethwani Maybe you can help @maneeshpm on that one?

MananJethwani:
this I suppose is a problem with jquery-ui. because the data is received in taskbar.js as we expect it to be, and to stop this behavior we might have to use [_renderItem](https://api.jqueryui.com/autocomplete/#method-_renderItem) property of autocomplete UI.

MananJethwani:
@maneeshpm `jquery-ui.js` is the library file from where we import suggestion system.

maneeshpm:
Thanks a lot for your help @MananJethwani, front end is not really my strong point :sweat_smile:  

Final result of the modification:
Screenshot from 2021-05-10 12-56-32


@kelson42 I believe this feature can be improved further by modifications in stemming and stopwords(openzim/libzim#324) to ensure all the matching terms are highlighted.

kelson42:
@maneeshpm Great to see this implemented and great you see the connection to the other ticket :)

src/reader.cpp
```diff
@@ -508,6 +508,7 @@ bool Reader::searchSuggestionsSmart(const string& prefix,
       suggestion.push_back(current->getTitle());
       suggestion.push_back(current->getPath());
       suggestion.push_back(kiwix::normalize(current->getTitle()));
+      suggestion.push_back(current.get_snippet());
```
mgautierfr, line 511: Maybe it is time to introduce a class/struct for the suggestion result instead of using a vector of strings.

mgautierfr (changes requested):

maneeshpm:
Rebasing temporarily on #526

include/reader.h
```diff
@@ -37,6 +37,38 @@ using namespace std;
...
+{
+  // Functions
+  public:
+    // Create a sugggestion item.
+    explicit SuggestionItem(std::string title, std::string normalizedTitle,
+                            std::string path, std::string snippet = "") :
+      title(title),
+      normalizedTitle(normalizedTitle),
+      path(path),
+      snippet(snippet) {}
```
mgautierfr, line 55: This is useless for the user to create a suggestion itself.
It is better to make the constructor private and add the `Reader` as friend class.

mgautierfr (changes requested):

kelson42:
@maneeshpm rebasing maybe?

kelson42:
@mgautierfr I have rebased this branch. We should really review/merge.

mgautierfr (approved):

mgautierfr:
Rebased-fixup on master (and fix conflicts). We are good.

PR #506, opened 2021-05-08
Empty query acts as a match-all query

### LINKED ISSUE #500: [REGRESSION] No (pattern) filter should provide everything
This is a screenshot of Kiwix-Desktop 2.0.5
image

This is how master HEAD (of everything) looks:
image

I believe the problem is the handling of filter with no values which give currently no value, but should give all values.

veloman-yunkan:
Should fix #500

After switching to Xapian-based search in the library/catalog, an empty query stopped acting as a match-all query. This commit restores the old behaviour in that regard.

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -327,7 +327,7 @@ bool willSelectEverything(const Xapian::Query& query)
 
 Xapian::Query buildXapianQueryFromFilterQuery(const Filter& filter)
 {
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan Thank you. I have tested with Kiwix Desktop and it seems to work fine. I have rebased on master as well.

PR #510, opened 2021-05-13
Add `get_zimId` method to Result

### LINKED ISSUE #107: Add get_zim_id to Result
Currently, the `Result` class has  `get_readerIndex` but no `get_zim_id `. 

Users have to keep an additional vector of zim file IDs somewhere to track which zim file a result is in. Users also cannot use the property `readers` of `Searcher` for this purpose, since it is protected (and it probably should be). 

But since `Searcher` already have the `readers`, it can save users some trouble by providing this info in `Result`. To go a bit further, we might just deprecate `get_readerIndex`, since providing the ID is probably more useful than index.

maneeshpm:
Fixes #107 

Must be merged after #526.

`get_zimId` method allows the user to get the uuid of the archive from which a result is retrieved directly from the search result itself. 

*Note*: With this, we might want to get rid of the method `get_readerIndex` later since we can directly identify the archive from the search result with this change.

Changes included in this pr:
- Add a new method `get_zimId` to the `Results`

**This PR depends on openzim/libzim#557**

### DIFF
--- a/include/searcher.h
+++ b/include/searcher.h
@@ -48,7 +48,6 @@ class Result
   virtual std::string get_content() = 0;
   virtual int get_wordCount() = 0;
   virtual int get_size() = 0;
-  virtual int get_readerIndex() = 0;
   virtual std::string get_zimId() = 0;
 };
 
--- a/include/searcher.h
+++ b/include/searcher.h
@@ -49,6 +49,7 @@ class Result
   virtual int get_wordCount() = 0;
   virtual int get_size() = 0;
   virtual int get_readerIndex() = 0;
+  virtual std::string get_zimId() = 0;
 };
 
 struct SearcherInternal;
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -77,9 +77,7 @@ std::string SearchRenderer::getHtml()
     result.set("title", p_result->get_title());
     result.set("url", p_result->get_url());
     result.set("snippet", p_result->get_snippet());
-    auto readerIndex = p_result->get_readerIndex();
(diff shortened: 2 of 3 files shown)


kelson42:
@maneeshpm I guess this is ready to review?

maneeshpm:
@kelson42 Yes it's ready for review. I couldn't test it due to the API change, but the the pr is a trivial one.

veloman-yunkan:
> I couldn't test it due to the API change, but the the pr is a trivial one.

@maneeshpm You can temporarily rebase your PR on #526

maneeshpm:
Thanks for the workaround @veloman-yunkan :smile:

mgautierfr:
CI is broken as we need to wait tomorrow for kiwix-build CI generating new dependency archive using just merged https://github.com/openzim/libzim/pull/557

> Note: With this, we might want to get rid of the method get_readerIndex later since we can directly identify the archive from the search result with this change.

Please do.

maneeshpm:
@mgautierfr Aaah that makes sense, thanks :D
I'll remove the `get_readerIndex` function.

maneeshpm:
I have removed all the direct usage of `get_readerIndex` in the last commit. But `Searcher` still maintains a vector of readers inside it. Modifying the class is more suited in #430.

kelson42:
@automactic Btw, would this PR simplify your life like expected

kelson42:
@mgautierfr We should really review/merge this, this seems trivial.

mgautierfr:
Rebased/fixup on master and fix conflicts.

maneeshpm:
Thanks @mgautierfr!

PR #528, opened 2021-05-17
Check if bookName is available in url parameters

### LINKED ISSUE #496: Kiwix-serve suggestion does not work from search results page
When we type anything in the search box from the search results page, we do not get suggestions. This is probably because the request is generated using the wrong parameters.
```
...
arguments :
 - content : 'search'
 - term : 'twin city'
Parsed : 
full_url: /suggest
url   : /suggest
acceptEncodingDeflate : 1
has_range : 0
is_valid_url : 1
.............
** running handle_suggest
Response :
httpResponseCode : 404
...
```
As seen in the log, the generated url takes the `content` parameter from the search page url(localhost:8080/**search**?content="foo"&pattern="bar"), hence the `content` parameter is invalid. 

```
..
arguments :
 - content : ''
 - term : 'twin'
Parsed : 
full_url: /suggest
url   : /suggest
acceptEncodingDeflate : 1
has_range : 0
is_valid_url : 1
.............
** running handle_suggest
Response :
httpResponseCode : 404
...
```
As seen above, This does not work from the main page as well since the url (localhost:8080/) does not have any content.

maneeshpm:
Fixes #496 

In certain pages like the search result page, `bookName` is not of the form `/bookName/endpoint?parameters` which is the assumption here:
https://github.com/kiwix/libkiwix/blob/51883558781008d006b983991bd4ff35e53f3e79/static/skin/taskbar.js#L7-L10
Rather it is available as a query parameter in `/endpoint?content=bookName&otherParameters`. From these pages, `bookName` should be assigned from parameters.

Changes included in this pr:
- Checks if the URL has a `content` parameter. If yes, assign `bookName` to this parameter.

### DIFF
--- a/static/skin/taskbar.js
+++ b/static/skin/taskbar.js
@@ -7,7 +7,10 @@ const jq = jQuery.noConflict(true);
 jq(document).ready(() => {
     (function ($) {
         const root = $( `link[type='root']` ).attr("href");
-        const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+
+        const bookName = (window.location.pathname == `${root}/search`)
+               ? (new URLSearchParams(window.location.search)).get('content') 
+               : window.location.pathname.split(`${root}/`)[1].split('/')[0];
     
         $( "#kiwixsearchbox" ).autocomplete({
     


kelson42:
@maneeshpm Thx, one small issue in the codefactor.

maneeshpm:
@kelson42 All checks are passing now. :+1:

veloman-yunkan (commented):
I don't know what kind of content ZIM archives and kiwix-serve may be used in the future (or, maybe, are already used). Nothing prevents from having a search string appended to regular article URLs (e.g. http://library.kiwix.org/phet_en_2021-05/A/index.html?content=abcd). If anyone decides to take advantage of that in frontend javascript and utilize the `content` parameter, then this logic will be tricked.

Currently, the book name is passed via the content parameter for the following targets:

1. `/meta` and `/suggest`: these do not have a taskbar  
2. `/random?content=bookname`: we shouldn't mind about this case since it immediately redirects to a URL of the form `/bookname/some_article` 
3. `/search?content=bookname&...`: as of today, this is the only case we should care about. Maybe it is better to apply the new logic only we are dealing with a URL of that kind?

maneeshpm:
@veloman-yunkan I agree, we should be performing this only for URL of that kind. I'll make that change.
Nevertheless, it would be interesting if we had a standard protocol for passing "content" across all the endpoints.

static/skin/taskbar.js
```diff
@@ -7,7 +7,14 @@ const jq = jQuery.noConflict(true);
 jq(document).ready(() => {
     (function ($) {
         const root = $( `link[type='root']` ).attr("href");
-        const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+        const urlParams = new URLSearchParams(window.location.search);
+
+        content = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+        if (window.location.href.search('\\/search\\?content\\=') != -1) {
```
veloman-yunkan, line 13: This fix pushes the imaginary corner case discussed in the previous review further but doesn't eliminate it. Suppose that there is a ZIM archive with an article named `search`. Besides, this fix will break if the `content` parameter stops being the first parameter in the query.

Why do you search for `search`?  Why don't you simply check that the path of the url is `${root}/search`?

static/skin/taskbar.js
```diff
@@ -7,7 +7,14 @@ const jq = jQuery.noConflict(true);
 jq(document).ready(() => {
     (function ($) {
         const root = $( `link[type='root']` ).attr("href");
-        const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+        const urlParams = new URLSearchParams(window.location.search);
```
veloman-yunkan, line 10: Move inside the `if` body.

static/skin/taskbar.js
```diff
@@ -7,7 +7,14 @@ const jq = jQuery.noConflict(true);
...
+
+        content = window.location.pathname.split(`${root}/`)[1].split('/')[0];
```
veloman-yunkan, line 12: Declare this variable with `let`. Alternatively you can get rid of both `content` and  `urlParams` auxiliary variables if you use the conditional operator `?`.

veloman-yunkan (changes requested):

static/skin/taskbar.js
maneeshpm: That will be better and robust! JS is something that I am not very familiar with, but I should be able to catch up soon. I'll make the change.

static/skin/taskbar.js
```diff
@@ -7,7 +7,10 @@ const jq = jQuery.noConflict(true);
 jq(document).ready(() => {
     (function ($) {
         const root = $( `link[type='root']` ).attr("href");
-        const bookName = window.location.pathname.split(`${root}/`)[1].split('/')[0];
+
+        const bookName = (window.location.pathname.search(`${root}/search`) != -1) ?
```
veloman-yunkan, line 11: I guess one of my previous comments was not straightforward enough. Don't `search` in the pathname, because that may match URLs that are different from `${root}/search`. Simply

```JS
const bookName = (window.location.pathname == `${root}/search`)
               ? (new URLSearchParams(window.location.search)).get('content') 
               : window.location.pathname.split(`${root}/`)[1].split('/')[0];
```

veloman-yunkan (changes requested):

veloman-yunkan:
And please fixup and rebase so that can merge without an extra iteration.

maneeshpm:
@veloman-yunkan Done! Now that I understand what you meant, it seems awkwardly trivial :sweat_smile: Thanks!

veloman-yunkan (approved):

Issue #386, opened 2020-07-23
Rename kiwix-lib to libkiwix


kelson42:
Always the same reasons:
* Stick to the formalism on Linux
* Better makes the difference with tools like kiwix-tools
* Unify naming with libzim.

automactic:
OK with me. I have actually always been calling it libkiwix. Anything I need to do form my side?

kelson42:
@automactic Thx for the positive feedback. Not for the moment.

rgaudin:
Same comment as @automactic.

mgautierfr:
While I agree why we would make that change, I am pretty cautious about this.

There is a lot of things depending of the "kiwix-lib" name. At least I can see (but maybe other):

- `kiwix-build` has to be changed to dl and build "libkiwix" instead of "kiwix-lib" (kiwix-build itself, but also the github and appveyor ci).
- `kiwix-lib` ci itself has it download the dependencies archive named after him.
- The bintray (https://bintray.com/kiwix/kiwix/kiwixlib) has to be changed (@macgills)
- All the release directory (https://download.kiwix.org/release/kiwix-lib/) will change.
- Packagers need to rename their packages and all the dependent packages. (@legoktm, @xvitaly, maybe @swills, others ?)
- I don't know what will happen to the github forks (https://github.com/kiwix/kiwix-lib/network/members)
- I don't know what will happen for webhooks and github apps and third apps (https://codecov.io/gh/kiwix/kiwix-lib)
- All links in "all" websites should be changed.

This should works "out of the box" but I mention it anyway : 
- No need to change the dependendent projects as we depend on "kiwix", not "kiwix-lib".
- Github will do the redirect for local clone repository. Although it is better for people to change the remote url, a `git pull`/`git push` will work.

xvitaly:
This is not good. Maintainers are not allowed to remove packages from already released Fedora versions. It will break updates of Kiwix stack for 1 year.

kelson42:
@xvitaly First of all this ticket is about the repo and Kiwix naming. You are still free to deal things like you want in RPMS. But "yes" would be good at some point to stick to the new name. For the rest, I'm sure there is std way to deal with that problem. Without having a clue, I would just transform the current package in a dummy one having the new package as dependence.

xvitaly:
> I'm sure there is std way to deal with that problem.

Standard procedure:
1. retire the old package (does not allowed for released Fedora versions);
2. review a new one;
3. fix all dependencies.

[...]

PR #529, opened 2021-05-18
Fix Libkiwix Github repository URLS

kelson42:
Fixes part of #386


kelson42:
@veloman-yunkan CI is OK now

mgautierfr:
There are other reference to `kiwix-lib` but on the github side we are good.

Issue #517, opened 2020-05-17
Kiwix-serve welcome page filter by category


kelson42:
We should be able to filter by category on kiwix-serve home page.

Issue #519, opened 2018-08-21
Kiwix-serve: Language filter of zim list on welcome page 


sharun-s:
I added a little bookmarklet to inject a language filter dropdown to filters zims on the library.kiwix.org site. It could probably work as is in a jquery ready function in home.html.tmpl. Unfamiliar with kiwix-tools and the build/test process to figure out how to submit this as a pr. So pasting code here...after library.kiwix.org loads just paste it into the dev console to see it in action... 

```

sel=$("<div><label for='langSelect'>Language</label> \
         <select id='langSelect'>\
         <option value='ar'>العربية</option>\
         ‎<option value='az'>azərbaycanca</option>\
         ‎<option value='bn'>বাংলা</option>\
         <option value='ca'>català</option>\
         <option value='da'>dansk</option>\
         <option value='de'>Deutsch</option>\
         <option value='el'>Ελληνικά</option>\
         <option value='en' selected>‎English</option> \
         <option value='es'> español</option> \
         <option value='fa'>فارسی</option> \
         <option value='fr'>français</option> \
         <option value='gl'>galego</option> \
         <option value='he'>עברית</option> \
         <option value='id'>Bahasa Indonesia</option> \
         <option value='it'>italiano</option> \
         <option value='ja'>日本語</option>\
         <option value='ka'>ქართული</option>\
         <option value='ko'>한국어</option>\
         <option value='ku-latn'>Kurdî (latînî)‎‎</option>\
         <option value='lt'>lietuvių</option>\
         <option value='ml'>മലയാളം</option>\
         <option value='ms'>Bahasa Melayu</option> \
         ‎<option value='nl'>Nederlands</option>\
         <option value='pl'>polski</option>\
         <option value='ps'>پښتو</option>\
         <option value='pt'>português</option>\
         <option value='ru'>русский</option> ‎\
         <option value='sd'>سنڌي</option> ‎\
         <option value='tr'>Türkçe</option> ‎\
         <option value='ur'>اردو</option> ‎\
         <option value='zh'>中文</option> ‎\
         ‎<option value='zh-cn'> 中文（中国大陆）‎</option> ‎\
         ‎<option value='zh-tw'>中文（台灣）‎</option></select></div>");

$(sel).insertBefore(".book__list")

catalog=$(".book__list a")

$("#langSelect").on("change",function (){
         lang = this.value;console.log(lang + " lang selected");
         catalog.hide();
[...]

PR #530, opened 2021-05-18
Kiwix Serve welcome page dynamic and subset loading (OPDS based)

### LINKED ISSUE #511: Kiwix serve welcome page should be dynamic
It should propose filters, make the [OPDS](https://wiki.kiwix.org/wiki/OPDS) call, parse the result with JS and show the result (cards) dynamically.

MananJethwani:
@kelson42 this resolves #511 and #514. Filters will be added in other PR.

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -658,7 +658,7 @@ std::vector<std::string>
 InternalServer::search_catalog(const RequestContext& request,
                                kiwix::OPDSDumper& opdsDumper)
 {
-    auto filter = kiwix::Filter().valid(true).local(true).remote(true);
+    auto filter = kiwix::Filter().valid(true).local(true);
     string query("<Empty query>");
     size_t count(10);
     size_t startIndex(0);
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -1,0 +1,76 @@
+(function() {
+    const root = $(`link[type='root']`).attr('href');
+    let isFetching = false;
+    const incrementalLoadingParams = {
+        start: 0,
+        count: viewPortToCount()
+    };
+
+    function queryUrlBuilder() {
+        let url = `${root}/catalog/search?`;
+        url += Object.keys(incrementalLoadingParams).map(key => `${key}=${incrementalLoadingParams[key]}`).join("&");
+        return url;
+    }
+
+    function viewPortToCount(){
+        return Math.floor(window.innerHeight/100 + 1)*(window.innerWidth>1000 ? 3 : 2);
+    }
+
+    function getInnerHtml(node, query) {
+        return node.querySelector(query).innerHTML;
+    }
+
+    function generateBookHtml(book) {
+        const link = book.querySelector('link').getAttribute('href');
+        const title =  getInnerHtml(book, 'title');
+        const description = getInnerHtml(book, 'summary');
+        const id = getInnerHtml(book, 'id');
+        const iconUrl = getInnerHtml(book, 'icon');
+        const articleCount = getInnerHtml(book, 'articleCount');
+        const mediaCount = getInnerHtml(book, 'mediaCount');
+
+        return `<a href='${link}' data-id='${id}'><div class='book'>
+            <div class='book__background' style="background-image: url('${iconUrl}');">
+            <div class='book__title' title='${title}'>${title}</div>
+            <div class='book__description' title='${description}'>${description}</div>
+            <div class='book__info'>${articleCount} articles, ${mediaCount} medias</div>
+            </div>
+        </div></a>`;
+    }
+
+    async function loadAndDisplayBooks() {
+        isFetching = true;
+        fetch(queryUrlBuilder()).then(async (resp) => {
+            const data = new window.DOMParser().parseFromString(await resp.text(), 'application/xml');
+            const books = data.querySelectorAll('entry');
+            let bookHtml = '';
+            books.forEach((book) => {bookHtml += generateBookHtml(book)});
+            document.querySelector('.book__list').innerHTML += bookHtml;
+            incrementalLoadingParams.start += books.length;
+            if (books.length < incrementalLoadingParams.count) {
+                incrementalLoadingParams.count = 0;
+            }
+            isFetching = false;
+        });
+    }
+
+    async function loadSubset() {
+        if (!isFetching &&
+            incrementalLoadingParams.count &&
+            window.innerHeight + window.scrollY >= document.body.offsetHeight
+        ) {
+            loadAndDisplayBooks();
+        }
+    }
+
+    window.addEventListener('resize', async () => {
+        incrementalLoadingParams.count = incrementalLoadingParams.count && viewPortToCount();
+        loadSubset();
+    });
+
+    window.addEventListener('scroll', loadSubset);
+
+    window.onload = async () => {
+        loadAndDisplayBooks();
+    }
+})();
--- a/static/templates/index.html
+++ b/static/templates/index.html
@@ -1,66 +1,96 @@
 <!DOCTYPE html>
 <html xmlns="http://www.w3.org/1999/xhtml">
-<head>
-  <meta charset="UTF-8" />
-  <title>Welcome to Kiwix Server</title>
-  <script type="text/javascript" src="{{root}}/skin/jquery-ui/external/jquery/jquery.js"></script>
-  <script type="text/javascript" src="{{root}}/skin/jquery-ui/jquery-ui.min.js"></script>
-  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.min.css" rel="Stylesheet" />
-  <link type="text/css" href="{{root}}/skin/jquery-ui/jquery-ui.theme.min.css" rel="Stylesheet" />
-  <style>
-        body {
-                background:
-                radial-gradient(#EEEEEE 15%, transparent 16%) 0 0,
-                radial-gradient(#EEEEEE 15%, transparent 16%) 8px 8px,
-                radial-gradient(rgba(255,255,255,.1) 15%, transparent 20%) 0 1px,
-                radial-gradient(rgba(255,255,255,.1) 15%, transparent 20%) 8px 9px;
-                background-color:#E8E8E8;
-                background-size:16px 16px;
-                margin-left: auto;
-                margin-right: auto;
-                max-width: 1100px;
-        }
-        .book__list { text-align: center; }
-        .book {
-                display: inline-block; vertical-align: bottom; margin: 8px; padding: 12px 15px; width: 300px;
-                border: 1px solid #ccc; border-radius: 8px;
-                text-align: left; color: #000; font-family: sans-serif; font-size: 13px;
-                background-color:#F1F1F1;
-                box-shadow: 2px 2px 5px 0px #ccc;
-        }
-        .book:hover { background-color: #F9F9F9; box-shadow: none;}
-        .book__background { background-repeat: no-repeat; background-size: 48px 48px; background-position: top right; }
-        .book__title {
-                padding: 0 55px 0 0;overflow: hidden; text-overflow: ellipsis; white-space: nowrap;
-                font-size: 18px; color: #0645ad; line-height: 1em;
-        }
-        .book__description {
-                padding: 5px 55px 5px 0px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap;
-                font-size: 15px; line-height: 1em;
-        }
-        .book__info { color: #777; font-weight: bold; font-size: 13px; line-height: 1em; }
-  </style>
-  <script type="text/javascript" src="{{root}}/skin/taskbar.js" async></script>
-</head>
-<body class="kiwix">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Welcome to Kiwix Server</title>
+    <script
+      type="text/javascript"
+      src="{{root}}/skin/jquery-ui/external/jquery/jquery.js"
+    ></script>
+    <script
+      type="text/javascript"
+      src="{{root}}/skin/jquery-ui/jquery-ui.min.js"
+    ></script>
+    <link
+      type="text/css"
+      href="{{root}}/skin/jquery-ui/jquery-ui.min.css"
+      rel="Stylesheet"
+    />
+    <link
+      type="text/css"
+      href="{{root}}/skin/jquery-ui/jquery-ui.theme.min.css"
+      rel="Stylesheet"
+    />
+    <style>
+      body {
+        background: radial-gradient(#eeeeee 15%, transparent 16%) 0 0,
+          radial-gradient(#eeeeee 15%, transparent 16%) 8px 8px,
+          radial-gradient(rgba(255, 255, 255, 0.1) 15%, transparent 20%) 0 1px,
+          radial-gradient(rgba(255, 255, 255, 0.1) 15%, transparent 20%) 8px 9px;
+        background-color: #e8e8e8;
+        background-size: 16px 16px;
+        margin-left: auto;
+        margin-right: auto;
+        max-width: 1100px;
+      }
+      .book__list {
+        text-align: center;
+      }
+      .book {
+        display: inline-block;
+        vertical-align: bottom;
+        margin: 8px;
+        padding: 12px 15px;
+        width: 300px;
+        border: 1px solid #ccc;
+        border-radius: 8px;
+        text-align: left;
+        color: #000;
+        font-family: sans-serif;
+        font-size: 13px;
+        background-color: #f1f1f1;
+        box-shadow: 2px 2px 5px 0px #ccc;
+      }
+      .book:hover {
+        background-color: #f9f9f9;
+        box-shadow: none;
+      }
+      .book__background {
+        background-repeat: no-repeat;
+        background-size: 48px 48px;
+        background-position: top right;
+      }
+      .book__title {
+        padding: 0 55px 0 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        font-size: 18px;
+        color: #0645ad;
+        line-height: 1em;
+      }
+      .book__description {
+        padding: 5px 55px 5px 0px;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        font-size: 15px;
+        line-height: 1em;
+      }
+      .book__info {
+        color: #777;
+        font-weight: bold;
+        font-size: 13px;
+        line-height: 1em;
+      }
+    </style>
+    <script type="text/javascript" src="{{root}}/skin/index.js" async></script>
+  </head>
+  <body class="kiwix">
+    <div class="kiwix">
+      <div class="book__list"></div>
+    </div>
 
-<div class="kiwix">
-  <div class='book__list'>
-    {{#books}}
-    <a href="{{root}}/{{name}}"><div class='book'>
-        <div class='book__background' style="background-image: url('{{root}}/meta?content={{#urlencoded}}{{{name}}}{{/urlencoded}}&name=favicon');">
-          <div class='book__title' title='{{title}}'>{{title}}</div>
-          <div class='book__description' title='{{description}}'>{{description}}</div>
-          <div class='book__info'>{{articleCount}} articles, {{mediaCount}} medias</div>
-        </div>
-    </div></a>
-    {{/books}}
-  </div>
-</div>
-
-<div id="kiwixfooter">
-  Powered by <a href="https://kiwix.org">Kiwix</a>
-</div>
-
-</body>
+    <div id="kiwixfooter">Powered by <a href="https://kiwix.org">Kiwix</a></div>
+  </body>
 </html>
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ subprojects/googletest-release*
 *.class
 build/
 .vscode/
+builddir/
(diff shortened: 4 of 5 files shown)


kelson42:
@MananJethwani It does not fix #517 and #519?

MananJethwani:
@kelson42 no, I will be create another PR for that soon.

kelson42:
@MananJethwani Ok, even better like this.

kelson42:
@rgaudin We should secure the frontend tech could be include in offspot. Your review is needed for this.

kelson42 (approved):
From a user perspective it works like expect. We need the technical review.

rgaudin:
> @rgaudin We should secure the frontend tech could be include in offspot. Your review is needed for this.

I don't think I can do that at the moment. Offspot requirements are not ready. We only have UI mockups at this stage and the extent of `could be include in offspot` is not defined neither and that could have drastic consequences.

kelson42 (changes requested):
There is one problem. If there is no URL associated with a ZIM file, the content is not in the OPDS and the welcome page is empty. The logic behind this is clear, but we need to have the welcome page working. Maybe we should add a parameter to the OPDS search feed to deliver as well the ZIM without URLs?

MananJethwani:
@kelson42 I agree will start implementing it.

kelson42:
@MananJethwani Or maybe we just deliver everything and this is the role of the kiwix user to do the necessary if he wants to make the content downloadable? @mgautierfr @veloman-yunkan what do you think?

MananJethwani:
@kelson42 I believe the second option is better we should deliver everything and we can then add a disabled download button if we don't get a URL and add a tooltip to show a message for the disabled download button.
also if we don't deliver everything then what would be the use case left for the `ZIM_PATH` method fr starting the server as we can't add URL in that case.

src/server/internalServer.cpp
```diff
@@ -658,6 +658,8 @@ std::vector<std::string>
 InternalServer::search_catalog(const RequestContext& request,
                                kiwix::OPDSDumper& opdsDumper)
 {
+    // added for discussion in #530 using the following method we can deliver zim file irrespective of a remote URl present for it.
+    // auto filter = kiwix::Filter().valid(true).local(true);
```
MananJethwani, line 662: @kelson42 I have added that method in the comments here.
comment the line below it and remove comments from this one to get your desired results.

kelson42: @MananJethwani `kiwix-serve --port=8082 *zim` still gives me an empty welcome page.

MananJethwani: @kelson42 are you sure you commented the correct line? because I tried it and it works for me.
you need to comment L663 and uncomment L662 ... it works for me that way.
here is the  proof you can clearly see the last link in this entry is for thumbnail but if I would have added a URL there would be another link there.
Screenshot from 2021-05-22 23-57-36

static/skin/index.js
```diff
@@ -0,0 +1,31 @@
...
+        const title =  getInnerHtml(book, 'title');
+        const description = getInnerHtml(book, 'summary');
+
+        bookHtml += `<a href='${link}' data-id='${getInnerHtml(book, 'id')}'><div class='book'>
+            <div class='book__background' style="background-image: url('${getInnerHtml(book, 'icon')}');">
+            <div class='book__title' title='${title}'>${title}</div>
+            <div class='book__description' title='${description}'>${description}</div>
+            <div class='book__info'>${getInnerHtml(book, 'articleCount')} articles, ${getInnerHtml(book, 'mediaCount')} medias</div>
+            </div>
+        </div></a>`;
```
veloman-yunkan: 1. Please extract this into a function `generateBookHtml(book)` that produces HTML for a single book.
2. I would prefer that all `getInnerHtml()` calls are made outside the template literal. Expressions like `data-id='${getInnerHtml(book, 'id')}'` are a little difficult to parse visually because of nested single-quote pairs.

static/skin/index.js
```diff
@@ -1,31 +1,65 @@
...
+    'start': 0,
+    'count': Math.floor(viewPortHeight/100 + 1) * 3
+};
+
+function queryUrlBuilder() {
+    let url = `${root}/catalog/search?`;
+    Object.keys(basicConfig).forEach((key, idx) => {
+        url += `${key}=${basicConfig[key]}${idx !== Object.keys(basicConfig).length - 1 ? '&' : ''}`;
+    });
+    return url;
```
veloman-yunkan, line 15: Using `map()` + `join()` is simpler:

```
url += Object.keys(basicConfig).map(key => `${key}=${basicConfig[key]}`).join("&")
```

static/skin/index.js
```diff
@@ -1,31 +1,65 @@
...
     });
-    document.querySelector('.book__list').innerHTML = bookHtml;
-  });
+    isFetching = false;
 }
 
+window.addEventListener('resize', async () => {
+    if (isFetching || isEnd) return;
+    viewPortHeight = window.innerHeight;
+    basicConfig['count'] = Math.floor(viewPortHeight/100 + 1) * 3;
```
veloman-yunkan: 1. Dependence of the `count` on the viewport size must be defined in a single place - introduce a helper function.
2. Shouldn't this dependence include the viewport width as well?

static/skin/index.js
```diff
@@ -1,31 +1,65 @@
...
+    basicConfig['start'] = basicConfig['start'] + basicConfig['count'];
```
veloman-yunkan, line 50: Isn't it better to update `basicConfig.start` at the end of `loadAndDisplayBooks()`? Then you can get rid of the `isEnd` variable by setting `basicConfig.count` to 0.

static/skin/index.js
```diff
@@ -1,31 +1,65 @@
 const root = $(`link[type='root']`).attr('href');
+let viewPortHeight = window.innerHeight;
+let isFetching = false;
+let isEnd = false;
+const basicConfig = {
```
veloman-yunkan, line 5: The name of this variable is not a good one

static/skin/index.js
```diff
@@ -1,31 +1,65 @@
...
-        </div></a>`;
+            bookHtml += `<a href='${link}' data-id='${getInnerHtml(book, 'id')}'><div class='book'>
+                <div class='book__background' style="background-image: url('${getInnerHtml(book, 'icon')}');">
+                <div class='book__title' title='${title}'>${title}</div>
+                <div class='book__description' title='${description}'>${description}</div>
+                <div class='book__info'>${getInnerHtml(book, 'articleCount')} articles, ${getInnerHtml(book, 'mediaCount')} medias</div>
+                </div>
+            </div></a>`;
+        });
+        document.querySelector('.book__list').innerHTML = (append ? document.querySelector('.book__list').innerHTML : '') + bookHtml;
```
veloman-yunkan, line 41: Do you really need the `append` flag? Can't you just always append? On load you will append to the initial empty content with the same net result.

veloman-yunkan (changes requested):
All the new global variables and functions in index.js increase the risk of interfering with variables/functions defined in other JS scripts (including those from the ZIM archives).

static/skin/index.js
MananJethwani: @veloman-yunkan I added this because earlier I used append or replace in case of filters but this won't matter now as using filter has changed most of the implementation and I will be always appending code.

static/skin/index.js
MananJethwani: @veloman-yunkan we can convert it to a oneliner like this 
```
return `${root}/catalog/search?${Object.keys(basicConfig).map(key => `${key}=${basicConfig[key]}`).join("&")}`;
```
or will it be too complex?

static/skin/index.js
MananJethwani: @veloman-yunkan what do you suggest would be a better name?
also, what should be done about the risk of interference with other JS files should I add a prefix like KSUI(for kiwix serve UI)?
naming would be then like KSUI_isFetching etc.

static/skin/index.js
kelson42: > All the new global variables and functions in index.js increase the risk of interfering with variables/functions defined in other JS scripts (including those from the ZIM archives).

Thank you @veloman-yunkan. I don't want to interfer much in the technical review, but considering that this welcome page has only our code (and not code coming from any ZIM file). There is no risk I can see of interference with other content JS.

MananJethwani: @kelson42 he might  be talking about `block-external.js` other than that I agree it can be easily maintained other than that there are no other js files which will interfere not even `taskbar.js` as I have removed them for filtering part.

static/skin/index.js
veloman-yunkan: I personally stick to "the max line length of 80 characters" rule and would instead take measures to shorten the line proposed in my previous comment, instead of pursuing a longer and a more incomprehensible one-liner.

static/skin/index.js
veloman-yunkan: > what do you suggest would be a better name?

`incrementalLoadingParams`?

veloman-yunkan: Regarding the risk of interference, @kelson42 already pointed out that we don't have such an issue currently. The general approach is to hide the variables in closures.

MananJethwani: ok then I am not adding any closure for now.

MananJethwani:
@veloman-yunkan changed as you asked!

static/skin/index.js
```diff
@@ -1,62 +1,68 @@
...
-                <div class='book__background' style="background-image: url('${getInnerHtml(book, 'icon')}');">
-                <div class='book__title' title='${title}'>${title}</div>
-                <div class='book__description' title='${description}'>${description}</div>
-                <div class='book__info'>${getInnerHtml(book, 'articleCount')} articles, ${getInnerHtml(book, 'mediaCount')} medias</div>
-                </div>
-            </div></a>`;
-        });
-        document.querySelector('.book__list').innerHTML = (append ? document.querySelector('.book__list').innerHTML : '') + bookHtml;
-        isEnd = !books.length;
+        books.forEach((book) => {bookHtml += generateBookHtml(book)});
```
veloman-yunkan: This is a good case for `map().join()`:

```
const bookHtml = books.map(generateBookHtml).join("");
```

static/skin/index.js
```diff
@@ -1,62 +1,68 @@
...
+        document.querySelector('.book__list').innerHTML += bookHtml;
+        incrementalLoadingParams['start'] = incrementalLoadingParams['start'] + incrementalLoadingParams['count'];
+        if (!books.length) {
+            incrementalLoadingParams['count'] = 0;
+        }
```
veloman-yunkan, line 51: You can set `count` to 0 sooner, if you check that `books.length < incrementalLoadingParams['count']`

static/skin/index.js
```diff
@@ -1,62 +1,68 @@
...
-                <div class='book__description' title='${description}'>${description}</div>
-                <div class='book__info'>${getInnerHtml(book, 'articleCount')} articles, ${getInnerHtml(book, 'mediaCount')} medias</div>
-                </div>
-            </div></a>`;
-        });
-        document.querySelector('.book__list').innerHTML = (append ? document.querySelector('.book__list').innerHTML : '') + bookHtml;
-        isEnd = !books.length;
+        books.forEach((book) => {bookHtml += generateBookHtml(book)});
+        document.querySelector('.book__list').innerHTML += bookHtml;
+        incrementalLoadingParams['start'] = incrementalLoadingParams['start'] + incrementalLoadingParams['count'];
```
veloman-yunkan, line 48: Better do
```
incrementalLoadingParams.start += books.length;
```

And are there any reasons for preferring the `obj['propname']` notation over `obj.propname` here and in several other places in this file?

static/skin/index.js
```diff
@@ -1,62 +1,68 @@
...
 
 window.addEventListener('resize', async () => {
-    if (isFetching || isEnd) return;
-    viewPortHeight = window.innerHeight;
-    basicConfig['start'] = basicConfig['start'] + basicConfig['count'];
-    basicConfig['count'] = Math.floor(viewPortHeight/100 + 1) * 3;
-    loadAndDisplayBooks(true);
+    if (isFetching || !incrementalLoadingParams['count'] || incrementalLoadingParams['count'] > viewPortToCount()) return;
+    incrementalLoadingParams['count'] = incrementalLoadingParams['count'] && viewPortToCount();
+    loadAndDisplayBooks();
```
veloman-yunkan, line 59: We need to load more books only if we are at the end of the book list after the resize.

veloman-yunkan (changes requested):

static/skin/index.js
MananJethwani: @veloman-yunkan map doesn't work for nodeList(books), but I will update my code in my PR for filters as I have used array for books there.

static/skin/index.js
MananJethwani: no such reason...will change it as well.

MananJethwani:
@veloman-yunkan I have pushed the changes,  I have added the closure as I tested `block-external` feature and it doesn't work as constant root from index.js was interfering with it root from `block-external.js`.

src/server/internalServer.cpp
```diff
@@ -658,7 +658,9 @@ std::vector<std::string>
 InternalServer::search_catalog(const RequestContext& request,
                                kiwix::OPDSDumper& opdsDumper)
 {
-    auto filter = kiwix::Filter().valid(true).local(true).remote(true);
+    // added for discussion in #530 using the following method we can deliver zim file irrespective of a remote URl present for it.
+    auto filter = kiwix::Filter().valid(true).local(true);
+    // auto filter = kiwix::Filter().valid(true).local(true).remote(true);
```
MananJethwani, line 663: please let me know if this is fine so that I will remove the comments.

static/skin/index.js
```diff
@@ -0,0 +1,75 @@
...
+        incrementalLoadingParams.count = incrementalLoadingParams.count && viewPortToCount();
+        if (window.innerHeight + window.scrollY >= document.body.offsetHeight) {
+            loadAndDisplayBooks();
+        }
+    });
+
+    window.addEventListener('scroll', async () => {
+        if (isFetching || !incrementalLoadingParams.count) return;
+        if (window.innerHeight + window.scrollY >= document.body.offsetHeight) {
+            loadAndDisplayBooks();
```
MananJethwani, line 68: these 2 lines and L60,61 are the same should I introduce another function? I am a bit confused as would it be over-doing things?

static/skin/index.js
```diff
@@ -1,71 +1,75 @@
...
-            incrementalLoadingParams['count'] = 0;
+    async function loadAndDisplayBooks() {
+        isFetching = true;
+        fetch(queryUrlBuilder()).then(async (resp) => {
+            const data = new window.DOMParser().parseFromString(await resp.text(), 'application/xml');
+            const books = data.querySelectorAll('entry');
+            let bookHtml = '';
+            books.forEach((book) => {bookHtml += generateBookHtml(book)});
+            document.querySelector('.book__list').innerHTML += bookHtml;
+            incrementalLoadingParams.start += incrementalLoadingParams.count;
```
veloman-yunkan, line 49: Currently, it doesn't make any difference, but it's more future-proof to advance `incrementalLoadingParams.start` by the count of items actually loaded.

static/skin/index.js
```diff
@@ -1,71 +1,75 @@
...
+            if (books.length < incrementalLoadingParams.count) {
+                incrementalLoadingParams.count = 0;
+            }
+            isFetching = false;
+        });
+    }
+
+    window.addEventListener('resize', async () => {
+        if (isFetching || !incrementalLoadingParams.count || incrementalLoadingParams.count > viewPortToCount()) return;
```
veloman-yunkan, line 58: What is the justification for the `incrementalLoadingParams.count > viewPortToCount()` check?

veloman-yunkan (changes requested):
Looks like this is the last iteration. After addressing the few comments included in this review, please clean-up your PR history, so that we can merge without an extra cycle.

static/skin/index.js
veloman-yunkan: You can move this check into the `loadAndDisplayBooks()` function.

veloman-yunkan:
>  I have added the closure

Great! I would welcome it even in the absence of any current conflict. The cost of doing it is minimal, while it can save a lot of debugging time in the future.

static/skin/index.js
MananJethwani: it is not a good solution as in the case of filters the process is different. I would like to keep it as it is, if it is fine with you:sweat_smile:

static/skin/index.js
MananJethwani: to check if the screen size is reduced no new data should be fetched in that case until and unless we are at the bottom of the page.

src/server/internalServer.cpp
MananJethwani: @veloman-yunkan what about this?

static/skin/index.js
veloman-yunkan: My point is that the `incrementalLoadingParams.count > viewPortToCount()` check is redundant. The bottom-of-the-page test is enough, isn't it? If so, then handling of resize and scroll events becomes identical, and then introducing another function is justified.

veloman-yunkan: > The bottom-of-the-page test is enough, isn't it?

Though this depends on when the resize event fires - before or after the layout is updated.

MananJethwani: oh sorry, I forgot about that part you are right it should work fine if we are including the bottom of the page test.
also, the event is fired after the layout is updated.

veloman-yunkan (approved):

src/server/internalServer.cpp
```diff
@@ -658,7 +658,7 @@ std::vector<std::string>
 InternalServer::search_catalog(const RequestContext& request,
                                kiwix::OPDSDumper& opdsDumper)
 {
-    auto filter = kiwix::Filter().valid(true).local(true).remote(true);
+    auto filter = kiwix::Filter().valid(true).local(true);
```
mgautierfr, line 661: Why removing this ?

OPDS stream list zim file a user can download. If we remove the `remote(true)` filter, we will return zim file without url to download from.

static/templates/index.html
```diff
@@ -1,66 +1,96 @@
...
-  <div class='book__list'>
-    {{#books}}
-    <a href="{{root}}/{{name}}"><div class='book'>
-        <div class='book__background' style="background-image: url('{{root}}/meta?content={{#urlencoded}}{{{name}}}{{/urlencoded}}&name=favicon');">
-          <div class='book__title' title='{{title}}'>{{title}}</div>
-          <div class='book__description' title='{{description}}'>{{description}}</div>
-          <div class='book__info'>{{articleCount}} articles, {{mediaCount}} medias</div>
-        </div>
-    </div></a>
-    {{/books}}
```
mgautierfr, line 57: If we don't use the `books` variable in the template, we should not set it in the cpp side.

mgautierfr (commented):
This implementation has race condition.

You can reproduce it by using the "adaptive view" in firefox developer tools :
- Set the bandwidth to regular 2G
- Set the windows size to make the js NOT download all available books.
- While it is loading, resize the window.
- Js loads (and display) several times the same book (and some books are missing)

src/server/internalServer.cpp
MananJethwani: @mgautierfr kelson discussed this point earlier in the PR and we agreed that we should show all zim files and show a disabled download button(and a tooltip) for those zim files that can't be downloaded.

static/templates/index.html
MananJethwani: I agree on this point I was going to open a separate issue for this, but if you want it removed from cpp side I will create a new PR and add that part in some time.

src/server/internalServer.cpp
mgautierfr: The problem is that ODPS stream is already used (and designed for) native client which list what zim files user can download.
By definition, the user can download a zim file if they have a associated url (they are "remote").
If you remove this filter, you break the API for existing client.

static/templates/index.html
mgautierfr: Why the next time. I don't "want" to remove the cpp time. I want to have the template and the cpp in sync.

src/server/internalServer.cpp
kelson42: @mgautierfr Can you please be specific? I'm not aware of anything broken.

mgautierfr: kiwix-desktop uses OPDS stream to propose zim file to download to the user. If OPDS stream now return one zim file without "url" attribute, kiwix-desktop will display to book to the user but he will not be able to download it.
(I've checked and kiwix-desktop reapply this filter internally before displaying books to the user, but other implementation (kiwix-android, kiwix-ios)) may not do the same).
(And this is maybe hypotical anyway as most (if not all) zim files in library.kiwix.org have a url, but if someone use a local opds stream, it will be broken)

This is a OPDS API break, we cannot change the content returned by a stream like that.
We may want to change this behavior or not (we can extend the catalog/v2 to handle this feature and use catalog/v2), but if we change it we have to be careful about this.

static/templates/index.html
MananJethwani: by remove, I meant that only😅.... I will bring them in sync.

src/server/internalServer.cpp
MananJethwani: @mgautierfr @kelson42 instead of creating a completely new endpoint wouldn't it be better if I just introduce a new search parameter that will check if we need all zim files and changes filter accordingly?

kelson42: @mgautierfr Kiwix-Desktop works perfectly, so there is no problem, there is no API break, there is no documentation to update. What you are talking about seems to be a potential problem if for some reason library.kiwix.org would deliver an OPDS stream with books without URLs. An "yes" I confirm, this is the resonsability of the person running the kiwix-serve to have a `library.xml` with URL if he wants that these books are downloadable. What is *concretly* wrong here or broken?

PR #535, opened 2021-05-22
Rename kiwix-lib in libkiwix

kelson42:
Change last occurences of "kiwix-lib" in "libkiwix" for #386.


veloman-yunkan (commented):
I am not sure if it is wise retrospectively renaming kiwix-lib in the change log. I think that it makes sense to note in the change log that this library was named `kiwix-lib` before version 10 and preserve its name for that period.

kelson42:
@veloman-yunkan OK, quite agnostic on this point. Will rollback and merge.

PR #537, opened 2021-05-24
Update libkiwix with search iterator rename in libzim

maneeshpm:
This PR braces libkiwix for openzim/libzim#563

Search iterator API in libzim has been shifted to use camel case naming. This has to be accommodated in libkiwix as well.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -507,9 +507,9 @@ bool Reader::searchSuggestionsSmart(const string& prefix,
          current != suggestions.end();
          current++) {
       std::vector<std::string> suggestion;
-      suggestion.push_back(current->getTitle());
-      suggestion.push_back(current->getPath());
-      suggestion.push_back(kiwix::normalize(current->getTitle()));
+      suggestion.push_back(current.getTitle());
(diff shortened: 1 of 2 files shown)


src/reader.cpp
```diff
@@ -507,9 +507,9 @@ bool Reader::searchSuggestionsSmart(const string& prefix,
          current != suggestions.end();
          current++) {
       std::vector<std::string> suggestion;
-      suggestion.push_back(current->getTitle());
-      suggestion.push_back(current->getPath());
-      suggestion.push_back(kiwix::normalize(current->getTitle()));
+      suggestion.push_back(current.getTitle());
+      suggestion.push_back(current.getPath());
+      suggestion.push_back(kiwix::normalize(current.getTitle()));
```
mgautierfr, line 512: Why this change ?
Does `current` have change type ?

mgautierfr: I've got my answer, you use directly the `getTitle` of the iterator instead of getting a pointer to the entry and then call `getTitle`. We are good here.

mgautierfr (approved):

PR #538, opened 2021-05-25
Revert "Kiwix Serve welcome page dynamic and subset loading (OPDS based)"

mgautierfr:
Reverts kiwix/libkiwix#530

The PR #530 has issues. It add potential race condition and change the opds behavior to return unwanted zim file.
@MananJethwani please open a new PR with the content of #530 and the fixes for my review.


kelson42:
@mgautierfr I don't understand your comment, it lacks details. Which race condition exactly? Which unwanted zim files?

mgautierfr:
> Which race condition exactly? 

I've explained in the review of https://github.com/kiwix/libkiwix/pull/530#pullrequestreview-665517455

> This implementation has race condition.
>
>You can reproduce it by using the "adaptive view" in firefox developer tools :
>
> - Set the bandwidth to regular 2G
> - Set the windows size to make the js NOT download all available books.
> - While it is loading, resize the window.
> - Js loads (and display) several times the same book (and some books are missing)


> Which unwanted zim files?

The OPDS stream is used by native client to know and display to the user the books he can download.
If the OPDS stream return "non remote" zim file, the client will display them.

PR #507, opened 2021-05-08
/catalog/search handles out-of-bounds pagination

veloman-yunkan:
Fixes #504

PR #539, opened 2021-05-25
Avoid windows header to define min/max macros.

mgautierfr:
PR #507 use std::min.
But on windows, the header define min and max macros and so the
compilation is broken.
Add `-NOMINMAX` define to avoid that.


veloman-yunkan (approved):

kelson42:
@mgautierfr Unfortunately, there is still a problem, the CI does not pass everywhere.

PR #540, opened 2021-05-25
added dynamic and subset loading of zim-files in kiwix-serve

### LINKED ISSUE #511: Kiwix serve welcome page should be dynamic

MananJethwani:
this fixes #511 and #514.

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -281,27 +281,6 @@ MustacheData InternalServer::get_default_data() const
   return data;
 }
 
-MustacheData InternalServer::homepage_data() const
-{
-  auto data = get_default_data();
-
-  MustacheData books{MustacheData::type::list};
-  for (auto& bookId: mp_library->filter(kiwix::Filter().local(true).valid(true))) {
-    auto& currentBook = mp_library->getBookById(bookId);
-
-    MustacheData book;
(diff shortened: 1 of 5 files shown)


src/server/internalServer.cpp
```diff
@@ -658,7 +637,7 @@ std::vector<std::string>
 InternalServer::search_catalog(const RequestContext& request,
                                kiwix::OPDSDumper& opdsDumper)
 {
-    auto filter = kiwix::Filter().valid(true).local(true).remote(true);
+    auto filter = kiwix::Filter().valid(true).local(true);
```
MananJethwani, line 640: @mgautierfr kelson discussed this point earlier in #530 and we agreed that we should show all zim files and show a disabled download button(and a tooltip) for those zim files that can't be downloaded.

mgautierfr: On the kiwix-serve side yes. But OPDS stream is also used (and designed for) native client. You cannot change the behavior here as you may break them.

kelson42: @mgautierfr ok, what is broken concretly? What would be a solution?

Issue #549, opened 2021-06-07
Use svelte for our js code ?


mgautierfr:
I've recently discovered [Svelte](https://svelte.dev/)
It is a framework to create UI in html/css/js.
In opposition to other framework (vue, react, jquery) svelte works at compilation time and generate output (css/js) only containing what we need and want.

Pro:
- No need to embed libraries (jquery, isotope (#534))
- Only what we need for fast loading
- High level way of programming (component, reactive ui, animation, ...)

Cons:
- Need to generate ("compile") the css/js code along with the cpp code. So we need to setup a node/svelte build env on developer machine (and CI). This could be mitigate by keeping the generated css/js in the git repository. Then, only frontend dev would have to setup de build env. Other dev would simply use the already generated files.

kelson42:
We have discussed this. This is an intetesting proposal. But the main problem we have is our dependence to jquery... and this dependence will vanish anyway.

To me the main question is: which practical improvement will that bring considering our actual Javascript code base?

mgautierfr:
- We would not embed isotope neither.
- As we don't embed a library, we ask user to load (and our server to send) less data. Our binaries don't embed it and so, are smaller.
- As we don't embed a library, we don't have to keep it up to date and we potentially have less security issue

kelson42:
@mgautierfr Ok, so the main advantage is to remove js isotop? If we do so the look&feel woukd be kept? Woukd you recode all the isotop library logic?

mgautierfr:
> so the main advantage is to remove js isotop? 

Yes, but also help us to implement all our frontend (and help us with new feature as incremental search or widget)

> If we do so the look&feel would be kept?

We just have to keep our current css

> Woukd you recode all the isotop library logic?

Mostly no. isotop is just a "flexbox" container with few animations. (and svelte handle animation pretty nicely)

---

Here a small POC of what it is possible to do : https://svelte.dev/repl/2b6fd14cbd454b339dc8c55ea32d67e0?version=3.48.0

Please not that is it is a POC made in less than two hours. Style is not identical. There is no download link and so.
But you can see that:
- It is working (fast)
- Code is pretty clear
- Tiling is made with flexbox, which is standard in css3
- Styling is a copy paste of our `style.css` (+ flexbox)
- No code for transition (but pretty easy to add)
- No history/cookies/... management
- No pagination
- The only files to embed would be the js and css generated

[...]

PR #559, opened 2021-06-18
Revamped Kiwix Serve Welcome page layout

MananJethwani:
@kelson42 this fixes kiwix/kiwix-tools#448 and fixes https://github.com/kiwix/kiwix-tools/issues/370
for now, I have changed the basic layout and updated the cards.


kelson42 (changes requested):
Lets go with the review:
* The "Download" button should point to the ZIM file (to download), not the the demonstration link like today. We should have a "Download" link and a "View" link.
* The style of the download/view links is not link in the mockup
* The size of the ZIM file is not displayed, it should (and it should be rounded properly)
* Each tile background should be grey, not white
* The tile in itself should not be clickable (only the "View" and "Download" links should be clickable).

MananJethwani:
@kelson42 I have updated everything except this -> 
 > Each tile background should be grey, not white

The background color is the same as that in the mock-up and I couldn't find a better color that suits the card background, If you think some other color would be better I am open to suggestions (kindly try to send some color code thx😅 ).

kelson42:
@MananJethwani Please rebase on master an request a new review if you are so far.

kelson42:
@MananJethwani Can you please rebase?

MananJethwani:
@kelson42 I think this PR is properly rebased.

kelson42 (changes requested):
* There is no "url", then the download button should not be displayed. It makes no sense to display a "download" label with link.
* If there is no "size", nothing should be displayed ("unknown" for the moment)
* If there is a "size", it should be below the title (not like in the mockup, which is wrong in my sense).
* Whatever there is a size or not, both title (and size) should centered verticaly on the icon height
* The tile background is not right like I said earlier already (you need to verify precisely). It's `#fcfcfc` and it should be `#fbfbfb`.
* The font for "Preview" and "Download" is not the right one

kelson42 (changes requested):
Looks better, but the icon size should be  fixed. See:
image


If the favicon has the wrong size it will break the rendering. The size of the icon should be like in the mockup, and the iconv given in the zim should be scaled on the fly by the browser if necessary.

Otherwise, I think it is time now to have a look to the styling of the selec boxes.

Few other remarks:
* "reset filter" link should not include the space before and the question mark.
* "Results: 1487 items" should be printed just on the top of the tiles (so not like in the mokcup) and only be "1487 items"
* Underscore replacement in tags, only applies once. "science in_the_bath" is replace with "science_in_the_bath", it should be "science in the bath"

kelson42 (changes requested):
New remarks:
* If no result, the "0 books" should be hidden
* The blue-grey color of the selector shoudl be replace with `#909090`
* The drop-down boxes are not styled at all. They should.
* The followin screenshot has 3 problems: 
image
 - the grey tobar height is too big
 - the category button is not aligned witht the search button
 - the vertical space above "12 books" is smaller than the one below (the one below should be like the one above).
* Please remove the "effect" on preview/download. Just color the background when we are over (but no filling effect).

MananJethwani:
@kelson42 I have fixed everything as asked, but the dropdown list styling doesn't work in the case of firefox and I can't seem to find a proper reason for that, might need some help here, I will keep on looking, for now, please check everything else.

kelson42 (changes requested):
Why "Category" selectbox is still not aligned with the "Search" button?
image

pulldown parts of select boxes are still not styled

New request: please make first letter uppercase for tags.

If there is only one book, it should not be "1 books", but "1 book"

MananJethwani:
@kelson42 regarding the misalignment of boxes please mention the exact width and height of the screen as when I test it on multiple screen sizes and this works fine and I can't seem to have this issue,
as for the dropdown styling please open it on chrome or brave browser it doesn't seem to work on Mozilla and I asked for help regarding that above already

MananJethwani:
found the alignment bug it was on screen size greater than 2100px sorry for that I checked and resolved it for smaller screen size.
still, only the dropdown styling problems persists.

mgautierfr:
I've quickly reviewed it. On the technical side I've nothing to say (but I no a js expert).

I would like to mention https://github.com/kiwix/libkiwix/issues/549. We start to generated all our html using custom js code and svelte is made especially for this.

mgautierfr (approved):

kelson42 (approved):

PR #561, opened 2021-06-21
Keep Kiwix Serve filter values over time

### LINKED ISSUE #557: Kiwix Serve filters on welcome page should be kept
Steps:
* Choose a few filters
* Click on a content and surf
* Comeback to the homepage

I get:
* The default filter

I want:
* The filters like I left them

MananJethwani:
@kelson42 this fixes #557

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -6,13 +6,14 @@
     };
     const filterTypes = ['lang', 'category', 'q'];
     const bookOrderMap = new Map();
+    const filterCookieName = 'filters'
     let footer;
     let fadeOutDiv;
     let iso;
     let isFetching = false;
     let noResultInjected = false;
-    let filters = getCookie('filters');
-    let params = new URLSearchParams(filters || '');
+    let filters = getCookie(filterCookieName);
+    let params = new URLSearchParams(window.location.search || filters || '');
     let timer;
 
     function queryUrlBuilder() {
@@ -180,7 +181,7 @@
         if (filterType) {
             params.set(filterType, filterValue);
             window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
-            setCookie('filters', params.toString());
+            setCookie(filterCookieName, params.toString());
         }
         await loadAndDisplayBooks(true);
     }
@@ -242,5 +243,6 @@
             langFilter.value = browserLang.length === 3 ? browserLang : iso6391To3[browserLang];
             langFilter.dispatchEvent(new Event('change'));
         }
+        setCookie(filterCookieName, params.toString());
     }
 })();
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -6,7 +6,8 @@
     };
     const filterTypes = ['lang', 'category', 'q'];
     const bookOrderMap = new Map();
-    const filterCookieName = 'filters'
+    const filterCookieName = 'filters';
+    const oneDayDelta = 86400000;
     let footer;
     let fadeOutDiv;
     let iso;
@@ -25,7 +26,7 @@
 
     function setCookie(cookieName, cookieValue) {
         const date = new Date();
-        date.setTime(date.getTime() + ( 24 * 60 * 60 * 1000));
+        date.setTime(date.getTime() + oneDayDelta);
         document.cookie = `${cookieName}=${cookieValue};expires=${date.toUTCString()};sameSite=Strict`;
     }
 
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -126,6 +126,7 @@
                 spanTag.getElementsByTagName('a')[0].onclick = (event) => {
                     event.preventDefault();
                     window.history.pushState({}, null, `${window.location.href.split('?')[0]}?lang=`);
+                    setCookie(filterCookieName, 'lang=');
                     resetAndFilter();
                     filterTypes.forEach(key => {document.getElementsByName(key)[0].value = params.get(key) || ''});
                 };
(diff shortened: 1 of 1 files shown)


kelson42 (changes requested):
Please rebase on master, without including the revamping of cards.

kelson42 (changes requested):
The problem now is that the URL given is not taken in account anyway. If I put localhost:8082/?lang=fra in the address bar it is automatically changed to the cookie values. The cookie values should apply only if no corresponding URL paramter is given.

kelson42 (changes requested):
If is set a filter with results, then set a language with no result, then click on reset filter. Then load a content and come back with the Home button. The filter which comes back is the one without result. So the "result filter" is not kept.

kelson42 (approved):
From a user perspective, seems to work as expected. Lets go forward with the code review.

static/skin/index.js
```diff
@@ -21,6 +23,23 @@
         return (url);
     }
 
+    function setCookie(cookieName, cookieValue) {
+        const date = new Date();
+        date.setTime(date.getTime() + ( 24 * 60 * 60 * 1000));
```
kelson42, line 28: space incoherence

static/skin/index.js
```diff
@@ -5,13 +5,15 @@
         count: viewPortToCount()
     };
     const filterTypes = ['lang', 'category', 'q'];
-    const bookMap = new Map();
+    const bookOrderMap = new Map();
```
mgautierfr, line 8: Why rename this to `bookOrderMap` ?

Are the books ordered ?

static/skin/index.js
mgautierfr: It would be nice to have a comment telling that `24*60*60*1000` is a day.
And even better, having a const name "1daydelta", computed once and doing the job of the comment.

static/skin/index.js
```diff
@@ -212,18 +233,17 @@
             const filterTag = document.getElementsByName(filter)[0];
             filterTag.addEventListener('change', () => {resetAndFilter(filterTag.name, filterTag.value)});
         });
+        if (filters) {
+            window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
```
mgautierfr, line 237: Do we really need this ? (Or at least, do we need it all the time ?)

It seems we pass here when we have loaded the page (so after a the user arrive to the page).
So the url should already be in the history.
We may want to pushstate if we modify the "url" internally but this is not the case here.

mgautierfr (commented):
I'm not a js expert so I may have missed things.

static/skin/index.js
MananJethwani: @mgautierfr yes the books are ordered and this is something that was requested by veloman in my previous PR but got missed so adding it here.

static/skin/index.js
MananJethwani: @mgautierfr this is needed here as in the case we get params from cookies the URL is not updated and 
L238 -> `params.forEach((value, key) => {document.getElementsByName(key)[0].value = value});`
doesn't dispatch an on change event it just changes the value if I call a dispatch event here it will mean we will be pushing into history for each separate filter so if the user wants to go back he would have to click 3-4 time which wouldn't be a good thing and as the filters were called together there should be only one step for them.

static/skin/index.js
mgautierfr: Haa, I've missed that the map is not a map between a book and a id but is a map between a book id and the order of the book.
We are good. (But it would have been nice to have this in a separated commit. It has nothing to do with "added cookies for filter effect". And the commit message could have explain why we've changed this.)

static/skin/index.js
mgautierfr: Couldn't we dispatch a event after L238 ?
Once we have changed all the values, we "apply" them.

---

It seems to me that if push a new history state here and the user go the url with cookies, we will have two history state:
- The url without param.
- The url with param (taken from cookies).

But the two are somehow the same state has the url without param will be initialized the same way than the second (and will insert a new state).
So if the user want to go back, he would have to click 2 times (fast enough to not let the page load the fist time)

MananJethwani: > Couldn't we dispatch a event after L238 ?
Once we have changed all the values, we "apply" them.

we can't dispatch a single event to change the URL and apply the filters all together as it calls `resetAndFilter` which only changes URL (by pushing into history only if there is a single filter change that is passed through it).
and even if we could it would still mean we would be pushing a new state into history.

>  It seems to me that if push a new history state here and the user go the url with cookies, we will have two history state:
The url without param.
The url with param (taken from cookies).
But the two are somehow the same state has the url without param will be initialized the same way than the second (and will insert a new state).
So if the user want to go back, he would have to click 2 times (fast enough to not let the page load the fist time)

agreed upon this part but at this moment I couldn't think of something else that would have a single back click as we are changing the URL through JS as always.
this similar thing happens when there is no cookie saved and user goes to the site and browser language is detected and a filter is applied in that case as well user has to click 2 times.

kelson42: I had a look and I believe to understand the situation. On my side I requested that @MananJethwani secures that the filter status is always synced with the URL params (I believe the reasons are obvious). This forces him to add an entry in the history to set the URL params properly based on the cookies. We get then 2 entries. Obviously it is possible to to load from the cookies and don't update the URL, but then my request would not be respected anymore. A choice has to be made. In my case I strongly prefer the current situation, ie. garanty that a user can copy the URL and will get the same result again.

kelson42:
I think we can merge like this now. Like always, if anything new is detected, we can fix it in a new PR.

PR #562, opened 2021-06-24
JNI search  methods for android

gouri-panda:
Created all search methods that we need for searching in `kiwix-android`
libkiwix already has filters and class that we want to search e.g [Book](https://github.com/kiwix/libkiwix/blob/master/src/wrapper/java/org/kiwix/kiwixlib/Book.java) and [Filter](https://github.com/kiwix/libkiwix/blob/master/src/wrapper/java/org/kiwix/kiwixlib/Filter.java) that we will reuse in android to filter all sort of things.
After this process will be in android
1. libkiwix will build the URL 
2. client shares the OPDS feed with libkiwix which parse it and in returns gives the list of books as an object

### DIFF
--- a/src/wrapper/java/org/kiwix/kiwixlib/Author.java
+++ b/src/wrapper/java/org/kiwix/kiwixlib/Author.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2019-2021 Kiwix
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


gouri-panda:
@kelson42 @mhutti1 @abdulwd I think that's what we need to replace to search the OPDS URL call and retrieve the OPDS feed.

kelson42:
@gouri-panda thx, but why this pr is here and not in repo openzim/java-libzim?

gouri-panda:
@kelson42 `openzim/java-libzim` isn't functional yet. When that will happen I will remove this work to that repo. Is it ok with you?

kelson42:
It isn't functional here neither because we can not publish to bintray anymore. What is the plan to fix that problem?

gouri-panda:
@kelson42 I am planning to build locally to work in android for a short amount of time.If it is possible I'll try to create a PR by the end of today.

kelson42:
Ok, but none of this will pass the CI, because it is not in bintray/maeven... so it won't be really mergeable.

gouri-panda:
>but none of this will pass the CI, because it is not in bintray/maeven.

@kelson42 I'm sorry, but what isn't in bintrary/maevn?

mgautierfr:
We should move to java-libkiwix the sooner possible. Having two repository in parallel will complicate a lot the process (if we merge this PR here, we will have to backport the change in java-libkiwix, with all the complexity of the a conflicting merge between two different repository)

PR #563, opened 2021-06-24
Added maven central configuration for kiwix-android

gouri-panda:
Tasks:- 
- [x] Added maven central dependencies
- [ ] Write/modify auto publication of pom.xml from bintray to maven
- [ ] See if it builds locally successfully
 After it builds successfully @kelson42 will do the remain stuff
- [ ] Create gpg-key for this
- [ ] Add gpg key and other sona types secret key to github secret environment


mgautierfr:
I think we should not add maven configuration to libkiwix but to https://github.com/kiwix/java-libkiwix

Issue #564, opened 2021-06-24
Do not expose mustache (and other depedency/tools) in public headers.


mgautierfr:
The public header otherTools.h include mustache.h (https://github.com/kiwix/libkiwix/blob/master/include/tools/otherTools.h#L27).
By doing thing, we defacto make mustache a direct dependency of projects using libkiwix.
If somehow `otherTools.h` is include by a project foo. The mustache header must be present even if foo never use mustache.
This is what happen here https://github.com/kiwix/kiwix-desktop/runs/2904840393?check_suite_focus=true (PR https://github.com/kiwix/kiwix-desktop/pull/656) where the compilation of kiwix-desktop fails as mustache.hpp is not present.

We must not include header of our private dependencies in public our headers.

More than just cleaning the mustache inclusion (and other tools, we also include `zim.h` and we should not), we should probably carrefully select what helper function are part of our API and move all other tools private.

legoktm:
> More than just cleaning the mustache inclusion (and other tools, we also include zim.h and we should not),

In the Debian package we currently depend on:
```
 libzim-dev (>= 6.0.0),
 libicu-dev,
 libpugixml-dev,
 libcurl4-gnutls-dev,
 libmicrohttpd-dev
```
since those are exposed as part of the public API. These are also listed in kiwix.pc.

PR #569, opened 2021-06-27
debian: Have libkiwix-dev depend on libkainjow-mustache-dev

legoktm:
Software trying to build against libkiwix-dev now need to have
libkainjow-mustache-dev installed since tools/otherTools.h includes it.

This should fix the PPA builds in kiwix-tools.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -27,7 +27,8 @@ Depends: libkiwix10 (= ${binary:Version}), ${misc:Depends}, python3,
  libicu-dev,
  libpugixml-dev,
  libcurl4-gnutls-dev,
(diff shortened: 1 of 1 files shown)


legoktm:
It looks like this was pretty recent, cdacc0caf183023a4519a307fe3f1a229a6bc18f, based on the discussion in that PR it seems like the dependency on mustache in the libkiwix headers was intentional.

kelson42:
This is probably invalidated by #564 and should therefore not be merged.

legoktm:
Yep, that would also fix it.

PR #570, opened 2021-06-27
[WIP] Extend libkiwix structures to use libzim

### LINKED ISSUE #430: Drop libzim wrapper

maneeshpm:
Fixes #430 
This PR depends upon #536 

This is the second stage to remove usage of wrapper structures from libkwix. In this step, we will extend the libkiwix structures to use libzim structures directly. This will still include the wrapper structures in the API but their usage will be derived from that of libzim structures.

Changes included in this PR:
- Add overloads for libzim structure in `kiwix::Book`

TODO:
- Add overloads in manager


maneeshpm:
Closing this in favour of kiwix/libkiwix#576.

PR #574, opened 2021-06-30
Fix public headers inclusion (+ small other fixes)

### LINKED ISSUE #564: Do not expose mustache (and other depedency/tools) in public headers.

mgautierfr:
Fix #564

This also fix some declaration of `SuggestionItem` methods.

~~On recent gcc, httplib raise warning and it break compilation with `-WError`.
This PR include a patch for this. Upstream patch is here : https://github.com/yhirose/cpp-httplib/pull/980~~
This has been removed from the PR as it fix a issue I've got only on my computer and the fix make macos compilation fail.


include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+ * real executable is different of the path of the archive.
+ * If `realPathOnly` is true, return the path of the real executable instead of the
+ * archive launched by the user.
+ *
+ * @param realPathOnly If we must return the real path of the executable.
+ * @return the path of the executable (utf8 encoded)
+ */
+std::string getExecutablePath(bool realPathOnly = false);
+
+/** Tell if the pat is a relative path.
```
veloman-yunkan, line 67: pat → path

include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+ *
+ * This function is provided as a small helper. It is probably better to use native tools
+ * to manipulate pathes.
```
veloman-yunkan, line 70: pathes → paths

include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+/** Append a path to another one.
+ *
+ * This function is provided as a small helper. It is probably better to use native tools
+ * to manipulate pathes.
+ *
+ * @param basePath the base path.
+ * @param relativePath a path to add to the base path, must be a relative path.
+ * @return The concatenation of the pathes, using the right separator.
+ */
+std::string appendToDirectory(const std::string& directoryPath, const std::string& filename);
```
veloman-yunkan, line 86: mismatch between the param names in the docstring and the function declaration

include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+
+/** Remove the last element of a path.
+ *
+ * This function is provided as a small helper. It is probably better to use native tools
+ * to manipulate pathes.
+ *
+ * @param path a path.
+ * @return The parent directory (or empty string is none).
```
veloman-yunkan, line 94: is → if

include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+
+/** Remove the last element of a path.
+ *
+ * This function is provided as a small helper. It is probably better to use native tools
+ * to manipulate pathes.
+ *
+ * @param path a path.
+ * @return The base name of the path or empty string is none (ending with a separator).
+ */
+std::string getLastPathElement(const std::string& path);
```
veloman-yunkan, line 106: 1. Mismatch between the function name and the first line of the docstring (a copy-paste error).
2. is → if (in the last line)

include/tools.h
```diff
@@ -20,4 +20,144 @@
...
+ *
+ * This function is provided as a small helper. It is probably better to use native tools.
+ *
+ * @param str The string to split.
+ * @param delims A string of potential delimiters.
+ * @param trimEmpty true if empty part must be trimmed from the result.
+ * @param keepDelim true if delimiter must be included from the result.
+ * @return a list of part (potentially containing delimiters)
+ */
+std::vector<std::string> split(const std::string& str, const std::string& delims, bool trimEmpty=true, bool keepDelim = false);
```
veloman-yunkan, line 150: 1. The effect of `trimEmpty` is not clear (probably because the term *trim* has a slightly different meaning in text processing related utilities). Maybe, renaming this parameter to `dropEmpty` will eliminate the confusion
2. Regarding `keepDelim` - where exactly in (rather than from) the result is the delimiter included

veloman-yunkan (changes requested):

include/tools.h
maneeshpm: Please note that this typo is present in other function docstrings as well.

mgautierfr: I don't know why my brain keep thinking that paths is written pathes. Even if **I** know it is not the case.

include/tools.h
mgautierfr: Done.
I've also add a small example.

include/tools.h
```diff
@@ -141,6 +141,13 @@ void sleep(unsigned int milliseconds);
  *
  * This function is provided as a small helper. It is probably better to use native tools.
  *
+ * Assuming text = "foo:;bar;baz,oups;"
+ *
+ * slit(text, ":;", true, true) => ["foo", ":", ";", "bar", ";", "baz,oups", ";"]
```
maneeshpm, line 146: A minor typo.

include/tools.h
```diff
@@ -150,11 +150,11 @@ void sleep(unsigned int milliseconds);
  *
  * @param str The string to split.
  * @param delims A string of potential delimiters.
- * @param trimEmpty true if empty part must be trimmed from the result.
+ * @param dropEmpty true if empty part must be dropped from the result.
  * @param keepDelim true if delimiter must be included from the result.
  * @return a list of part (potentially containing delimiters)
  */
-std::vector<std::string> split(const std::string& str, const std::string& delims, bool trimEmpty=true, bool keepDelim = false);
+std::vector<std::string> split(const std::string& str, const std::string& delims, bool dropEmpty=true, bool keepDelim = false);
```
maneeshpm, line 157: The example suggests that each character of the string `delims` is a potential delimiter. It can be easily confused with the entire string being a single delimiter. I think we should mention this explicitly in the parameter docstring.

maneeshpm (changes requested):
Just a couple of comments for better docstring

mgautierfr:
Rebase on master (fixing the conflict (new archiveTools.h)).

Fix last review comments of @maneeshpm

maneeshpm (approved):

kelson42:
This branch needs to be rebased

kelson42:
IMO we can merge now.

PR #575, opened 2021-06-30
Patch httplib.h for gcc `maybe-uninitialized` warning.

mgautierfr:
This is the removed commit from #574.

(Upstream issue : https://github.com/yhirose/cpp-httplib/pull/980)

This patch is not compatible with macos.


kelson42:
@mgautierfr What shoukd we do with this PR?

PR #583, opened 2021-07-12
Modal download  box on Kiwix Serve welcome page

### LINKED ISSUE #579: Kiwix Serve welcome page Download overlay
To propose, based on the metalink XML:
*  Magnet links
* Hashes
* Bittorrent links
* Usual HTTP download link

This should be displayed if metalink is available and if someone clicks on a tile download button.

The parsing of the metalink file should be smart, ie. we download/parse only the part of interset in the XML and then stops. We are not interested in all the hashes.

MananJethwani:
@kelson42 this fixes #579


kelson42 (changes requested):
In case of the magnet, can you please just read the file in js and load it directly? So the user will automatically get an invitation to open it via its favourit Bittorent reader.

MananJethwani:
@kelson42 I have done the changes as you asked regarding the working of magnet link but there is something that bothers me a bit, when we click on a magnet link our js will fetch the actual magnet link from it using fetch and apply window.location but sometimes this fetch process takes longer than expected and the browser sits idle appearing like the button didn't work, should we add a loader of some sort for that case?

legoktm:
@MananJethwani did you create the images yourself or were they copied from somewhere? If you could indicate what the source is that would be appreciated!

MananJethwani:
@legoktm I don't have the exact link, but I remember I used free icon packs from flaticon or icon8.

legoktm:
Thanks for the quick response! Unfortunately, I don't think either of those license terms are acceptable at a quick glance...

https://www.freepikcompany.com/legal#nav-flaticon-agreement has a bunch of restrictions on use, and https://intercom.help/icons8-7fb7577e8170/en/articles/5534926-universal-multimedia-licensing-agreement-for-icons8 has restrictions on derivative works.

I'm sure we can find equivalent icons from Wikimedia Commons that are really freely licensed...

legoktm:
Actually, I take that back, all of these icons are pretty basic geometric shapes that are likely ineligible for copyright. And Commons already treats the BitTorrent logo like that (c.f. https://commons.wikimedia.org/wiki/File:BitTorrent_logo.svg). I think we're all OK here, if Debian has concerns I'll comment again.

PR #587, opened 2021-07-19
added nofilter attribute and iframe conditions

### LINKED ISSUE #585: kiwix-serve (ZIM?) should propose widgets
_From @kelson42 on June 19, 2016 15:28_

People using kiwix-serve within integrated solution like Rachel, XSCE, Ideascube use it with other content coming from other sources. They need at some point to add a menu entry to point to that content. It might be interesting for kiwix-serve to propose an easy to integrate widget to make their life easier. This widget beside describing the content would for example provide a searchbox which in one click would allow to find a content.

_Copied from original issue: kiwix/kiwix-xulrunner#245_

MananJethwani:
this resolves #585

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -16,6 +16,8 @@
     let noResultInjected = false;
     let filters = getCookie(filterCookieName);
     let params = new URLSearchParams(window.location.search || filters || '');
+    let noFilter = params.has('nofilter') && params.get('nofilter');
+    let inIframe = false;
     let timer;
 
     function queryUrlBuilder() {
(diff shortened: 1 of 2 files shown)


MananJethwani:
To include a widget on the MediaWiki server we will require https://www.mediawiki.org/wiki/Extension:Iframe, and to make it a bit responsive we need to use `width:100%` for some reason vh and vw don't work for the MediaWiki based servers.

mgautierfr:
@MananJethwani Is this PR a draft or ready to be reviewed ?

MananJethwani:
@mgautierfr it's still a draft, @kelson42 asked me too test it on a wiki but for some reason I am not able to use widget extension for mediawiki as of now....will have a look in some time after solving some other tickets.

kelson42:
@juuz0 Please consider this PR. We will have to decide if hidding completly the top searchbar should be done like this via a parameter, or this can be done easily, just by customising the CSS of iframe HTML (to me sounds preferable).

PR #591, opened 2021-07-25
Allow kiwix-serve to get suggestions of custom range

### LINKED ISSUE #242: Allow kiwix-serve to retrieve suggestions in the range provided by users
**Reported by kelson42 on 2013-05-13 11:33 UTC**
To allow to scroll without a end in the suggestion list.

maneeshpm:
Fixes #242 

This PR will allow kiwix serve to retrieve suggestions of a given range that can be specified in the front end call. These changes will open up the backend to implement *continuous scroll* feature as specified in the issue. But that is rather working on the front end side and this PR is only "allowing" to enable the feature.

Changes included in this PR:
- check for two new arguments `start`, `suggestionLength` in the handle_suggest API that will retrieve suggestions in the range given by (`start`, `start+suggestionLength`).

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -345,13 +345,13 @@ std::unique_ptr<Response> InternalServer::build_homepage(const RequestContext& r
 
 // TODO: retrieve searcher from caching mechanism
 SuggestionsList_t getSuggestions(const zim::Archive* const archive,
-                  const std::string& queryString, int suggestionCount)
+                  const std::string& queryString, int start, int suggestionCount)
 {
   SuggestionsList_t suggestions;
   auto searcher = zim::SuggestionSearcher(*archive);
   if (archive->hasTitleIndex()) {
     auto search = searcher.suggest(queryString);
-    auto srs = search.getResults(0, suggestionCount);
+    auto srs = search.getResults(start, suggestionCount);
 
     for (auto it : srs) {
       SuggestionItem suggestion(it.getTitle(), kiwix::normalize(it.getTitle()),
@@ -439,8 +439,6 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
 
   std::string content;
   std::string mimeType;
-  unsigned int maxSuggestionCount = 10;
-  unsigned int suggestionCount = 0;
 
   std::string bookName;
   std::string bookId;
@@ -453,6 +451,20 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
     archive = mp_library->getArchiveById(bookId);
   } catch (const std::out_of_range&) {
     return Response::build_404(*this, request, bookName, "");
+  }
+
+  auto start = 0;
+  try {
+    start = request.get_argument<unsigned int>("start");
+  } catch (const std::exception&) {}
+
+  unsigned int count = 10;
+  try {
+    count = request.get_argument<unsigned int>("count");
+  } catch (const std::exception&) {}
+
+  if (count == 0) {
+    count = 10;
   }
 
   if (archive == nullptr) {
@@ -468,7 +480,7 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
   bool first = true;
 
   /* Get the suggestions */
-  SuggestionsList_t suggestions = getSuggestions(archive.get(), queryString, maxSuggestionCount);
+  SuggestionsList_t suggestions = getSuggestions(archive.get(), queryString, start, count);
   for(auto& suggestion:suggestions) {
     MustacheData result;
     result.set("label", suggestion.getTitle());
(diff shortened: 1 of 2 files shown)


src/server/internalServer.cpp
```diff
@@ -353,6 +353,7 @@ SuggestionsList_t getSuggestions(const zim::Archive* const archive,
     // TODO: This case should be handled by libzim
     std::vector<std::string> variants = getTitleVariants(queryString);
     int currCount = 0;
+    int suggestionCount = end - start;
```
maneeshpm, line 356: Getting results of a certain range in non-Xapian based search is not supported as of now, so we will simply consider the count of suggestions here. 
This will anyway have to be implemented on libzim side after openzim/libzim#574

src/server/internalServer.cpp
```diff
@@ -442,6 +441,19 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
...
+  auto start = 0;
+  try {
+    start = request.get_argument<unsigned int>("start");
+  } catch (const std::exception&) {}
+
+  unsigned int suggestionLength = 10;
+  try {
+    suggestionLength = request.get_argument<unsigned int>("suggestionLength");
+  } catch (const std::exception&) {}
+  if (suggestionLength == 0) suggestionLength = 10;
```
mgautierfr, line 453: Just a matter of style but I prefer :
```cpp
if (suggestionLength == 0) {
    suggestionLength = 10;
}
```

As for `pageLength` in `handle_search`

(Yes, I must write a coding style guide)

mgautierfr (changes requested):
Code is good but I'm a bit disturbed by the idea of adding a somehow hidden feature, not tested and not used.

It would be nice to have this change associated with a change in the front end to use it (and some tests of the feature)

kelson42:
> Code is good but I'm a bit disturbed by the idea of adding a somehow hidden feature, not tested and not used.
> 
> It would be nice to have this change associated with a change in the front end to use it (and some tests of the feature)

I'm supportive of this comment. IMO, we should have a dedicated automated test able that way for example to retrieve by range of 10 suggestions, around 100 suggestions. The case of stopping before the end, and reaching the end of available suggestions should be tested. Cases of giving out-of-bands values to the ABI should be properly tested as well.

maneeshpm:
Thanks for the suggestion @kelson42 @mgautierfr. It revealed a nasty error #595. Working on a fix, shouldn't take much time.

maneeshpm:
@kelson42 Added some unit tests for checking the added changes. 

Rebasing the PR just to show the solution employed to our earlier tackled issue is employed here as well.

src/server/internalServer.cpp
maneeshpm: Aah right, you mentioned this before :sweat_smile: 

Casually tagging openzim/libzim#513, we should really document our coding style.

kelson42:
@MananJethwani Would you please verify that this feature would effectively allow us for example to provide an "infinite" scroll possibility in the suggestion selecbox?

maneeshpm:
@kelson42 The last review pointed out a lack of unit testing in this pr. I have added some new tests to demonstrate the features added. And hence, a new review is necessary.

kelson42:
@maneeshpm Can you please rebase this branch?

mgautierfr (changes requested):
It seems the two first commits are already merged in master.
They are keep in the PR by the rebase because they are using tab instead of space. We should simply remove them.

The update fix only one of my remarks. It add unit-test.
But we are adding a feature that no one is using. It would be nice to
make the kiwix-serve "frontend" uses the feature provided by the backend.
But this is js development and it should probably be made by a different developer. @kelson42 I let you decide on this. Having unit-test is enough for me.

maneeshpm:
Rebased to fix the commit history. 
I agree with @mgautierfr, will let anyone with experience in js build the front end. Till now, we have "enabled" the option from the backend.

src/server/internalServer.cpp
```diff
@@ -455,6 +453,20 @@ std::unique_ptr<Response> InternalServer::handle_suggest(const RequestContext& r
...
   }
 
+  auto start = 0;
+  try {
+    start = request.get_argument<unsigned int>("start");
+  } catch (const std::exception&) {}
+
+  unsigned int suggestionLength = 10;
+  try {
+    suggestionLength = request.get_argument<unsigned int>("suggestionLength");
```
mgautierfr, line 463: Just a small new remark.
`suggestionLength` seems a bit too long and error prone (different cases).
`count` is probably better.

mgautierfr (changes requested):
Sorry, I'm adding a late remark.

src/server/internalServer.cpp
maneeshpm: done!

mgautierfr (approved):

PR #596, opened 2021-07-30
improved browser lang filter working

### LINKED ISSUE #594: Smarter default filter
On first visit of kiwix-serve, the language filter is set to the language of the browser.

I think most frequently than not, the library might not have any content for that filter, offering a deceptive first contact.

Ideally, if the library has no content for that browser language, it should fallback to no language filter.

Cookie behavior is fine IMO.

MananJethwani:
@kelson42 this fixes #594

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -170,6 +170,14 @@
         })
     }
 
+    async function getBookCount(query) {
+        const url = `${root}/catalog/search?${query}`;
+        return await fetch(url).then(async (resp) => {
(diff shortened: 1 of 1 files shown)


kelson42 (approved):
@MananJethwani Thank you very much for your quick move on this. LGTM, even if it would have been nicer to avoid two HTTP requests in case of everything is OK with the number of books.

rgaudin (approved):
👍 haven't actually tested though

PR #597, opened 2021-07-30
Fix usage of zim::Searcher::getResults() in libkiwix

### LINKED ISSUE #595: Improper usage of `Searcher::getResults(start, end)`
Following openzim/libzim#596, we have some fixes to make in the internal server

eg. We always calculate end = start + pageLength, and then call getResults(start, end). But it should rather be getResults(start, pageLength)

maneeshpm:
Fixes #595 

The correct usage does not require the user to calculate an `end` using the `pageLength`. We can directly use `getResults(start, pageLength)`

### DIFF
--- a/include/searcher.h
+++ b/include/searcher.h
@@ -85,12 +85,12 @@ class Searcher
    *
    * @param search The search query.
    * @param resultStart the start offset of the search results (used for pagination).
-   * @param resultEnd the end offset of the search results (used for pagination).
+   * @param maxResultCount Maximum results to get from start (used for pagination).
    * @param verbose print some info on stdout if true.
    */
   void search(const std::string& search,
               unsigned int resultStart,
-              unsigned int resultEnd,
+              unsigned int maxResultCount,
               const bool verbose = false);
 
   /**
@@ -104,12 +104,12 @@ class Searcher
    * @param longitude The longitude of the center point.
    * @param distance The radius of the disc.
    * @param resultStart the start offset of the search results (used for pagination).
-   * @param resultEnd the end offset of the search results (used for pagination).
+   * @param maxResultCount Maximum number of results to get from start (used for pagination).
    * @param verbose print some info on stdout if true.
    */
   void geo_search(float latitude, float longitude, float distance,
                   unsigned int resultStart,
-                  unsigned int resultEnd,
+                  unsigned int maxResultCount,
                   const bool verbose = false);
 
   /**
@@ -148,14 +148,14 @@ class Searcher
   zim::SearchResultSet getSearchResultSet();
 
   unsigned int getResultStart() { return resultStart; }
-  unsigned int getResultEnd() { return resultEnd; }
+  unsigned int getMaxResultCount() { return maxResultCount; }
 
  protected:
   std::string beautifyInteger(const unsigned int number);
   void closeIndex();
   void searchInIndex(string& search,
                      const unsigned int resultStart,
-                     const unsigned int resultEnd,
+                     const unsigned int maxResultCount,
                      const bool verbose = false);
 
   std::vector<Reader*> readers;
@@ -163,7 +163,7 @@ class Searcher
   std::string searchPattern;
   unsigned int estimatedResultCount;
   unsigned int resultStart;
-  unsigned int resultEnd;
+  unsigned int maxResultCount;
 
  private:
   void reset();
--- a/test/searcher.cpp
+++ b/test/searcher.cpp
@@ -32,7 +32,7 @@ TEST(Searcher, incrementalRange) {
 
   int suggCount = 0;
   for (int i = 0; i < 10; i++) {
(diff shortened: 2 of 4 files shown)


mgautierfr (changes requested):
This PR is a "big" and tricky API change.
And it also change the behavior in a way that either it was buggy before or it is buggy now.

This PR change the function arguments from "two" unsigned int (start, end) to "two" other unsigned int (start, number).
So ABI is keep but the API is totally changed. It means that if we forget one call, we will not detect it at compilation and we will introduce a bug difficult to see.

On top of that, we were using `getResults` passing the start/end couple.
Now we are passing start/number.
But end=start+number. If start is 0, it doesn't change anything. But if we are getting result in the middle of the search, it is broken.

This change follow the semantic change of https://github.com/openzim/libzim/pull/597. But it is only a semantic change. Nothing is really changed on libzim side. So we should not change things here.

So there is something wrong somewhere. I've the feeling that previous code was wrong as xapian't `get_mset` takes maxitem and not end (https://xapian.org/docs/apidoc/html/classXapian_1_1Enquire.html#a1f2e35f79bc731bebfbdaceda17b9892).

It would be nice to have a unit test showing that :
- Either previous code is buggy, then we add a new (failing) test in a commit and fix it in a following commit(s).
- Either this PR is buggy, then we add a new (passing) test in a commit. This commit make the test failing, and we add a new fixup! commit fixing this.

maneeshpm:
@mgautierfr Showing that the previous code is buggy is easy on libzim side. But on libkiwix side, our `SearchRenderer` which takes a `SearchResultSet` sort of masks this issue in ft search by selecting only `pageLength` number of results. I was only able to discover it because of #242 where getResult(5,10) returned 15 suggestions. What kind of failing test case do you suggest?

mgautierfr:
> I was only able to discover it because of #242 where getResult(5,10) returned 15 suggestions. What kind of failing test case do you suggest?

It seems that the sentence just before the question is the answer to the question :)

test/searcher.cpp
```diff
@@ -6,6 +6,16 @@ namespace kiwix
...
 TEST(Searcher, search) {
+	// test usage of zim::Searcher::getResult()
+	zim::Archive archive("./test/zimfile.zim");
+	zim::Searcher ftsearcher(archive);
+	zim::Query query;
+	query.setQuery("ray", false);
+	auto search = ftsearcher.search(query);
+
+	auto srs = search.getResults(5, 10);
+	ASSERT_EQ(srs.size(), 5);	// Fails! usage of getResults is wrong.
```
mgautierfr: It is better to have the test made with the expected result (and have the test failing) and explain in the commit that this is this specific test that will be fixed in next commit.

And please put this test in a new test, this is specific enough to have a test to check incremental searching.

include/searcher.h
```diff
@@ -85,12 +85,12 @@ class Searcher
...
    * @param search The search query.
    * @param resultStart the start offset of the search results (used for pagination).
-   * @param resultEnd the end offset of the search results (used for pagination).
+   * @param resultMax Maximum results to get from start (used for pagination).
    * @param verbose print some info on stdout if true.
    */
   void search(const std::string& search,
               unsigned int resultStart,
-              unsigned int resultEnd,
+              unsigned int resultMax,
```
mgautierfr, line 93: Maybe rename this to `maxResults` or `resultCount`.

`resultMax` make me think about the "result maximum (highest)". So a single result, which is semantically wrong.

mgautierfr (changes requested):
Two small change that should be easy to fix.

include/searcher.h
kelson42: or `resultMaxCount`

test/searcher.cpp
```diff
@@ -31,11 +31,11 @@ TEST(Searcher, incrementalRange) {
 
 	int suggCount = 0;
 	for (int i = 0; i < 10; i++) {
-		auto srs = search.getResults(i*5, (i + 1)*5); // get 5 results
+		auto srs = search.getResults(i*5, 5); // get 5 results
 		ASSERT_EQ(srs.size(), 5);
-    suggCount += srs.size();
+    		suggCount += srs.size();
```
mgautierfr, line 36: There is a mix between spaces and tab. (Please use spaces)

mgautierfr (changes requested):
Just a small change to be pedantic.
Else we are good.

mgautierfr (approved):

test/searcher.cpp
maneeshpm: fixed in both the commits.

include/searcher.h
maneeshpm: renamed to `maxResultCount`

test/searcher.cpp
maneeshpm: done!

PR #600, opened 2021-08-04
Use OPDS API to populate categories/languages select boxes on Kiwix Serve welcome page

### LINKED ISSUE #555: Use language/category API to populate Kiwix Serve filter select boxes
@veloman-yunkan has implemented category and language API end points, see:
* language: https://github.com/kiwix/libkiwix/pull/553
* category: https://github.com/kiwix/libkiwix/pull/492

Current Kiwix Serve filter select boxes choces are hardcoded, we should use the newly created end-points to:
* Avoid code duplication
* Provide less useless choices (the end points provide values matching ZIM files in its catalog)
* Benefit of language localisation (at least for category end-point), one time this will be implemented (see #59)

MananJethwani:
fixes #555 
@kelson42 instead of using DOM for XML parsing this PR uses string manipulation as for some reason xml produced by language endpoint couldn't be parsed on latest version of firefox.


veloman-yunkan:
> instead of using DOM for XML parsing this PR uses string manipulation as for some reason xml produced by language endpoint couldn't be parsed on latest version of firefox.

That's not a good reason for using custom string manipulation instead of XML parsing. We need to find out what's wrong with the XML produced by the /catalog/v2/ OPDS API. Can you please identify the issue in the XML?

MananJethwani:
> That's not a good reason for using custom string manipulation instead of XML parsing. We need to find out what's wrong with the XML produced by the /catalog/v2/ OPDS API. Can you please identify the issue in the XML?

I tried Identifying the issue but couldn't find any possible reason for this.
here is a JS code snippet, if you run this while yourserver is hosted at port `8000` using firefox would give you `null` while chorme gives `<entry></entry>` : 
```
async function fun() {
  await fetch('http://localhost:8000/catalog/v2/languages').then(async (resp) => {
    const data = new window.DOMParser().parseFromString(await resp.text(), 'application/xhtml+xml');
    console.log(data.querySelector('entry'));
  })
}

fun();
```

though still trying to find the exact problem.

MananJethwani:
@veloman-yunkan the problem seems to be with `:` in `dc:language` and `thr:count` firefox DOMPareser seems to give error for that.
@kelson42 should it be changed on backend side or should I handle it in JS?

kelson42:
The part before the character : is called the DOM node namespace. XML namespaces are fully supported in Firefox, see https://developer.mozilla.org/en-US/docs/Web/API/Element/namespaceURI. The OPDS norm specifies the namespecies striclty I believe and therefore can not be changed randomly. The problem is probably linked to the namespace (if your analysis is proper),  but can probably be fixed properly. Have you checked that the XML schema definition is proper in the « xml » node?

MananJethwani:
@veloman-yunkan, @kelson42 and I investigated a bit more and the problem seems to be exactly with `thr:count` the namespace for `thr` is not provided and should be provied in the XML can you please have a look and create PR to update XML provided by language endpoint.

MananJethwani:
@veloman-yunkan I have added thr namespace as suggested by @kelson42 ... please have a look.

static/skin/index.js
```diff
@@ -363,15 +373,20 @@
         fadeOutDiv = document.getElementById('fadeOut');
         loader = document.querySelector('.loader');
         await loadAndDisplayBooks();
-        await loadAndDisplayOptions('#languageFilter', langList);
-        await loadAndDisplayOptions('#categoryFilter', categoryList);
+        await loadAndDisplayOptions('#languageFilter', '/catalog/v2/languages', 'lang');
+        await loadAndDisplayOptions('#categoryFilter', '/catalog/v2/categories');
```
veloman-yunkan, line 377: Don't forget about `root`

static/skin/index.js
```diff
@@ -209,11 +213,17 @@
...
-        document.querySelector(nodeQuery).innerHTML += Object.keys(query)
-            .map((option) => {return `<option value='${option}'>${htmlEncode(query[option])}</option>`})
-        .join('');
+    async function loadAndDisplayOptions(nodeQuery, query, type = '') {
+        await fetch(query).then(async (resp) => {
+            const data = new window.DOMParser().parseFromString(await resp.text(), 'application/xml');
+            let optionStr = '';
+            data.querySelectorAll('entry').forEach(entry => {
+                const title = getInnerHtml(entry, 'title');
+                const value = type === 'lang' ? getInnerHtml(entry, 'language') : title;
```
veloman-yunkan, line 222: Special handling of languages doesn't feel right. Why don't you make the `type` parameter of `loadAndDisplayOptions()` mandatory and pass it to `getInnerHtml()` unmodified?

veloman-yunkan (changes requested):
Please squash the two commits into one. Then it makes sense to extract the `xmlns:thr` fix into a commit of its own.

static/skin/index.js
MananJethwani: @veloman-yunkan that doesn't seem right to me.....there is no meaning of adding `type` to `getInnerHtml()` it's all because the value for langugae options need to be iso-3 value of language while for categories title and value are same.
can you be more specific please.

kelson42:
Just rebased the feature branch

kelson42 (changes requested):
I hope I don't have broken it, but we my test:
* The tiles don't display
* Both categories/languages select boxes have only the "all" option
image

static/skin/index.js
veloman-yunkan: I meant having

```
const value = getInnerHtml(entry, type);
```

in `loadAndDisplayOptions()` and calling the latter as follows:

```
await loadAndDisplayOptions('#languageFilter', `${root}/catalog/v2/languages`, 'language');
await loadAndDisplayOptions('#categoryFilter', `${root}/catalog/v2/categories`, 'title');
```

Then the `type` parameter of `loadAndDisplayOptions()` should be renamed too (`entry_node_name`?).

MananJethwani:
@kelson42 it's broken because of the rebase, @veloman-yunkan I have a doubt at this point the new entry block looks somethinglike this - 
```
<entry>
    <id>urn:uuid:4fa9d7d9-49e7-5dd3-f3e9-b4425b9b68ba</id>
    <title>Geographie von Wikipedia</title>
    <summary>Eine Auswahl von Wikipedia Artikeln zum Thema Geographie</summary>
    <language>deu</language>
    <updated>2021-07-15T00:00:00Z</updated>
    <name>wikipedia_de_geography</name>
    <flavour>nopic</flavour>
    <category>wikipedia</category>
    <tags>wikipedia;_category:wikipedia;_pictures:no;_videos:no;_details:yes;_ftindex:yes</tags>
    <articleCount>12930</articleCount>
    <mediaCount>402</mediaCount>
    <link rel="http://opds-spec.org/image/thumbnail"
          href="/meta?name=Illustration_48x48@1&amp;content=wikipedia_de_geography_nopic_2021-07"
          type="image/png;width=48;height=48;scale=1"/>
    <link type="text/html" href="/wikipedia_de_geography_nopic_2021-07" />
    <author>
      <name>Wikipedia</name>
    </author>
    <publisher>
      <name>Kiwix</name>
    </publisher>
    
  </entry>
```
what is the exact method I should use to separate various kind of links like here links are being used fot 2 purpose....1 for icon and 1 for href link to open preview, should I use type and assume that `text/html` will only be used once and only for our preview url link, and all other will be icon tags?

kelson42:
@MananJethwani IMO, but @veloman-yunkan is the expert. You should filter by attribute `rel=http://opds-spec.org/image/thumbnail` to get all the illustrations and then use the `type` attribute to get the size/pixeldepth you need.

kelson42 (approved):
From a user perspective this PR works fine. Thx @MananJethwani. Lets wait final technical review from @veloman-yunkan to merge.

veloman-yunkan (commented):
Final code looks fine but the commit history needs to be improved. Please squash all commit into one and then break the resulting change into smaller logical independent commits.

MananJethwani:
@veloman-yunkan hope this works for you.

veloman-yunkan (approved):

PR #602, opened 2021-08-08
OPDS feed with partial entries

### LINKED ISSUE #209: Use OPDS Feed with partial entries

veloman-yunkan:
Fixes #209

Things to note:

1. The id of the entry in OPDS feed has the format urn:uuid:*entryid*, however when accessing individual entries via the catalog v2 OPDS API the `urn:uuid:` prefix must **not** be included.

### DIFF
--- a/include/opds_dumper.h
+++ b/include/opds_dumper.h
@@ -64,6 +64,14 @@ class OPDSDumper
   std::string dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query) const;
 
   /**
+   * Dump the OPDS complete entry document.
+   *
+   * @param bookId the id of the book
+   * @return The OPDS complete entry document.
+   */
+  std::string dumpOPDSCompleteEntry(const std::string& bookId) const;
+
+  /**
    * Dump the categories OPDS feed.
    *
    * @return The OPDS feed.
--- a/include/opds_dumper.h
+++ b/include/opds_dumper.h
@@ -59,9 +59,10 @@ class OPDSDumper
    *
    * @param bookIds the ids of the books to include in the feed
    * @param query the query used to obtain the list of book ids
+   * @param partial whether the feed should include partial or complete entries
    * @return The OPDS feed.
    */
-  std::string dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query) const;
+  std::string dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query, bool partial) const;
 
   /**
    * Dump the OPDS complete entry document.
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -75,7 +75,7 @@ kainjow::mustache::object getSingleBookData(const Book& book)
                                ? MustacheData(false)
                                : MustacheData(book.getUrl());
     return kainjow::mustache::object{
-      {"id", "urn:uuid:"+book.getId()},
+      {"id", book.getId()},
       {"name", book.getName()},
       {"title", book.getTitle()},
       {"description", book.getDescription()},
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -134,20 +134,22 @@ string OPDSDumper::dumpOPDSFeed(const std::vector<std::string>& bookIds, const s
   return render_template(RESOURCE::templates::catalog_entries_xml, template_data);
 }
 
-string OPDSDumper::dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query) const
+string OPDSDumper::dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query, bool partial) const
 {
   const auto bookData = getBookData(library, bookIds);
 
+  const char* const endpoint = partial ? "/partial_entries" : "/entries";
   const kainjow::mustache::object template_data{
      {"date", gen_date_str()},
      {"endpoint_root", rootLocation + "/catalog/v2"},
-     {"feed_id", gen_uuid(libraryId + "/entries?"+query)},
+     {"feed_id", gen_uuid(libraryId + endpoint + "?" + query)},
      {"filter", query.empty() ? MustacheData(false) : MustacheData(query)},
      {"query", query.empty() ? "" : "?" + urlEncode(query)},
      {"totalResults", to_string(m_totalResults)},
      {"startIndex", to_string(m_startIndex)},
      {"itemsPerPage", to_string(m_count)},
-     {"books", bookData }
+     {"books", bookData },
+     {"dump_partial_entries", MustacheData(partial)}
   };
 
   return render_template(RESOURCE::templates::catalog_v2_entries_xml, template_data);
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -51,7 +51,7 @@ namespace
 {
 
 typedef kainjow::mustache::data MustacheData;
-typedef kainjow::mustache::list BookData;
+typedef kainjow::mustache::list BooksData;
 typedef kainjow::mustache::list IllustrationInfo;
 
 IllustrationInfo getBookIllustrationInfo(const Book& book)
@@ -95,15 +95,26 @@ kainjow::mustache::object getSingleBookData(const Book& book)
     };
 }
 
-BookData getBookData(const Library* library, const std::vector<std::string>& bookIds)
+std::string getSingleBookEntryXML(const Book& book, bool withXMLHeader, const std::string& endpointRoot, bool partial)
 {
-  BookData bookData;
+  auto data = getSingleBookData(book);
+  data["with_xml_header"] = MustacheData(withXMLHeader);
+  data["dump_partial_entries"] = MustacheData(partial);
+  data["endpoint_root"] = endpointRoot;
+  return render_template(RESOURCE::templates::catalog_v2_entry_xml, data);
+}
+
+BooksData getBooksData(const Library* library, const std::vector<std::string>& bookIds, const std::string& endpointRoot, bool partial)
+{
+  BooksData booksData;
   for ( const auto& bookId : bookIds ) {
     const Book& book = library->getBookById(bookId);
-    bookData.push_back(getSingleBookData(book));
+    booksData.push_back(kainjow::mustache::object{
+        {"entry", getSingleBookEntryXML(book, false, endpointRoot, partial)}
+    });
   }
 
-  return bookData;
+  return booksData;
 }
 
 std::string getLanguageSelfName(const std::string& lang) {
@@ -119,7 +130,7 @@ std::string getLanguageSelfName(const std::string& lang) {
 
 string OPDSDumper::dumpOPDSFeed(const std::vector<std::string>& bookIds, const std::string& query) const
 {
-  const auto bookData = getBookData(library, bookIds);
+  const auto booksData = getBooksData(library, bookIds, "", false);
   const kainjow::mustache::object template_data{
      {"date", gen_date_str()},
      {"root", rootLocation},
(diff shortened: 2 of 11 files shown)


kelson42:
@veloman-yunkan Should we remove the WIP in the topic of the PR?

veloman-yunkan:
> @veloman-yunkan Should we remove the WIP in the topic of the PR?

@kelson When I changed the status of this PR from Draft to Ready the meaning of WIP changed from "Work In Progress" to "Want It Peer-reviewed" :smiley: . But since it can be confusing, I guess we indeed better drop that abbreviation from the title. I already did.

kelson42:
@veloman-yunkan OK, just put a small comment helps in such case to avoid confusion.

test/server.cpp
```diff
@@ -735,7 +735,7 @@ TEST_F(LibraryServerTest, catalog_search_by_phrase)
   EXPECT_EQ(maskVariableOPDSFeedData(r->body),
     OPDS_FEED_TAG
     "  <id>12345678-90ab-cdef-1234-567890abcdef</id>\n"
-    "  <title>Filtered zims (q=&quot;ray charles&quot;)</title>\n"
+    "  <title>Filtered zims (?q=&quot;ray charles&quot;)</title>\n"
```
mgautierfr, line 738: Do we really want this change ?

(But not sure it is really important, we don't use the feed title)

test/server.cpp
```diff
@@ -1176,3 +1190,69 @@ TEST_F(LibraryServerTest, catalog_v2_entries_filtered_by_search_terms)
...
+    "          href=\"/catalog/v2/entries/raycharles\"\n"
+    "          type=\"application/atom+xml;type=entry;profile=opds-catalog\"\n"
+    "          title=\"Ray Charles\"/>\n"
+    "  </entry>\n"
+    "  <entry>\n"
+    "    <id>urn:uuid:raycharles_uncategorized</id>\n"
+    "    <link rel=\"alternate\"\n"
+    "          href=\"/catalog/v2/entries/raycharles_uncategorized\"\n"
+    "          type=\"application/atom+xml;type=entry;profile=opds-catalog\"\n"
+    "          title=\"Ray (uncategorized) Charles\"/>\n"
```
mgautierfr, line 1254: I'm a bit disturb by this attribute here, I would have put the title as a node in the `entry`.
But it is what we have in the example in the spec 🤷🏽‍♂️

mgautierfr (changes requested):
Nothing to say on the code itself. It is good.

But I'm not sure about the provided API :
- As said in the review comment, I understand from the spec that we should have a endpoint which return the full entry xml.
- So we should probably have a `catalog/v2/entry` endpoint for one entry (in opposition to `catalog/v2/entries` which list entries, potentially only one complete entry).

https://specs.opds.io/opds-1.2#512-partial-and-complete-catalog-entries says that we should have `atom:category`, `atom:contributor`, `atom:rights`, `opds:price`. It make no really sense to have these in our use case. But maybe we could add more information than the `id` in the partial feed. At least the `updated` node to allow client if the entry has been updated since last time it get the entry information. Maybe the `name` could also be useful.

I'm also a bit afraid that the mustache template become a bit to complex
with all the conditional part. But it can be easily fixed by using two template if it appears it is too complex to maintain. So no problem here for now.

test/server.cpp
```diff
@@ -1176,3 +1190,69 @@ TEST_F(LibraryServerTest, catalog_v2_entries_filtered_by_search_terms)
...
+    "  <link rel=\"up\"\n"
+    "        href=\"/catalog/v2/entries\"\n"
+    "        type=\"application/atom+xml;profile=opds-catalog;kind=acquisition\"/>\n"
+    "\n"
+    "  <title>Single Entry</title>\n"
+    "  <updated>YYYY-MM-DDThh:mm:ssZ</updated>\n"
+    "\n"
+    RAY_CHARLES_CATALOG_ENTRY
+    "</feed>\n"
+  );
```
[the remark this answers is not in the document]
veloman-yunkan: Good catch. You are right, complete entries can be served as separate XML (non-feed) documents.

test/server.cpp
veloman-yunkan: Added the `<title>` node without removing the `title` attribute from the link node. In the example, the latter is used to make it explicit that the link leads to the complete entry (note the word "Complete" in the title). I think it's safe to drop the title attribute from `<link rel="alternate">`

veloman-yunkan:
> But I'm not sure about the provided API :
> 
>   * As said in the review comment, I understand from the spec that we should have a endpoint which return the full entry xml.
> 
>   * So we should probably have a `catalog/v2/entry` endpoint for one entry (in opposition to `catalog/v2/entries` which list entries, potentially only one complete entry).

Done. But I don't like that a complete entry document (which is now not a feed document anymore) is still produced by `OPDSDumper::dumpOPDSFeedV2()`.

> 
> 
> https://specs.opds.io/opds-1.2#512-partial-and-complete-catalog-entries says that we should have `atom:category`, `atom:contributor`, `atom:rights`, `opds:price`. It make no really sense to have these in our use case. But maybe we could add more information than the `id` in the partial feed. At least the `updated` node to allow client if the entry has been updated since last time it get the entry information. Maybe the `name` could also be useful.

Now including `<updated>` and `<title>` in the partial entries.

> I'm also a bit afraid that the mustache template become a bit to complex
> with all the conditional part. But it can be easily fixed by using two template if it appears it is too complex to maintain. So no problem here for now.

I had the same concerns.

This PR definitely needs to be cleaned up. I will do it once we agree that the API and the output format are good.

test/server.cpp
mgautierfr: Yes, we can remove the title from the link. It is a duplicated value and it is better in a `<title>` node.

mgautierfr:
> Done. But I don't like that a complete entry document (which is now not a feed document anymore) is still produced by OPDSDumper::dumpOPDSFeedV2().

Yes, it should be done in another method.

>>    I'm also a bit afraid that the mustache template become a bit to complex
>>    with all the conditional part. But it can be easily fixed by using two template if it appears it is too complex to maintain. So no problem here for now.

> I had the same concerns.

With the new "single_entry_mode" it is even worth.

> This PR definitely needs to be cleaned up. I will do it once we agree that the API and the output format are good.

I agree with the output format (assuming we remove the title from the link) and the API.

---

I wonder if we could homogenize a bit the code with `libxml_dumper` by having only one "dumper" which can work with different modes (by switching the template used). Libxml, complete feed and partial feed would only differentiate by the used template.
`Book` could also generate its own "BookData" `mustache::object` and the dumper would simply pass a list of BookData to the template. For single entry the server could simply get this data and pass it to a "single mode" template.

But having a unique "dumper" is probably not for this PR.

(If you have a idea for the parsing side, I'm a bit disturbed by the idea of a `Manager` parsing the xml/opds and having yet a parsing part in the Book constructors. But I haven't figure out how to better organize this)

kelson42:
@veloman-yunkan Any news on this PR?

include/opds_dumper.h
```diff
@@ -59,9 +59,10 @@ class OPDSDumper
    *
    * @param bookIds the ids of the books to include in the feed
    * @param query the query used to obtain the list of book ids
+   * @param partial whether the feed should include partial or complete entries
    * @return The OPDS feed.
    */
-  std::string dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query) const;
+  std::string dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const std::string& query, bool partial = false) const;
```
mgautierfr, line 65: I would prefer to not have a default value for `partial`.

Partial/complete feed are somehow at equal level. Complete feed is not really the default/normal way to dump feed.
And as we use dumpOPDSFeedV2 only twice, there is no real reason to keep some kind of compatibility.
This is indeed a API break, but this release will have API break anyway.

static/templates/catalog_v2_entries.xml
```diff
@@ -49,6 +53,7 @@
     {{#url}}
     <link rel="http://opds-spec.org/acquisition/open-access" type="application/x-zim" href="{{{url}}}" length="{{{size}}}" />
     {{/url}}
+{{/dump_partial_entries}}
```
mgautierfr, line 56: I wonder if we could go a bit further and reuse the output of the template `catalog_v2_complete_entry.xml` here instead of almost copying the same template.

mgautierfr (commented):
We are globally good here.

My two comments are somehow a nice to have but not mandatory (and the second may be a bit tricky to implement).

include/opds_dumper.h
veloman-yunkan: Done

static/templates/catalog_v2_entries.xml
veloman-yunkan: Done.

src/server/internalServer.h
```diff
@@ -75,7 +75,8 @@ class InternalServer {
     std::unique_ptr<Response> handle_catalog(const RequestContext& request);
     std::unique_ptr<Response> handle_catalog_v2(const RequestContext& request);
     std::unique_ptr<Response> handle_catalog_v2_root(const RequestContext& request);
-    std::unique_ptr<Response> handle_catalog_v2_entries(const RequestContext& request);
+    std::unique_ptr<Response> handle_catalog_v2_entries(const RequestContext& request, bool partial = false);
```
mgautierfr, line 78: I've put my last comment about the default value for partial on the wrong line but I think it should be changed also here.

src/opds_dumper.cpp
```diff
@@ -95,6 +92,25 @@ BookData getBookData(const Library* library, const std::vector<std::string>& boo
...
+std::string getSingleBookEntryXML(const Book& book, bool withXMLHeader, const std::string& endpointRoot, bool partial)
+{
+  auto data = getSingleBookData(book);
+  data["with_xml_header"] = MustacheData(withXMLHeader);
+  data["dump_partial_entries"] = MustacheData(partial);
+  data["endpoint_root"] = endpointRoot;
+  return render_template(RESOURCE::templates::catalog_v2_entry_xml, data);
+}
+
+BookData getBookData(const Library* library, const std::vector<std::string>& bookIds, const std::string& endpointRoot, bool partial)
```
mgautierfr, line 107: Make this plural ? (`BooksData getBooksData(...)`)

mgautierfr (changes requested):
Two small easy fixes and we are good.

src/opds_dumper.cpp
veloman-yunkan: Done

src/server/internalServer.h
veloman-yunkan: Done

mgautierfr (approved):
We are good !

Please rebase-fixup on last master and we merge.

veloman-yunkan:
> We are good !
> 
> Please rebase-fixup on last master and we merge.

Done

PR #604, opened 2021-08-10
fixed books availablity on larger screens and added zoom level support

### LINKED ISSUE #603: kiwix-serve does not populate enough cards to scroll on a large screen
On a 4k screen I get 4 rows of 7 cards. I have 130 zim files loaded and it's impossible to scroll further. In order to see more I have to resize the window (either a bit larger or smaller) to get additional cards to load, and then I am able to scroll.

MananJethwani:
@kelson42 this resolves #603

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -62,7 +62,8 @@
     }
 
     function viewPortToCount(){
-        return Math.floor(window.innerHeight/300 + 1)*(window.innerWidth>1000 ? 4 : 3);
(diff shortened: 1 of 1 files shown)


kelson42:
@jmontleon Does that PR fixes the problem you have reported?

jmontleon:
@kelson42 Yes, it works as expected now. Thank you and @MananJethwani for the quick fix!

Issue #571, opened 2021-06-28
Kiwix Serve welcome page title should be customizable


kelson42:

kelson42:
Probably this should be done via allowing a custom template for the home page!?

mgautierfr:
Related to https://github.com/kiwix/kiwix-tools/issues/17

kelson42:
@MananJethwani Do you think you could make a new option `--customIndexTemplate` which would take an alternative to our `index.html` template?

PR #607, opened 2021-08-20
added cutomIndexTemplate option to kiwix-serve

MananJethwani:


MananJethwani:
@mgautierfr @veloman-yunkan, where can I add test for `customIndexTemplate`, and exactly how, are there any tests for other arguments like `withLibraryButton` which I can refer to.

veloman-yunkan:
> @mgautierfr @veloman-yunkan, where can I add test for `customIndexTemplate`, and exactly how, are there any tests for other arguments like `withLibraryButton` which I can refer to.

@MananJethwani  I don't think there is such a test. I also believe that the best way to test this functionality would be in kiwix-tools, by running the kiwix-serve binary via pytest. Please see kiwix/kiwix-tools#375 for how I had intended to do something similar, before @kelson42 fairly pointed out that that feature better be tested in libkiwix via googletest.

kelson42:
I want indeed to have the httpd daemon primitive tested within libkiwix and I don't understand what stop us to do it.

kelson42:
I want indeed to have the httpd daemon primitive tested within libkiwix and I don't understand what stop us to do it.

MananJethwani:
@kelson42 and @veloman-yunkan can you please have a look at the tests I have created, thanks.

kelson42:
@MananJethwani We should add a document in the README about the template variables which are exposed.

mgautierfr:
I wonder if this is not a bit too much.
The main need in #571 is to customize the style/text of what is displayed.
With this PR we allow the user to change almost everything. We the constraints that he must correctly do everything.
If he changes the js, he also must do the correctly js call to our server (or it own using our API). It introduces de facto a new public API on our side.

Maybe it is enough to let the user set the index title and allowing it to simply set a custom css ?
For now, this PR make me feel we are opening a Pandora's box.

src/server/internalServer.cpp
```diff
@@ -336,7 +351,20 @@ InternalServer::get_matching_if_none_match_etag(const RequestContext& r) const
...
+  //                             computeAbsolutePath(getCurrentDirectory(), m_customIndexTemplate) :
+  //                             m_customIndexTemplate;
+  //   resourceString = getFileContent(m_customIndexTemplate);
+  //   if (resourceString.empty() && m_verbose.load()) {
+  //     printf('Invalid customIndexTemplate path, falling back to default welcome page template.');
+  //     resourceString = RESOURCE::templates::index_html;  
+  //   }
+  // } else {
+  //   resourceString = RESOURCE::templates::index_html;
+  // }
```
mgautierfr, line 366: Please remove the commented code.

src/server/internalServer.cpp
```diff
@@ -137,10 +138,24 @@ InternalServer::InternalServer(Library* library,
...
+{
+  if (!m_customIndexTemplate.empty()) {
+    m_customIndexTemplate = kiwix::isRelativePath(m_customIndexTemplate) ?
+                              computeAbsolutePath(getCurrentDirectory(), m_customIndexTemplate) :
+                              m_customIndexTemplate;
+    m_indexTemplateString = getFileContent(m_customIndexTemplate);
+    if (m_indexTemplateString.empty() && m_verbose.load()) {
+      printf("Invalid path for customIndexTemplate, falling back to default template.\n");
+    }
+  }
```
mgautierfr, line 154: I'm not sure we simply want to print a message and go with(out) it.

If a user ask to use a specific template and we cannot find it, it is probably a error.

---

In fact, I wonder if the user may directly pass the template string to the constructor and not the path to a file. This simplify a lot the things on libkiwix side and allow use case where the template is not stored in a file (generated template, ...)
It would be to kiwix-serve/kiwix-desktop/wrapper to load the template and pass it to the server's constructor.

mgautierfr (changes requested):

src/server/internalServer.cpp
MananJethwani: @mgautierfr any example or some resource I can refer for this....not exactly a cpp dev.
I guess we should have a discussion regarding this with @kelson42 before going ahead.

mgautierfr: You have pretty everything you need.
Instead of adding a argument `customIndexTemplate(path)`, directly add the argument `indexTemplateString`. (And use `RESOURCE::templates::index_html` if the string is empty)

And move the code loading the template from `customIndexTemplate` (you 've already implemented) in kiwix-serve.

But yes, we may agree with @kelson42 on the feature before going ahead.

kelson42:
@MananJethwani Hi, any feedback here to Matthieu's comments?

MananJethwani:
@kelson42 extremely sorry for the delay will update the PR by tomorrow

src/server/internalServer.cpp
```diff
@@ -336,7 +338,7 @@ InternalServer::get_matching_if_none_match_etag(const RequestContext& r) const
 
 std::unique_ptr<Response> InternalServer::build_homepage(const RequestContext& request)
 {
-  return ContentResponse::build(*this, RESOURCE::templates::index_html, get_default_data(), "text/html; charset=utf-8", true);
+  return ContentResponse::build(*this, !m_indexTemplateString.empty() ? m_indexTemplateString : RESOURCE::templates::index_html, get_default_data(), "text/html; charset=utf-8", true);
```
mgautierfr, line 341: This could be made once at `InternalServer` construction.
No need to check every time if m_indexTemplateString is empty as it will never change.

src/server/internalServer.h
```diff
@@ -54,7 +54,8 @@ class InternalServer {
                    bool verbose,
                    bool withTaskbar,
                    bool withLibraryButton,
-                   bool blockExternalLinks);
+                   bool blockExternalLinks,
+                   std::string indexTemplateString = "");
```
mgautierfr, line 58: Do we really need a default argument here ?
This is internal, so no api compatibility to keep.

include/tools.h
```diff
@@ -167,5 +167,7 @@ std::vector<std::string> split(const std::string& str, const std::string& delims
  * @throw std::out_of_range if iso2 code is not known.
  */
 std::string converta2toa3(const std::string& a2code);
+
+std::string getFileContent(const std::string& path);
```
mgautierfr, line 171: Please add a doc string.
And making this function part of the public api could (should) be moved in a separated commit.

mgautierfr (changes requested):
See my comments in the code.

You've adapted the test code but we don't really test the feature itself as we never try to run a test with a custom template.
A basic test would be to pass a custom template with a specific word/sentence and check that this word is correctly present in the output.

include/tools.h
MananJethwani: @mgautierfr I will change the commit history once PR is ready to get merged

MananJethwani:
@mgautierfr any updates?

kelson42:
I have rebased the branch on `origin/master`

src/server/internalServer.cpp
```diff
@@ -138,11 +138,12 @@ InternalServer::InternalServer(Library* library,
   m_withTaskbar(withTaskbar),
   m_withLibraryButton(withLibraryButton),
   m_blockExternalLinks(blockExternalLinks),
-  m_indexTemplateString(indexTemplateString),
   mp_daemon(nullptr),
   mp_library(library),
   mp_nameMapper(nameMapper ? nameMapper : &defaultNameMapper)
-{}
+{
+  m_indexTemplateString = !indexTemplateString.empty() ? indexTemplateString : RESOURCE::templates::index_html;
```
mgautierfr, line 145: It is better to have a "positive" test.
We can always set `m_indexTemplateString` to `indexTemplateString` in the initializer list and then, in the constructor body, test if `m_indexTemplateString` is empty and then set it to `RESOURCE::templates::index_html`

mgautierfr (changes requested):
We still need a bit more testing. To be sure we actually use the custom template index passed.

I think you can also simplify the git history. We probably need two commits :
- One to add the handling off the customIndexTemplate (and the associated tests)
- One to expose the 3 helper functions in `tools.h`

test/server.cpp
```diff
@@ -211,6 +214,53 @@ ResourceCollection all200Resources()
...
+  const int PORT = 8001;
+  const ZimFileServer::FilePathCollection ZIMFILES {
+    "./test/zimfile.zim",
+    "./test/corner_cases.zim"
+  };
+
+  std::unique_ptr<ZimFileServer>   zfs_;
+  zfs_.reset(new ZimFileServer(PORT, ZIMFILES, ""));
+  EXPECT_EQ(200, zfs_->GET("/")->status);
+  zfs_.reset();
```
mgautierfr, line 227: This can be better written like :
```cpp
auto zfs_ = sd::unique_ptr<ZimFileServer>(new ZimFileServer(PORT, ZIMFILES, ""));
EXPECT_EQ(200, zfs_->GET("/")->status);
```
No need to call reset at the end as the unique_ptr will delete the allocated ZimFileServer (this is the only purpose of unique_ptr, it would be sad to not us it correctly)

(Same for the test `indexTemplateCheck`)

mgautierfr (changes requested):

test/server.cpp
```diff
@@ -211,6 +214,53 @@ ResourceCollection all200Resources()
...
+  "</body>"
+"</html>"));
+  EXPECT_EQ("<!DOCTYPE html><html xmlns='http://www.w3.org/1999/xhtml'><head>"
+    "<title>Welcome to kiwix library</title>"
+    "<link type=\"root\" href=\"\">"
+    "</head>"
+      "<body>"
+        "<div class='kiwixHomeBody'>"
+          "<h3 class='kiwixHomeBody__results'></h3>"
+          "<div class='book__list'></div>"
```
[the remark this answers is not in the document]
MananJethwani: @mgautierfr do you want me to add back the script and check if zim files are present as well, according to me the test is not about if zim files are rendered but about if template is loaded so according to me no need to add a script right?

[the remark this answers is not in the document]
mgautierfr: Yes, rendering the template is already tested with other tests. Here we are testing if we correctly use the provided template. We can keep it like this.
(But we main reduce a bit the template to some kind of minimal content, as we don't care about the content let's be even more explicit.)

PR #622, opened 2021-10-19
Provide HTTP URL for the server

### LINKED ISSUE #613: Fournish HTTP URL helper for the server
Kiwix Serve, Kiwix Desktop and Kiwix for Android provide a HTTP daemon feature based on the libkiwix server primitive.

In an attempt ot increase usability Kiwix Desktop and Kiwix for Android provide a `http://192.158.1.42:80/` like of URL the software user can easy share with a broader audience connected to the same local (WiFi) network. This is handy and we want a similar feature  for Kiwix Serve as well, see https://github.com/kiwix/kiwix-tools/issues/473.

For the moment Kiwix Desktop and Kiwix Android have each their own implementation to figure out this HTTP URL, but to avoid to code for a third time such a primitive in Kiwix Serve. I would like to have in libkiwix and would basically be:
* List all interfaces/IPs
* Choose the best IP (heuristic) if not specified
* Build URL (with the port).

We had an implementation in Kiwix Desktop 1.0 which seems a good start point:
https://github.com/kiwix/kiwix-xulrunner/blob/master/src/common/networkTools.cpp

Does not handle IPV5 though... but not sure this is really necessary in a LAN context.

juuz0:
Fixes #613.


src/server/networkTools.cpp
```diff
@@ -0,0 +1,137 @@
...
+#endif
+    ifreq=(struct ifreq*)((char*)ifreq+len);
+    i+=len;
+  }
+#endif
+  return interfaces;
+}
+
+std::string kiwix::getBestPublicIp() {
+  std::map<std::string, std::string> interfaces = kiwix::getNetworkInterfaces();
```
mgautierfr, line 101: It is better to use `auto interfaces = kiwix::getNetworkInterfaces();`

src/server/networkTools.cpp
```diff
@@ -0,0 +1,137 @@
...
+#ifndef _WIN32
+  const char* const prioritizedNames[] =
+      { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
+  const int count = (sizeof prioritizedNames) / (sizeof prioritizedNames[0]);
+  for (int i = 0; i < count; ++i) {
+    std::map<std::string, std::string>::const_iterator it =
+        interfaces.find(prioritizedNames[i]);
+    if (it != interfaces.end())
+      return it->second;
+  }
```
mgautierfr, line 112: This would be better written :
```cpp
for(auto name: prioritizedNames) {
  auto it = interfaces.find(name);
  if(it != interfaces.end()) {
    return it->second;
  }
}
```

src/server/networkTools.cpp
```diff
@@ -0,0 +1,137 @@
...
+#endif
+
+  for (std::map<std::string, std::string>::iterator iter = interfaces.begin();
+      iter != interfaces.end(); ++iter) {
+    std::string interfaceIp = iter->second;
+    if (interfaceIp.length() >= 7 && interfaceIp.substr(0, 7) == "192.168")
+      return interfaceIp;
+  }
```
mgautierfr, line 120: It is better to write:
```cpp
for(auto& interface: interfaces) {
  auto interfaceIp = interface.second;
  if (interfaceIp.find("192.168") == 0) {
    return interfaceIp;
  }
}
```

(Same for other tests)

src/server/networkTools.cpp
```diff
@@ -0,0 +1,137 @@
...
+    if (interfaceIp.length() >= 7 && interfaceIp.substr(0, 7) == "172.16.")
+      return interfaceIp;
+  }
+
+  for (std::map<std::string, std::string>::iterator iter = interfaces.begin();
+      iter != interfaces.end(); ++iter) {
+    std::string interfaceIp = iter->second;
+    if (interfaceIp.length() >= 3 && interfaceIp.substr(0, 3) == "10.")
+      return interfaceIp;
+  }
```
mgautierfr, line 134: The three independent loop could be combine in two imbricated loop (one looping on 3 prefix and the child loop going on the interfaces)

src/server/networkTools.h
```diff
@@ -0,0 +1,48 @@
...
+#include <netdb.h>
+#endif
+
+#include <iostream>
+#include <vector>
+#include <string>
+#include <map>
+
+namespace kiwix {
+  std::map<std::string, std::string> getNetworkInterfaces();
```
mgautierfr, line 44: This function is not called outside of `networkTools.cpp`. No need to declared it in the header.

src/server/internalServer.h
```diff
@@ -35,6 +35,7 @@ extern "C" {
 
 #include "server/request_context.h"
 #include "server/response.h"
+#include "networkTools.h"
```
mgautierfr, line 38: We should include this in the `cpp` file, not the header.

mgautierfr (changes requested):
There is a lot of (small) things to change here.

I know that most of the code here is not written by you but as soon as you submit it in your PR you must assume for it.
It would be acceptable to just copy the code if it was a "external" code we don't want to maintain (as we've made for `test/httplib.h`, but here the code is "our" code and we will totally integrate it in our code base).

(I may have not find all things to change. If you find other things to fix while you are rewriting the parts I've found, do not hesitate to fix it directly 🙂).
(And then add you name in the copyright holders).

src/server/networkTools.cpp
juuz0: I used itr instead of `interface` here as `interface` seems to be already reserved  in Windows.

juuz0:
Apart from the changes requested &  the naming issue related to `interface` (wrote in comment), I removed some unnecessary header files.
About  adding name in copyright, not much sure how to write there :D

kelson42:
@juuz0 Would you be able to implement https://github.com/kiwix/kiwix-tools/issues/473 ? So i can verify that what is done on this PR works fine?

juuz0:
@kelson42 The current version of this PR also implements https://github.com/kiwix/kiwix-tools/issues/473

kelson42:
@juuz0 I have tested ans from a user perspective it works fine. Just a small remark, there should be no space before the colon in the message. This is the typographie rule in English.

juuz0:
@kelson42 Thanks! Fixed :)

kelson42:
@juuz0 You are right, the problem is the ZIM file itself. But I have discovered an other bug... if I launch:
```
$ ./kiwix-desktop wikipedia_fr_all_mini_2021-10.zim wikipedia_es_history_maxi_2021-10.zim 
```
Only the first one is open. I guess the two of them should be open right?

src/server/networkTools.cpp
```diff
@@ -0,0 +1,132 @@
...
+std::string kiwix::getBestPublicIp() {
+  auto interfaces = getNetworkInterfaces();
+
+#ifndef _WIN32
+  const char* const prioritizedNames[] =
+      { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
+  for(auto name: prioritizedNames) {
+    auto it = interfaces.find(name);
+    if(it != interfaces.end())
+      return it->second;
```
mgautierfr, line 117: Same here, please add a `{}` couple.

mgautierfr (changes requested):
This is far better now.
I've got only few changes (mostly cosmetic) and we are good.

juuz0:
Should squash now then?

mgautierfr:
> Should squash now then?

Yes, I think you can squash the two last commits.
It is nice to keep the "copy" of the original code and the changes you've made on it in two separated commits.

One missing thing is your copyright.
Simply copy the line `* Copyright 2012 Emmanuel Engelhart <kelson@kiwix.org>` in the files and modify the copied line with the correct year and your name (and your mail if you wish)

src/server/internalServer.cpp
```diff
@@ -187,6 +189,9 @@ bool InternalServer::start() {
   auto server_start_time = std::chrono::system_clock::now().time_since_epoch();
   m_server_id = kiwix::to_string(server_start_time.count());
   m_library_id = m_server_id;
+  std::string url = "http://" + m_addr + ":" + std::to_string(m_port) + "/";
+  std::cout << "The Kiwix server is running and can be accessed in the local network at: "
+            << url << std::endl;
```
mgautierfr, line 194: We must not print the message from libkiwix:
- It is to kiwix-serve to print the message to the user.
- It would not be usefull when the server is run from kiwix-desktop as user doesn't have access to the console output.

We need to define a new method to get the url and make the tools using `libkiwix`'s server display the information to the user.

mgautierfr (changes requested):

src/server/internalServer.cpp
juuz0: So, 
I should remove this helper line.
make a method to get URL. 
And use that URL to implement https://github.com/kiwix/kiwix-tools/issues/473 separately

mgautierfr: Yes. And also do the same in kiwix-desktop (we already display a IP address but we should use the new method in libkiwix)

juuz0:
Done, I think. 

Unrelated & out of curiosity, why is there a distinction between Internal Server & Server class?

kelson42:
GetNetworkInterfaces() shoukd better be exposed to allow https://github.com/kiwix/kiwix-desktop/issues/322.

mgautierfr (changes requested):

include/server.h
```diff
@@ -55,6 +55,11 @@ namespace kiwix
         */
        std::string getHttpUrl();
 
+       /**
+        * Get all available network interfaces
+        */
+       std::map<std::string, std::string> getNetworkInterfaces();
```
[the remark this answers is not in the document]
juuz0: I actually have no idea whats the correct way to do that.
Any hints/help please!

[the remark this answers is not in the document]
mgautierfr: Just remove it :) (The method is already exposed (in your PR) in `networkTools.h`)
 
If we really want to have it a `Server` level we must define the method as static (https://en.cppreference.com/w/cpp/language/static)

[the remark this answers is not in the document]
juuz0: But if it isn't available at a Server level (after removing), how would I use it if needed (like in https://github.com/kiwix/kiwix-desktop/pull/735)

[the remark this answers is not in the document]
juuz0: So, is this approach fine? 
Remove the includes (`networkTools.h`) and functions from internalServer
include `networkTools.h` in `server.cpp`
define `getNetworkInterfaces()` and `getBestPublicIp` as static  (?)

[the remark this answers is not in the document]
mgautierfr: > But if it isn't available at a Server level (after removing), how would I use it if needed

We simply keep the "plain" function in `networkTools.h`.
Then in kiwix-desktop, we simply have to include `networkTools.h` and use the function :
```cpp
#include <kiwix/networkTools.h>
[...]
const auto interfacesMap = kiwix::getNetworkInterfaces();
```

We CAN also define `getNetworkInterfaces` as a static method but it doesn't provide any useful value  (except than explicitly telling that the methode is associated to the "global" server feature).
But we MUST not define `getNetworkInterfaces` as a classic method. Simply because we want the user to be able to use this function before creating a server.

The simpler is to just declare it as a plain function in `networkTools.h` and remove it from the server class.

For `getBestPublicIp` it is different. Is is the best public IP to contact the server. So it must be associated to a server instance. And the returned value may change depending of how the server has been started.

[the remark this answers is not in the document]
juuz0: Thanks for this amazing explaination! 

One more query, for networkTool.h to be available via `#include <kiwix/networkTools.h`, it should be moved to `include/` directory, right?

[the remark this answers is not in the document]
mgautierfr: No, my bad.

The public functions must be declared in `include/tools.h`. `networkTools.h` (and friends) are for private helpers. You can take example on the functions already declared in `tools.h` and implemented in `<foo>Tools.cpp`.
And kiwix-desktop will `#include <kiwix/tools.h>`

juuz0:
Changed name to `ipAddressTools.(h/cpp)` as another `networkTools.(h/cpp)`. If there is a better name, I will quickly change it :)

kelson42:
@juuz0 Feel free to rebase on `origin/master HEAD`.

kelson42:
@mgautierfr It is important to review this PR in priority as many other PR in kiwix-tools and kiwix-desktop depend on it.

include/server.h
```diff
@@ -21,6 +21,7 @@
 #define KIWIX_SERVER_H
 
 #include <string>
+#include <map>
```
mgautierfr, line 24: This is not need now you've moved `getNetworkInterface` out of `server.h`

include/server.h
```diff
@@ -49,6 +50,11 @@ namespace kiwix
         * Stop the daemon.
         */
        void stop();
+       
+       /**
+        *  Get the best public IP from available interfaces given by getNetworkInterfaces()
+        */
+       std::string getBestPublicIp();
```
mgautierfr, line 57: This should not be the "best public IP".
This should be the best public IP where we can connect to the server.

We need a `getBestPublicIp` which get the best IP from networkInterface (At least internally. But it could be public to help UI decide what is the best default value) but it is not what we need here.
If the server is started with a specific address, we want to have this specific address back (else it it useless).
If the server is started without address, we indeed want to the "best" address.

src/server.cpp
```diff
@@ -23,6 +23,7 @@
 #include "name_mapper.h"
 
 #include <string>
+#include <map>
```
mgautierfr, line 26: Do we need this ?

src/server/internalServer.h
```diff
@@ -32,6 +32,7 @@ extern "C" {
 
 #include <atomic>
 #include <string>
+#include <map>
```
mgautierfr, line 35: Do we need this ?

mgautierfr (changes requested):
I think there is a missunderstand about what we want to expose in the API. We want:
- A `getNetworkInterfaces` function to allow UI to propose a choice to the user.
- A `setAddress`/`setPort` on the server to allow the configuration of the server before we start it (we already have this).
- A `getAddress`/`getPort` on the server to know on which address the user can connect to the server (if the server is configured with a specific address/port we must return this. If it is not configured, we must return the "best" address).
- A `getBestPublicIp`, a choice from `getNetworkInterfaces` to allow UI to propose a default choice. (But maybe the best default choice is ANY).

`getNetworkInterfaces` and `getBestPublicIp` are not associated to any server.

---

Why renaming `networkTools.h` to `ipAddressTools.h` ?
You've wrote that a `networkTools.h` already exists but which one ?

---

I've made a review based on the final change. There are a lot of commit now in this PR that should be removed. (For example, all the changes around the message printed to the user should be removed)
Consider your git history as a recipe for other developers (first do this, then do this...) and not as a changelog of what you have done (I've tried this, then that, revert this, change ...)
It is nice for the reviewer to know what I have changed since the last review. To do this,
you can add new commit with the title "fixup! <title of the previous commit to change>" to tell the reviewer (and git) that this commit will change the previous commit but it is not a new step in the recipe.

juuz0:
> Why renaming networkTools.h to ipAddressTools.h ?
You've wrote that a networkTools.h already exists but which one ?

It is in `src/tools/`

> I think there is a missunderstand ....... getNetworkInterfaces and getBestPublicIp are not associated to any server.

Thanks for this, clears a lot for me. 
Can you also tell whats the correct way to define a function for server (getAddress for example). Should it be defined in both `server` and `internalServer` (and the function in `server` calls the `internalServer` one) ? 

> I've made a review based on the final change ............. he previous commit but it is not a new step in the recipe.

Noted, will follow from all next commits :)

mgautierfr:
> It is in src/tools/

Why not simply add the functions declaration/definition in the existing `networkTools.(h|cpp)` ?

> Should it be defined in both server and internalServer (and the function in server calls the internalServer one) ?

Yes

juuz0:
All done, rebased

mgautierfr (approved):
Be careful, your second commit move the code from `src/server/networkTools.cpp` to `src/tools/networkTools.cpp`, but it also change the code itself.
It make the review (and potential bug detection) pretty difficult.
I'm assuming the final code is ok with what I already reviewed.

Anyway, I agree with the API and approve the PR.
Please rebase on new master so we can merge.
It misses some docstring now we have a API documentation generated. But we can do it in another PR.

juuz0:
Thanks for checking and the patience :D
I will add doc strings in another PR (will have to learn how to write them first)

juuz0:
Code factor was failing, fixed now.

Issue #625, opened 2021-10-25
Opds dumper need to read all archives to have illustration size.


mgautierfr:
https://github.com/kiwix/libkiwix/blob/master/src/opds_dumper.cpp#L57-L70 make the opds dumper read all zim file in the library to create the opds stream.

The normal behavior should be that we trust the `library.xml` (when loading from it) and do not read the file to update the metadata information we know about the books.
This kind of method should be on the library side (`Manager::readBookFromPath` or `Book::update`)

On my side, with 14 zim files on usb extern hard drive, the first opds request from the home page takes more than 1 minutes.

mgautierfr:
To be exact, we don't read all archives in the library but all archives returned by the opds stream. If the stream is paged (limited to N books) we read only N archives, but if not, we read all archives.
But, another issue, we do this every time, for each requests, as there is no cache. The fs cache help a bit but that all.

veloman-yunkan:
> https://github.com/kiwix/libkiwix/blob/master/src/opds_dumper.cpp#L57-L70 make the opds dumper read all zim file in the library to create the opds stream.
> 
> The normal behavior should be that we trust the `library.xml` (when loading from it) and do not read the file to update the metadata information we know about the books. This kind of method should be on the library side (`Manager::readBookFromPath` or `Book::update`)
> 
> On my side, with 14 zim files on usb extern hard drive, the first opds request from the home page takes more than 1 minutes.

@kelson42 @mgautierfr 

The said change in `OPDSDumper` was introduced by #577 while fixing #533. The current format of the library XML doesn't allow for multiple illustrations, that's why the needed information is read from the ZIM archives. To address this issue, we must enhance the library XML format:

1. allow multiple illustrations per book
2. include information about dimensions of each illustration

Tools that generate library XMLs must be updated as well.

In this context, I wonder why we need a separate XML format for the library. Can't we replace the `<book>` nodes in the library XML with (probably slightly customized) OPDS `<entry>` nodes, so that we can eliminate some functional duplication between [`kiwix::Book::updateFromXml()`](https://github.com/kiwix/libkiwix/blob/9479c0685ddaa71ac5de45b42a962d5b65bfe077/src/book.cpp#L111) and [`Book::updateFromOpds()`](https://github.com/kiwix/libkiwix/blob/9479c0685ddaa71ac5de45b42a962d5b65bfe077/src/book.cpp#L155)?

mgautierfr:
[...]

PR #626, opened 2021-10-28
Better, simpler handle_meta.

### LINKED ISSUE #616: Expose counter on meta endpoint
Currently, we can't access `Counter` metadata via kiwix-serve. I've set this as *bug* because it's a regression but might have been removed on purpose.

I know that @kelson42 also relies on this.

https://github.com/kiwix/libkiwix/blob/3b942bb7458af7daae00505b4a660a47eb33628f/src/server/internalServer.cpp#L404-L427

Side questions:
- Why the change in case? Metadata are title case and having to switch feels unnatural.
- Why not offering all metadata using this endpoint? I see that all of the supported ones have a dedicated getter here. Maybe we should provide a generic one as well, to be used for others?

mgautierfr:
- Serve all metadata available in the zim archive.
- Serve (almost) only metadata available in the zim archive
  (no default value).
- Provide the correct mimytype all the time
  (not only for favicon/Illustration)
- Provide a small compatibility layer for zim archive without favicon
  (Mainly because we are also providing a compatibility layer elsewhere)

Fix #616

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -385,12 +385,11 @@ SuggestionsList_t getSuggestions(const zim::Archive* const archive,
 std::unique_ptr<Response> InternalServer::handle_meta(const RequestContext& request)
 {
   std::string bookName;
-  std::string bookId;
   std::string meta_name;
   std::shared_ptr<zim::Archive> archive;
   try {
     bookName = request.get_argument("content");
-    bookId = mp_nameMapper->getIdForName(bookName);
+    auto bookId = mp_nameMapper->getIdForName(bookName);
     meta_name = request.get_argument("name");
     archive = mp_library->getArchiveById(bookId);
   } catch (const std::out_of_range& e) {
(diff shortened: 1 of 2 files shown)


src/server/internalServer.cpp
```diff
@@ -401,36 +400,25 @@ std::unique_ptr<Response> InternalServer::handle_meta(const RequestContext& requ
...
+    // will try to access 48px illustration even if we actually have a favicon.
+    if (const unsigned illustrationSize = parseIllustration(meta_name)) {
+      std::string content, mimetype;
+      getArchiveFavicon(*archive, illustrationSize, content, mimetype);
+      auto response = ContentResponse::build(*this, content, mimetype);
+      response ->set_cacheable();
+      return std::move(response);
+    } else {
+      return Response::build_404(*this, request, bookName, "");
+    }
```
veloman-yunkan, line 420: I think that this piece of code can be rewritten, by expanding `getArchiveFavicon()`, as follows:

```c++
  zim::Item meta_item;
  try {
    if (const unsigned illustrationSize = parseIllustration(meta_name)) {
      // Some (old) zim have no Illustration metadata but has a favicon in the content.
      // But the api will return the favicon as a 48px Illustration. And so kiwix-serve
      // will try to access 48px illustration even if we actually have a favicon.
      meta_item = archive.getIllustrationItem(size);
    } else {
      meta_item = archive->getMetadataItem(meta_name);
    }
  } catch (const zim::EntryNotFound& e) {
    return Response::build_404(*this, request, bookName, "");
  }
  auto response = ContentResponse::build(*this, meta_item.getData(), meta_item.getMimetype());
  response->set_cacheable();
  return std::move(response);
```

It may also make sense to extract the body of the `try` statement into a free helper function `zim::Item getIllustrationOrMetadataItem(const std::string& meta_name)`. Then you can restore the backward compatibility of the API (handling of the previously lower-case names of metadata entries as well as the now dropped `favicon` metadata without introducing any clutter to `InternalServer::handle_meta()`.

veloman-yunkan (changes requested):
This PR introduces a change in the metadata API:

* the names of metadata fields must now start with an uppercase letter
* the `favicon` metadata item is no longer supported

Is it deliberate? I am asking this, since the name of the PR suggests that it should be a non-functional change.

src/server/internalServer.cpp
veloman-yunkan: BTW, the code in my previous comment should also address @rgaudin's [concern](https://github.com/kiwix/libkiwix/pull/626#pullrequestreview-791864405)

rgaudin:
> This PR introduces a change in the metadata API:
> 
>     * the names of metadata fields must now start with an uppercase letter
> 
>     * the `favicon` metadata item is no longer supported
> 
> 
> Is it deliberate? I am asking this, since the name of the PR suggests that it should be a non-functional change.

Both have been discussed with @mgautierfr and @kelson42 ; sorry it wasn't clear in the PR. We want this API to do less and be more transparently. As you now, metadata starts with an uppercase letter and the libzim7 rewrite introduced the all-lowercase version. We want to come back to requesting what's stored.

Regarding favicon, we figured this shouldn't be supported as it is already properly available via OPDS.

mgautierfr:
> Is it deliberate? I am asking this, since the name of the PR suggests that it should be a non-functional change.

Yes, I should have be clearer in my description.

This endpoint is not widely used for now. I've created it only for the favicon as we need to put a url in the opds stream. I've added also some other metadata but it was "just" because it was a nice to have.
Now that OPDS stream are created using the `Illustration_*" name, favicon name is not used.
And we think it is better to rewrite this endpoint (and use it) without fancy name rewriting. If this rewriting appear to be necessary we may add it in the future. But if we have them right now it will become a feature we need to maintain.

The only compatibility we need to have is with "old" zim files.

---

> As you now, metadata starts with an uppercase letter and the libzim7 rewrite introduced the all-lowercase version

libzim7 didn't make any change on this side. The spec (https://wiki.openzim.org/wiki/Metadata) tells it is still a "Titlecase" key. And libzim itself do not modify the name of the metadata you provide.

mgautierfr:
Well, in fact I think this PR is globally wrong. Or more exactly is goes in a wrong direction.

This PR, as the previous code, get the data to send back from the `zim::Archive`. But we should take them from the `kiwix::Book`.
Else we will have the same issue than in https://github.com/kiwix/libkiwix/issues/625. All client listing zim files and display icons to the user will trigger a loading of the archive on the opds server. This is definitively something we don't want, we must follow the "Trust the library" directive.

kelson42:
We agree to follow an other approach. Closing.

PR #628, opened 2021-10-30
Clickable kiwix-serve tiles

### LINKED ISSUE #627: [Server] Make zim card clickable.
The new UI of the kiwix-serve front-end is a bit disturbing.

- The "card" (displaying information about a book) reacts to mouse over (small zoom) but it is not clickable. It make the user think something is wrong as the "system" reacts to his moves but doesn't work.
- The same way, the lang tag in the cards looks like a button but is not. One may think that if the book exists in different language, we will have different button on the same card. On so it may also try to click on the lang "button" to open the zim.

There is the `preview` link to open the book. It somehow makes sense along the download link, but without it (as we can see in http://dev.library.kiwix.org) it feel a bit alone and odd. And remember that on local server, there will probably not have any download button.

What could be made :
- Have the whole card clickable.
- Remove the preview link (the card itself is the preview link).
- Change the styling of the lang tag and make it more like the "category" tag.
- Keep the download button. Either unchanged or maybe more like the current lang tag (like a button). I'm not sure if this is a good idea to change the styling. To test.

kelson42:
Fixes #627


kelson42:
@mgautierfr Your point about the language code is clear but I have mixed feelings about that so I made a really light change which I hope goes in the right direction. But, I have no clue why macOS CI fails.
@MananJethwani Please let me know if I don't have broken something. After a bit of usage we have decided to slightly change things on the tiles.

mgautierfr (changes requested):
This is far better than previously.

- The code factor error must be (easily) fixed.
- I would add a small round corner (1px should be enough) around the download button. I found the current version a bit rough.

I confirm the removing of the border around the tag lang, it make the tag a bit less like a button.

mgautierfr (approved):

rgaudin:
Fixed the codefactor issue and added the rounded border suggested by @mgautierfr

PR #630, opened 2021-10-31
Support for multiple illustrations in a Book

### LINKED ISSUE #625: Opds dumper need to read all archives to have illustration size.

veloman-yunkan:
Fixes #625.

Note that this fix only adds support for multiple illustrations inside the `kiwix::Book` class, **without** enhancing the library format. As a result, ZIM files with multiple illustrations will not appear as such in the OPDS output if the kiwix server is started with a trusted library.

### DIFF
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -108,9 +108,16 @@ void Book::update(const zim::Archive& archive) {
   m_mediaCount = getArchiveMediaCount(archive);
   m_size = static_cast<uint64_t>(getArchiveFileSize(archive)) << 10;
 
-  Illustration& favicon = getMutableDefaultIllustration();
-  getArchiveFavicon(archive, 48, favicon.data, favicon.mimeType);
-  // XXX: isn't favicon.url neglected here?
+  m_illustrations.clear();
+  for ( const auto illustrationSize : archive.getIllustrationSizes() ) {
+    const auto illustration = std::make_shared<Illustration>();
+    const zim::Item illustrationItem = archive.getIllustrationItem(illustrationSize);
+    illustration->width = illustration->height = illustrationSize;
+    illustration->mimeType = illustrationItem.getMimetype();
+    illustration->data = illustrationItem.getData();
+    // NOTE: illustration->url is left uninitialized
+    m_illustrations.push_back(illustration);
+  }
 }
 
 #define ATTR(name) node.attribute(name).value()
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -104,6 +104,7 @@ void Book::update(const zim::Archive& archive) {
 
   Illustration& favicon = getMutableDefaultIllustration();
   getArchiveFavicon(archive, 48, favicon.data, favicon.mimeType);
+  // XXX: isn't favicon.url neglected here?
 }
 
 #define ATTR(name) node.attribute(name).value()
@@ -182,6 +183,7 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
     }
     if (rel == "http://opds-spec.org/image/thumbnail") {
       Illustration& favicon = getMutableDefaultIllustration();
+      // XXX: shouldn't favicon.data be cleared()?
       favicon.url = urlHost + linkNode.attribute("href").value();
       favicon.mimeType = linkNode.attribute("type").value();
     }
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -102,7 +102,8 @@ void Book::update(const zim::Archive& archive) {
   m_mediaCount = getArchiveMediaCount(archive);
   m_size = static_cast<uint64_t>(getArchiveFileSize(archive)) << 10;
 
-  getArchiveFavicon(archive, 48, m_illustration.data, m_illustration.mimeType);
+  Illustration& favicon = getMutableDefaultIllustration();
+  getArchiveFavicon(archive, 48, favicon.data, favicon.mimeType);
 }
 
 #define ATTR(name) node.attribute(name).value()
@@ -129,9 +130,10 @@ void Book::updateFromXml(const pugi::xml_node& node, const std::string& baseDir)
   m_articleCount = strtoull(ATTR("articleCount"), 0, 0);
   m_mediaCount = strtoull(ATTR("mediaCount"), 0, 0);
   m_size = strtoull(ATTR("size"), 0, 0) << 10;
-  m_illustration.data = base64_decode(ATTR("favicon"));
-  m_illustration.mimeType = ATTR("faviconMimeType");
-  m_illustration.url = ATTR("faviconUrl");
+  Illustration& favicon = getMutableDefaultIllustration();
+  favicon.data = base64_decode(ATTR("favicon"));
+  favicon.mimeType = ATTR("faviconMimeType");
+  favicon.url = ATTR("faviconUrl");
   try {
     m_downloadId = ATTR("downloadId");
   } catch(...) {}
@@ -179,8 +181,9 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       m_size = strtoull(linkNode.attribute("length").value(), 0, 0);
     }
     if (rel == "http://opds-spec.org/image/thumbnail") {
-      m_illustration.url = urlHost + linkNode.attribute("href").value();
-      m_illustration.mimeType = linkNode.attribute("type").value();
+      Illustration& favicon = getMutableDefaultIllustration();
+      favicon.url = urlHost + linkNode.attribute("href").value();
+      favicon.mimeType = linkNode.attribute("type").value();
     }
  }
 
@@ -213,25 +216,36 @@ void Book::setPath(const std::string& path)
    : path;
 }
 
+const Book::Illustration& Book::getDefaultIllustration() const
+{
+  return m_illustration;
+}
+
+Book::Illustration& Book::getMutableDefaultIllustration()
+{
+  return m_illustration;
+}
+
 const std::string& Book::getFavicon() const {
-  if (m_illustration.data.empty() && !m_illustration.url.empty()) {
+  const Illustration& favicon = getDefaultIllustration();
+  if (favicon.data.empty() && !favicon.url.empty()) {
     try {
-      m_illustration.data = download(m_illustration.url);
+      favicon.data = download(favicon.url);
     } catch(...) {
-      std::cerr << "Cannot download favicon from " << m_illustration.url;
+      std::cerr << "Cannot download favicon from " << favicon.url;
     }
   }
-  return m_illustration.data;
+  return favicon.data;
 }
 
 const std::string& Book::getFaviconUrl() const
 {
-  return m_illustration.url;
+  return getDefaultIllustration().url;
 }
 
 const std::string& Book::getFaviconMimeType() const
 {
-  return m_illustration.mimeType;
+  return getDefaultIllustration().mimeType;
 }
 
 std::string Book::getTagStr(const std::string& tagName) const {
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -55,27 +55,7 @@ bool Book::update(const kiwix::Book& other)
   if (m_id != other.m_id)
     return false;
 
-  m_readOnly = other.m_readOnly;
-  m_path = other.m_path;
-  m_pathValid = other.m_pathValid;
-  m_title = other.m_title;
-  m_description = other.m_description;
-  m_language = other.m_language;
-  m_creator = other.m_creator;
-  m_publisher = other.m_publisher;
-  m_date = other.m_date;
-  m_url = other.m_url;
-  m_name = other.m_name;
-  m_flavour = other.m_flavour;
-  m_tags = other.m_tags;
-  m_category = other.m_category;
-  m_origId = other.m_origId;
-  m_articleCount = other.m_articleCount;
-  m_mediaCount = other.m_mediaCount;
-  m_size = other.m_size;
-  m_illustrations = other.m_illustrations;
-  m_downloadId = other.m_downloadId;
-
+  *this = other;
   return true;
 }
 
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -73,11 +73,7 @@ bool Book::update(const kiwix::Book& other)
   m_articleCount = other.m_articleCount;
   m_mediaCount = other.m_mediaCount;
   m_size = other.m_size;
-  m_illustrations.clear();
-  for ( const auto& ill : other.m_illustrations ) {
-    m_illustrations.push_back(std::make_shared<Illustration>(*ill));
-  }
-
+  m_illustrations = other.m_illustrations;
   m_downloadId = other.m_downloadId;
 
   return true;
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -40,8 +40,6 @@ Book::Book() :
   m_pathValid(false),
   m_readOnly(false)
 {
-  const auto illustration = std::make_shared<Illustration>();
-  m_illustrations.assign(1, illustration);
 }
 
 /* Destructor */
@@ -241,12 +239,6 @@ const Book::Illustration& Book::getDefaultIllustration() const
     }
   }
   throw std::runtime_error("No default illustration");
-}
-
-Book::Illustration& Book::getMutableDefaultIllustration()
-{
-  const Book* const const_this = this;
-  return const_cast<Illustration&>(const_this->getDefaultIllustration());
 }
 
 const std::string& Book::Illustration::getData() const
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -231,6 +231,8 @@ void Book::setPath(const std::string& path)
    : path;
 }
 
+const Book::Illustration Book::missingDefaultIllustration;
+
 const Book::Illustration& Book::getDefaultIllustration() const
 {
   for ( const auto& ilPtr : m_illustrations ) {
@@ -238,7 +240,7 @@ const Book::Illustration& Book::getDefaultIllustration() const
       return *ilPtr;
     }
   }
-  throw std::runtime_error("No default illustration");
+  return missingDefaultIllustration;
 }
 
 const std::string& Book::Illustration::getData() const
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -47,6 +47,11 @@ Book::~Book()
 {
 }
 
+Book::Illustrations Book::getIllustrations() const
+{
+  return m_illustrations;
+}
+
 bool Book::update(const kiwix::Book& other)
 {
   if (m_readOnly)
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -40,7 +40,10 @@ Book::Book() :
   m_pathValid(false),
   m_readOnly(false)
 {
+  const auto illustration = std::make_shared<Illustration>();
+  m_illustrations.assign(1, illustration);
 }
+
 /* Destructor */
 Book::~Book()
 {
@@ -72,7 +75,10 @@ bool Book::update(const kiwix::Book& other)
   m_articleCount = other.m_articleCount;
   m_mediaCount = other.m_mediaCount;
   m_size = other.m_size;
-  m_illustration = other.m_illustration;
+  m_illustrations.clear();
+  for ( const auto& ill : other.m_illustrations ) {
+    m_illustrations.push_back(std::make_shared<Illustration>(*ill));
+  }
 
   m_downloadId = other.m_downloadId;
 
@@ -220,12 +226,14 @@ void Book::setPath(const std::string& path)
 
 const Book::Illustration& Book::getDefaultIllustration() const
 {
-  return m_illustration;
+  assert(m_illustrations.size() == 1);
+  return *m_illustrations.front();
 }
 
 Book::Illustration& Book::getMutableDefaultIllustration()
 {
-  return m_illustration;
+  const Book* const const_this = this;
+  return const_cast<Illustration&>(const_this->getDefaultIllustration());
 }
 
 const std::string& Book::Illustration::getData() const
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -227,10 +227,13 @@ const Book::Illustration& Book::getDefaultIllustration() const
 const std::string& Book::Illustration::getData() const
 {
   if (data.empty() && !url.empty()) {
-    try {
-      data = download(url);
-    } catch(...) {
-      std::cerr << "Cannot download favicon from " << url;
+    const std::lock_guard<std::mutex> l(mutex);
+    if ( data.empty() ) {
+      try {
+        data = download(url);
+      } catch(...) {
+        std::cerr << "Cannot download favicon from " << url;
+      }
     }
   }
   return data;
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -72,9 +72,7 @@ bool Book::update(const kiwix::Book& other)
   m_articleCount = other.m_articleCount;
   m_mediaCount = other.m_mediaCount;
   m_size = other.m_size;
-  m_favicon = other.m_favicon;
-  m_faviconMimeType = other.m_faviconMimeType;
-  m_faviconUrl = other.m_faviconUrl;
+  m_illustration = other.m_illustration;
 
   m_downloadId = other.m_downloadId;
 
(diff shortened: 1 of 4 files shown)


src/book.cpp
```diff
@@ -104,7 +113,16 @@ void Book::update(const zim::Archive& archive) {
...
+  m_illustrations.clear();
+  for ( const auto illustrationSize : archive.getIllustrationSizes() ) {
+    const auto illustration = std::make_shared<Illustration>();
+    const zim::Item illustrationItem = archive.getIllustrationItem(illustrationSize);
+    illustration->width = illustration->height = illustrationSize;
+    illustration->mimeType = illustrationItem.getMimetype();
+    illustration->data = illustrationItem.getData();
+    // XXX: illustration->url is left uninitialized
+    m_illustrations.push_back(illustration);
+  }
```
mgautierfr, line 125: This doesn't guaranty that first illustration will be the default one (48x48@1). It even doesn't guaranty we will have a illustration at all.

mgautierfr (changes requested):
This goes in the right direction but it is not enough :
- "Default" book should have no illustration.
- A book/zim may have no default illustration or even no illustration at all. The spec says we must create a zim file with a illustration but nothing really prevent someone to create a zim not following the spec (And it is what we done in old zim files)

I would move to :
- Use `Illustrations` vector all the time. (const Illustration).
- Create illustration only when we are sure we have one.
- I'm not sure we need a `getDefaultIllustration` at all. 

And we must make evolved the library xml format to support multiple illustrations (which is not a easy part as we also need to keep (forward at least) compatibility)

veloman-yunkan:
>   * "Default" book should have no illustration.

This contradicts `Book`'s (old) C++ API - there is a `getFavicon()` method and no `hasFavicon()` method. Should we drop all 3 `Book::getFavicon*()` functions? Or throw an exception if they are called on a Book that doesn't have any illustrations?

mgautierfr:
Yes, the `getFavicon()` is returning an empty string if there is no Favicon. This is a wrong API and we should drop it.
This feature will probably come in a major release, it is ok to break the (C++) api. What we don't wont is to break the web/opds api as user may use a old client with a up to date kiwix-serve used at library.kiwix.org

However, in the same time, we should keep the android API (at least until we have moved the wrapper in https://github.com/kiwix/java-libkiwix and have a proper wrapper there).

On my side it is ok to drop the `Book::getFavicon*()` methods from public c++ api. And either port the wrapper to the new methods or keep `getFavicon()` as internal helper.

veloman-yunkan:
> I would move to :
> 
>     * Use `Illustrations` vector all the time. (const Illustration).
> 
>     * Create illustration only when we are sure we have one.
> 
>     * I'm not sure we need a `getDefaultIllustration` at all.

Done. `getDefaultIllustration()` was preserved in order to avoid dealing with the removal of `get/setFavicon*()` API at this point. The only change with respect to that deprecated API is that if no 48x48 illustration is available in the Book then the `getFavicon*()` functions will throw.

src/book.cpp
```diff
@@ -41,11 +41,17 @@ Book::Book() :
...
 }
+
 /* Destructor */
 Book::~Book()
 {
 }
 
+Book::Illustrations Book::getIllustrations() const
+{
+  return Illustrations(m_illustrations.begin(), m_illustrations.end());
```
mgautierfr, line 52: We don't need to explicitly create a `Illustrations` now. Copy constructor should be enough.

src/book.cpp
```diff
@@ -72,9 +78,10 @@ bool Book::update(const kiwix::Book& other)
   m_articleCount = other.m_articleCount;
   m_mediaCount = other.m_mediaCount;
   m_size = other.m_size;
-  m_favicon = other.m_favicon;
-  m_faviconMimeType = other.m_faviconMimeType;
-  m_faviconUrl = other.m_faviconUrl;
+  m_illustrations.clear();
+  for ( const auto& ill : other.m_illustrations ) {
+    m_illustrations.push_back(std::make_shared<Illustration>(*ill));
+  }
```
mgautierfr, line 84: I wonder if we still need this.
As we are storing `std::shared<const Illustration>` it should be ok to share the illustration between several books.
And so we can use the copy constructor.

src/book.cpp
```diff
@@ -208,22 +227,67 @@ std::string Book::getHumanReadableIdFromPath() const
...
 
+void Book::setFavicon(const std::string& favicon)
+{
+  getMutableDefaultIllustration().data = favicon;
+}
+
+void Book::setFaviconMimeType(const std::string& faviconMimeType)
+{
+  getMutableDefaultIllustration().mimeType = faviconMimeType;
+}
```
mgautierfr, line 238: We use them only in tests (so we don't really use them).
We probably don't want to keep this old API to set the illustration.
If we really want to provide such API (I'm not sure we should), it should be at least some kind of `setDefaultIllustration`.

src/book.cpp
```diff
@@ -208,22 +227,67 @@ std::string Book::getHumanReadableIdFromPath() const
...
+      std::cerr << "Cannot download favicon from " << url;
     }
   }
-  return m_favicon;
+  return data;
+}
+
+const std::string& Book::getFavicon() const {
+  return getDefaultIllustration().getData();
+}
```
mgautierfr, line 281: As this now throw an exception, it is a change in the API.
At least libxml_dumper expect a empty string if there is no favicon and do not catch exception.
It appears that the java wrapper already catch any exception.

mgautierfr (changes requested):

src/book.cpp
veloman-yunkan: We don't store `std::shared_ptr<const Illustration>` in `m_illustrations` (we use `std::shared_ptr<Illustration>` instead). `Book::getMutableDefaultIllustration()` may return a reference to an existing entry in `m_illustrations`. The `const_cast` therein is justified by the fact that the const `getDefaultIllustration()` member function is called from a non-const member function.

Sharing the default illustration between books will misbehave when a book is copied from another via `Book::update()`, and then one of them is updated by `Book::updateFromXml()` or `Book::updateFromOpds()` - the default illustration will change in both books. Though this scenario doesn't seem likely, we better not open up possibilities for such bugs.

src/book.cpp
veloman-yunkan: See https://github.com/kiwix/libkiwix/pull/630#discussion_r744215633

src/book.cpp
veloman-yunkan: > It appears that the java wrapper already catch any exception.

I don't think that is correct. At least [Book's getters](https://github.com/kiwix/libkiwix/blob/9479c0685ddaa71ac5de45b42a962d5b65bfe077/src/wrapper/java/book.cpp#L52-L80) do not handle exceptions.

I added handling of favicon-less books in `LibXMLDumper`, but I am not sure how to proceed with the java wrapper.

veloman-yunkan: I reverted the change to `LibXMLDumper`, as well as the earlier change in the API of `Book::getFavicon()`. Now instead of throwing, `Book::getDefaultIllustration()` returns a reference to a static `Illustration` object representing the case of a missing favicon. Thus books without a favicon don't have it included in the result of `getIllustrations()`. Yet the `getFavicon*()` methods keep returning empty strings.

One question is this: if the book has one or more illustrations but none with 48x48 dimensions, maybe calling `getDefaultIllustration()` on it should return one of those existing illustrations rather than the placeholder `missingDefaultIllustration` object?

mgautierfr: > I don't think that is correct. At least Book's getters do not handle exceptions.

You're right. I was reading [`Reader` code](https://github.com/kiwix/libkiwix/blob/9479c0685ddaa71ac5de45b42a962d5b65bfe077/src/wrapper/java/kiwixreader.cpp#L221-L239)

> I am not sure how to proceed with the java wrapper.

The easier is probably to catch the exception in the wrapper and make it return a empty string if there is no favicon (as before).

> I reverted the change to LibXMLDumper, as well as the earlier change in the API of Book::getFavicon(). Now instead of throwing, Book::getDefaultIllustration() returns a reference to a static Illustration object representing the case of a missing favicon. Thus books without a favicon don't have it included in the result of getIllustrations(). Yet the getFavicon*() methods keep returning empty strings.

I dislike this. The previous version was better. We must design our API following the logic we want to expose and then implement the method with the constraints we have (compatibility and transition are part of the constraints)
The illustration idea is that we can have several illustrations in a zim file (or none). We somehow mandate creator to at least add a 48x48@1 illustration but we cannot force them, and previous zim had no illustration (and sometime no favicon), so we must be prepare to this.

So :
- `Book` have a list of constant illustrations.
- If there is no illustration, this list is empty.
- If a "user" want to get a illustration with a specific size and the illustration is not present, he gets an error (smart choice is a open question. We have this https://github.com/openzim/libzim/issues/541 on libzim side)
- We need to make evolve our data structure (library.xml and opds) to "store" several illustations.

We must handle compatibility with :
- Old zim (this is already handled by libzim itself)
- Old library.xml or opds stream. Here we must read the data and "convert" it to our new scheme.
- Old clients :
  - For cpp clients, we can simply break the API (at least now, because next release will break the API anyway) and ask client to update (and we must do it on kiwix-desktop side)
  - For opds clients, we must ensure that parsing the new opds stream will be ok. As current implementation simply loop on thumbnail and overwrite the favicon data, putting the 48x48@1 image at the end should make things compatible.
  - For xml client it is bit trickier. As the current implementation simply doesn't check the library version (https://github.com/kiwix/libkiwix/blob/master/src/manager.cpp#L58) we must keep a (temporally at least) compatibility by keeping a favicon base64 encoded in the book attributes.
  - For android/java wrapper, we are in the (beginning) middle of moving the wrapper is a separated repository. I suggest we don't change the API here (and so add a compatibility in the wrapper layer). When the wrapper will have moved, we will handle it as a simple cpp client.


Keeping this `getDefaultIllustration` make me think we stick in the middle of a transition where we keep some kind of technical debt.

mgautierfr: > I don't think that is correct. At least Book's getters do not handle exceptions.

You're right. I was reading [`Reader` code](https://github.com/kiwix/libkiwix/blob/9479c0685ddaa71ac5de45b42a962d5b65bfe077/src/wrapper/java/kiwixreader.cpp#L221-L239)

> I am not sure how to proceed with the java wrapper.

The easier is probably to catch the exception in the wrapper and make it return a empty string if there is no favicon (as before).

> I reverted the change to LibXMLDumper, as well as the earlier change in the API of Book::getFavicon(). Now instead of throwing, Book::getDefaultIllustration() returns a reference to a static Illustration object representing the case of a missing favicon. Thus books without a favicon don't have it included in the result of getIllustrations(). Yet the getFavicon*() methods keep returning empty strings.

I dislike this. The previous version was better. We must design our API following the logic we want to expose and then implement the method with the constraints we have (compatibility and transition are part of the constraints)
The illustration idea is that we can have several illustrations in a zim file (or none). We somehow mandate creator to at least add a 48x48@1 illustration but we cannot force them, and previous zim had no illustration (and sometime no favicon), so we must be prepare to this.

So :
- `Book` have a list of constant illustrations.
- If there is no illustration, this list is empty.
- If a "user" want to get a illustration with a specific size and the illustration is not present, he gets an error (smart choice is a open question. We have this https://github.com/openzim/libzim/issues/541 on libzim side)
- We need to make evolve our data structure (library.xml and opds) to "store" several illustations.

We must handle compatibility with :
- Old zim (this is already handled by libzim itself)
- Old library.xml or opds stream. Here we must read the data and "convert" it to our new scheme.
- Old clients :
  - For cpp clients, we can simply break the API (at least now, because next release will break the API anyway) and ask client to update (and we must do it on kiwix-desktop side)
  - For opds clients, we must ensure that parsing the new opds stream will be ok. As current implementation simply loop on thumbnail and overwrite the favicon data, putting the 48x48@1 image at the end should make things compatible.
  - For xml client it is bit trickier. As the current implementation simply doesn't check the library version (https://github.com/kiwix/libkiwix/blob/master/src/manager.cpp#L58) we must keep a (temporally at least) compatibility by keeping a favicon base64 encoded in the book attributes.
  - For android/java wrapper, we are in the (beginning) middle of moving the wrapper is a separated repository. I suggest we don't change the API here (and so add a compatibility in the wrapper layer). When the wrapper will have moved, we will handle it as a simple cpp client.


Keeping this `getDefaultIllustration` make me think we stick in the middle of a transition where we keep some kind of technical debt.

veloman-yunkan: > I dislike this. 

> Keeping this `getDefaultIllustration` make me think we stick in the middle of a transition where we keep some kind of technical debt.

`Book::getDefaultIllustration()` is not part of the public API. It is just an internal function, helping to temporarily preserve the deprecated `getFavicon*()` API. Keeping the `getFavicon*()` API but changing its behavior (throwing an exception when there is no favicon in the ZIM archive) doesn't make any sense - we should either get rid of it completely or ensure perfect backward compatibility. The current state of this PR is along the lines of the latter option.

mgautierfr: > Keeping the `getFavicon*()` API but changing its behavior (throwing an exception when there is no favicon in the ZIM archive) doesn't make any sense

I'm in favor of removing `getFavicon*()` from the API. Or, if we not removing it, we .keep it as a compatibility layer, using the "correct" API and not changing its behavior.

I'm saying that we must design our API without `getFavicon*()` in mind and then add a compatibility layer.
But if the "normal" API returns a `missingDefaultIllustration` to help implement `getFavicon*()` compatibility functions, there is something wrong.
If there is no illustration, the normal API must return nothing (or empty list, or throw a exception). 

> `Book::getDefaultIllustration()` is not part of the public API

Yes, but it makes me think we are building the implementation around this default illustration. But there is no fundamental default illustration. There are illustrations, with different size, and by default we use to look for the 48x48@1 one. The favicon thing is just a compatibility thing with old behavior.

veloman-yunkan: > I'm saying that we must design our API without `getFavicon*()` in mind and then add a compatibility layer.

That's what I've done. The proposed API is the newly introduced public method `Book::getIllustrations()` which returns *only existing illustrations*.  

> > `Book::getDefaultIllustration()` is not part of the public API
> 
> Yes, but it makes me think we are building the implementation around this default illustration.

`Book::getDefaultIllustration()` only serves as the compatibility layer for the deprecated `getFavicon*()` API.

src/book.cpp
mgautierfr: > Sharing the default illustration between books will misbehave when a book is copied from another via Book::update(), and then one of them is updated by Book::updateFromXml() or Book::updateFromOpds() - the default illustration will change in both books. Though this scenario doesn't seem likely, we better not open up possibilities for such bugs.

In my idea, all illustrations are const. So a `Book::updateFrom*` always clear the `m_illustrations` vector and recreate the illustration(s) from what is in the xml/opds/archive. We don' share the default illustration because there would be no default illustration.

src/book.cpp
mgautierfr: > That's what I've done. The proposed API is the newly introduced public method Book::getIllustrations() which returns only existing illustrations.

I've re-read your PR and your right. I've went with a general tirade about public API but we were ok on this. My bad.

> Book::getDefaultIllustration() only serves as the compatibility layer for the deprecated getFavicon*() API

I think this is on the `Book::getMutableDefaultIllustration` we disagree on.
When we parse the xml/opds, we simply have to clear the vector and create a (48x48) Illustration and add it to the vector. Illustration don't have to be mutable (except temporarily at creation).
The `Book::getDefaultIllustration` would simply be a helper to find the 48x48 illustration and throw a exception if not found.

veloman-yunkan: > When we parse the xml/opds, we simply have to clear the vector and create a (48x48) Illustration and add it to the vector. Illustration don't have to be mutable (except temporarily at creation).

OK

> The `Book::getDefaultIllustration` would simply be a helper to find the 48x48 illustration and throw a exception if not found.

Do you strongly insist on throwing an exception? What is so bad with the current hack of returning the fake `missingDefaultIllustration` object which helps to avoid the extra `try {} catch {}` blocks in the 3 `getFavicon*()` functions?

src/book.cpp
veloman-yunkan: Deal

veloman-yunkan:
The only remaining arguable point seems to be [whether `getDefaultIllustration()` should throw](https://github.com/kiwix/libkiwix/pull/630#discussion_r745665321). Also, the history of this PR has to be cleaned up before it is merged.

src/book.cpp
mgautierfr: > Do you strongly insist on throwing an exception?

No. It is not so important. `Book::getDefaultIllustration` returning a fake illustration is ok.

src/book.cpp
```diff
@@ -215,15 +212,42 @@ void Book::setPath(const std::string& path)
    : path;
 }
 
-const std::string& Book::getFavicon() const {
-  if (m_favicon.empty() && !m_faviconUrl.empty()) {
+const Book::Illustration Book::missingDefaultIllustration;
+
+const Book::Illustration& Book::getDefaultIllustration() const
```
mgautierfr, line 217: Working on the implementation of the catalog illustration endpoint (which would be base on this PR), I wonder if this method may be buggy.
The method may return a reference to the (existing) illustration but nothing guaranty the reference is keep alive.

One could call `book.getData()` (which internally call `getDefaultIllustration`) and in the same time we update the book (this is especially true when we will implement signal handling to reload the library). 
As `getDefaultIllustration` is internal and we don't keep the reference to the illustration out of Book's methods the only way to trigger this would be to have a race condition.

But in the context of kiwix-serve, we will have several thread serving the illustration and a potential reload that could appear at any time.
What do you think ? If you confirm, it is probably better to fix this now as it would be a bug pretty difficult to debug after.
Or we may define that `Library`/`Book` is not thread safe and we will have to handle this in https://github.com/kiwix/kiwix-tools/issues/243

mgautierfr (commented):
I've got one last issue. I think it would be better to handle it now but I'm ok if you think we should not.

Else we are good.

mgautierfr:
I've got a segmentation fault when I run kiwix-desktop compiled with this PR.
@veloman-yunkan, can you do the test and confirm this ?

veloman-yunkan:
> I've got a segmentation fault when I run kiwix-desktop compiled with this PR. @veloman-yunkan, can you do the test and confirm this ?

@mgautierfr No segfault with my `native_dyn` build (I don't have a `native_static` build for kiwix-desktop). Does kiwix-desktop crash on start-up?

src/book.cpp
veloman-yunkan: > One could call `book.getData()` ...

I guess you meant `book.getFavicon()`, didn't you? The issue you describe is rather a problem with the  `Book::getFavicon*()` API. So the answer depends on the what and when we are going to do with those legacy functions.

mgautierfr:
> Does kiwix-desktop crash on start-up?

Some time (I'm not sure now).
But it crashes when it tries to parse opds stream (when you click on "All files" in the library view)

veloman-yunkan:
The only remaining arguable point seems to be [whether `getDefaultIllustration()` should throw](https://github.com/kiwix/libkiwix/pull/630#discussion_r745665321). Also, the history of this PR has to be cleaned up before it is merged.

> But it crashes when it tries to parse opds stream (when you click on "All files" in the library view)

@mgautierfr My build doesn't crash

src/book.cpp
mgautierfr: Yes `book.getFavicon()`.
I'm in favor to drop `Book::getFavicon()`. We don't have to keep the cpp API, the next libkiwix release will break it anyway. We must use this slot to design the AP we want. Conceptually we don't have favicon now, we only and illustrations.

mgautierfr:
I rebuild all the projects and it doesn't crash now. Probably some kind of ABI change conflict with a `.o` not rebuild.

src/book.cpp
mgautierfr: Agree

veloman-yunkan:
The PR is now cleaned-up and rebased.

src/book.cpp
```diff
@@ -144,10 +144,11 @@ void Book::updateFromXml(const pugi::xml_node& node, const std::string& baseDir)
   m_articleCount = strtoull(ATTR("articleCount"), 0, 0);
   m_mediaCount = strtoull(ATTR("mediaCount"), 0, 0);
   m_size = strtoull(ATTR("size"), 0, 0) << 10;
-  Illustration& favicon = getMutableDefaultIllustration();
-  favicon.data = base64_decode(ATTR("favicon"));
-  favicon.mimeType = ATTR("faviconMimeType");
-  favicon.url = ATTR("faviconUrl");
+  const auto favicon = std::make_shared<Illustration>();
```
mgautierfr, line 126: We probably miss a `m_illustrations.clear()` here.
As long as we have zim files with only one illustration/favicon, it is ok.
But if we have several illustratrions we should clear the vector as we will overwrite the first illustration (without any guaranty that the first illustration (in the archive) is the same size than the "default" illustration put in the xml)

src/book.cpp
```diff
@@ -181,8 +175,11 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       m_size = strtoull(linkNode.attribute("length").value(), 0, 0);
     }
     if (rel == "http://opds-spec.org/image/thumbnail") {
-      m_faviconUrl = urlHost + linkNode.attribute("href").value();
-      m_faviconMimeType = linkNode.attribute("type").value();
+      const auto favicon = std::make_shared<Illustration>();
```
mgautierfr, line 178: Same comment than in `updateFromXml`

src/book.cpp
```diff
@@ -215,15 +212,42 @@ void Book::setPath(const std::string& path)
...
     try {
-      m_favicon = download(m_faviconUrl);
+      data = download(url);
     } catch(...) {
-      std::cerr << "Cannot download favicon from " << m_faviconUrl;
+      std::cerr << "Cannot download favicon from " << url;
     }
   }
-  return m_favicon;
+  return data;
```
mgautierfr, line 236: We probably want to protect this from race condition. Two "simultaneous" `Illustration::getData` calls may lead to two downloads.

mgautierfr (changes requested):
There is still few things I've found.
The `m_illustrations.clear()` is pretty easy to add so it would be nice to have it right now.

The race condition may be addressed now or with the other potential race condition in `Book::getDefaultIllustration`

src/book.cpp
veloman-yunkan: > We probably miss a `m_illustrations.clear()` here.

No, we don't. [`std::vector::assign()`](https://en.cppreference.com/w/cpp/container/vector/assign) replaces entire contents of the container.

src/book.cpp
veloman-yunkan: Same response as for `updateFromXml`

src/book.cpp
mgautierfr: Ho, I was thinking `assign` only change one item.
So we are good here.
Do you want to fix the race condition in `Illustration::getData` now or we merge ?

veloman-yunkan: I will fix it shortly.

src/book.cpp
veloman-yunkan: Should I use a per-illustration mutex, or a global shared mutex is fine?

veloman-yunkan: I went with a shared mutex.

veloman-yunkan: Another issue with `Book::Illustration::getData()` is that it will never stop trying to download the data if the URL is wrong or inaccessible. Shouldn't we set a limit on the count of download attempts or a minimum time interval between such attempts?

mgautierfr: A per-illustration mutex is better. If user code does multi-threading we probably don't want to introduce a single sync point, where we do only one download at the time.

veloman-yunkan: Fixed in an amended commit.

mgautierfr (approved):

Issue #631, opened 2021-11-04
Define some endpoints for metadata (and other)


mgautierfr:
Following my comment here https://github.com/kiwix/libkiwix/pull/626#issuecomment-959293854 I would like to open a discussion on what we need to serve and how.

I think there is two different use case :
- One is to get information about books available in the catalog (illustration and other metadata)
- One is to get information about what is in the zim, for some QA or other (It seems that it is what is wanted here https://github.com/kiwix/libkiwix/issues/616)

It seems that we already a way to get information about book with recent implementation of partial entries (https://github.com/kiwix/libkiwix/pull/602), we can get all the metadata of a book by simply asking it to the catalog.
What is missing is a entry point to download the illustration (from what we have in the library/catalog).
I propose the entrypoint `/catalog/v2/illustration/<bookId>?size=<size>`
The size parameter setting the width and height. Assuming a scale=1. If (when) we extend the illustration to non scare illustration and different scale we will add the corresponding parameter.
Illustration would be taken from what is in the catalog (in library.xml if kiwix-serve is started with one), never from the zim file.

So the `/meta` endpoint is ok to read the zim file as it would correspond to the second use case.
But I we recreate it, I would like to take the occasion to go a bit further.
As the "meta" endpoint will return the metadata unchanged, we can also extend it to also return raw content/item in the whole zim file.
Something like:
- `/raw/<zimName>/meta/<metaName>` to get the metadata in the zim file. If present. Unchanged. No fallback.
- `/raw/<zimName>/content/<contentPath>` to get the content of a entry in the zim files. If present. Unchanged (no chrome, no header bar added, no url rewritten). No fallback (at least not more than what libzim is making internally to handle compatibility)

The `/raw/<zimName>/content/` endpoint may help with warc2zim where the ServiceWorker need to access the raw content as it is adding its own chrome. (https://github.com/openzim/warc2zim/issues/16 https://github.com/kiwix/libkiwix/pull/403)

What do you think about this ?
We don't have to implement everything in the same time. But `/catalog/v2/illustration` and `/raw/<zimName>/meta/` (if we agree on them) are needed shortly.

rgaudin:
Interesting!

#616 was about restoring access to an information we used to get. Gentle reminder, whenever you expose something, people will start relying on it.

Behind #616 is a dead-simple access to metadata of a book/zim. I suppose we could return what's in the catalog and users would be happy. `Counter` is probably not going to be rewritten in the CMS anyway…
[...]

PR #645, opened 2021-12-08
Add illustration API

### LINKED ISSUE #644: Parsing library.xml add a dummy illustration if the library doesn't contain one.
The code is here :

https://github.com/kiwix/libkiwix/blob/master/src/book.cpp#L126-L130

`ATTR` returns a empty string if the attribute is not here.

mgautierfr:
Following the discussion at #631 this PR add a `catalog/v2/illustration` endpoint.
It also fixes #644 and make the OPDS stream handling the root prefix (it was forgotten for the entry links)

### DIFF
--- a/src/server/internalServer_catalog_v2.cpp
+++ b/src/server/internalServer_catalog_v2.cpp
@@ -66,6 +66,8 @@ std::unique_ptr<Response> InternalServer::handle_catalog_v2(const RequestContext
     return handle_catalog_v2_categories(request);
   } else if (url == "languages") {
     return handle_catalog_v2_languages(request);
+  } else if (url == "illustration") {
+    return handle_catalog_v2_illustration(request);
   } else {
     return Response::build_404(*this, request.get_full_url(), "", "");
   }
@@ -144,6 +146,20 @@ std::unique_ptr<Response> InternalServer::handle_catalog_v2_languages(const Requ
              opdsDumper.languagesOPDSFeed(),
              "application/atom+xml;profile=opds-catalog;kind=navigation"
   );
+}
+
+std::unique_ptr<Response> InternalServer::handle_catalog_v2_illustration(const RequestContext& request)
+{
+  try {
+    const auto bookName  = request.get_url_part(3);
+    const auto bookId = mp_nameMapper->getIdForName(bookName);
+    auto book = mp_library->getBookByIdThreadSafe(bookId);
+    auto size = request.get_argument<unsigned int>("size");
+    auto illustration = book.getIllustration(size);
+    return ContentResponse::build(*this, illustration->getData(), illustration->mimeType);
+  } catch(...) {
+    return Response::build_404(*this, request.get_full_url(), "", "");
+  }
 }
 
 } // namespace kiwix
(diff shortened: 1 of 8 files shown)


src/server/internalServer_catalog_v2.cpp
```diff
@@ -146,4 +148,18 @@ std::unique_ptr<Response> InternalServer::handle_catalog_v2_languages(const Requ
   );
 }
 
+std::unique_ptr<Response> InternalServer::handle_catalog_v2_illustration(const RequestContext& request)
+{
+  try {
+    const auto bookName  = request.get_url_part(3);
+    const auto bookId = mp_nameMapper->getIdForName(bookName);
+    auto book = mp_library->getBookById(bookId);
```
veloman-yunkan, line 156: Since library reloading is now supported we should use library in a thread-safe way. `Library::getBookById()` is not a thread-safe method. Use `Library::getBookByIdThreadSafe()` instead.

veloman-yunkan (changes requested):
Please make codecov happy by creating a unit test for `InternalServer::handle_catalog_v2_illustration()`

test/server.cpp
```diff
@@ -162,52 +161,53 @@ std::ostream& operator<<(std::ostream& out, const Resource& r)
...
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=description" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=language" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=name" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=tags" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=date" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=creator" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=publisher" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=favicon" },
+  { WITH_ETAG, "/ROOT/meta?content=zimfile&name=Illustration_48x48@1" },
+  { NO_ETAG, "/ROOT/catalog/v2/illustration/zimfile?size=48" },
```
mgautierfr, line 204: This should probably be `WITH_ETAG`
But https://github.com/kiwix/libkiwix/blob/master/src/server/internalServer.cpp#L320-L328 explicitly prevent catalog endpoints to be cacheable.
@veloman-yunkan want do you think, should we made (some) catalog endpoints cacheable ? Maybe in another PR ?

kelson42: @mgautierfr @veloman-yunkan This is the topic of https://github.com/kiwix/libkiwix/issues/650 which expects your both feedbacks.

veloman-yunkan: > @veloman-yunkan want do you think, should we made (some) catalog endpoints cacheable ? Maybe in another PR ?

> @mgautierfr @veloman-yunkan This is the topic of #650 which expects your both feedbacks.

Agree. Let's leave it for #650.

veloman-yunkan (approved):

PR #646, opened 2021-12-08
Raw endpoints

mgautierfr:


test/server.cpp
```diff
@@ -310,6 +317,28 @@ TEST_F(ServerTest, BookMainPageIsRedirectedToArticleIndex)
...
+  EXPECT_EQ(p->body, std::string("Ray Charles"));
+
+  p = zfs1_->GET("/raw/zimfile/meta/Creator");
+  EXPECT_EQ(200, p->status);
+  EXPECT_EQ(p->body, std::string("Wikipedia"));
+
+  // The raw content of Ray_Charles doesn't contain a taskbar...
+  p = zfs1_->GET("/raw/zimfile/content/A/Ray_Charles");
+  EXPECT_EQ(200, p->status);
+  EXPECT_TRUE(p->body.find("taskbar") == std::string::npos);
```
veloman-yunkan, line 334: Isn't this check too weak? It may be fine for whitebox testing, but for blackbox testing you should validate that the content returned by raw endpoint hasn't undergone *any* transformations.

veloman-yunkan (changes requested):

test/server.cpp
mgautierfr: I somehow agree. But we are doing whitebox testing :)
This is why I also add a test on "A/Ray_Charles" path to be sure that "taskbar" is in the returned content.
It is not easy to be sure that we have not do any transformation on a content, even more if target a blackbox testing.

veloman-yunkan: > It is not easy to be sure that we have not do any transformation on a content, even more if target a blackbox testing.

Why don't you just compare `p->body` with the same article fetched from the ZIM file directly?

mgautierfr: With stupid question, we have stupid answer :) Thanks. Will do.

mgautierfr: You were right to ask for a correct test. The code was buggy. Fixed.

src/server/internalServer.cpp
```diff
@@ -878,4 +881,61 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+    const std::string bookId = mp_nameMapper->getIdForName(bookName);
+    archive = mp_library->getArchiveById(bookId);
+  } catch (const std::out_of_range& e) {}
+
+  if (archive == nullptr) {
+    const std::string error_details = "No such book: " + bookName;
+    return Response::build_404(*this, request.get_full_url(), bookName, "", error_details);
+  }
+
+  auto itemPath = request.get_url().substr(bookName.size()+kind.size()+6);
```
veloman-yunkan, line 916: 6 is a magic number here. Please link it to `"/raw/" + "/"` in a comment .

src/server/response.h
```diff
@@ -78,9 +78,27 @@ class Response {
...
-    static std::unique_ptr<ContentResponse> build(const InternalServer& server, const std::string& template_str, kainjow::mustache::data data, const std::string& mimetype, bool isHomePage = false);
+    ContentResponse(
+      const std::string& root,
+      bool verbose,
+      bool raw,
+      bool withTaskbar,
+      bool withLibraryButton,
+      bool blockExternalLinks,
+      const std::string& content,
+      const std::string& mimetype);
```
veloman-yunkan, line 89: This starts looking too ugly but given that it is used only twice I can't suggest a simple enough improvement that can be justified.

mgautierfr: I agree.
We had the same problem with the filter feature in the library. I've fixed it by introducing the `Filter` class.
But, as you said, not sure it worth it as it is pretty internal and use only twice.

src/server/internalServer.cpp
mgautierfr: Done (and fixed the value)

src/server/internalServer.cpp
```diff
@@ -913,7 +913,11 @@ std::unique_ptr<Response> InternalServer::handle_raw(const RequestContext& reque
...
 
-  auto itemPath = request.get_url().substr(bookName.size()+kind.size()+6);
+  // Remove the beggining of the path:
+  // /raw/<bookName>/<kind>/foo
+  // ^^^^^          ^      ^
+  //   5      +     1  +   1   = 7
+  auto itemPath = request.get_url().substr(bookName.size()+kind.size()+7);
   if (itemPath[0] == '/') {
     itemPath = itemPath.substr(1);
   }
```
veloman-yunkan, line 923: Is this still needed? If yes, then one can also argue in favour of replacing the `if` with a `while`.

src/server/internalServer.cpp
```diff
@@ -894,11 +894,12 @@ std::unique_ptr<Response> InternalServer::handle_raw(const RequestContext& reque
      bookName = request.get_url_part(1);
      kind = request.get_url_part(2);
   } catch (const std::out_of_range& e) {
-     return Response::build_404(*this, request, bookName, "", "");
+     return Response::build_404(*this, request.get_full_url(), bookName, "", "");
   }
 
   if (kind != "meta" && kind!= "content") {
-    return Response::build_404(*this, request, bookName, "", "");
+    const std::string error_details = kind + " is not a valid request for raw content.";
```
veloman-yunkan, line 901: This variable should be used on the next line, but it isn't

src/server/internalServer.cpp
```diff
@@ -878,4 +881,65 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+  //   5      +     1  +   1   = 7
+  auto itemPath = request.get_url().substr(bookName.size()+kind.size()+7);
+  if (itemPath[0] == '/') {
+    itemPath = itemPath.substr(1);
+  }
+
+  try {
+    if (kind == "meta") {
+      auto item = archive->getMetadataItem(itemPath);
+      return ItemResponse::build(*this, request, item, true);
```
veloman-yunkan, line 928: Please prepend `/*raw=*/` to the last argument

src/server/internalServer.cpp
```diff
@@ -878,4 +881,65 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+    } else {
+      auto entry = archive->getEntryByPath(itemPath);
+      if (entry.isRedirect()) {
+        return build_redirect(bookName, entry.getItem(true));
+      }
+      return ItemResponse::build(*this, request, entry.getItem(), true);
```
veloman-yunkan, line 934: Please prepend `/*raw=*/` to the last argument

src/server/response.cpp
```diff
@@ -368,14 +393,14 @@ ItemResponse::ItemResponse(bool verbose, const zim::Item& item, const std::strin
...
-std::unique_ptr<Response> ItemResponse::build(const InternalServer& server, const RequestContext& request, const zim::Item& item)
+std::unique_ptr<Response> ItemResponse::build(const InternalServer& server, const RequestContext& request, const zim::Item& item, bool raw)
 {
   const std::string mimetype = get_mime_type(item);
   auto byteRange = request.get_range().resolve(item.getSize());
   const bool noRange = byteRange.kind() == ByteRange::RESOLVED_FULL_CONTENT;
   if (noRange && is_compressible_mime_type(mimetype)) {
     // Return a contentResponse
-    auto response = ContentResponse::build(server, item.getData(), mimetype);
+    auto response = ContentResponse::build(server, item.getData(), mimetype, false, raw);
```
veloman-yunkan, line 403: Please prepend `/*isHomePage=*/` before `false`.

test/server.cpp
```diff
@@ -310,6 +317,32 @@ TEST_F(ServerTest, BookMainPageIsRedirectedToArticleIndex)
...
+  auto p = zfs1_->GET("/raw/zimfile/meta/Title");
+  EXPECT_EQ(200, p->status);
+  EXPECT_EQ(p->body, std::string("Ray Charles"));
+
+  p = zfs1_->GET("/raw/zimfile/meta/Creator");
+  EXPECT_EQ(200, p->status);
+  EXPECT_EQ(p->body, std::string("Wikipedia"));
+
+  // The raw content of Ray_Charles returned by the server is
+  // the same than the one in the zim file.
```
veloman-yunkan, line 332: than → as

veloman-yunkan (changes requested):
A few last minor fixes. You are welcome to rebase&fixup for the last iteration.

src/server/internalServer.cpp
mgautierfr: No, I've removed it.
The itemPath must be exact (and nothing prevent a zim file to have a item starting with a `/`)

veloman-yunkan (approved):

src/server/internalServer.cpp
```diff
@@ -878,4 +881,62 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
...
+  if (archive == nullptr) {
+    const std::string error_details = "No such book: " + bookName;
+    return Response::build_404(*this, request.get_full_url(), bookName, "", error_details);
+  }
+
+  // Remove the beggining of the path:
+  // /raw/<bookName>/<kind>/foo
+  // ^^^^^          ^      ^
+  //   5      +     1  +   1   = 7
+  auto itemPath = request.get_url().substr(bookName.size()+kind.size()+7);
```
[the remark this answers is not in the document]
kelson42: @mgautierfr This (late) comment desserves a reponse IMO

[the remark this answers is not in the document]
mgautierfr: Yes, `request.get_url()` returns the url relative to the root location (so without it).

PR #648, opened 2021-12-11
Searcher::add_reader() rejects duplicate readers

### LINKED ISSUE #108: Prevent the same zim file from being added twice
Currently the implementation of `void Searcher::add_reader` simply store `Reader` in a vector. User can perform the following actions:

- add the same `Reader` object multiple times
- add different  `Reader` instance of the same zim file 

which may result in multiple xapian database of the same zim file being added to the search. (I am not sure if `zim::Search` handles such cases).

It might be better to figure out if the `Reader` of a specific zim file is already added to the search. If already added, simply return and do nothing. We might be able to achieve this by using a `std::unordered_map<std::string, std::shared_ptr<kiwix::Reader>> readers` internally.

veloman-yunkan:
Fixes #108

The simplest solution relying on linear search among previously added readers is implemented on the assumption that `Searcher::add_reader()` will hardly be the bottleneck if we need to search in hundreds of ZIM files.

### DIFF
--- a/src/searcher.cpp
+++ b/src/searcher.cpp
@@ -94,6 +94,12 @@ bool Searcher::add_reader(Reader* reader)
   if (!reader->hasFulltextIndex()) {
       return false;
   }
+
+  for ( const Reader* const existing_reader : readers ) {
+    if ( existing_reader->getZimFilePath() == reader->getZimFilePath() )
+      return false;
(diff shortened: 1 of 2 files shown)


kelson42:
@mgautierfr Any feedback on this?

PR #651, opened 2021-12-11
Less confusing 404 errors from /random, /meta and /suggest endpoints

veloman-yunkan:
Related to kiwix/kiwix-tools#506 (but not sure if it can be considered a fix)

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -386,23 +386,23 @@ std::unique_ptr<Response> InternalServer::handle_meta(const RequestContext& requ
 {
   std::string bookName;
   std::string bookId;
-  std::string meta_name;
   std::shared_ptr<zim::Archive> archive;
   try {
     bookName = request.get_argument("content");
     bookId = mp_nameMapper->getIdForName(bookName);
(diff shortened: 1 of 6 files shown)


kelson42:
@veloman It's nice to have a dedicated message now, but 2 buttons of 3 are still away!
image

I probably don't understand everything but the approach of `Response::build_404` seems correct ot if:
* We have a proper message (you made one)
* The mainpage/random buttons are still there is the bookname is valid.

veloman-yunkan:
> @veloman-yunkan It's nice to have a dedicated message now, but 2 buttons of 3 are still away!

@kelson42 Fixed that too

kelson42 (approved):
@veloman-yunkan Thank for the last fix From the user perspective it works like expected and to my opinion it fixes the issue. Do you haad something else in mind?

Regarding `/meta` and `/suggest` end-points I don't have tested anything because I don't know what problem is fixed and how.

@mgautierfr Please have a look to the code

mgautierfr (approved):
I approve.

I am bit disturbed by making the "term" optional. But it is not so important. The search for a empty term will return something anyway and the API is used "internally".

PR #664, opened 2021-12-25
Prepare kiwix-serve front page for missing metadata

### LINKED ISSUE #624: Front page must be prepared for zim file with missing metadata
Pretty old zim file may have no favicon at all.
In this case, the opds stream doesn't contain any node for the favicon.
(The node `<link rel="http://opds-spec.org/image/thumbnail" [...]/>`)
Opds stream is :
```
<feed xmlns="http://www.w3.org/2005/Atom" xmlns:opds="http://opds-spec.org/2010/catalog">
  <id>f2f68963-ed3a-f761-c43a-c94cdfaae6c2</id>
  <title>Filtered zims (count=17.515151515151516&amp;start=0)</title>
  <updated>2021-10-21T14:34:07Z</updated>
  <totalResults>1</totalResults>
  <startIndex>0</startIndex>
  <itemsPerPage>1</itemsPerPage>
  <link rel="self" href="" type="application/atom+xml" />
  <link rel="search" type="application/opensearchdescription+xml" href="/catalog/searchdescription.xml" />
  <entry>
    <id>urn:uuid:9bcd90f8-db78-5d68-1a81-2528ca25e806</id>
    <title>wikipedia en wp1-0.7 2009-05</title>
    <updated>T00:00:00Z</updated>
    <summary></summary>
    <language></language>
    <name></name>
    <flavour></flavour>
    <category></category>
    <tags>_ftindex:no;_pictures:yes;_videos:yes;_details:yes</tags>
    <articleCount>90284</articleCount>
    <mediaCount>0</mediaCount>
    
    <link type="text/html" href="/wikipedia_en_wp1-0.7_2009-05" />
    <author>
      <name></name>
    </author>
    <publisher>
      <name></name>
    </publisher>
    
  </entry>
</feed>
```

For now, the front page's js ends with a "undefined" url for the favicon and set the html image src to undefined. And so the browser do a request with "undefined" query.
Front page should not set a icon at all if there is none in the opds stream.

### LINKED ISSUE #657: Should invalid ZIM be loaded?
Wasted some time trying to understand why my kiwix-serve homepage was broken. It was due to a ZIM in my library not having the [mandatory `Language` metadata](https://wiki.openzim.org/wiki/Metadata): JS code assumes a language code and crashes if it's missing.

As this metadata (and a few others) are mandatory, shouldn't libkiwix/kiwix-serve refuse to load them with an explicit message instead of accepting whatever is in the library?

juuz0:
Fixes #624
Fixes #657


kelson42:
@juuz0 This PR should fix the lack of illustration as well. What happens to you if you delete the illustration entry in the OPDS feed for example?

juuz0:
@kelson42 If I remove favicon data from `library.xml`, it still loads the thumbnail. So, I'm unable to reproduce that image bug, do you have any ideas on how to reproduce that?

kelson42:
@juuz0 I believe you don't work at the right level. The right level is the OPDS level. The kiwix-serve welcome page is fed by the OPDS and populated in javascript. What happens in your case is that the library.xml is missing an illustration, the kiwix-serve read again the ZIM file to populate its internal library... and the the OPDS. You shoukd either tweak the OPPDS ouput to remove the illutration XML node... or just tweak the javascript yo do as you would not have an illustration entry in the OPDS. Then you will have a reproduction case and be able to fix the html output.

juuz0:
@kelson42 Many thanks! That helps a lot. I'll check it 👍

juuz0:
Currently looks like 
image

kelson42:
@juuz0 Looks good!

@mgautierfr @Popolechien @rgaudin Would you like to se a kind of defaukt placeholders? For language? For (lack of) illustration? If "yes", which one?

Popolechien:
@kelson42 Language should always be there rather than a (supposedly empty) placeholder, but yes if there is no image there should also be a placeholder so that the overall look is the same as other entries.
Other than that I can think of anything.

kelson42:
@Popolechien what should be displayed if no language is given?

rgaudin:
Don't think that fixes the problem reported in #657. A placeholder for icon is fine but I'm doubtful for Language. Many ZIMs are *obviously* for a certain language, based on the content so assigning a default `mul` or `eng` might create more confusion. I know there is an ISO language code for *unkown* but I can't recall what it is. It could be a good placeholder.

Popolechien:
@kelson42 I was thinking about it and thought that not having a language was not an option. But if for some reason that is possible then my first question is: how will the zim file be filtered by kiwix-serve? Present no matter what language is default selected, or only visible if language is set to `all`?

As for display hopefully the title will give the language away, so no need to put some placeholder that would only confirm that the service has no idea about what it is showing (ie info that is not actionable).

kelson42:
@juuz0 So, we are almost good IMO. Woukd you be able to display utf8 character https://emojipedia.org/prohibited/ (in grey?) in place of the missing icon?

rgaudin:
> @juuz0 So, we are almost good IMO. Woukd you be able to display utf8 character https://emojipedia.org/prohibited/ (in grey?) in place of the missing icon?

Prohibited has a very different meaning from *missing*…
Also, since this place is usually filled with a fixed size icon, I suggest we replace it with a fixed-size placeholder icon to prevent UI side effects

Popolechien:
Yeah, was going to say that `prohibited` is pretty agressive. Would suggest something along the lines of https://emojipedia.org/white-question-mark/

kelson42:
@rgaudin There is a small technical challenge with having a harcoded png because we have no end pont IMO to deliver it... but its doable. This is why I proposed a Unicode character but I understand your feedback (even if I only partly agree), maybe you have a better one in mind?

kelson42:
https://emojiguide.org/white-question-mark ?

rgaudin:
> @rgaudin There is a small technical challenge with having a harcoded png because we have no end pont IMO to deliver it... but its doable. This is why I proposed a Unicode character but I understand your feedback (even if I only partly agree), maybe you have a better one in mind?

Where don't you agree? If you're confident you can easily control and maintain the display of an emoji in place of a static PNG icon in all platform/resolution/zoom levels, then go ahead. I know I wouldn't.

If there's no endpoint and the image is chosen wisely, a base64 payload can surely do it. Probably won't be displayed a hundred times in a given library anyway.

juuz0:
Tried the above suggestions (all are PNG right now)
image
image

And here maybe this can be considered as a placeholder too :D
image

kelson42:
@juuz0 Lets go for the question mark then, but try to reduce its size. You are free to harcode a png in base64 (in the javascript) or use the unicode character (but @rgaudin is right, it will be easier with a PNG to secure the display).

juuz0:
> https://emojiguide.org/white-question-mark ?

image
^ emoji (font size 48px)

juuz0:
> juuz0 Lets go for the question mark then, but try to reduce its size. You are free to harcode a png in base64 (in the javascript) or use the unicode character (but rgaudin is right, it will be easier with a PNG to secure the display).

@kelson42 Sure, which question mark though?

rgaudin:
> > https://emojiguide.org/white-question-mark ?
> 
> image ^ emoji (font size 48px)

A flat (no bezel) version would be smaller in size and more neutral. On the page @Popolechien linked above, I think the first one was OK

kelson42:
@juuz0 the question mark introduction woukd benefit as well to be done in CSS, so no manual introduction of an IMG tag. See for example https://www.hongkiat.com/blog/css-encoded-image/.

juuz0:
This also fixes the crash that was happening in #657 now

static/skin/index.js
```diff
@@ -67,7 +67,12 @@
     }
 
     function getInnerHtml(node, query) {
-        return node.querySelector(query).innerHTML;
+        const queryNode = node.querySelector(query);
+        console.log(queryNode);
+        if(queryNode != null) {
```
kelson42, line 72: Space missing after “if”

kelson42: Better use a ternary operator

kelson42: Remove the log entry

static/skin/index.js
```diff
@@ -103,15 +108,17 @@
         if (sort) {
             divTag.setAttribute('data-idx', bookOrderMap.get(id));
         }
+        const languageAttr = `${(language == "")  ? `style="background-color: transparent"`:``}`;
```
kelson42, line 107: * global bacticks are uselss
* spaces shoule be out properly
* I don’t understand why we test language differently than iconURl
* please use backticks for interpolation only, otherwise please use single quote
* rmove parenthesis, they are not necessary

static/skin/index.js
```diff
@@ -103,15 +108,17 @@
...
+        const faviconAttr = `${ (iconUrl != undefined) ? `style="background-image: url('${root}${iconUrl}');"` : ``}`;
```
kelson42, line 108: * similar remarks like a line above
* Semi-colon is not necessary

kelson42 (changes requested):

static/skin/index.js
juuz0: iconURI is given by `getAttribute` which returns `undefined` so I am using that

kelson42 (approved):
@juuz0 Good, I have added a small beautification commit. Please check that everything is OK, Merged the commits appropriatly... and we can merge.

juuz0:
Small bug with language, fixing

kelson42:
@juuz0 OK, let me know when reqdy to merge

juuz0:
@kelson42  Rebased all fixups, ready to merge now :+1:

PR #665, opened 2021-12-29
Dependences version

kelson42:
Depends on https://github.com/openzim/libzim/pull/656

### DIFF
--- a/include/version.h
+++ b/include/version.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011 Emmanuel Engelhart <kelson@kiwix.org>
+ * Copyright 2021 Emmanuel Engelhart <kelson@kiwix.org>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU  General Public License as published by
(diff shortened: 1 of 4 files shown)


kelson42:
@veloman-yunkan This is ready to review, basically the same construt like in openzim/libzim.

src/version.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * MA 02110-1301, USA.
+ */
+
+#include <iostream>
+#include <sstream>
+
+#include <version.h>
+#include <zim/zim.h>
+#include <kiwix_config.h>
+#include <unicode/uvernum.h>
```
veloman-yunkan, line 26: Didn't this header cause problems under Windows?

veloman-yunkan (commented):
Also the commit message is inaccurate

src/version.cpp
kelson42: @veloman-yunkan No, the problem with libzim was the libicu should not be included if libzim is not compiled with libxapian (which is the case for Windows CI). That said,  `unicode/uversion.h` seems slightly better `unicode/uvernum.h` but without any impact of any sort (both work fine).

kelson42:
@veloman-yunkan Yes, have put a new message to the commit. Looks good now?

include/version.h
```diff
@@ -27,6 +27,7 @@ namespace kiwix
 {
   typedef std::vector<std::pair<std::string, std::string>> LibVersions;
   LibVersions getVersions();
+  std::string getVersionsString();
```
veloman-yunkan, line 30: What about slightly enhancing the `printVersions()` function instead?

```
void printVersions(std::ostream& out = std::cout);
```

Then `getVersionsString()` can be replaced with

```
std::ostringstream oss;
zim::printVersions(oss);
kiwix::printVersions(oss);
std::string versionsString = oss.str();
```

kelson42: Fixes in last commit

veloman-yunkan (approved):
Please don't forget to clean up the commit history before merging

PR #666, opened 2022-01-01
Make aria2 secret a random value

### LINKED ISSUE #425: SECURITY: Public secret for Aria2c RPC command module
Simply put, the secret for the aria2 command is in the code, publicly visible (Unless I am mistaken in my C++ reading, and that default is never used).  That means any program capable of sending requests to localhost can order the Aria2 server to do whatever it wants.  That (AFAIK) includes javascript executed by websites, meaning a user with kiwix installed can be made to download arbitrary files just by visiting a website.

The RPC module of Aria2 is complex, and I don't have the time at the moment to examine it in great detail.  While it doesn't permit you to directly run arbitrary commands as download hooks, it is highly complicated and could have an arbitrary code execution issue.  Even if there isn't one, an attacker still has access to a powerful tool, and could probably force all kinds of nasty behaviors.

The fix to this is fairly easy: preprocess the secret in.  If the server is shut down on every update, simply set it to a compile-time random constant,  which would make the attack far harder.  I am not a security person, so I'm no expert on keeping these keys secret, but I do think that it should be at least more than just a string literal.

juuz0:
Fixes #425

### DIFF
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -3,12 +3,14 @@
 #include "aria2.h"
 #include "xmlrpc.h"
 #include <iostream>
+#include <algorithm>
 #include <sstream>
 #include <thread>
 #include <chrono>
 #include "tools.h"
 #include "tools/pathTools.h"
 #include "tools/stringTools.h"
+#include "tools/otherTools.h"
 #include "downloader.h" // For AriaError
 
 #ifdef _WIN32
@@ -30,7 +32,7 @@ namespace kiwix {
 Aria2::Aria2():
   mp_aria(nullptr),
   m_port(42042),
-  m_secret("kiwixariarpc"),
+  m_secret(getNewRpcSecret()),
   m_curlErrorBuffer(new char[CURL_ERROR_SIZE]),
   mp_curl(nullptr)
 {
@@ -193,6 +195,13 @@ std::string Aria2::tellStatus(const std::string& gid, const std::vector<std::str
     }
   }
   return doRequest(methodCall);
+}
+
+std::string Aria2::getNewRpcSecret()
+{
+  std::string uuid = gen_uuid("");
+  uuid.erase(std::remove(uuid.begin(), uuid.end(), '-'));
+  return uuid.substr(0, 9);
 }
 
 std::vector<std::string> Aria2::tellActive()
--- a/src/aria2.h
+++ b/src/aria2.h
@@ -37,6 +37,7 @@ class Aria2
 
     std::string addUri(const std::vector<std::string>& uri, const std::vector<std::pair<std::string, std::string>>& options = {});
     std::string tellStatus(const std::string& gid, const std::vector<std::string>& statusKey);
+    static std::string getNewRpcSecret();
     std::vector<std::string> tellActive();
     std::vector<std::string> tellWaiting();
     void saveSession();


kelson42:
@juuz0 I guess this needs a PR on kiwix/kiwix-desktop side as well and that this libkiwix PR is insufficient to adapt kiwix-desktop.

juuz0:
@kelson42 kiwix-desktop works fine without any changes though.. Not sure if this affects kiwix-desktop in any way.

kelson42:
@juuz0 I'm sorry, I didn't had a look in detail but I though we talk about https://aria2.github.io/manual/en/html/aria2c.html?highlight=secret#cmdoption-rpc-secret... so it shoukd have an impact on kiwix-desktop side!?

juuz0:
@kelson42 No, I don't think so. In Kiwix desktop we use the Downloader class (which uses Aria2 class and methods) so I don't think there should be any changes needed there

src/aria2.h
```diff
@@ -37,6 +37,7 @@ class Aria2
 
     std::string addUri(const std::vector<std::string>& uri, const std::vector<std::pair<std::string, std::string>>& options = {});
     std::string tellStatus(const std::string& gid, const std::vector<std::string>& statusKey);
+    std::string makeAriaSecretToken();
```
kelson42, line 40: Better call something linke `getNewRpcSecret()`

src/aria2.cpp
```diff
@@ -195,6 +197,14 @@ std::string Aria2::tellStatus(const std::string& gid, const std::vector<std::str
   return doRequest(methodCall);
 }
 
+std::string Aria2::makeAriaSecretToken()
+{
+  unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
+  std::string letters("0123456789abcdefghijklmnopqrstuvqxyzABCDEFGHIJKLMNOPQRSTUVWXYZ");
```
kelson42, line 203: For me this does not work, somehow the aria2c operation get stuck in Kiwix-Desktop. After testing around, it seems that the string is too long! If I put `0123456789abcdefghijklmnopqrstuvqxyzA`. It works fine.

kelson42 (changes requested):
@juuz0 See my two comments, but if these details are fixed, then we should be good.

src/aria2.h
```diff
@@ -37,6 +37,7 @@ class Aria2
 
     std::string addUri(const std::vector<std::string>& uri, const std::vector<std::pair<std::string, std::string>>& options = {});
     std::string tellStatus(const std::string& gid, const std::vector<std::string>& statusKey);
+    std::string getNewRpcSecret();
```
veloman-yunkan, line 40: Declare this as a static member function

kelson42 (approved):
LGTM

src/aria2.cpp
```diff
@@ -195,6 +197,14 @@ std::string Aria2::tellStatus(const std::string& gid, const std::vector<std::str
   return doRequest(methodCall);
 }
 
+std::string Aria2::getNewRpcSecret()
+{
+  unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
+  std::string letters("0123456789abcdefghijklmnopqrstuvqxyzABCDEFGHIJKLMNOPQRSTUVWXYZ");
+  std::shuffle(letters.begin(), letters.end(), std::default_random_engine(seed));
+  return letters.substr(0, 9);
```
veloman-yunkan, line 205: This is not much better than simply converting `seed` to a string. The output is fully determined by the `seed` value (unless the underlying random engine also adds some host-dependent salt).

I think that returning the stringified `seed` value provides an adequate level of security. The extra code only creates a false impression of higher security.

kelson42: @veloman-yunkan  I don't understand your comment. `seed` is the clock, so it is not random and can not be used as random value. The shuffled string is not directly depending on `seed` but on the pseudo random generator of `std::default_random_engine` (itself seeded by `seed`). So, "yes" this is pseudo-random, but this is the usual way of doing to get a random value.

veloman-yunkan: @kelson42 As long as you know the value of `seed` you can find out the result of `std::shuffle()`, can't you?

kelson42: @veloman-yunkan Now I understand your comment. To answer your question, you might know or not. I don't know the details of implementation of the pseudo-random generator. But I still prefer this way as just putting the unix date (even if I understand your point). Maybe you have an idea to easily improve the quality of the seed?

veloman-yunkan: > But I still prefer this way as just putting the unix date

That's my concern. The result of the `shuffle()` is no better than the timestamp used as a seed value. If you are uncomfortable with the unix date returned directly you can reverse or square it (but that is intended only to illustrate my point).

> Maybe you have an idea to easily improve the quality of the seed?

I don't believe that this is the case where we should go into such lengths. Security measures should be proportional to the risks. I think that we have a lot of similar vulnerabilities in other places.

veloman-yunkan: BTW, why not taking advantage of the [`gen_uuid()`](https://github.com/kiwix/libkiwix/blob/b70c92cadefdb10295ad8bb34c992407519d193a/src/tools/otherTools.cpp#L368) function from `src/tools/otherTools.cpp`?

juuz0: Does `gen_uuid()` not suffer from the same problem of creating a false impression?
If we can get the random string, we can also get its uuid?

veloman-yunkan: If called with an empty string, `gen_uuid()` generates a random UUID based on the wall time and process CPU time values. Even though it is not an ideal source of random strings, the advantage of reusing it is that we don't create a new way of producing random strings. If we ever decide that our random string generation must be improved we will have to do it in a single place. 

Note that you may need to drop the dashes from the result (if they are not allowed in the token strings accepted by aria2).

src/aria2.cpp
```diff
@@ -3,12 +3,14 @@
 #include "aria2.h"
 #include "xmlrpc.h"
 #include <iostream>
+#include <algorithm>
```
veloman-yunkan, line 6: This header is not needed either

veloman-yunkan (changes requested):
This is almost an approval. Please rebase-fixup for the next iteration to get the formal approval.

src/aria2.cpp
juuz0: It is for std::remove

veloman-yunkan: Indeed.

veloman-yunkan (approved):

src/aria2.h
kelson42: done

kelson42:
Rebasing and merging...

PR #668, opened 2022-01-03
Introduce kiwix::fileReadable

### LINKED ISSUE #619: Function `kiwix::fileExists` does not behave properly if file exists but is not readable
During the review of https://github.com/kiwix/kiwix-tools/pull/477, I discovered a wrong behaviour:
```
$ ls -la index.html 
-rwx------ 1 root root 2649 Okt 11 15:14 index.html
$ kiwix-serve --port=8081 --customIndex=index.html  phzh_core-greek-one_el_2020-12.zim 
ERROR: No such file exist /home/kelson/code/kiwix-tools/index.html
```

The problem is that `kiwix::fileExists` returns `false`... and actually it should return `true`.

I see two solutions to that problem:
* `kiwix::fileExists` is renamed to `kiwix::fileExistsAndReadable`
* An additional function `kiwix::fileReadable` is created.

... and then all usages of this code should be checked to see if the test is appropriate.

We should also consider to have this/these method(s) generating the error message so it can be reused... but now sure how to do that properly.

juuz0:
Fixes #619
Part 2 - https://github.com/kiwix/kiwix-tools/pull/517

### DIFF
--- a/src/tools/pathTools.cpp
+++ b/src/tools/pathTools.cpp
@@ -305,16 +305,18 @@ std::string kiwix::getFileContent(const std::string& path)
 bool kiwix::fileExists(const std::string& path)
 {
 #ifdef _WIN32
-  return PathFileExistsW(Utf8ToWide(path).c_str());
+  return (_waccess_s(Utf8ToWide(path).c_str(), 0) == 0);
+#else  
+  return (access(path.c_str(), F_OK) == 0); 
+#endif
+}
+
+bool kiwix::fileReadable(const std::string& path)
+{
+#ifdef _WIN32
+  return (_waccess_s(Utf8ToWide(path).c_str(), 4) == 0);
 #else
-  bool flag = false;
-  std::fstream fin;
-  fin.open(path.c_str(), std::ios::in);
-  if (fin.is_open()) {
-    flag = true;
-  }
-  fin.close();
-  return flag;
+  return (access(path.c_str(), R_OK) == 0);
 #endif
 }
 
--- a/src/tools/pathTools.cpp
+++ b/src/tools/pathTools.cpp
@@ -306,8 +306,8 @@ bool kiwix::fileExists(const std::string& path)
 {
 #ifdef _WIN32
   return (_waccess_s(Utf8ToWide(path).c_str(), 0) == 0);
-#else  
-  return (access(path.c_str(), F_OK) == 0); 
+#else
+  return (access(path.c_str(), F_OK) == 0);
 #endif
 }
 
--- a/test/pathTools.cpp
+++ b/test/pathTools.cpp
@@ -20,6 +20,8 @@
 #include "gtest/gtest.h"
 #include <string>
 #include <vector>
+#include <unistd.h>
+#include <fcntl.h>
 #include "../include/tools.h"
 #include "../src/tools/pathTools.h"
 
@@ -190,6 +192,34 @@ TEST(pathTools, appendToDirectory)
             P5("a","b","c",".","foo.xml"));
 }
 
+TEST(pathTools, fileExists)
+{
+  ASSERT_TRUE(kiwix::fileExists(P3(".","test","example.zim")));
+  ASSERT_FALSE(kiwix::fileExists(P3(".","test","noFile.zim")));
+}
+
+TEST(pathTools, fileReadable)
+{
+  ASSERT_TRUE(kiwix::fileReadable(P3(".","test","example.zim")));
+  ASSERT_FALSE(kiwix::fileReadable(P3(".","test","noFile.zim")));
+#ifdef _POSIX_SOURCE
+  std::string path = P3(".","test","example.zim");
+  auto myprivs = geteuid();
+  if (myprivs != 0) {
+    int fd = open(path.c_str(), O_RDONLY);
+    if (fd != -1) {
+      int wResp = fchmod(fd, ~(S_IRWXU | S_IRWXG | S_IRWXO)); // remove all permissions
+      if (wResp == 0) {
+        EXPECT_FALSE(kiwix::fileReadable(P3(".","test","example.zim")));
+      }
+      int resetResp = fchmod(fd, S_IRUSR | S_IRGRP | S_IROTH | S_IWUSR | S_IWGRP); // reset back permissions to -rw-rw-r--
+      if (resetResp == 0) {
+        EXPECT_TRUE(kiwix::fileReadable(P3(".","test","example.zim")));
+      }
+    }
+  }
+#endif
+}
 
 TEST(pathTools, goUp)
 {
--- a/include/tools.h
+++ b/include/tools.h
@@ -180,12 +180,21 @@ std::string getFileContent(const std::string& path);
 
 /** checks if file exists.
  *
- * This function returns boolean stating if file exists or not.
+ * This function returns boolean stating if file exists.
  *
  * @param path The absolute path provided in string format.
  * @return Boolean representing if file exists or not.
  */
 bool fileExists(const std::string& path);
+
+/** checks if file is readable.
+ * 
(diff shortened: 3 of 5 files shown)


src/kiwixserve.cpp
```diff
@@ -39,7 +39,7 @@ void KiwixServe::run()
     std::string kiwixServeCmd = appendToDirectory(
         removeLastPathElement(getExecutablePath(true)),
         KIWIXSERVE_CMD);
-    if (fileExists(kiwixServeCmd)) {
+    if (fileExistsAndReadable(kiwixServeCmd)) {
```
mgautierfr, line 42: Here we don't need that `kiwixServeCmd` is readable.
Just that the file exists. (If we want further check, we should check it is executable)

src/tools/pathTools.cpp
```diff
@@ -302,7 +302,7 @@ std::string kiwix::getFileContent(const std::string& path)
   return content;
 }
 
-bool kiwix::fileExists(const std::string& path)
+bool kiwix::fileExistsAndReadable(const std::string& path)
 {
 #ifdef _WIN32
   return PathFileExistsW(Utf8ToWide(path).c_str());
```
mgautierfr, line 308: On windows, we only check that the file exists. Not it is readable.

mgautierfr (changes requested):
Few things to change.
Not really important by itself as this PR doesn't change any behavior. But we need something consistent. Else we will have issue in the future as we will miss use the function.

src/tools/pathTools.cpp
juuz0: So, both functions should be kept? `fileExists` and `fileExistsAndReadable` (renamed to `fileReadable` maybe)

src/kiwixserve.cpp
juuz0: same with `ARIA2_CMD`?

mgautierfr: yes

src/tools/pathTools.cpp
mgautierfr: The function should be fixed.
It doesn't do the same thing on windows and linux.
When it was called `fileExists`, it was also checking the file was readable on linux but we could consider that a implementation details.

But if we rename it to `fileExistsAndReadable` we MUST check that the file is readable (on all platforms).

src/tools/pathTools.cpp
```diff
@@ -304,18 +304,14 @@ std::string kiwix::getFileContent(const std::string& path)
...
-  std::fstream fin;
-  fin.open(path.c_str(), std::ios::in);
-  if (fin.is_open()) {
-    flag = true;
-  }
-  fin.close();
-  return flag;
-#endif
+  std::fstream fin(path.c_str());
+  return !fin.fail();
```
mgautierfr, line 308: Is this working if the file exists but don't have read right ?

mgautierfr (changes requested):
It is important to keep the windows code apart.
On windows, we need to convert our utf8 encoded path into a wstring and use the `fooW` functions. Else we may fail to open file with "exotic" char in the path.

src/tools/pathTools.cpp
juuz0: Yes
image

mgautierfr: I don't know the code of `app` but it seems it return different result depending if the file is readable or not.
But as we use this check in `fileExists` it should return the same result.

juuz0: Oops, that was a bad screenshot, my bad.
image

app's code:

```
#include <fstream>
#include <iostream>
int main()
{
  std::string path = "lol.txt";
  std::fstream fin(path.c_str());
  std::cout << !fin.fail() << std::endl;
  return 0;
}
```

juuz0: Oh, correct. I got confused.

mgautierfr: > About Windows code, what is the way to check if a file is readable? 

It is not so easy. I've found :
 - [GetNamedSecurityInfoW](https://docs.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getnamedsecurityinfowhttps://docs.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getnamedsecurityinfow) but it seems pretty complex to use and I'm not even sure it provides what we need
 - [CreateFile](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew) Use with the right option, it doesn't create a file but create a handle to a existing file if it exist and has the right needed. It is somehow the same that creating a c++ stream, but as c++ stream creators have no version for widestring, we need to use specific windows functions (https://stackoverflow.com/a/2112336/292425)

To check that a file exists (and just exists, whatever the rights we have). The best is probably to use [`stat`](https://www.man7.org/linux/man-pages/man2/stat.2.html) ([`_wstat`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170) on windows)

juuz0: I'll check those, https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-170 this looks interesting also

juuz0:
@mgautierfr 
Have used acess/waccess, seems to work good for both platforms :}

mgautierfr:
Seems good to me.
Please rebase/fixup on last master.

Please also rewrite the commit message of the last commit. Use the first line as a title and the rest of the message as the details. But you must have a empty line between the first line and the rest.

The security issue raised by code factor doesn't seem to apply here. We use access only to check the readability of the file. We take no real exploitable decision based on this.

I wonder if we could also have a `isExecutable` function, but it would not easy to implement on Windows (at leas we would have to spend time searching on how to to do it). We can go as it is and add it later if it is really needed.

kelson42:
@juuz0 Please close the discussions which are fixed/resolved (and a rebase can only help ;)

juuz0:
@kelson42 
*facepalm moment* I did rebase but forgot to pull latest changes before it lol
Thanks!

kelson42 (commented):
On the principle this looks good and I agree with @mgautierfr last comments. I would even add that IMO this two functions desserve a proper unit testing (at least on UNIX, not sure how to change the permission of a file easily with C++ on Windows).

juuz0:
No idea how unit tests are done, will read into them.

P.S. - Changed PR title to better match current state

mgautierfr (approved):

test/pathTools.cpp
```diff
@@ -190,6 +190,26 @@ TEST(pathTools, appendToDirectory)
...
+TEST(pathTools, fileExists)
+{
+  ASSERT_TRUE(kiwix::fileExists(P3(".","test","example.zim")));
+  ASSERT_FALSE(kiwix::fileExists(P3(".","test","noFile.zim")));
+#ifdef __linux__
+  system("chmod -r ./test/example.zim");
+  ASSERT_TRUE(kiwix::fileExists(P3(".","test","example.zim")));
+  system("chmod +r ./test/example.zim");
+#endif
+}
```
kelson42, line 202: * If the test crashs before setting permission back, next run will have an error
* Actually there is an elegant way to change the permission, see https://stackoverflow.com/questions/592448/how-to-set-file-permissions-cross-platform-in-c (not a must... but would be nice, I let you decide what you want to do).

kelson42 (changes requested):

test/pathTools.cpp
juuz0: Do you mean the `std::filesystem` answer of that link, I'll check that?  (the best chosen answer wouldn't work here I think, it cannot revoke permissions - there's also a similar `chmod` function in sys/stat.h but that also suffers from the same problem)

kelson42 (approved):
LGTM

PR #669, opened 2022-01-05
Remove the meta endpoint in the server.

mgautierfr:
This is the last PR to finalize #631. Now we have `/raw` and `catalog/v2/illustration` endpoints we don't need to keep the meta endpoint.

### DIFF
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -296,6 +296,9 @@ const char* urls404[] = {
   "/ROOT/catch/external",
   "/ROOT/zimfile/A/non-existent-article",
 
+  "/ROOT/raw/non-existent-book/meta/Title",
+  "/ROOT/raw/zimfile/wrong-kind/Foo",
+
   // zimfile has no Favicon nor Illustration_48x48@1 meta item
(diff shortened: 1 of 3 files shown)


test/server.cpp
```diff
@@ -296,6 +295,10 @@ const char* urls404[] = {
   "/ROOT/suggest?content=non-existent-book&term=abcd",
   "/ROOT/catch/external",
   "/ROOT/zimfile/A/non-existent-article",
+
+  // zimfile has no Favicon nor Illustration_48x48@1 meta item
+  "/ROOT/raw/zimfile/meta/Favicon",
+  "/ROOT/raw/zimfile/meta/Illustration_48x48@1",
```
veloman-yunkan, line 301: This suggestion should be made in #646, but I only reviewed the changes and didn't think about testing of the HTTP 404 error functionality (which didn't show up in the diff). It makes sense to test how the `/raw` endpoint reacts to
- non-existent ZIM file
- invalid `kind` (neither `meta` nor `content`)

veloman-yunkan (approved):

PR #673, opened 2022-01-11
UrlEncode the `content_id`.

### LINKED ISSUE #670: OPDS stream is malformed when zim name contains `&`
I've got a zim file with a `&` in my library.
The urls we've added recently on the opds stream are not url encoded and so, the xml generated is wrong.

```xml
<entry>
    <id>urn:uuid:78b0dd21-c983-d979-c969-8e5656ff022f</id>
    <title>Programming Puzzles &amp; Code Golf Stack Exchange</title>
    <updated>2017-04-01T00:00:00Z</updated>
    <summary>Q&amp;A for programming puzzle enthusiasts and code golfers</summary>
    <language>en</language>
    <name></name>
    <flavour></flavour>
    <category></category>
    <tags>_ftindex:no;_pictures:yes;_videos:yes;_details:yes</tags>
    <articleCount>191336</articleCount>
    <mediaCount>59594</mediaCount>
    <link rel="http://opds-spec.org/image/thumbnail"
          href="/catalog/v2/illustration/programming_puzzles_&_code_golf_stack_exchange_en_all_2017-04/?size=48"
          type="image/png;width=48;height=48;scale=1"/>
    <link type="text/html" href="/programming_puzzles_&_code_golf_stack_exchange_en_all_2017-04" />
    <author>
      <name>Programming Puzzles &amp; Code Golf Stack Exchange</name>
    </author>
    <publisher>
      <name>dattaz</name>
    </publisher>
    
  </entry>
```

mgautierfr:
The HumanReadableId can contains special char (`&`/`=`/...)
As it is used as to create a url in the opds template,
we must url encode it.

- We don't need to encode the book id as it is a uuid, it never contains
special char.
- We don't need to encode the book url as it is read from the library and
the url must already be correctly encoded in the library.xml.
(tests modified accordingly)

Fix #670

### DIFF
--- a/test/data/zimfile&other.zim
+++ b/test/data/zimfile&other.zim
@@ -1,0 +1,1 @@
+zimfile.zim
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -81,7 +81,7 @@ kainjow::mustache::object getSingleBookData(const Book& book)
       {"title", book.getTitle()},
       {"description", book.getDescription()},
       {"language", book.getLanguage()},
-      {"content_id",  book.getHumanReadableIdFromPath()},
+      {"content_id",  urlEncode(book.getHumanReadableIdFromPath(), true)},
       {"updated", book.getDate() + "T00:00:00Z"},
       {"category", book.getCategory()},
       {"flavour", book.getFlavour()},
--- a/test/data/library.xml
+++ b/test/data/library.xml
@@ -35,8 +35,8 @@
       ></book>
   <book
         id="charlesray"
-        path="./zimfile.zim"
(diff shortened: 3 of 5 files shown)


test/data/zimfile&other.zim
```diff
@@ -0,0 +1 @@
+zimfile.zim
```
veloman-yunkan, line 1: Is this (an invalid new test ZIM file) intentional or a result of some mistake/error?

mgautierfr: This is a git link to zimfile.zim. It will be a symlink on linux and whatever kind of link on windows :)

veloman-yunkan (approved):

PR #674, opened 2022-01-12
Deprecate old methods

### LINKED ISSUE #654: Make many useless methods deprecated
With libzim7 and libkiwix10 we have moved a bit of logic to libzim. Libkiwix to a certain extend is simpler and now many metods are simple wrappers over one libzim method. Beside keeping API compatibility, this as no value and ultimatively such methods will disappear. The sooner the better. We should mark these methods as deprecated and print warnings if they are used to inform users and push developers to update their code.

mgautierfr:
Fix #654


include/entry.h
```diff
@@ -176,6 +176,10 @@ class Entry
 
   private:
     zim::Entry entry;
+
+  private:
+    Entry(zim::Entry entry, bool marker);
```
veloman-yunkan, line 181: Explain in a comment the purpose of this constructor

veloman-yunkan (changes requested):
Just one comment about a missing comment :)

include/entry.h
mgautierfr: Done

veloman-yunkan (approved):

veloman-yunkan:
Must be rebased/fixed-up before merge

PR #675, opened 2022-01-14
Revert removing of deprecated methods used by android wrapper.

mgautierfr:


veloman-yunkan (approved):
I didn't check that this PR accurately reverts the 4 methods of `kiwix::Reader` that were dropped in #674 (e5eeb08206f231531e7686f4f2e5fcb32e26ecdf). My approval only means that I have no objections.

PR #680, opened 2022-01-18
Add a new private constructor not deprecated for Reader.

mgautierfr:
As we still create a `Reader` in the deprecated code of `Library`,
we need a way to create a reader without raising a deprecated warning.
So we create a another constructor with a dummy argument and we use it.

The constructor is used in `Library::getReaderById`. It was used internally by `make_shared` which silently ignore the warning.
When compiling on Windows with cl, the warning is not ignored and so we have an error.

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -219,7 +219,7 @@ std::shared_ptr<Reader> Library::getReaderById(const std::string& id)
   if ( !archive )
     return nullptr;
 
-  const auto reader = make_shared<Reader>(archive);
+  const shared_ptr<Reader> reader(new Reader(archive, true));
   std::lock_guard<std::mutex> lock(m_mutex);
   m_readers[id] = reader;
   return reader;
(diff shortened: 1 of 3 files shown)


src/library.cpp
```diff
@@ -219,7 +219,7 @@ std::shared_ptr<Reader> Library::getReaderById(const std::string& id)
   if ( !archive )
     return nullptr;
 
-  const auto reader = make_shared<Reader>(archive);
+  const shared_ptr<Reader> reader(new Reader(archive, true));
```
veloman-yunkan, line 222: Why didn't you just add the extra argument?

```
const auto reader = make_shared<Reader>(archive, true);
```

mgautierfr: Because the Reader would be constructed by the `make_shared` function and so we would need to make the constructor public.

veloman-yunkan: Indeed, I missed that.

veloman-yunkan (approved):

PR #686, opened 2022-01-21
Catalog search with zero count

### LINKED ISSUE #684: URGENT: Kiwix-desktop 2.1.0 can't download anything online (with latest kiwix-serve at library.kiwix.org)
I have deployed kiwix-serve `3.1.2` on library.kiwix.org.

As a consequence kiwix-desktop can not list anything on the online library. It makes the OPDS requests, but there is always 0 books listed.

veloman-yunkan:
Fixes #684

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -720,7 +720,8 @@ InternalServer::search_catalog(const RequestContext& request,
     const auto totalResults = bookIdsToDump.size();
     const size_t count = request.get_optional_param("count", 10UL);
     const size_t startIndex = request.get_optional_param("start", 0UL);
-    bookIdsToDump = subrange(bookIdsToDump, startIndex, count);
+    const size_t intendedCount = count > 0 ? count : bookIdsToDump.size();
+    bookIdsToDump = subrange(bookIdsToDump, startIndex, intendedCount);
     opdsDumper.setOpenSearchInfo(totalResults, startIndex, bookIdsToDump.size());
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan No sure about your opinion, but I would add a code comment explaining that this is temporary, because of #684 and that this should be removed in a near future.

veloman-yunkan:
> @veloman-yunkan No sure about your opinion, but I would add a code comment explaining that this is temporary, because of #684 and that this should be removed in a near future.

I don't see why this fix should be temporary. The change in the handling of `count=0` by the `/catalog/search` endpoint was unintentional (just a by-product of some code refactoring in 07252a127af39d09edb0f57086c957b27e99cb73 #492). I don't see any problem in reversing that change in behaviour.

kelson42:
@veloman-yunkan OK, not sure this is the best approach, but considering that requesting 0 books ins anyway quite a non-sense, I won't argue about that. So, thx for that fix.

What about the differrence of behaviour between empty string and undefined? Is that on purpose? What are the rationals?

veloman-yunkan:
> What about the differrence of behaviour between empty string and undefined? Is that on purpose? What are the rationals?

@kelson42 The default value for count was set to 10 and that was on purpose. Handling of `count=` (empty) is unspecified and I don't think that it *should* be specified (just like we don't specify the handling of, for example, `count=ten` or `count=half_of_all_books`).

PR #687, opened 2022-01-21
Add undefined check for humanFriendlyTitle

### LINKED ISSUE #685: http://library.kiwix.org does not achieve to populate its cateogry select box
Look at http://library.kiwix.org and the "Category" select box is empty.

If I test it localy on my computer it seems to work fine.

Category API seems to work fine as well:
https://library.kiwix.org/catalog/v2/categories

I believe this is because this javascript error I see which probably makes the whole javascript engine to stop:
```
Uncaught (in promise) TypeError: title[0] is undefined
    humanFriendlyTitle https://library.kiwix.org/skin/index.js:57
    loadAndDisplayOptions https://library.kiwix.org/skin/index.js:241
    loadAndDisplayOptions https://library.kiwix.org/skin/index.js:238
```

juuz0:
Fixes #685

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -53,8 +53,12 @@
     };
 
     const humanFriendlyTitle = (title) => {
-        title = title.replace(/_/g, ' ');
-        return htmlEncode(title[0].toUpperCase() + title.slice(1));
+        if (typeof title === 'string' && title.length > 0) {
+            title = title.replace(/_/g, ' ');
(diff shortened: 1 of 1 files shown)


kelson42 (changes requested):
If title is undefined, the the line 56 is useless or might even fail IMO. I believe we should test first then do the stuff… return empty string otherwise.

PR #693, opened 2022-01-27
Add non-minified version of isotope.pkgd.js

legoktm:
Debian wants to have the source files for minified scripts. Otherwise
same rationale as #368 which was for jquery-ui.

I downloaded this from <https://unpkg.com/isotope-layout@3.0.6/dist/isotope.pkgd.js>.


veloman-yunkan:
I am not sure what I can review about this PR

PR #697, opened 2022-01-28
Add title with full language name

juuz0:
Fixes #692

PR #701, opened 2022-02-05
Add mappings for languages not given by libicu

### LINKED ISSUE #611: Ladakhi does not appears properly in the list of (kiwix-serve) language
If I load https://mirror.download.kiwix.org/zim/videos/dse-ladak_hi_class-3rd_2021-08.zim with Kiwix-serve, I get this on the Welcome Page:
image

I should see the language listed as "Ladakhi" in place of "lbj".

See more info about Ladakhi language at https://en.wikipedia.org/wiki/Ladakhi_language

juuz0:
Fixes #611 

There are 7,893 languages under ISO 639-3. I could not find a mapping between lang code and native names (there's this [langCode vs english names](https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3_Name_Index.tab) but we display native names).
Instead, this just adds the languages missed by libicu in library.kiwix.org. If this way is *approved*, I would like to ask:
* Should the map be rather moved to C++?

### DIFF
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -127,43 +127,43 @@ BooksData getBooksData(const Library* library, const std::vector<std::string>& b
 }
 
 std::map<std::string, std::string> iso639_3 = {
-  {"atj","atikamekw"},
-  {"azb","آذربایجان دیلی"},
-  {"bcl","central bikol"},
-  {"bgs","tagabawa"},
-  {"bxr","буряад хэлэн"},
-  {"cbk","chavacano"},
-  {"cdo","閩東語"},
-  {"diq","dimli"},
-  {"dty","डोटेली"},
-  {"eml","emiliân-rumagnōl"},
-  {"fbs","српскохрватски"},
-  {"kbp","kabɩyɛ"},
-  {"kld","Gamilaraay"},
-  {"lbe","лакку маз"},
-  {"lbj","ལ་དྭགས་སྐད་"},
-  {"map","Austronesian"},
-  {"mhr","марий йылме"},
-  {"mnw","ဘာသာမန်"},
-  {"myn","mayan"},
-  {"nah","nahuatl"},
-  {"nai","north American Indian"},
-  {"nds","plattdütsch"},
-  {"nrm","bhasa narom"},
-  {"olo","livvi"},
-  {"pih","Pitcairn-Norfolk"},
-  {"pnb","Western Panjabi"},
-  {"rmr","Caló"},
-  {"rmy","romani shib"},
-  {"roa","romance languages"},
-  {"twi","twi"},
-  {"ido","ido"},
-  {"dag", "Dagbani"}
+  {"atj", "atikamekw"},
+  {"azb", "آذربایجان دیلی"},
+  {"bcl", "central bikol"},
+  {"bgs", "tagabawa"},
+  {"bxr", "буряад хэлэн"},
+  {"cbk", "chavacano"},
+  {"cdo", "閩東語"},
+  {"dag", "Dagbani"},
+  {"diq", "dimli"},
+  {"dty", "डोटेली"},
+  {"eml", "emiliân-rumagnōl"},
+  {"fbs", "српскохрватски"},
+  {"ido", "ido"},
+  {"kbp", "kabɩyɛ"},
+  {"kld", "Gamilaraay"},
+  {"lbe", "лакку маз"},
+  {"lbj", "ལ་དྭགས་སྐད་"},
+  {"map", "Austronesian"},
+  {"mhr", "марий йылме"},
+  {"mnw", "ဘာသာမန်"},
+  {"myn", "mayan"},
+  {"nah", "nahuatl"},
+  {"nai", "north American Indian"},
+  {"nds", "plattdütsch"},
+  {"nrm", "bhasa narom"},
+  {"olo", "livvi"},
+  {"pih", "Pitcairn-Norfolk"},
+  {"pnb", "Western Panjabi"},
+  {"rmr", "Caló"},
+  {"rmy", "romani shib"},
+  {"roa", "romance languages"},
+  {"twi", "twi"}
 };
 
 std::once_flag fillLanguagesFlag;
 
-void fillLanguagesMap() 
+void fillLanguagesMap()
 {
   for (auto icuLangPtr = icu::Locale::getISOLanguages(); *icuLangPtr != NULL; ++icuLangPtr) {
     auto lang = *icuLangPtr;
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -157,7 +157,8 @@ std::map<std::string, std::string> iso639_3 = {
   {"rmy","romani shib"},
   {"roa","romance languages"},
   {"twi","twi"},
-  {"ido","ido"}
+  {"ido","ido"},
+  {"dag", "Dagbani"}
 };
 
 std::once_flag fillLanguagesFlag;
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -126,13 +126,62 @@ BooksData getBooksData(const Library* library, const std::vector<std::string>& b
   return booksData;
 }
 
+std::map<std::string, std::string> iso639_3 = {
+  {"atj","atikamekw"},
+  {"azb","آذربایجان دیلی"},
+  {"bcl","central bikol"},
+  {"bgs","tagabawa"},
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
@juuz0 From a a user perspective it works as expected. I also appreciate you made an effort to identify the other languages impacted by a similar problem. I'll let @veloman-yunkan review from a code perspective, but I believe this would be more efficient to merge `iso6391To3.js` and `iso639_3.js`, this is a bit a waste of resources to have to make two HTTP requests IMO.

veloman-yunkan (changes requested):
I think you have missed the fact that `loadAndDisplayOptions()` is used for populating not only the list of languages but also the list of categories. Your (already merged) PR #697 suffers from the same problem (and introduces a potential bug). Please try to take the mentioned fact into account adjust this PR accordingly (as well as identify and fix the potential bug introduced by #697).

juuz0:
> I believe this would be more efficient to merge iso6391To3.js and iso639_3.js, this is a bit a waste of resources to have to make two HTTP requests IMO.

Done.

> I think you have missed the fact that loadAndDisplayOptions() is used for populating not only the list of languages but also the list of categories...

Correct, done.

kelson42:
> Fixing `getLanguageSelfName()` in `src/opds_dumper.cpp` will be a more coherent implementation.

I agree with this comment.

veloman-yunkan (changes requested):

src/opds_dumper.cpp
```diff
@@ -124,13 +124,51 @@ BooksData getBooksData(const Library* library, const std::vector<std::string>& b
...
   const icu::Locale locale(lang.c_str());
   icu::UnicodeString ustring;
   locale.getDisplayLanguage(locale, ustring);
   std::string result;
   ustring.toUTF8String(result);
-  return result;
+  try {
+    return iso639_3.at(lang);
+  } catch (...) {
+    return result;
```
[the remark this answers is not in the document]
kelson42: A small comment (I already made on Slack) on my side is that if we could call `iso639_3.at(lang)` only if necessary, this would be better as well. This is a bit more code, but would make it slighly more performant, and this routine is really called often.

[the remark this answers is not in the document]
veloman-yunkan: Such decisions should be substantiated by benchmarking data. But my bet is that it shouldn't matter much since `getLanguageSelfName()` is hardly going to be a bottleneck at all. Benchmarking can help to eliminate any concern with respect to the performance of  `getLanguageSelfName()` and just use the simplest solution (which you have already implemented).

veloman-yunkan (changes requested):

src/opds_dumper.cpp
```diff
@@ -242,10 +237,27 @@ std::string OPDSDumper::categoriesOPDSFeed() const
...
+    iso639_3.insert({iso3LangCode, result});
+    ++isoLanguages;
+  }
+}
+
 std::string OPDSDumper::languagesOPDSFeed() const
 {
   const auto now = gen_date_str();
   kainjow::mustache::list languageData;
+  fillLanguagesMap();
```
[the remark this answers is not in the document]
juuz0: Does it make sense to have this iso639_3 map be available by a method in kiwix/tools.h?
That way it can be used by Kiwix-Desktop also (helping in this: https://github.com/kiwix/kiwix-desktop/issues/681)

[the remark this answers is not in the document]
kelson42: This is a tricky question... If I understand properly what implies your question is: Should Kiwix Desktop use the REST API or directly use the libkiwix primitive to populate its own language list? I'm in favour of the second for performance/pragmatical reasons. The advantages of having the language list pulled from the online REST API at https://library.kiwix.org/ seems not big enough to me. But interested in @veloman-yunkan's opinion on this.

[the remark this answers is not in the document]
veloman-yunkan: As long as the language mapping API is well thought out I don't have any argument against exposing it via `kiwix/tools.h`. However I am not sure why you may need it in kiwix-desktop. If you are trying to close with it a gap present in the libkiwix API, maybe it can be fixed in a better way.

veloman-yunkan (commented):
The latest changes are fine, but there is still one unresolved comment.

src/server/internalServer.cpp
```diff
@@ -182,6 +181,7 @@ bool InternalServer::start() {
   auto server_start_time = std::chrono::system_clock::now().time_since_epoch();
   m_server_id = kiwix::to_string(server_start_time.count());
   m_library_id = m_server_id;
+  kiwix::fillLanguagesMap();
```
veloman-yunkan, line 184: This is still not the best way of filling the languages map only once. In particular it assumes that `libkiwix` will be used in a way where a `kiwix::Server` object is created only once. Do you know about the [Singleton](https://en.wikipedia.org/wiki/Singleton_pattern) design pattern and/or about [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once)?

veloman-yunkan (changes requested):
There is no need to use fixup commit's for this small PR. Please submit the next request for review as a single-commit history.

src/server/internalServer.cpp
juuz0: I don't but learning what it is :+1:

src/opds_dumper.cpp
```diff
@@ -206,10 +239,29 @@ std::string OPDSDumper::categoriesOPDSFeed() const
   );
 }
 
+std::once_flag fillLanguagesFlag;
+
+void fillLanguagesMap() 
+{
+  std::call_once(fillLanguagesFlag, []() {
```
veloman-yunkan, line 247: I think it is better to call `fillLanguagesMap()` via `std::call_once()` instead of wrapping its entire body in `std::call_once()`. Such a point of view is supported by the function's name and the fact that the languages map object (`iso639_3`) is publicly accessible (i.e. can be cleared or otherwise modified by anyone). Therefore, the request to "fill the languages map" must be performed by the respective function unconditionally. However, for the current code logic, that that request must be made only once.

BTW, why did you choose the solution based on `std::call_once`? An alternative was to make `iso639_3` a static variable inside `getLanguageSelfName()` that is initialized from the return value of a function `getLanguagesMap()`:

```c++
std::map<std::string, std::string> getLanguagesMap() {
  std::map<std::string, std::string> langMap = {"atj","atikamekw"},
    ...
  }
  // add libicu data to langMap
  ...
  return langMap;
}

std::string getLanguageSelfName(const std::string& lang) {
  static const auto iso639_3 = getLanguagesMap();
  ...
}
```

veloman-yunkan (changes requested):

src/opds_dumper.cpp
juuz0: > BTW, why did you choose the solution based on std::call_once?

Because I didn't fully get `Singleton` and how it could be used :)

>  An alternative was to make iso639_3 a static variable inside getLanguageSelfName() that is initialized from the return value of a function getLanguagesMap():

This is neat, which one should I prefer?

veloman-yunkan: Whichever you like more for this PR

veloman-yunkan: I mean it, really. Still advising to explain your choice in a written form. It will help to shape your own taste in software engineering.

juuz0: Okay then, thanks for the freedom!
I searched up differences about the ways and they look fully identical, so purely `call_once` is written *call once* (more intuitve) I will go by it

src/opds_dumper.cpp
```diff
@@ -206,10 +240,26 @@ std::string OPDSDumper::categoriesOPDSFeed() const
...
+    const icu::Locale locale(lang);
+    icu::UnicodeString ustring;
+    locale.getDisplayLanguage(locale, ustring);
+    std::string displayLanguage;
+    ustring.toUTF8String(displayLanguage);
+    std::string iso3LangCode = locale.getISO3Language();
+    iso639_3.insert({iso3LangCode, displayLanguage});
+  }
+}
+
```
veloman-yunkan, line 257: Please move this into the unnamed namespace just under the definition of `iso639_3` so that

1. all code responsible for the languages map is in one place
2. global namespace is not polluted (which is important in a library)

veloman-yunkan (changes requested):

src/opds_dumper.cpp
juuz0: Done, thanks

veloman-yunkan (approved):

kelson42 (approved):

PR #705, opened 2022-02-07
Add book titles to search results

### LINKED ISSUE #704: Add book titles to search results
When seaching across multiple zim archives it would be useful to be able to see which archive each search result came from to avoid confusion. 

For instance, in my case I have Wikipedia, archwiki, and python documentation amongst others set up with kiwix-serve. If I seach for something technical like `Python` across all these books, it can be really confusing which book each result came from.

Most public search engines have a bit of text under each search result that shows what site results come from. By adding the book title at the bottom of each result, we can emulate a behaviour lots of users have come to expect.

One possible downside is that the book titles will clutter the search results, but a few words per search result don't seem to distract too much.

thavelick:
Implements/fixes #704 and fixes https://github.com/kiwix/kiwix-tools/issues/478

This is my first pull request here, so appologies if I've missed any process or coding style guidlines or best practices. Happy to revise anything or provide additional context.

It would be good as well to have a corresponding PR in kiwix/kiwix-tools and kiwix/kiwix-desktop to see how it works from a Kiwix port / end-user perspective.

### DIFF
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -26,6 +26,8 @@
 #include "library.h"
 #include "name_mapper.h"
 
+#include "tools/archiveTools.h"
+
 #include <zim/search.h>
 
 #include <mustache.hpp>
@@ -45,10 +47,11 @@ SearchRenderer::SearchRenderer(Searcher* searcher, NameMapper* mapper)
       resultStart(searcher->getResultStart())
 {}
 
-SearchRenderer::SearchRenderer(zim::SearchResultSet srs, NameMapper* mapper,
+SearchRenderer::SearchRenderer(zim::SearchResultSet srs, NameMapper* mapper, Library* library,
                       unsigned int start, unsigned int estimatedResultCount)
     : m_srs(srs),
       mp_nameMapper(mapper),
+      mp_library(library),
       protocolPrefix("zim://"),
       searchProtocolPrefix("search://?"),
       estimatedResultCount(estimatedResultCount),
@@ -90,6 +93,10 @@ std::string SearchRenderer::getHtml()
     std::ostringstream s;
     s << it.getZimId();
     result.set("resultContentId", mp_nameMapper->getNameForId(s.str()));
+    std::shared_ptr<zim::Archive> archive;
+    try {
+       result.set("bookTitle", mp_library->getBookById(s.str()).getTitle());
+    } catch (const std::out_of_range& e) {}
 
     if (it.getWordCount() >= 0) {
       result.set("wordCount", kiwix::beautifyInteger(it.getWordCount()));
--- a/include/search_renderer.h
+++ b/include/search_renderer.h
@@ -22,6 +22,7 @@
 
 #include <string>
 #include <zim/search.h>
+#include <library.h>
 
 namespace kiwix
 {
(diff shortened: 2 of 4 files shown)


thavelick:
I realized after the fact that by changing the constructor of `SearchRenderer`, this may break 3rd party software that uses libkiwix. I could see either:

1. Adding a new constructor that adds the libary but keeping the original one OR
2. Refactoring  this to set the library with a setter instead of with a constructor 

I happy to make one of the above changes but I want to make sure there's enough interest in this enhancement being merged in to the main rpeo before I spend a lot of time making it perfect. It works for my own purposes as is.

kelson42:
@thavelick Thank you for your first PR. We will review it and give you a feedback soon.

src/search_renderer.cpp
```diff
@@ -90,6 +93,11 @@ std::string SearchRenderer::getHtml()
     std::ostringstream s;
     s << it.getZimId();
     result.set("resultContentId", mp_nameMapper->getNameForId(s.str()));
+    std::shared_ptr<zim::Archive> archive;
+    try {
+       archive = mp_library->getArchiveById(s.str());
+       result.set("bookTitle", archive ? getArchiveTitle(*archive) : "");
```
veloman-yunkan, line 99: An empty string for `bookTitle` won't disable the mustache block marked with `{{#bookTitle}}...{{/bookTitle}}`. You will need to pass a `kainjow::mustache::data(false)` instead.

veloman-yunkan (changes requested):
The two commits should be combined into one.

Regarding the non-backward-compatible change in the public API of `SearchRenderer`, I personally wouldn't mind about it, but the final word should belong to @mgautierfr.

kelson42:
@thavelick Any fix in view regarding @veloman-yunkan feedbacks?

thavelick:
@kelson42 I've been busy with work stuff and haven't had a chance to look the feedback in any depth yet. I think I should be able to address everything here this weekend.

src/search_renderer.cpp
thavelick: Updated that so it uses mustache's false.  Really this should be pertty much impossible. Both the `{#bookTitle}` wrapper and this null check are meant as guards against future bugs.

thavelick:
@veloman-yunkan @kelson42 I've squashed my commits to one and addressed your comment

@mgautierfr looking forward to hearing your thoughts on the constructor.

thavelick:
Note, the ability to have mulitlpe seach results is currently a little hidden I've been testing this by starting up `kiwix-serve` and then hitting `/search?pattern=emulator` which will trigger resutls across archives if you have both wikipedia and wikihow zimfiles.

kelson42:
@thavelick Thx for the update. It is true that the multizim search feature is not proeminent. This is because it does not scale in appropriate manner. This is a problem @maneeshpm is working on and I hope the problem will be fixed in one of the next minor releases.

src/search_renderer.cpp
```diff
@@ -90,6 +93,11 @@ std::string SearchRenderer::getHtml()
     std::ostringstream s;
     s << it.getZimId();
     result.set("resultContentId", mp_nameMapper->getNameForId(s.str()));
+    std::shared_ptr<zim::Archive> archive;
+    try {
+       archive = mp_library->getArchiveById(s.str());
+       result.set("bookTitle", archive ? getArchiveTitle(*archive) : kainjow::mustache::data(false));
```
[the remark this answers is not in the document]
thavelick: @veloman-yunkan Your right, I don't think there is a situation in which archive can be null here. I've switched to just using `mp_library->getBookById(s.str()).getTitle()` as you've suggested

thavelick:
@veloman-yunkan @kelson42 Let me know if you need anything else form me to get this one over the line

veloman-yunkan (approved):

PR #711, opened 2022-02-11
Add tag filtering in kiwix serve

### LINKED ISSUE #582: Add ability to filter by tags in Kiwix-serve welcome page
It is possible via /catalog/v2/entries?q=tag:tagname but the exact workflow is still to be decided.

What is sure is that each displayed tag should be a link to allow a filter by tag.

But we probably don't want an additonal tag filter select box (and I would not know how to populate it anyway).

juuz0:
Fixes #582

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -77,6 +77,13 @@
         return queryNode != null ? queryNode.innerHTML : "";
     }
 
+    function generateTagLink(tagValue) {
+        tagValue = tagValue.toLowerCase();
+        const humanFriendlyTagValue = humanFriendlyTitle(tagValue);
+        const tagMessage = `Filter by tag "${humanFriendlyTagValue}"`;
+        return `<span class='tag__link' aria-label='${tagMessage}' title='${tagMessage}' data-tag=${tagValue}>${humanFriendlyTagValue}</span>`
+    }
+
     function generateBookHtml(book, sort = false) {
         const link = book.querySelector('link[type="text/html"]').getAttribute('href');
         let iconUrl;
@@ -91,9 +98,9 @@
         const langCode = getInnerHtml(book, 'language');
         const language = languages[langCode];
         const tags = getInnerHtml(book, 'tags');
-        let tagHtml = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
-            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)})
-            .join(' | ').replace(/_/g, ' ');
+        const tagList = tags.split(';').filter(tag => {return !(tag.startsWith('_'))});
+        const tagFilterLinks = tagList.map((tagValue) => generateTagLink(tagValue));
+        const tagHtml = tagFilterLinks.join(' | ');
         let downloadLink;
         let zimSize = 0;
         try {
@@ -113,17 +120,21 @@
         }
         const faviconAttr = iconUrl != undefined ? `style="background-image: url('${iconUrl}')"` : '';
         const languageAttr = langCode != '' ? `title="${language}" aria-label="${language}"` : 'style="background-color: transparent"';
-        divTag.innerHTML = `<a class="book__link" href="${link}" data-hover="Preview">
+        divTag.innerHTML = `
             <div class="book__wrapper">
+            <a class="book__link" href="${link}" data-hover="Preview">
+            <div class="book__link__wrapper">
             <div class="book__icon" ${faviconAttr}></div>
             <div class="book__header">
                 <div id="book__title">${title}</div>
                 ${downloadLink ? `<div class="book__download"><span data-link="${downloadLink}">Download ${humanFriendlyZimSize ? ` - ${humanFriendlyZimSize}</span></div>`: ''}` : ''}
             </div>
             <div class="book__description" title="${description}">${description}</div>
+            </div>
+            </a>
             <div class="book__languageTag" ${languageAttr}>${getLanguageCodeToDisplay(langCode)}</div>
             <div class="book__tags"><div class="book__tags--wrapper">${tagHtml}</div></div>
-            </div></div></a>`;
+            </div></div>`;
         return divTag;
     }
 
@@ -333,6 +344,7 @@
                 insertModal(downloadButton);
             }
         });
+        refreshTagLinks();
     }
 
     async function resetAndFilter(filterType = '', filterValue = '') {
@@ -376,10 +388,45 @@
             }
         });
     }
+    
+    function addTagElement(tagValue, resetFilter) {
+        const tagElement = document.getElementsByClassName('tagFilterLabel')[0];
+        tagElement.style.display = 'inline-block';
+        const humanFriendlyTagValue = humanFriendlyTitle(tagValue);
+        tagElement.innerHTML = `${humanFriendlyTagValue}`;
+        const tagMessage = `Stop filtering by tag "${humanFriendlyTagValue}"`;
+        tagElement.setAttribute('aria-label', tagMessage);
+        tagElement.setAttribute('title', tagMessage);
+        if (resetFilter)
+            resetAndFilter('tag', tagValue);
+    }
+
+    function refreshTagLinks() {
+        const tagLinks = document.getElementsByClassName('tag__link');
+        [...tagLinks].forEach(elem => {
+            if (!elem.getAttribute('click-listener')) {
+                elem.addEventListener('click', () => addTagElement(elem.dataset.tag, true));
+                elem.setAttribute('click-listener', 'true');
+            }
+        });
+    }
+
+    function removeTagElement(resetFilter) {
+        const tagElement = document.getElementsByClassName('tagFilterLabel')[0];
+        tagElement.style.display = 'none';
+        if (resetFilter)
+            resetAndFilter('tag', '');
+    }
 
     function updateVisibleParams() {
         document.querySelectorAll('.filter').forEach(filter => {filter.value = params.get(filter.name) || ''});
         updateFilterColors();
+        const tagKey = params.get('tag');
+        if (tagKey !== null && tagKey.trim() !== '') {
+            addTagElement(tagKey, false);
+        } else {
+            removeTagElement(false);
+        }
     }
 
     window.addEventListener('resize', (event) => {
@@ -417,6 +464,8 @@
         document.querySelectorAll('.filter').forEach(filter => {
             filter.addEventListener('change', () => {resetAndFilter(filter.name, filter.value)});
         });
+        const tagElement = document.getElementsByClassName('tagFilterLabel')[0];
+        tagElement.addEventListener('click', () => removeTagElement(true));
         if (filters) {
             const currentLink = window.location.search;
             const newLink = `?${params.toString()}`;
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -358,19 +358,13 @@
             window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
             setCookie(filterCookieName, params.toString());
         }
-        document.querySelectorAll('.filter').forEach(filter => {
-            if (filter.value) {
-                filter.style = 'background-color: #858585; color: #fff';
-            } else {
-                filter.style = 'background-color: #ffffff; color: black';
-            }
-        });
+        updateFilterColors();
         await loadAndDisplayBooks(true);
     }
 
     window.addEventListener('popstate', async () => {
         await resetAndFilter();
-        document.querySelectorAll('.filter').forEach(filter => {filter.value = params.get(filter.name) || ''});
+        updateVisibleParams();
     });
 
     async function loadSubset() {
@@ -382,6 +376,21 @@
                 fadeOutDiv.style.display = 'none';
             }
         }
+    }
+
+    function updateFilterColors() {
+        document.querySelectorAll('.filter').forEach(filter => {
+            if (filter.value) {
+                filter.style = 'background-color: #858585; color: #fff';
+            } else {
+                filter.style = 'background-color: #ffffff; color: black';
+            }
+        });
+    }
+
+    function updateVisibleParams() {
+        document.querySelectorAll('.filter').forEach(filter => {filter.value = params.get(filter.name) || ''});
+        updateFilterColors();
     }
 
     window.addEventListener('resize', (event) => {
@@ -422,12 +431,7 @@
         if (filters) {
             window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
         }
-        params.forEach((value, key) => {
-            const selectBox = document.getElementsByName(key)[0];
-            if (selectBox) {
-                selectBox.value = value
-            }
-        });
+        updateVisibleParams();
         document.getElementById('kiwixSearchForm').onsubmit = (event) => {event.preventDefault()};
         if (!window.location.search) {
             const browserLang = navigator.language.split('-')[0];
@@ -438,13 +442,7 @@
                 langFilter.dispatchEvent(new Event('change'));
             }
         }
-        document.querySelectorAll('.filter').forEach(filter => {
-            if (filter.value) {
-                filter.style = 'background-color: #858585; color: #fff';
-            } else {
-                filter.style = 'background-color: #ffffff; color: black';
-            }
-        });
         setCookie(filterCookieName, params.toString());
     }
 })();
+
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -282,7 +282,7 @@
                     kiwixHomeBody.setAttribute('style', 'display: flex; justify-content: center; align-items: center');
                     divTag.getElementsByTagName('a')[0].onclick = (event) => {
                         event.preventDefault();
-                        window.history.pushState({}, null, `${window.location.href.split('?')[0]}?lang=`);
+                        window.history.pushState({}, null, `?lang=`);
                         setCookie(filterCookieName, 'lang=');
                         resetAndFilter();
                         document.querySelectorAll('.filter').forEach(filter => {
@@ -355,7 +355,7 @@
         params = new URLSearchParams(window.location.search);
         if (filterType) {
             params.set(filterType, filterValue);
-            window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
+            window.history.pushState({}, null, `?${params.toString()}`);
             setCookie(filterCookieName, params.toString());
         }
         updateFilterColors();
@@ -429,7 +429,7 @@
             filter.addEventListener('change', () => {resetAndFilter(filter.name, filter.value)});
         });
         if (filters) {
-            window.history.pushState({}, null, `${window.location.href.split('?')[0]}?${params.toString()}`);
+            window.history.pushState({}, null, `?${params.toString()}`);
         }
         updateVisibleParams();
         document.getElementById('kiwixSearchForm').onsubmit = (event) => {event.preventDefault()};
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -277,23 +277,9 @@
                 setTimeout(() => {
                     const divTag = document.createElement('div');
                     divTag.setAttribute('class', 'noResults');
-                    divTag.innerHTML = `No result. Would you like to <a href="/?lang=">reset filter</a>?`;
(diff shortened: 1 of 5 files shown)


kelson42 (changes requested):
In the principle, this is what I meant. But we have to improve a few things:
* "Search" button size is somehow changed now, should be the same size like before
* The tag link should be a real link, and there should be clear visually that we can click on it
* One link per tag, it seems it can only take two tags together (if we have two tags)

There is as well a few other visual details, but we can have a look to that later.

juuz0:
> The tag link should be a real link, and there should be clear visually that we can click on it

Do you mean the real link as `<a>` tag?
Note that the whole book div is covered by another `<a>`, HTML specification doesn't allow nesting of anchor tags

juuz0:
> @juuz0 Any chance we could try to reduce the surface of the tile link? Any way this is not a blocker if we can not have a real link.

Reducing surface could be neat, I'll try thanks

juuz0:
Rebased too

kelson42:
@juuz0 I consider to be an acceptable first version of this feature.
@veloman-yunkan Would you be able please to make a code review?

kelson42 (approved):
LGTM

static/skin/index.js
```diff
@@ -372,6 +383,16 @@
...
     }
 
+    function addTagElement(tagValue) {
+        const tagElement = document.getElementsByClassName('tagFilterLabel')[0];
+        tagElement.style.display = 'inline-block';
+        tagElement.innerHTML = `&#127991; ${tagValue}`;
+        tagElement.addEventListener('click', (e) => {
+            tagElement.style.display = 'none';
+            resetAndFilter('tag', '');
+        });
```
veloman-yunkan, line 393: This will keep accumulating event listeners associated with the tag label button. You should do this only once rather than on every call of `addTagElement()`

veloman-yunkan (changes requested):

static/skin/index.js
```diff
@@ -407,6 +424,12 @@
         document.querySelectorAll('.filter').forEach(filter => {
             filter.addEventListener('change', () => {resetAndFilter(filter.name, filter.value)});
         });
+        document.querySelectorAll('.tagFilterLabel').forEach(tagElement => {
+            tagElement.addEventListener('click', (e) => {
+                tagElement.style.display = 'none';
+                resetAndFilter('tag', '');
+            });
+        });
```
veloman-yunkan, line 432: This code is a little confusing. It creates a currently wrong impression that there can be multiple tag elements.
It would be better to have a function `removeTagFilter()` right below `addTagElement()`. Then, instead of dynamically binding an event listener to the (only) tag label, you can specify the handler through the `onclick` attribute in HTML.

veloman-yunkan (changes requested):

static/skin/index.js
```diff
@@ -91,9 +91,16 @@
         const langCode = getInnerHtml(book, 'language');
         const language = languages[langCode];
         const tags = getInnerHtml(book, 'tags');
-        let tagHtml = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
-            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)})
-            .join(' | ').replace(/_/g, ' ');
+        const tagList = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
+            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)});
+        const tagFilterLinks = tagList.map((tagValue) => {
+            tagValue = tagValue.toLowerCase().replace(/\s/g, '_');
```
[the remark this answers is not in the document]
mgautierfr: Yes, the only forbidden char is `;` https://wiki.openzim.org/wiki/Tags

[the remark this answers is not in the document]
veloman-yunkan: Then the next question is - if whitespace is replaced with an underscore here, where is the effect of this transformation reversed?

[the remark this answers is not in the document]
juuz0: in line 442,
using `humanFriendlyTitle` function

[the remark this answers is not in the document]
veloman-yunkan: But the transformed tag value infiltrates the link (see the several lines below). I think that's incorrect.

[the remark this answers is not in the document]
juuz0: Sorry, I don't fathom how it infiltrates.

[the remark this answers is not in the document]
juuz0: Sorry, but I can't fathom how it infiltrates.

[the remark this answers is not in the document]
veloman-yunkan: Let me explain it as inline comments in the `generateTagLink()` function:

```javascript
function generateTagLink(tagValue) {
    tagValue = tagValue.toLowerCase().replace(/\s/g, '_');
    // From now on tagValue has whitespace replaced with underscores 
    let tempParams = new URLSearchParams(window.location.search);
    tempParams.set('tag', tagValue);
    // In tempParams the parameter 'tag' has a value where whitespace has been replaced with underscores
    const currentLink = `${window.location.href.split('?')[0]}?${tempParams.toString()}`;
    // currentLink was generated with a tag value where whitespace has been replaced with underscores
    // Therefore underscores (not originally present in the tag value) have infiltrated the link. QED.
    return `<a class='tag__link' href='${currentLink}'>${humanFriendlyTitle(tagValue)}</a>`
}
```

veloman-yunkan (changes requested):
Please fix the last issue still remaining from the previous iteration (see https://github.com/kiwix/libkiwix/pull/711#discussion_r841208347)

static/skin/index.js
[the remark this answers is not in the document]
juuz0: That's intentional (replacing whitespaces with underscores in the link).
A tag like `science in the bath` can't be used like: `tag=science in the bath`. For that, I changed it to `tag=science_in_the_bath` which works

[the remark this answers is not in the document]
veloman-yunkan: > A tag like `science in the bath` can't be used like: `tag=science in the bath`. For that, I changed it to `tag=science_in_the_bath` which works

Is the treatment of underscores in the tag value a document feature? Which code does implement it? Is there a unit-test for it?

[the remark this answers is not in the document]
juuz0: My bad, I didn't realise the tag value had underscores previously too so `.replace(/\s/g, '_');` is useless.
The tag was always `science_in_the_bath`

veloman-yunkan (changes requested):
This time I played with the code instead of reviewing it. Here are a couple of observations:

1. When no books match the selection criteria "No result. Would you like to **reset filter**?" is shown. If filtering by tag was enabled (and the tag label was displayed), clicking the **reset filter** link doesn't remove the tag label (though filtering by tag gets disabled).
2. With tag filtering enabled, when I type text in the search box and click the search button, filtering by tag is preserved. However, if I type text in the search box and hit Enter, filtering by tag is removed.
3. Clicking the tag link resets any language and/or category filters.
4. History tracking doesn't work well (and that doesn't seem to be restricted to tags)

kelson42:
@juuz0 Any feedback?

juuz0:
Soon :+1:

kelson42:
@juuz0 Any news? Are you blocked?

juuz0:
Not ready for review yet (aware of the CI too), but will be soon :+1:

kelson42:
Any chance we can conclude soon? I woukd really like to bring it in the next release.

juuz0:
> Clicking the tag link resets any language and/or category filters.

Fixing this required (and also having real links) would require to reload all books (current behaviour of kiwix-serve only added/removed new books on updating lang/category). Instead, I removed links (no `<a>` tags now). 

> When no books match the selection criteria.......

Fixed

> With tag filtering enabled, when I type text in the search box............

Fixed

> History tracking doesn't work well...........

Should be fixed (atleast related to tags), but happy to have an example where it isn't working

juuz0:
@kelson42 Please review it again since I removed links :)

kelson42 (changes requested):
Seems to work fine but:
* Please put a "cursor" pointer of the tag "link"
* Put a `title` and a `aria-label` tag (for usability) on the links/buttons (don't forget they should be translated). See the top taskbar loaded for a ZIM file as example
* Remove the strange artifact on the left in the tag buttong 
image

kelson42:
@juuz0 Any update?

juuz0:
@kelson42 Yes, doing this today :+1:

static/skin/index.js
```diff
@@ -285,6 +297,7 @@
                         window.history.pushState({}, null, `${window.location.href.split('?')[0]}?lang=`);
                         setCookie(filterCookieName, 'lang=');
                         resetAndFilter();
+                        removeTagElement(true);
```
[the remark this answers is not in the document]
juuz0: >  Maybe the click event handler of the newly created div element deserves a separate named function too. This should be done in a separate refactoring commit.

Went against this because I feel the click handling is very much only for this div. Will change it to a named function if you do not like though :)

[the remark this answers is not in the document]
veloman-yunkan: It looks like the first part of my comment was not addressed.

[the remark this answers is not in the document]
veloman-yunkan: Sorry, that was a false observation. Got lost between the old and new version of this PR.

static/skin/index.js
```diff
@@ -267,35 +279,38 @@
         });
     }
 
+    function noResultsContent(kiwixHomeBody) {
```
veloman-yunkan, line 282: Function names and signatures must not be misleading. This function name in its current form suggests that it returns some content, whereas it modifies its parameter and some global state too (`loader`). Therefore a better name would be `setNoResultsContent()`. Besides, passing `kiwixHomeBody` as a parameter doesn't make sense - since that creates a wrong impression that all side-effect of that function is limited to the `kiwixHomeBody` element.

static/skin/index.js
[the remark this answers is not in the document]
veloman-yunkan: > > Maybe the click event handler of the newly created div element deserves a separate named function too. This should be done in a separate refactoring commit.
> 
> Went against this because I feel the click handling is very much only for this div. Will change it to a named function if you do not like though :)

The utility of named functions is that - when named correctly - they help to understand the logic of the code easier without having to decipher it.

But now I have another question regarding the click handler - why do we need it at all? Won't the browser's following that link have the same effect?

static/skin/index.js
```diff
@@ -267,35 +279,38 @@
...
+            resetAndFilter();
+            removeTagElement(false);
+            document.querySelectorAll('.filter').forEach(filter => {
+                filter.value = params.get(filter.name) || '';
+                if (filter.value) {
+                    filter.style = 'background-color: #858585; color: #fff';
+                } else {
+                    filter.style = 'background-color: #ffffff; color: black';
+                }
+            })
```
veloman-yunkan, line 301: Why is this at all needed?

veloman-yunkan (commented):
I tried to figure out what changed since my last review, but failed since the changes were not made as fixup commits. I will test and fully review this PR again tomorrow.

static/skin/index.js
juuz0: I don't know it fully myself but after a series of tests, this is helping in back-forward actions.
Without this, if "reset filters" link is pressed, it correctly clears all filters. If then history back button is pressed, it correctly goes back (although I have to press it twice) but now I can't press the history forward button (gets disabled) to get back to the page with filters removed.

static/skin/index.js
```diff
@@ -279,26 +279,31 @@
         });
     }
 
-    function noResultsContent(kiwixHomeBody) {
+    function resetAllFiltersAndRedirect() {
+        window.history.pushState({}, null, `${window.location.href.split('?')[0]}?lang=`);
```
veloman-yunkan, line 283: 1. Under Firefox this can be shortened to a clearer `window.history.pushState({}, null, '?lang=');`. If other browsers handle relative urls in `window.history.pushState()` equally well, then it makes sense to simplify all usages of that method in this file.
2. Also it makes sense to introduce a small helper function `pushFilterState()` that combines `window.history.pushState()` with `setCookie(filterCookieName, ...)`.

static/skin/index.js
```diff
@@ -267,35 +279,43 @@
...
+        resetAndFilter();
+        removeTagElement(false);
+        document.querySelectorAll('.filter').forEach(filter => {
+            filter.value = params.get(filter.name) || '';
+            if (filter.value) {
+                filter.style = 'background-color: #858585; color: #fff';
+            } else {
+                filter.style = 'background-color: #ffffff; color: black';
+            }
+        });
```
veloman-yunkan, line 294: Querying `params` (and subsequent logic) in a function that is supposed to reset all filters looks a little strange. However, I see that this piece of code is almost duplicated in `resetAndFilter()` and the handler installed on `window.onload`. So instead of optimizing this instance, it makes more sense to deduplicate.

static/skin/index.js
veloman-yunkan: I will have to deal with history tracking in #716, so I am now going to dive into https://html.spec.whatwg.org/multipage/history.html. If I find out the reason for the issues you observed with simpler code and ways to solve them easily, I will let you know.

juuz0:
In fixup commit,
I removed all the reset link logic, now it's only a simple link.
In the onload handler, we are checking if the link with params is same as current, helps not to push the same thing twice.
Seems to work well for all 3 you mentioned

1. open the kiwix-serve welcome/library page
2. click on a book or enter another (unrelated) URL (e.g. openzim.org) in the address bar
3. go back. Once you return to the kiwix-serve library, you cannot navigate forward in browsing history.

veloman-yunkan:
> In fixup commit, I removed all the reset link logic, now it's only a simple link. In the onload handler, we are checking if the link with params is same as current, helps not to push the same thing twice. Seems to work well for all 3 you mentioned
> 
>     1. open the kiwix-serve welcome/library page
> 
>     2. click on a book or enter another (unrelated) URL (e.g. openzim.org) in the address bar
> 
>     3. go back. Once you return to the kiwix-serve library, you cannot navigate forward in browsing history.

Great!

Now you have to make sure that the color of the filter fields is also properly updated during history navigation. For example, consider the following scenario:

1. Open the library without any filter applied. All fields are white.
2. Enter some garbage in the search field so that no book matches it. You end up with a gray search field and "No result. Would you like to reset filter" instead of a list of books.
3. Navigate back. The search field becomes empty but stays gray.
4. Navigate forward. The text in the search field is restored, but the field becomes white.

juuz0:
> Now you have to make sure that the color of the filter fields is also properly updated......

Done :heavy_check_mark:

veloman-yunkan (commented):
Please also make the simplification recommended in the first part of https://github.com/kiwix/libkiwix/pull/711#discussion_r904944870 (as a separate clean-up commit) and rebase-fixup this PR for the last iteration.

static/skin/index.js
```diff
@@ -279,36 +279,24 @@
...
+    function setNoResultsContent() {
+        const kiwixHomeBody = document.querySelector('.kiwixHomeBody');
+        const divTag = document.createElement('div');
+        divTag.setAttribute('class', 'noResults');
+        divTag.innerHTML = `No result. Would you like to <a href="/?lang=">reset filter</a>?`;
+        kiwixHomeBody.append(divTag);
+        kiwixHomeBody.setAttribute('style', 'display: flex; justify-content: center; align-items: center');
+        loader.setAttribute('style', 'position: absolute; top: 50%');
+    }
+
```
veloman-yunkan, line 291: The message of this commit must explain why this function is shorter than the inline code that it is supposed to replace. Or, even better, you can make this change in two steps:

1. Drop the `onclick` handler for the "reset filter" link (explaining why it was needed & what was wrong with that workaround)
2. Extract the `setNoResultsContent()` function

veloman-yunkan (changes requested):
Please rewrite the history by moving all cleanup & refactoring changes to the beginning.

static/skin/index.js
juuz0: Done, thank you!

static/skin/index.js
```diff
@@ -280,20 +280,6 @@
                     divTag.innerHTML = `No result. Would you like to <a href="/?lang=">reset filter</a>?`;
```
veloman-yunkan: Just noticed that dropping the `onclick` handler exposes a bug in the URL which doesn't account for the `--urlRootLocation` option of kiwix-serve. If you fix it in this commit, don't forget to update the `setNoResultsContent()` in the next commit too.

static/skin/index.js
```diff
@@ -429,7 +415,11 @@
             filter.addEventListener('change', () => {resetAndFilter(filter.name, filter.value)});
         });
         if (filters) {
-            window.history.pushState({}, null, `?${params.toString()}`);
+            const currentLink = window.location.search;
+            const newLink = `?${params.toString()}`;
+            if (currentLink != newLink) {
+                window.history.pushState({}, null, newLink);       
+            }
```
veloman-yunkan, line 475: Please elaborate the description of the commit with the negative side-effects of `window.history.pushState()` being called unconditionally (the worst of which is the discarded forward history when the current page is loaded as a result of navigating back).

static/skin/index.js
```diff
@@ -91,9 +98,10 @@
         const langCode = getInnerHtml(book, 'language');
         const language = languages[langCode];
         const tags = getInnerHtml(book, 'tags');
-        let tagHtml = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
-            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)})
-            .join(' | ').replace(/_/g, ' ');
+        const tagList = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
+            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)});
```
veloman-yunkan, line 102: Looks like post-processing the `tagList` with `.map()` is no longer needed.

static/skin/index.js
```diff
@@ -91,9 +98,10 @@
         const langCode = getInnerHtml(book, 'language');
         const language = languages[langCode];
         const tags = getInnerHtml(book, 'tags');
-        let tagHtml = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
-            .map((tag) => {return tag.charAt(0).toUpperCase() + tag.slice(1)})
-            .join(' | ').replace(/_/g, ' ');
+        const tagList = tags.split(';').filter(tag => {return !(tag.split(':')[0].startsWith('_'))})
```
veloman-yunkan, line 101: `tag.split(':')[0].startsWith('_')` can be simplified to `tag.startsWith('_')`

veloman-yunkan (changes requested):
Almost at the finish line!

Yet, please address the new review comments in fixup commits.

static/skin/index.js
juuz0: Did this while editing the commit message and to update setNoResultsContent() there's a fixup commit

static/skin/index.js
```diff
@@ -277,23 +277,10 @@
                 setTimeout(() => {
                     const divTag = document.createElement('div');
                     divTag.setAttribute('class', 'noResults');
-                    divTag.innerHTML = `No result. Would you like to <a href="/?lang=">reset filter</a>?`;
+                    const resetLink = window.location.pathname + '?lang=';
+                    divTag.innerHTML = `No result. Would you like to <a href="${resetLink}">reset filter</a>?`;
```
veloman-yunkan: Simply removing the '/' from the link would do the trick.

static/skin/index.js
veloman-yunkan: Reading the updated commit message I see that you don't fully realize the problem with the unconditional `window.history.pushState()` here. It is independent of the handling of the "reset filter" link as demonstrated in https://github.com/kiwix/libkiwix/pull/711#discussion_r905029154. Therefore I recommend that this commit is also split into two changes:

1. Fix the problem with `window.history.pushState()` in `window.onload` with a proper explanation (you can include the example from https://github.com/kiwix/libkiwix/pull/711#discussion_r905029154).
2. Drop the `onclick` handler of the "reset filter" link noting that it was a workaround of a more general issue now fixed by the previous commit.

static/skin/index.js
juuz0: Neat

static/skin/index.js
juuz0: Correct, thank you.

veloman-yunkan (approved):
Thank you for your patience!

Issue #650, opened 2021-12-11
Update Kiwix-serve cache strategy


kelson42:
Kiwix serve delivers many types of content:
1.  ZIM content
2. Kiwix-serve static content
3. Kiwix-serve dynamic content (API, search, etc.)

For each type, here is how we want to cache:
1. ZIM content: cache as long as the (concerned) content does not change
2. Kiwix-serve static content: cache as long as we don't run a different version of kiwix-serve
3. Kiwix-serve dynamic content: cache as long as the (concerned) library has not been updated

Which implies 3 different strategies...

## ZIM content

ZIM content does give us no freedom about the URLs calls and we have no way to garanty that what is available at the URL at two different times is the same content. Therefore the content can be cached but must be revalidate (with `Etag` issued from the ZIM ID). To avoid a systematic revalidation we can set a max-age of one hour like `Cache-Control: max-age=3600, must-revalidate`.

## Kiwix-serve static content

This is the easier case, we need to "cache busting", for example by adding `?UNIQUE_ID` at the end of the URL of any of the resources. This `UNIQUE_ID` should be fixed at compilation time. `Cache-Control: max-age=31536000, immutable` then should be enough (so no other related HTTP headers, like  `Etag`, `Last-Modified` or `Expires`).

## Kiwix-serve dynamic content

The dynamic content is consistent and can be cached as long as the library is not updated or the server restarted. I would propose to force systematically revalidation with an `Etag` which is changed each time the internal library is changed. The cache policy would look like `Cache-Control: max-age=0, must-revalidate`.

kelson42:
@veloman-yunkan I want to make a new attempt to sortout this problem of cache. Because the problem is broader than the API or Etag, I have close #598, but thank you very much again for your explanation which are perfectly correct I think now. Would you be able please to assess this ticket and tell me if you disagree with something? Or maybe you would have an other approach to do it better?

kelson42:
While you are waiting and interacting with @mgautierfr on https://github.com/kiwix/libkiwix/pull/679 and I suspect it might take weeks ; I would really appreciate to have you looking at this ticket.

veloman-yunkan:
> While you are waiting and interacting with @mgautierfr on #679 and I suspect it might take weeks ; I would really appreciate to have you looking at this ticket.

@kelson42 OK, will do.

veloman-yunkan:
> == Kiwix-serve static content ==
> 
[...]

PR #712, opened 2022-02-14
Static resource versioning

veloman-yunkan:
Addresses the static resource part of #650

HTML templates are now preprocessed by a new script `kiwix-resources` that detects links marked with `?KIWIXCACHEID` and substitutes `cacheid=CACHEID` in lieu of the `KIWIXCACHEID` placeholder. The `CACHEID` value of a resource is obtained by taking the first 8 characters of its SHA1 hashsum hex digest. Inter-resource dependency is correctly handled by computing the cacheid for the preprocessed (rather than the source) version of a resource.

### DIFF
--- a/scripts/kiwix-resources
+++ b/scripts/kiwix-resources
@@ -53,26 +53,26 @@ def set_cacheid(resource_matchobj):
     cacheid = 'cacheid=' + get_resource_revision(resource)
     return path + '?' + cacheid + extra_query
 
-def preprocess_line(line):
-    if 'KIWIXCACHEID' in line:
-        line = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, line)
-        assert not 'KIWIXCACHEID' in line
-    return line
+def preprocess_text(s):
+    if 'KIWIXCACHEID' in s:
+        s = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, s)
+        assert not 'KIWIXCACHEID' in s
+    return s
 
 def get_preprocessed_resource(srcpath):
-    modified_line_count = 0
-    preprocessed_lines = []
+    """Get the transformed content of a resource
+
+    If the resource at srcpath is modified by preprocessing then this function
+    returns the transformed content of the resource. Otherwise it returns None.
+    """
     try:
-        with open(srcpath, 'r') as source:
-            for line in source:
-                ppline = preprocess_line(line)
-                if ppline != line:
-                    modified_line_count += 1
-                preprocessed_lines.append(ppline)
-        return "".join(preprocessed_lines), modified_line_count
+        with open(srcpath, 'r') as resource_file:
+            content = resource_file.read()
+            preprocessed_content = preprocess_text(content)
+        return preprocessed_content if preprocessed_content != content else None
     except UnicodeDecodeError:
         # It was a binary resource
-        return None, 0
+        return None
 
 
 def symlink_resource(src, resource_path):
@@ -91,8 +91,8 @@ def preprocess_resource(resource_path):
     outpath = os.path.join(OUT_DIR, resource_path)
     if os.path.exists(outpath):
         os.remove(outpath)
-    preprocessed_content, modified_line_count = get_preprocessed_resource(srcpath)
-    if modified_line_count == 0:
+    preprocessed_content = get_preprocessed_resource(srcpath)
+    if preprocessed_content is None:
         symlink_resource(srcpath, outpath)
     else:
         with open(outpath, 'w') as target:
--- a/scripts/kiwix-resources
+++ b/scripts/kiwix-resources
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+
+'''
+Copyright 2022 Veloman Yunkan <veloman.yunkan@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 3 of the License, or any
+later version.
+
+This program is distributed in the hope that it will be useful, but
+WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+02110-1301, USA.
+'''
+
+import argparse
+import hashlib
+import os.path
+import re
+
+def read_resource_file(resource_file_path):
+    with open(resource_file_path, 'r') as f:
+        return [line.strip() for line in f]
+
+def list_resources(resource_file_path):
+    for resource_path in read_resource_file(resource_file_path):
+        print(resource_path)
+
+def get_resource_revision(base_dir, resource_path):
+    with open(os.path.join(base_dir, resource_path), 'rb') as f:
+        return hashlib.sha1(f.read()).hexdigest()[:8]
+
+resource_revisions = {}
+
+def fill_resource_revisions(resource_file_path):
+    base_dir = os.path.dirname(os.path.realpath(resource_file_path))
+    for resource in read_resource_file(resource_file_path):
+        resource_revisions[resource] = get_resource_revision(base_dir, resource)
+
+RESOURCE_WITH_CACHEID_URL_PATTERN=r'"([^"?]+)\?KIWIXCACHEID([^"]*)"'
+
+def set_cacheid(resource_matchobj):
+    path = resource_matchobj.group(1)
+    resource = path
+    root_prefix = '{{root}}/'
+    if resource.startswith(root_prefix):
+        resource = resource[len(root_prefix):]
+    extra_query = resource_matchobj.group(2)
+    cacheid = 'cacheid=' + resource_revisions[resource]
+    return f'"{path}?{cacheid}{extra_query}"'
+
+def preprocess_line(line):
+    if 'KIWIXCACHEID' in line:
+        line = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, line)
+        assert not 'KIWIXCACHEID' in line
+    return line
+
+def preprocess_template(srcpath, dstpath):
+    preprocessed_lines = []
+    with open(srcpath, 'r') as source:
+        for line in source:
+            preprocessed_lines.append(preprocess_line(line))
+
+    with open(dstpath, 'w') as target:
+        print("".join(preprocessed_lines), end='', file=target)
+
+def symlink_resource(src, resource_path):
+    if os.path.exists(resource_path):
+        if os.path.islink(resource_path) and os.readlink(resource_path) == src:
+            return
+        os.remove(resource_path)
+    os.symlink(src, resource_path)
+
+def preprocess_resource(srcdir, resource_path, outdir):
+    resource_dir = os.path.dirname(resource_path)
+    if resource_dir != '':
+        os.makedirs(os.path.join(outdir, resource_dir), exist_ok=True)
+    srcpath = os.path.join(srcdir, resource_path)
+    outpath = os.path.join(outdir, resource_path)
+    if resource_path.startswith('templates/'):
+        preprocess_template(srcpath, outpath)
+    else:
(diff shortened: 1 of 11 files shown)


veloman-yunkan:
@mgautierfr @kelson42 

Before merging this PR I would like to functionally cover with unit-tests entire code that produces responses from mustache templates so that we have a guarantee that nothing breaks. However, since that can be a lot of work, I think that this PR should be reviewed now because if the taken approach must be changed then an early decision can be helpful.

kelson42:
@veloman-yunkan Fine to me. Appreciate your approach to improve tests before merging features.

scripts/kiwix-resources
```diff
@@ -0,0 +1,106 @@
...
+    base_dir = os.path.dirname(os.path.realpath(resource_file_path))
+    for resource in read_resource_file(resource_file_path):
+        resource_revisions[resource] = get_resource_revision(base_dir, resource)
+
+src_or_href_pattern=r'(src|href)="\{\{root\}\}/(skin/[^"]+)"'
+def add_resource_revision(resource_matchobj):
+    attr = resource_matchobj.group(1)
+    resource = resource_matchobj.group(2)
+    if resource in resource_revisions:
+        uniqueid = '?' + resource_revisions[resource]
```
mgautierfr, line 51: If the matched link already contains a query string, this will create a invalid link.
I don't think it is important for now as the resource links are pointing to "file" but I prefer to mention it now.

mgautierfr (commented):
I agree with the global idea. 

I've few comments:
- `static/skin/index.js` also have link to resources we should cache. We must not only process the templates but also other resources
- Instead of `/skin/foo.png&<md5id>`, I would prefer a `/skin/foo.png&cacheid=<md5id>`. It is a bit more verbose and convey correctly the intent of the "random numbers" at the end.
- I wonder if we should also store the md5id somewhere in the cpp code if we have to generate links to resource on the cpp side. I don't know if it is really useful, just questioning.

veloman-yunkan:
@mgautierfr 

> * `static/skin/index.js` also have link to resources we should cache. We must not only process the templates but also other resources

I introduced an explicit `KIWIXCACHEID` placeholder in URLs where the cacheid must be injected. It seemed to be the easiest way to safely inject the cacheid in any type of a resource. The downside is that the URLs must now be manually marked (when adding in files under `static/' new URLs to static resources).

> * I wonder if we should also store the md5id somewhere in the cpp code if we have to generate links to resource on the cpp side. I don't know if it is really useful, just questioning.

If we ever need that then we can generate the link via a mustache template; hence no cacheid will need to be stored in the source code (excluding the generated file `kiwixlib-resources.cpp`).

scripts/kiwix-resources
veloman-yunkan: The hack with `KIWIXCACHEID` solves this potential issue as well

scripts/kiwix-resources
```diff
@@ -0,0 +1,126 @@
...
+    resource = 'skin/' + resource_matchobj.group(3)
+    extra_query = resource_matchobj.group(4)
+    cacheid = 'cacheid=' + resource_revisions[resource]
+    return '"' + path + '?' + cacheid + extra_query + '"'
+
+def preprocess_line(line):
+    if 'KIWIXCACHEID' in line:
+        line = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, line)
+        assert not 'KIWIXCACHEID' in line
+    return line
```
mgautierfr, line 59: I think this could be better rewriten with something link:

```python
RESOURCE_WITH_CACHEID_URL_REG = re.compile(r'(?P<pre_url>".*/(?P<resource>skin/[^"?]+)\?.*)KIWIXCACHEID(?P<post_url>[^"]*")'

def set_cacheid(resource_matchobj):
    resource = resource_matchopj.group('resource')
    cacheid = 'cacheid='+resource_revisions[resource]
    return resource_matchobj.group('pre_url') + cacheid + resource_matchobj.group('post_url')

def process_line(line):
    if 'KIWIXCACHEID' in line:
        RESOURCE_WITH_CACHEID_URL_REG.sub(set_cacheid, line)
        assert not 'KIWIXCACHEID' in line
    return line
```
(not tested)

scripts/kiwix-resources
```diff
@@ -0,0 +1,126 @@
...
+    modified_line_count = 0
+    preprocessed_lines = []
+    try:
+        with open(srcpath, 'r') as source:
+            for line in source:
+                ppline = preprocess_line(line)
+                if ppline != line:
+                    modified_line_count += 1
+                preprocessed_lines.append(ppline)
+        return "".join(preprocessed_lines), modified_line_count
```
mgautierfr, line 71: Why process each line separately ?
It should be possible to pass all the content to `re.sub`(and `re.subn` returns the number of occurrences replaced)

kelson42:
@veloman-yunkan Should we remove the WIP?

mgautierfr:
@veloman-yunkan do you plan to fix my comment ?
As I said, they are not blocker. If you don't plan to work on this anymore I will rebase on master and merge.

veloman-yunkan:
> @veloman-yunkan do you plan to fix my comment ? As I said, they are not blocker. If you don't plan to work on this anymore I will rebase on master and merge.

@mgautierfr As I have [stated earlier in this PR](#issuecomment-1041747579) I expected more work to be done related to testing. Some of the sought coverage improvement has already been merged into master by a few PRs extracted from #679 but there are still some pages that are not tested. If you feel comfortable having this PR merged before achieving 100% coverage of the functionally affected areas, then I can clean it up for final review.

kelson42:
@veloman-yunkan I'm supportive to 100% code coverage for this feature... Even if this takes a bit longer. That said, 1 month without update is a really long time. I understand the current situation is complex and probably exceptional (many fronts, many big PRs, etc.), that said, this is the kind of situation to avoid. Preferable are few PRs which move quickly forward to their respective goals.

veloman-yunkan:
The current implementation contains a bug: it doesn't take into account the dependence of static resource on each other. If resource A refers to B and B refers to C, then a change in C would result in its cache id being updated in the preprocessed version of B. However the cache id of B won't change since the cache id is derived from the source rather than from the preprocessed output. Will fix that issue next week.

veloman-yunkan:
> * I wonder if we should also store the md5id somewhere in the cpp code if we have to generate links to resource on the cpp side. I don't know if it is really useful, just questioning.

@mgautierfr In #744 I ran against a situation where this must be addressed (see https://github.com/kiwix/libkiwix/pull/744#issuecomment-1089083287)

kelson42:
@veloman-yunkan Is that the next PR to polish/merge?

veloman-yunkan:
> @veloman-yunkan Is that the next PR to polish/merge?

@kelson42 Yes. I will rebase it today.

veloman-yunkan:
> > @veloman-yunkan Is that the next PR to polish/merge?
> 
> @kelson42 Yes. I will rebase it today.

Done. Other than resolving a conflict with a change coming from #697, I only slightly changed the order of commits and their descriptions. @mgautierfr's comments left on the early WIP version are still unaddressed.

kelson42:
@veloman-yunkan OK, so waiting to you to fix this last things before @mgautierfr can review this PR.

veloman-yunkan:
> @veloman-yunkan OK, so waiting to you to fix this last things before @mgautierfr can review this PR.

@kelson42 There is nothing to fix that I am aware of. @mgautierfr's comments were just suggestions for (arguable) improvements. I will consider making them if the feedback provided on the undrafted version of this PR doesn't lead to changes that cancel the opportunity for such improvements.

kelson42:
@mgautierfr Can you please review this PR? It seems to be ready now!

mgautierfr (commented):
I'm globally ok with the code of this PR.
The questioned commit is acceptable (at least, while we have very few use case for search problem. If we have more need to generate link, we will need a proper solution)

**But**, I recently read this page https://medium.com/@hutyxxx/how-does-the-browser-know-what-to-cache-edfd5d63c239. At cache busting section it says that using a query-string is generally discouraged (Mainly because it seems that some proxies deactivate cache for resource with a query string). As we use a lot kiwix-serve behind a proxy, it would be nice to be sure we are ok
@rgaudin can you confirm/test/check this ?
If we don't go with a query string, we will have to change the url `foo.js?cacheid=abcdef` to `foo-abcdef.js` and store a mapping `cacheid_url -> generic_url` to allow the server to answer the request (and it may help us to fix the questioned commit)

rgaudin:
> **But**, I recently read this page https://medium.com/@hutyxxx/how-does-the-browser-know-what-to-cache-edfd5d63c239. At cache busting section it says that using a query-string is generally discouraged (Mainly because it seems that some proxies deactivate cache for resource with a query string). As we use a lot kiwix-serve behind a proxy, it would be nice to be sure we are ok @rgaudin can you confirm/test/check this ? If we don't go with a query string, we will have to change the url `foo.js?cacheid=abcdef` to `foo-abcdef.js` and store a mapping `cacheid_url -> generic_url` to allow the server to answer the request (and it may help us to fix the questioned commit)

I think this might be outdated information. Haven't tested with Squid which [reportedly changed that back in 2008](https://blog.giantgeek.com/?p=1557) but I've tested with varnish (the one we use) we use and query-stringed paths are cached properly. Otherwise would have been a shame considering our illustrations are only available though query strings.

I prefer query string as it leaves paths properly named and *permanent-friendly* (although resources don't have same expectations as content).

mgautierfr:
Then we can merge
I've rebased on master (and remove the `[???]` in the commit message.

scripts/kiwix-resources
veloman-yunkan: Done but in a way that still requires `KIWIXCACHEID` to immediately follow the `?` symbol. This is done for safety and doesn't create any obstacles.

scripts/kiwix-resources
veloman-yunkan: Done in a very slightly different way. But I doubt that the result is a net optimization (if that was the point of this suggestion).

mgautierfr: It was not a optimization. More a simplification of the code. This code is not really a critical path, I prefer to have simpler code

scripts/kiwix-resources
```diff
@@ -53,23 +53,18 @@ def set_cacheid(resource_matchobj):
...
     return path + '?' + cacheid + extra_query
 
-def preprocess_line(line):
-    if 'KIWIXCACHEID' in line:
-        line = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, line)
-        assert not 'KIWIXCACHEID' in line
-    return line
+def preprocess_text(s):
+    if 'KIWIXCACHEID' in s:
+        s = re.sub(RESOURCE_WITH_CACHEID_URL_PATTERN, set_cacheid, s)
```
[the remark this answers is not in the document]
veloman-yunkan: I also approached this change as a simplification. Returning a single value instead of a pair of values that must be consistent with each other is inline with that goal. A simple comparison of the source content and the result of transforming it makes `preprocess_text()` easier to enhance should such a need arise.

scripts/kiwix-resources
```diff
@@ -53,23 +53,18 @@ def set_cacheid(resource_matchobj):
...
-                    modified_line_count += 1
-                preprocessed_lines.append(ppline)
-        return "".join(preprocessed_lines), modified_line_count
+        with open(srcpath, 'r') as resource_file:
+            content = resource_file.read()
+            preprocessed_content = preprocess_text(content)
+        return preprocessed_content, preprocessed_content != content
     except UnicodeDecodeError:
         # It was a binary resource
         return None, 0
```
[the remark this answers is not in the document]
veloman-yunkan: You are right. It just missed my attention. Thanks for noticing it. I took this opportunity to make this function return a single value too.

veloman-yunkan:
During rebase & fixup the commit message of the commit currently titled "An optimization (?) of resource preprocessing" must be rewritten.

mgautierfr (approved):
LGTM

> During rebase & fixup the commit message of the commit currently titled "An optimization (?) of resource preprocessing" must be rewritten.

Please rebase-fixup and rewrite the message. Then we can merge.

scripts/kiwix-resources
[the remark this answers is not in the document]
mgautierfr: Yes. And this part is not critical at all. Perfect is the enemy of good. Let's move on.

veloman-yunkan:
> Please rebase-fixup and rewrite the message. Then we can merge.

Done

PR #714, opened 2022-02-17
Optimized kiwix-serve welcome page

ankur12-1610:
Fixes: [kiwix-tools/#538](https://github.com/kiwix/kiwix-tools/issues/538)

Before: 
153743002-48e1f36a-d9a6-43a6-83f6-93532dab6150

After:
kiwix-media

### DIFF
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -129,7 +129,7 @@ body {
 }
 
 .kiwixHomeBody {
-    padding: 20px 15vw;
+    padding: 20px 5vw 20px 15vw;
     min-height: calc(100vh - 170px);
     position: relative;
 }
@@ -440,7 +440,7 @@ body {
 @media screen and (max-width: 1100px) {
 
     .kiwixHomeBody {
-        padding: 20px 10vw;
+        padding: 20px 5vw 20px 10vw;
         min-height: calc(100vh - 170px);
     }
 
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -129,7 +129,7 @@ body {
 }
 
 .kiwixHomeBody {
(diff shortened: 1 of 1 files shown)


ankur12-1610:
@kelson42 can you please review?

kelson42 (changes requested):
@ankur12-1610 Thank you for your PR. It indeed works better now, but I'm not really happy with the solution because there is still an asymetry between the left and the right. The tiles should maximize the space usage and be distributed in the middle horizontaly. So the whole block of the book list should be centered horizontally.

ankur12-1610:
> @ankur12-1610 Thank you for your PR. It indeed works better now, but I'm not really happy with the solution because there is still an asymetry between the left and the right. The tiles should maximize the space usage and be distributed in the middle horizontaly. So the whole block of the book list should be centered horizontally.

@kelson42 got it. So the current padding is `20px 5vw 20px 10vw`, the soultion whould be that I change it to `20px 5vw 20px 5vw` right?

ankur12-1610:
@kelson42 I've made the required changes kindly review :)

kelson42:
@ankur12-1610 This is getting better, but it seems not propoer for all screens. See here, it is not centered properly
image

kelson42:
@ankur12-1610 Any news on this PR?

ankur12-1610:
Hey @kelson42 sorry for the late reply

I'm working on it and will update the PR bt a day or two :')

kelson42:
@ankur12-1610 thx!

ankur12-1610:
@kelson42 I did the changes, although in a little unorthodox way :sweat_smile:, by testing every float value :)

kelson42:
Superseeded by #786

Issue #702, opened 2022-02-05
Implement OPDS "search" sorting


kelson42:
Currently there is no sorting at all.

Once we will have introduced the attribute of popularity this will be necessary, see #489

But there is another need. I wanted to be informed about the latest ZIM published via OPDS and I remarked that:
* There is not sorting by date (descending)
* There is not even a way to filter with a creation date range

veloman-yunkan:
Before implementing this enhancement we need to make sure that it plays well with the [OPDS spec](https://specs.opds.io/opds-1.2#62-sorting-relations). The spec mentions three fields that have to do with dates:

> OPDS Catalog Entries must include an `atom:updated` element indicating when the OPDS Catalog Entry was last updated. A `dc:issued` element should be used to indicate the first publication date of the Publication and must not represent any date related to the OPDS Catalog Entry.
>
> OPDS Catalog Entries may use `atom:published` to indicate when the OPDS Catalog Entry was first accessible.

Thus

- `atom:updated` is the date corresponding to the OPDS Entry (rather than the publication associated with it). In our case this should be the time when the book was added to the library during library loading or when the entry was updated during library reloading. 
- `atom:published` is the earliest value that `atom:updated` had for this OPDS feed entry. In our case this should be the time when the book was added to the library during library loading.
- `dc:issued` is the time when the actual publication was issued. It is unambiguous for publications that have only one of the hardcopy or digital representations. However, if we consider a paper publication that was then digitized or a book that was first published online and printed on paper later should we treat the hardcopy and the digital version as different representations of the same publication or as two different publications? I think we can use it to represent the creation date of ZIM files (though if a ZIM file represents a single real-world publication then we hit the mentioned interpretation problem). 

While doing this small research I found out that in our OPDS streams we populate the `atom:updated` field with the book creation date (which is against the spec):

https://github.com/kiwix/libkiwix/blob/dc4f9a4939eef6e227fae81cb5fb46e527157b9d/src/opds_dumper.cpp#L73-L97

Now the question is - should we fix the inconsistency with the usage of the `atom:updated` field and put the ZIM file creation date in a `dc:issued` node instead?

kelson42:
@veloman-yunkan Thank you for this research work, this is englighting!

> Before implementing this enhancement we need to make sure that it plays well with the [OPDS spec](https://specs.opds.io/opds-1.2#62-sorting-relations). The spec mentions three fields that have to do with dates:
> 
[...]

PR #715, opened 2022-02-19
Added <dc:issued> field to OPDS entries

veloman-yunkan:
As agreed in https://github.com/kiwix/libkiwix/issues/702#issuecomment-1030916357

As of this PR the `<updated>` field in OPDS entries is not affected in any way. The same data is simply duplicated in a `<dc:issued>` field. The `<updated>` field will be changed later, after 2 conditions are met:

1. All respective client code is switched to use the `<dc:issued>` field
2. library.xml format is enhanced to contain information about book metadata timestamps

### DIFF
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -161,7 +161,9 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
   m_language = VALUE("language");
   m_creator = node.child("author").child("name").child_value();
   m_publisher = node.child("publisher").child("name").child_value();
-  m_date = fromOpdsDate(VALUE("updated"));
+  const std::string dcIssuedDate = VALUE("dc:issued");
+  m_date = dcIssuedDate.empty() ? VALUE("updated") : dcIssuedDate;
+  m_date = fromOpdsDate(m_date);
   m_name = VALUE("name");
   m_flavour = VALUE("flavour");
   m_tags = VALUE("tags");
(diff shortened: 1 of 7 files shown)


veloman-yunkan:
> > All respective client code is switched to use the dc:issued field
> 
> Most of the client use libkiwix to parse the OPDS stream. It would be nice to also update the OPDS parsing part in this pr also. It seems that the only part to change is https://github.com/kiwix/libkiwix/blob/master/src/book.cpp#L164

@mgautierfr Done

src/book.cpp
```diff
@@ -161,7 +161,7 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
   m_language = VALUE("language");
   m_creator = node.child("author").child("name").child_value();
   m_publisher = node.child("publisher").child("name").child_value();
-  m_date = fromOpdsDate(VALUE("updated"));
+  m_date = fromOpdsDate(VALUE("dc:issued"));
```
mgautierfr, line 164: Maybe we could add a fallback to `updated` if `dc::issued` if we have new library reading old opds stream.

veloman-yunkan: Done

mgautierfr:
Agree. Please rebase-fixup before we merge.

mgautierfr:
> Agree. Please rebase-fixup before we merge.

Done

PR #639, opened 2021-12-01
No jquery

mgautierfr:
Fixes #610
Fixes #719
Fixes #369

PR #716, opened 2022-02-20
Iframe-based content viewer

### LINKED ISSUE #394: Use iframe instead of introduce taskbar.
## Rational

A number of issues has been raise recently about the "taskbar".
Some are pretty relative to the warc2zim project : #391
Some of them are more generic and came from other (old) discussion/issues : #392, #322, https://github.com/kiwix/kiwix-tools/issues/17, (https://github.com/kiwix/kiwix-tools/issues/59)

I personally always have been a bit disturbed by the fact we insert UI/chrome content in the main content itself.
- On a performance side : kiwix-serve has to modify content (and so, more memory allocation)
- On the UI side, mixing two kind of content (especially when one of them is not under our control) can also merge the CSS rules, element positioning... (https://github.com/kiwix/kiwix-tools/issues/308, https://github.com/kiwix/kiwix-tools/issues/313, https://github.com/kiwix/kiwix-tools/issues/209)

All this make me think that maybe introduce thing not related to the content in the content is not the good idea. As the issue title imply, using a iframe inside a UI to display the (raw) content is probably better than changing the content to insert UI.

## The change

I've already worked on that on ideascube side (https://framagit.org/ideascube/ideascube/-/merge_requests/590) but never finish this.

The main idea is to use a "proxy page" (a viewer) to display the article content in the iframe.
The viewer uses js script to keep track of the iframe url and so the navigation is not broken (page title, history, ...)

I haven't think totally on the url organization (that's also why there is this issue) but we could have :

- `https://example.com/zim_name/foo.html` : kiwix-serve return the viewer for this. The js viewer parse the url and set the iframe to `https:/example.com/raw?content=zim_name&path=foo.html` (return the raw article without modification). It would not change the article url, but would break the caching system.
- `https://example.com/viewer#content=zim&path=foo.html` would be the viewer (browser/server cache efficient as the frament is not send to the server/taken into account in the cache browser). The js viewer parses the frament and call `https://example.com/zim_name/foo.html`.
- Any other combination we can find.

The second has my preference because of the following. Although it change the url (previous url are now raw, they work but there is no topbar anymore).

One avantage of a iframe hover #392 is that it doesn't introduce any change in the zim file itself
(except if zim file provides its own viewer (warc))

### Compatibility with warc base zim :
They could have a metadata telling they provide they own player, kiwix-serve would use the viewer in the zim file instead of the integrated one.
As the random url served by kiwix-serve will be a redirection to `https://example.com/viewer#...`, random would work. Same for the search/suggestion (but it would be to the viewer to provide the ui anyway).

### Customization :
While the vendor proxy could catch the `https://example.com/viewer` request and provide its own, it would break the warc zim files (who need a specific viewer).
Instead, the viewer page could load a wellknown js url. The vendor could catch this request and serve it. It would be to this js to modify the viewer dom (add logo, change colors, home button's url).

### NO JS fallback.
The viewer can have a `noscript` element providing a link to the raw main page. From now, all link in the page would be point to the "raw" url. But, as said @tim-moody we can almost assume that all browser has js/iframe support.

### Drawbacks

The only one I see is that the "basic" url are now raw url.
- It make the url less user friendly (but understandable).
- It change the meaning of the url. It should not be a problem for classic navigation as the kiwix-serve home page would point the viewer instead of the direct url. But already existing link/bookmark will now point to the raw page.

We could mitigate this by having the `https:/example.com/raw?content=zim_name&path=foo.html` to serve the raw content and https://example.com/zim_name/foo.html` redirect to the viewer. But break the nojs browsing. Maybe cookies can help us to track the situation.

### LINKED ISSUE #671: Don't introduce taskbar for non-front articles
When creating a ZIM, HTML Items with `FRONT_ARTICLE=false` hints are considered non-front articles. As such, the taskbar should not be introduced for them.

I'm not sure about how it should be implemented as Hints are create-time only. I suppose it would rely on the index…

This should allow us to stop depending on `text/html;raw=true` mime types.

veloman-yunkan:
Will fix #394
Will fix #671

* Added `/content` endpoint. Book content is served both under `/content/<book>/<path>/<to>/<entry>` and (for backward compatibility) under `/<book>/<path>/<to>/<entry>`.
* Added a new iframe-based content viewer at '/viewer'
   * `/viewer#<book>/<path>/<to>/<entry>` displays in the iframe the content of `<path>/<to>/<entry>` of the book `<book>`.
   * `/viewer#search?<search_query>` displays in the iframe the search results for `<search_query>`. Enhanced book filtering in `<search_query>` is not supported (a single book must be specified via the `content` query parameter).
   * Clicking links in the iframe updates the URL in the address bar (except when the link leads to an external resource).
   * Manually editing the URL in the address bar updates the content in the iframe.
   * History maintenance works.
* Removed support for taskbar injection in the backend.
* PR history contains some hacks that were introduced at the prototype stage when the '/raw' endpoint was used. The hacks have not been reverted.

Questions:
  - How should we handle following an external link in the iframe?
  - Is `/viewer#search?<search_query>` an acceptable URL scheme for search URLs? A subtle problem with it is that it assumes that a book cannot be named `search`. An alternative is `/viewer/search#<search_query>` (or maybe you have other suggestions).


veloman-yunkan:
@kelson42 A prototype version of the iframe-based viewer is ready. I invite you to play with it by building kiwix-serve with this PR included, running it and opening the `/viewer` endpoint in your browser. If you would like to disable the new viewer you will have to clear cookies.

Known issues:

- Some glitches in history tracking
- Errors inside the viewer come with their own taskbar (two kiwix taskbars are displayed)
- Links inside the viewer point to the '/raw' versions of the resources (a very minor issue, IMO)
- Opening of external links was not addressed at all

kelson42:
@veloman-yunkan Great, I will have a look.

kelson42:
@veloman-yunkan I have tested and it looks promising. I have not much to say beside the known issues you have listed:
* The fulltext results links should be modified (to use the viewer)
* Page for SW based ZIM files using "Try Loading HTTPS URL" should be adapted
* 404 page is not implemented at all
* Does the redirects works fine?

@rgaudin @mgautierfr Would be good you could give a try as well but to me it looks promising and the basics seem to work fine (with Wikipedia and LowTechMagazin SW file). We should probably implement what is missing to start to check that everything works fine.

veloman-yunkan:
@kelson42 

> @veloman-yunkan I have tested and it looks promising. I have not much to say beside the known issues you have listed:
> 
>   * The fulltext results links should be modified (to use the viewer)

This was done. I simply slightly messed up when publishing the viewer for public testing. The newly pushed fixup commit should make it work.
   
> 
>   * Page for SW based ZIM files using "Try Loading HTTPS URL" should be adapted

What is a SW-based ZIM file? I have never dealt with such files before.

> 
>   * 404 page is not implemented at all

Can you please try again with the mentioned fix?  If you still have the same issue, please provide more details about it.

> 
>   * Does the redirects works fine?

That was another known issue that I forgot to mention.

kelson42:
Regarding SW based ZIM files. These ZIM files are made by openzim/warc2zim. All the HTML is played by a page using Service-Workers (SW), therefore they work slightly differently from traditional ZIM. Many of them are available at https://download.kiwix.org/zim/zimit/. SW has a lot of security contstraints and basically works only through HTTPS or with `localhost`. Looking twice to the problem, it seems that the string is in the ZIM itself and actually this is not a problem for this PR.

kelson42:
404 somehow works now but:
* Double taskbar
* The URL given in the error message is not the one of the URL bar (I understand why, but might be disturbing... not sure)
image

I think you are on the good way and I could recommend to fix the last glitches so we can start a complete review. I would also recommend to make this feature as an cmd line option (so not based on a cookie) if this is not too complex to deal legacy and new taskbars together. Otherwise, we will probably have to take the risk to just move on with this new iframe based taskbar.

veloman-yunkan:
> I would also recommend to make this feature as an cmd line option (so not based on a cookie) if this is not too complex to deal legacy and new taskbars together. Otherwise, we will probably have to take the risk to just move on with this new iframe based taskbar.

@kelson42 What is the preferred way of deploying the new taskbar? The two extreme scenarios are:

1. Keep for a while the old and new taskbars functional *in the same running instance of kiwix-serve* so that users have the option to switch to the old version in case of any issues in the new version. When all discovered issues are fixed (and the rate of new bug reports vanishes) the old taskbar is obsoleted and disabled. Any traces of the old implementation are gradually eliminated from the code.

    Pros:
   - This approach can be done in multiple small PRs.
   - The new taskbar can be rolled out gradually (randomly enabled for more and more users with the freedom to opt out). Entire user-base of kiwix will participate in the testing.
   - Safer

    Cons:
    - More work.

3. Develop the new taskbar in the cleanest way in a branch (one big PR). Any code supporting the old taskbar will be mercilessly deleted/replaced when it poses barriers to the new implementation. Testing will need to be performed by dedicated testers. When the quality of the new implementation is assured, the PR is merged and the old taskbar is gone once and for all.

    Pros(?):
   - Likely, less total work (but in one big chunk). However this will strongly require that this course of action is carried out quickly, so that we avoid any potential extra work of addressing conflicts with other changes merged into master.
   - All-or-nothing result (if at some point a conceptual problem is discovered that will make us drop the idea completely, the code in master won't be polluted with attempts of implementing the wrong solution)

    Cons:
    - Higher risk (less testers, one big change)

Your recommendation (of enabling the new mode via a command line option of kiwix-serve) is between these two edge cases. What are the justifications for such a compromise?

kelson42:
@veloman-yunkan To me, if the overall overhead od #1 is not over one day, then #1 is probably preferable. Otherwise #2.

veloman-yunkan:
> @veloman-yunkan To me, if the overall overhead of #1 is not over one day, then #1 is probably preferable. Otherwise #2.

Can you please clarify what you mean by overhead of one day? 24 man*hours?

kelson42:
> > @veloman-yunkan To me, if the overall overhead of #1 is not over one day, then #1 is probably preferable. Otherwise #2.
> 
> Can you please clarify what you mean by overhead of one day? 24 man*hours?

No, around 8 hours.

veloman-yunkan:
> > > @veloman-yunkan To me, if the overall overhead of #1 is not over one day, then #1 is probably preferable. Otherwise #2.
> > 
> > 
> > Can you please clarify what you mean by overhead of one day? 24 man*hours?
> 
> No, around 8 hours.

That probably means that we will have to go with option 2. But I think that we better have @mgautierfr's input too.

veloman-yunkan:
Here are some of my thoughts regarding the final result of switching to an iframe-based viewer:

Book resources accessed through the `/{book}/{path}/{to}/{item}` URLs should be returned unmodified (as if served by the `/raw` endpoint). The `/raw/{book}/content` endpoint becomes redundant unless we anticipate that we may still have to apply some transformation to the resources served from the shorter/public URLs). Maybe we can take advantage of this major change and consider introducing a '/content` endpoint. Current hierarchy of kiwix-serve's URLs places all content/books at the top level, which means that one cannot have books with any of the following names:
    - catalog
    - catch
    - random
    - raw
    - search
    - skin
    - suggest
   
Though that is a minor issue but solving it will give us more freedom when in need of adding new endpoints.

src/server/internalServer.cpp
```diff
@@ -630,7 +630,9 @@ std::unique_ptr<Response> InternalServer::handle_search(const RequestContext& re
     renderer.setProtocolPrefix(m_root + "/");
     renderer.setSearchProtocolPrefix(m_root + "/search?");
     renderer.setPageLength(pageLength);
-    auto response = ContentResponse::build(*this, renderer.getHtml(), "text/html; charset=utf-8");
+    const bool rawMode = rawModeWasRequested(request);
```
mgautierfr, line 638: In #731 I introduce the `format=xml` for opensearch results. (without being sure that `format` is the right name).
I wonder if there is something to define in common or not.

mgautierfr:
Lot of things here. 

There is this PR https://github.com/kiwix/libkiwix/pull/639 which remove jquery from the taskbar. I haven't finish it because of time and I'm not sure of some js/css feature implemented.
It would be nice to have this two PR compatibles. (As we somehow rewrite here the frontend using iframe instead of inserting the taskbar, I think my questions in #639 may be far more simpler)

About the plan, I'm in favor of .2, we would need better testing (and there is somehow already conflict between open PRs) but it would greatly simplify the code. We have a lot of code everywhere to serve "raw" content or not depending of the context, changing path and so. Having only one version would simplify this : never insert taskbar (remove the code), (almost) always return raw content... (And I'm speaking of simplification compared from what we already have, not what we would have if we go for .1)


Another point about the search.
This PR for the `/search?content=foo&pattern=stuff` to `/viewer?search&content=foo&pattern=stuff`
There is no technical reason to use a iframe as we don't render zim content here (and the main purpose of iframe is to separe chrome from content).
So we can keep the initial endpoint. (And if we go with .2 it should not be to complex to implement).
We may (but in another PR) make the search page a "full feature" page which js doing request to the backend and using the xml output (once #639 is finished and merge) to render the search result.


> Current hierarchy of kiwix-serve's URLs places all content/books at the top level, which means that one cannot have books with any  [....]

Totally agree with you. We should have a better "REST" API and stop serving content using the `/<bookName>` endpoint.

veloman-yunkan:
> About the plan, I'm in favor of .2, we would need better testing (and there is somehow already conflict between open PRs) but it would greatly simplify the code. We have a lot of code everywhere to serve "raw" content or not depending of the context, changing path and so. Having only one version would simplify this : never insert taskbar (remove the code), (almost) always return raw content... (And I'm speaking of simplification compared from what we already have, not what we would have if we go for .1)

To justify taking the approach no. 2, we must properly prioritize the work on the iframe-based taskbar. For example, in the refactoring portion of #679 (currently spun off as #732) a noticeable effort has been put into handling the taskbar. Now it is quite disappointing to acknowledge that all that effort was kind of in vain (yet it shouldn't prevent us from doing the right thing).

@kelson42 @mgautierfr Can we come up with a short-term plan for our work in libkiwix?

kelson42:
@veloman-yunkan @mgautierfr It is indeed really unfortunate if work is done and then thrown away. We don't have the resources here to allow us this kind of waste. I'm really thankful to any initiative to avoid this or improve the overall code througput. But, I have no strong opinion or plan in which orders things/tickets should be implemented __within a milestone__. To a large extend I trust you (the developers). I also trust you to talk to each other and find the optimal path to do things.

On my side, when I see a few tickets have been achieved and the current DOING pipe is a bit empty, I use to bring your attention on this or that ticket. Usually these tickets are urgent (incident) tickets or tickets in the current milestone. @veloman-yunkan Beside this, your work since a year is strongly tight to the "Better Library" project https://github.com/orgs/kiwix/projects/8.

On Friday, we had to release `10.1.0` quickly because of the XSS regression, but I have taken all what was left in the milestone and created a new https://github.com/kiwix/libkiwix/milestone/5 `10.2.0` milestone. These are the tickets which should be fixed in priority. Obviously I always have an open hear to your inputs if something could/should be changed in the milestone.

mgautierfr:
Indeed it is a pity that unnecessary work has been done. When you start working on internationalization, I did plan that you would change so much things.
The switch to a iframe is issue open for a long time. I know about it but it is sadly normal that you miss it when you implement internationalization.

I see three solutions to avoid (or minimize) such problem:
- Spending more time (mainly @kelson42 and me) on task specification, the scope of those modification and issue that can be related.
- Developers (@veloman-yunkan, me and others) raising potential issue the sooner. Some of them are obvious to identify, some other less. Changing a lot of files and the global design of a system is a such of potential issue (I'm not saying that such modifications are not welcomed, I like what you've done in #732)
- We should have a public roadmap describing what we want to do and in which order. We could all refer to it when implementing something to check if we will make some change in the future in conflict with what we are doing. We would still have to discuss what to do but it would help us for the two first points. This roadmap could be in the git repository, it would allow everyone to see the changes and comment on them.

kelson42:
@veloman-yunkan If I follow things properly, it’s time now to rebase and complete this PR right?

kelson42:
> > @veloman-yunkan If I follow things properly, it’s time now to rebase and complete this PR right?
> 
> If we go with plan 2, we must wait till there are no concurrent big changes going on in libkiwix at the same time. The search enhancement (#729 and friends) is one such big chunk of work. Let's compile a full list of such planned changes and decide an optimal implementation order for them.

#729 and friends are part of [10.2.0 milestone (to be released ASAP)](https://github.com/kiwix/libkiwix/milestone/6) and what we talk here is part of next [10.3.0](https://github.com/kiwix/libkiwix/milestone/7). I can only recommend to have a look to the whole [10.3.0](https://github.com/kiwix/libkiwix/milestone/7) milestone to make you an opinion. To me, and AFAIK, bringing the iframe is a fondamental work to do to allow a proper cache system. Therefore things should be done in this order. What runs as well, is the removal of JQuery which is almost over, you have here to agree with @mgautierfr when to bring this. All the rest of what is in 10.3.0 milestone is less prioritary and could/might be done later. #479 and #754 might need an early architecture discussion to avoid later bad suprises.

src/server/internalServer.cpp
```diff
@@ -107,6 +108,13 @@ unsigned int getCacheLength(const char* name, unsigned int defaultVal) {
 
   return defaultVal;
 }
+
+struct CustomizedResourceData
+{
+  std::string mimeType;
+  std::string resourceFilePath;
+};
```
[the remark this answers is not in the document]
veloman-yunkan: I extracted this commit into a small PR (#779) of its own. The new functionality in its current form doesn't fully take over the old solution that's why I didn't remove the old solution from `kiwix-compile-resources` in that PR. If we decide to enhance #779 and simplify `kiwix-compile-resources` instead, then we should also try to do it in a way that eliminates the need for #607 & kiwix/kiwix-tools#477 (`--customIndexTemplate` becoming a special case of a more general solution).

kelson42:
@veloman-yunkan Any news? I'm worried to not have seen things pushed forward since two weeks.

veloman-yunkan:
> @veloman-yunkan Any news? I'm worried to not have seen things pushed forward since two weeks.

I am back to this PR. Will try un-WIP it over the weekend.

veloman-yunkan:
Regarding the [proposed API revamping](https://github.com/kiwix/libkiwix/pull/716#issuecomment-1076430466) and removal of the old viewer if we go with plan 2 and a major release of libkiwix - how do we deal with our [users' attitude on this](https://github.com/kiwix/libkiwix/issues/394#issuecomment-665209028)?

kelson42:
@veloman-yunkan There is no hard constraint to keep current URL schema. But, If this has to happen, this has to be discussed first. As far as I understand, /{book}/{path}/{to}/{item} like URLs will/should deliver the Iframe HTML (with correspondong proper /raw content in the Iframe). This is IMO looks better than an URL like /viewer?... and fit to what already have.

Removal of the book names aliases (without dates) might be possible. To be clarified with @rgaudin.

If there is anyhow a conflict between API URLs and user genereated URLs, then the API should have the priority and a warning should be displayed at start of kiwix-serve.

For the rest I believe no additonal API end-point should be necessary and from a human user perspective it should be very similar to what we have already.

There is already an option to run kiwix-serve without taskbar. Such a feature should be still possible. I'm in favour of keeping this command line option but in worse case this shoukd be possible with https://github.com/kiwix/kiwix-tools/issues/17.

BTW, I expect the implementation of this ticket to fix #671 too.

@veloman-yunkan Hope this comment helps you. AFAIK the conceptional/architectural work is mostly completed here. Therefore it is important to me we move forward quickly to a first fully working PR so we can start the reviewing/polishing work. Let us know ASAP if you see important things which still need clarifications.

src/server/internalServer.cpp
```diff
@@ -916,9 +919,9 @@ std::unique_ptr<Response> InternalServer::handle_raw(const RequestContext& reque
       auto item = archive->getMetadataItem(itemPath);
       return ItemResponse::build(*this, request, item, /*raw=*/true);
     } else {
-      auto entry = archive->getEntryByPath(itemPath);
-      if (entry.isRedirect()) {
-        return build_redirect(bookName, entry.getItem(true));
+      auto entry = getEntryFromPath(*archive, itemPath);
+      if (entry.isRedirect() || itemPath.empty()) {
+        return build_redirect(bookName, entry.getItem(true), /*raw=*/true);
```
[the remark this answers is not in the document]
veloman-yunkan: This change can be undone, provided that the support for old taskbar is dropped.

src/server/internalServer.cpp
veloman-yunkan: If we drop the old taskbar, I would prefer to undo the `?mode=raw` hacks for the `/random` and `/search` endpoints

mgautierfr:
Nice PR. Nothing to say on the code itself.

> Added /content endpoint. Book content is served both under /content/<book>/<path>/<to>/<entry> and (for backward compatibility) under /<book>/<path>/<to>/<entry>.

Would it be better to have `/book/path/to/entry` redirect to `/content/book/path/to/entry` ? This way we have a clear advertising that now the content has to be served using `/content` endpoint

> /viewer#search?<search_query> displays in the iframe the search results for <search_query>. Enhanced book filtering in <search_query> is not supported (a single book must be specified via the content query parameter).
> Is /viewer#search?<search_query> an acceptable URL scheme for search URLs? A subtle problem with it is that it assumes that a book cannot be named search. An alternative is /viewer/search#<search_query> (or maybe you have other suggestions).

It would be better to keep the `search` endpoint to do the search. However, it is the simpler solution to have a taskbar on search feature now that we have removed the taskbar insertion in the backend
(and avoid taskbar duplication in the frontend).
This may be fixed using svelte (see below)

> Clicking links in the iframe updates the URL in the address bar (except when the link leads to an external resource).
> How should we handle following an external link in the iframe?

We already have a feature to catch external link and display a message to the user before opening it
(https://github.com/kiwix/libkiwix/blob/master/static/skin/block_external.js redirect external link to https://github.com/kiwix/libkiwix/blob/master/static/templates/captured_external.html)
We should keep the functionnality, but maybe move in the viewer

> Removed support for taskbar injection in the backend.

This is a code we are happy and a bit sad to see go away.

> Robust maintenance of history

I suggest to do this when (if) we move to svelte (see below)

> The suggestions drop-down list doesn't disappear when search is performed
  for the entered text (rather than an item from the list is selected).

May be fixed by https://github.com/kiwix/libkiwix/pull/639 which remove jquery drop-down version with `autoComplete` library


> - Handle the case when book filtering results in a single book match
>  // XXX: Now this has to be handled by the iframe-based viewer which
      // XXX: has to resolve if the book selection resulted in a single book.
      /*

Indeed. Even more, I would say that it should depend of the context and only the search client know it:
- If user ask for a search in a single zim, search result (js client) must include a taskbar.
- If user ask for a multizim search, the search result must not include a taskbar (or include a extended version of it to select zim files). Even if the zim filter results in only one zim
- Searches issued from a external client must managed their ui themselves


---

The more I think about it, the more I think we should use svelte (or any high level js framework, but I like svelte) in the front end.
There is a lot of things that are shared among different part of our ui:
- History management
- Taskbar
- IFrame management
- Book cards (in the home page)
- Fulltext search

Using svelte component/store, we would be able to compose everything as needed and avoid code duplication.
Even more, as we will introduce a widget feature (#585) which will compose all the features depending of the user context.

If there is no opposition, I would be in favor of merging this PR soon and start working on #549 just after.
There is a lot of issue related to the frontend (mainly handle by @juuz0 for now), moving to svelte the sooner will avoid unnecessary work (let's not do the same mistake of working taskbar handling to remove taskbar)

veloman-yunkan:
> > Added /content endpoint. Book content is served both under /content//// and (for backward compatibility) under ////.
> 
> Would it be better to have `/book/path/to/entry` redirect to `/content/book/path/to/entry` ? This way we have a clear advertising that now the content has to be served using `/content` endpoint

Done

> > Clicking links in the iframe updates the URL in the address bar (except when the link leads to an external resource).
> > How should we handle following an external link in the iframe?
> 
> We already have a feature to catch external link and display a message to the user before opening it (https://github.com/kiwix/libkiwix/blob/master/static/skin/block_external.js redirect external link to https://github.com/kiwix/libkiwix/blob/master/static/templates/captured_external.html) We should keep the functionnality, but maybe move in the viewer

I will address external link blocking next. The challenge is in intercepting URL change events in the iframe. Currently I do that only *after* the URL has changed and the page was loaded (via the `onload` event). Also note that if the origin of the document inside iframe is not the same as the origin of the parent document then it is impossible to find out the URL of the iframe content. I guess I will have to use a service worker.
 
> > Robust maintenance of history
> 
> I suggest to do this when (if) we move to svelte (see below)

This is leftover from the early versions. I don't know of any issue with history maintenance due to the viewer (but there is one known issue in the welcome page).

> > The suggestions drop-down list doesn't disappear when search is performed
> > for the entered text (rather than an item from the list is selected).
> 
> May be fixed by #639 which remove jquery drop-down version with `autoComplete` library

I simply didn't even try to fix it because of its low severity.

mgautierfr:
> I will address external link blocking next.

I agree as long as we fix it before doing a release

> The challenge is in intercepting URL change events in the iframe. Currently I do that only after the URL has changed and the page was loaded (via the onload event). Also note that if the origin of the document inside iframe is not the same as the origin of the parent document then it is impossible to find out the URL of the iframe content. I guess I will have to use a service worker.

We should avoid service worker:
- It will conflict with WARC zim file which have their own service worker
- It will force our use to use https.

As the iframe and the parent document have the same origin, it would be possible to access to the document and inject js to catch the url change. Either we open the link with target attribute to `_blank` or `_top`. Or we send a message to the parent with [Window.postMessage()](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) and handle it there.

kelson42:
@veloman-yunkan Playing with the branch and here a few things which don't work IMO:
* The content of the search-box does not change, even if we click on links. IMO, it should work like before (emptied when new page loaded)
* `--nolibrarybutton` option is not respected. Using it has no impact.
* `--blockexternal` option seems always activated (already reported)
* `--nosearchbar` is not respected either.
* Probably a rigourous check should be run for all options of kiwix-serve
* There is a problem around the cache mgmt. CTRL+SHIFT+R seems to not apply to the content of the iframe

veloman-yunkan:
> @veloman-yunkan Playing with the branch and here a few things which don't work IMO:
> 
>   * The content of the search-box does not change, even if we click on links. IMO, it should work like before (emptied when new page loaded)

Fixed

>   * `--nolibrarybutton` option is not respected. Using it has no impact.

Fixed

>   * `--blockexternal` option seems always activated (already reported)

Fixed

>   * `--nosearchbar` is not respected either.

[Fixed partially](https://github.com/kiwix/libkiwix/pull/716/commits/674f7dc7f9da209ab8d08010fc99dce02d08a03d). BTW, isn't it time to rename the `--nosearchbar` option?
 
>   * Probably a rigourous check should be run for all options of kiwix-serve

Do you mean automatic testing?

>   * There is a problem around the cache mgmt. CTRL+SHIFT+R seems to not apply to the content of the iframe

Can you describe a scenario for reproducing this?

kelson42:
> > * `--nosearchbar` is not respected either.
> 
> [Fixed partially](https://github.com/kiwix/libkiwix/pull/716/commits/674f7dc7f9da209ab8d08010fc99dce02d08a03d). BTW, isn't it time to rename the `--nosearchbar` option?

Why not. What is the problem, do you have an alternative in mind?

> > * Probably a rigourous check should be run for all options of kiwix-serve
> 
> Do you mean automatic testing?

No, I mean manually, to secure we don't introduce a regression.

> > * There is a problem around the cache mgmt. CTRL+SHIFT+R seems to not apply to the content of the iframe
> 
> Can you describe a scenario for reproducing this?

If I remember properly, load an article of a ZIM, then restart the kiwix-serve with an other version of the ZIM with an other version of the article or even no article at all. Then CTRL+SHIFT+R does not show that the article is not present in the new ZIM.

veloman-yunkan:
> > BTW, isn't it time to rename the `--nosearchbar` option?
> 
> Why not. What is the problem, do you have an alternative in mind?

`--nosearchbar` is used to disable the entire taskbar. I guess when taskbar was first introduced it consisted of a searchbar only and the option was name adequate. Now it better be changed to `--notaskbar` or `--notoolbar`. 

> > > * There is a problem around the cache mgmt. CTRL+SHIFT+R seems to not apply to the content of the iframe
> > 
> > 
> > Can you describe a scenario for reproducing this?
> 
> If I remember properly, load an article of a ZIM, then restart the kiwix-serve with an other version of the ZIM with an other version of the article or even no article at all. Then CTRL+SHIFT+R does not show that the article is not present in the new ZIM.

Thanks. I reproduced it but I am not sure if it can be fixed. Looks like the user's instruction to override the cache isn't propagated to the loading of the page in the iframe which is initiated as a separate action by javascript code after the page is loaded in the top window.

mgautierfr (changes requested):
Yet another long review :)

Nice to see so much code being removed in the server side. 
As we have dropped the old taskbar feature, we should also remove the `raw` mode now.
Almost all content (if not ALL content) should be raw now.

There is also the handling of the book names in the search to handle:
- `books.name` instead of `content`
- Multizim search handling

As/When you rebase this PR, you can squash existing fixup commit. No need to keep them.

veloman-yunkan:
The previous push was a rebase with elimination of fixup commits. Commits addressing new review comments coming after this comment.

test/server.cpp
```diff
@@ -1013,15 +961,42 @@ TEST_F(ServerTest, RandomPageRedirectsToAnExistingArticle)
...
+  {
+  auto g = zfs1_->GET("/ROOT/zimfile/A/index");
+  ASSERT_EQ(302, g->status);
+  ASSERT_TRUE(g->has_header("Location"));
   ASSERT_EQ("/ROOT/content/zimfile/A/index", g->get_header_value("Location"));
   }
 
+  {
+  // XXX: Should this be a redirect or we better return an error immediately
+  // XXX: when the target URL is invalid?
```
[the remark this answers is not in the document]
veloman-yunkan: > The error is because sever may not contain the requested book.

I wrote that comment because of a concern about a different case - when the top-level component is **not** intended to designate a book. Then the text of the 404 Not Found page can be somewhat misleading. For example:

- usage of an obsolete endpoint (e.g. `/meta`, converted to `/content/meta` in the 404 error)
- a typo in the manually typed API endpoint (e.g. `/ramdon`, converted to `/content/ramdon` in the 404 error)
- usage of a non-empty root location for a kiwix-server started without the `--urlRootLocation` option (`/ROOT/random` is converted to `/content/ROOT/random` in the 404 error)

Anyway, I've removed that comment for now.

static/templates/catalog_v2_entry.xml
```diff
@@ -18,7 +18,7 @@
     {{#icons}}<link rel="http://opds-spec.org/image/thumbnail"
           href="{{root}}/catalog/v2/illustration/{{{content_id}}}/?size={{icon_size}}"
           type="{{icon_mimetype}};width={{icon_size}};height={{icon_size}};scale=1"/>
-    {{/icons}}<link type="text/html" href="{{root}}/{{{content_id}}}" />
+    {{/icons}}<link type="text/html" href="{{root}}/content/{{{content_id}}}" />
```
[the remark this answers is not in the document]
veloman-yunkan: I am pretty much sure that links in the OPDS stream must point to the content rather than to some viewer showing that content.

veloman-yunkan:
> As we have dropped the old taskbar feature, we should also remove the `raw` mode now.
> Almost all content (if not ALL content) should be raw now.

Done

> There is also the handling of the book names in the search to handle:
> 
>     * `books.name` instead of `content`

Done

> 
>     * Multizim search handling

Do you mean adding support for multizim searches via the viewer UI? That's a new feature that is not present in the old taskbar either.

kelson42:
> * Multizim search handling
>> Do you mean adding support for multizim searches via the viewer UI? That's a new feature that is not present in the old taskbar either.

I'm not in favour of building anything new (feature) regarding this. We should open anew ticket.

veloman-yunkan:
@kelson42 I'll do that after @mgautierfr looks at the changes since his last review.

mgautierfr:
> - Multizim search handling
Do you mean adding support for multizim searches via the viewer UI? That's a new feature that is not present in the old taskbar either.

No, we don't have to add multizim searches via the UI. I was speaking about handling the case of a multizim search initiated by a direct request made by someone using the opensearch api. But saying that, I realize that it will probably use the "raw" search api and never the search through the viewer.
So I think we can assume we will not have to handle thing (at least in a near future) and we are good on this point.

static/templates/index.html
```diff
@@ -41,6 +41,7 @@
     </style>
     <script src="{{root}}/skin/isotope.pkgd.min.js?KIWIXCACHEID" defer></script>
     <script src="{{root}}/skin/iso6391To3.js?KIWIXCACHEID"></script>
+    <script>const enableViewer = {{enable_viewer}};</script>
```
mgautierfr, line 44: Why not using the `viewer_setting.js` here ? (which could be renamed to `setting.js`)
It would be a central place indicating the configuration of the server.

static/templates/viewer_settings.js
```diff
@@ -0,0 +1 @@
+const linkBlockingEnabled = {{enable_link_blocking}};
```
mgautierfr, line 1: Wouldn't be better to move this in a `setting` object ?
It would avoid introduction of magic global variables which are always to know from where they come from few months later.

And would it worth to move this in json ?

mgautierfr (commented):
I realize a important point.
`--nosearchbar` and `--no-librarybutton` options were introduced to allow vendor of kiwix-serve (iiab, olip, ...) to integrate it in there UI.
This was because we are introducing the UI in the content, and so we need a way to reduce what we introduce in the content.

But with this PR, we follow a different path: The content is not decorated and we insert it inside a "player" (the home page, the viewer now and probably in the future a js based search view). So probably those options became somehow obsolete (but don't remove them please)

And this is even more emphasized with the introduction of Widgets where we provide a "specific part only viewer". It will provide three ways to access content in kiwix-serve:
- As raw content using `/content` endpoint
- With a partial viewer with `/widget` endpoint
- A full feature viewer with `/`, `/viewer`, `/search` endpoints

I think we need to take a step back in think about the big pictures and provide a coherent way to integrate kiwix-serve in different systems. (HowTo guide, what endpoints we publicly support, which options we keep, ...)

(Not in this PR, but we have to think about that before doing the next release, at least enough to know where we are going)

static/templates/index.html
veloman-yunkan: This is safer because it doesn't depend on the order of loading of `viewer_settings.js` and `index.js` (in which `enableViewer` is used).

static/templates/viewer_settings.js
veloman-yunkan: > Wouldn't be better to move this in a `setting` object ?

Indeed. Done.

> And would it worth to move this in json ?

I can see disadvantages (will have to load the json with a `fetch()` vs the browser loading the .js file automatically) but no advantages.

static/templates/index.html
mgautierfr: Ok

static/templates/viewer_settings.js
mgautierfr: > will have to load the json with a fetch() vs the browser loading the .js file automatically)

Indeed. If we need a full API for a js application loading the setting, we will be able to do it when needed

mgautierfr:
How it behaves with warc2zim zim files ? The embedded player is expecting some specific url organization. I suppose it should still work (simply embedded in a iframe) but we have to test it.

veloman-yunkan:
> How it behave with warc2zim zim files ? The embedded player is expecting some specific url organization. I suppose it should still work (simply embedded in a iframe) but we have to test it.

Not only did I not test this PR on warc2zim files, but I have never dealt with warc2zim files either. Still, wouldn't the desired behaviour for the viewer be to disable itself if the ZIM file is known to come with its own viewer?

veloman-yunkan:
@mgautierfr Regarding the need to resolve the conflict introduced by #796 - is it OK now to rebase the PR now? What clean up actions would you suggest during that rebase?

mgautierfr:
> Not only did I not test this PR on warc2zim files, but I have never dealt with warc2zim files either. Still, wouldn't the desired behaviour for the viewer be to disable itself if the ZIM file is known to come with its own viewer?

This is probably a good behavior but I've not play a lot with warc2zim files neither. @kelson42 may have a better idea.
zim with service worker can be found with : http://library.kiwix.org/catalog/search?tag=_sw:yes

> is it OK now to rebase the PR now?

Yes

>  What clean up actions would you suggest during that rebase?

I would say remove all the `[HACK]...` commits and put directly the final solution. And remove commits which use `kiwix_serve_customized_resources`. But it may be too complex to rebase everything. Do your best.

veloman-yunkan:
> > What clean up actions would you suggest during that rebase?
> 
> I would say remove all the `[HACK]...` commits and put directly the final solution. And remove commits which use `kiwix_serve_customized_resources`. But it may be too complex to rebase everything. Do your best.

I extracted introduction of the `/content` endpoint into a [separate PR](https://github.com/kiwix/libkiwix/pull/806). Then this PR can be rewritten on top of it in a much cleaner way.

kelson42:
After discussing with @mgautierfr, it seems that we all agree on the principle and the code of this PR, but there is work do to in the git revision log. @veloman-yunkan Do you agree? Is the necessary work left to get final approval clear?

veloman-yunkan:
> After discussing with @mgautierfr, it seems that we all agree on the principle and the code of this PR, but there is work do to in the git revision log. @veloman-yunkan Do you agree? Is the necessary work left to get final approval clear?

@kelson42 I am not aware of any concerns regarding the git revision log. If you were referring to the history that was shown on this page - that was simply a github bug; on the commits tab the revision log was correct. After rebasing on top of the merged #806 the revision log on this page looks fine too.

kelson42:
@mgautierfr Can you please make a new review itetation?

kelson42:
FYI, I have tested with a service worker based ZIM file and it seems to work perfectly.

mgautierfr:
> [Last commit message]
>  `--nosearchbar` option of `kiwix-serve` (despite its misleading name)
was used to disable the entire taskbar. This commit accounts for the
existence of that option only partially:
>
>1. Links to books on the welcome/library page are affected - by default
   books are displayed in the viewer, but in a kiwix-serve instance run
   with --nosearchbar books are loaded in the top window.
>
>2. The `/viewer` endpoint is enabled unconditionally, so if anyone
   enters the viewer URL in the address bar they will see books in the
   viewer.

I somehow disagree with this. External link blocking is inserted by the viewer. If we don't use it and load the content directly on top window, the blocking of external link will be disabled. User may want to have external url blocking without a viewer.

We should load all content in a IFrame and having the viewer displaying itself or not depending of the `--nosearchbar` option.
However this PR is pretty big. I think we can simply add a documentation telling that `--nosearchbar` is disabling all the viewer (and so external link blocking) and fix this in a next PR.

veloman-yunkan:
> I somehow disagree with this. External link blocking is inserted by the viewer. If we don't use it and load the content directly on top window, the blocking of external link will be disabled. User may want to have external url blocking without a viewer.

Good catch!

> We should load all content in a IFrame and having the viewer displaying itself or not depending of the `--nosearchbar` option.
> However this PR is pretty big. I think we can simply add a documentation telling that `--nosearchbar` is disabling all the viewer (and so external link blocking) and fix this in a next PR.

I don't mind fixing it in this PR. @mgautierfr Should I do it?

mgautierfr:
@veloman-yunkan you can do it here if you prefer

veloman-yunkan:
> @veloman-yunkan you can do it here if you prefer

While working on that, I found out that autohiding of the toolbar on narrow-screens was broken. I was able to fix it only partially. Need to figure out how to deal during navigation with the state tracking the scroll status.

mgautierfr:
It seems that now `prependToFirstOccurence` and `appendToFirstOccurence` are not used anymore. We could remove them. (And as prependToFirstOccurence` is not explicitly tested, it would increase our coverage).

`handle_viewer_settings` seems not tested. It would be nice to have a unittest fetching `viewer_settings.js` and check that values are correct.

Else, I think we are good. The issue with narrow-screen can be handled later.

kelson42:
Please open ASAP all the tickets necessary regarding regressions or new problems.

kelson42:
So, do I’m right? Here are the things left to do:
* Remove unused code
* Fix/improve automated testing to pass the CI
* open tickets for the regressions/problems which still need to be fixed

kelson42:
@veloman-yunkan We are eager to move one, but nothing seems to move here although this is the priority of the moment. Something wrong somewhere I should know?

test/server_testing_tools.h
```diff
@@ -54,10 +54,24 @@ class ZimFileServer
   typedef std::shared_ptr<httplib::Response>  Response;
   typedef std::vector<std::string> FilePathCollection;
 
+  enum Options
+  {
+    NO_TASKBAR_NO_LINK_BLOCKING = 0,
+    NO_TASKBAR                  = 0,
```
mgautierfr, line 60: Do we really need this `NO_TASKBAR` when it is the same thing than `NO_TASKBAR_NO_LINK_BLOCKING` ?

test/server_testing_tools.h
```diff
@@ -174,16 +171,7 @@ class ServerTest : public ::testing::Test
     zfs1_.reset(new ZimFileServer(SERVER_PORT, options, ZIMFILES));
   }
 
-  ZimFileServer& taskbarlessZimFileServer()
-  {
-    if ( ! taskbarlessZfs_ ) {
-      taskbarlessZfs_.reset(new ZimFileServer(SERVER_PORT+1, ZimFileServer::NO_TASKBAR, ZIMFILES));
```
mgautierfr: This was the only place where we used `NO_TASKBAR`, so we should remove its definition in the enum.

mgautierfr (changes requested):
Only a small change.

> Please tell me if the issue with out-of-date cacheids in the old commits has to be fixed.

I don't think we need to fix the out-of-date cacheids.
It will break automatic git-bisect but manual bisect will be easily fixable. Let's move on.

test/server_testing_tools.h
veloman-yunkan: Good catch. Fixed by moving the commit that drops `taskbarlessZimFileServer()` before the commit that introduces `ZimFileServer::Options`.

test/server_testing_tools.h
veloman-yunkan: This concern is dropped from the agenda due to `NO_TASKBAR` becoming unused and therefore not in any way justified.

veloman-yunkan:
I think that before merging the code has to go through independent testing by some users.

mgautierfr (approved):

mgautierfr:
> I think that before merging the code has to go through independent testing by some users.

I think we can merge now. With master we have nightlies user can test with. And if there is bug, it can be fixed by another (smaller) PR instead of adding to this long PR. Of course we need to test master before doing the release.

rgaudin:
I am also in favor of merging now and using our regular process. It will make this larger testing a lot easier IMO.

mgautierfr:
Time to merge this huge PR 🎉

kelson42:
@veloman-yunkan @mgautierfr The tickets corresponding to the work left, questionning, regressions have not been opne. please to that immediatly before we forget.

rgaudin:
Should be deployed on dev with nightly tomorrow ; will announce on slack once it is

veloman-yunkan:
> @veloman-yunkan @mgautierfr The tickets corresponding to the work left, questionning, regressions have not been opne. please to that immediatly before we forget.

@kelson42 The only issue that I was aware about is now filed as #820.

PR #717, opened 2022-02-24
Add condition for archive being nullptr

juuz0:
Fixes the crashing of kiwix-desktop on using Fulltext search

### DIFF
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -95,7 +95,7 @@ std::string SearchRenderer::getHtml()
     result.set("resultContentId", mp_nameMapper->getNameForId(s.str()));
     std::shared_ptr<zim::Archive> archive;
     try {
-       result.set("bookTitle", mp_library->getBookById(s.str()).getTitle());
+       result.set("bookTitle", archive ? mp_library->getBookById(s.str()).getTitle() : kainjow::mustache::data(false));
     } catch (const std::out_of_range& e) {}
 
     if (it.getWordCount() >= 0) {


kelson42 (approved):
Works fine now

src/search_renderer.cpp
```diff
@@ -95,7 +95,7 @@ std::string SearchRenderer::getHtml()
     result.set("resultContentId", mp_nameMapper->getNameForId(s.str()));
     std::shared_ptr<zim::Archive> archive;
     try {
-       result.set("bookTitle", mp_library->getBookById(s.str()).getTitle());
+       result.set("bookTitle", archive ? mp_library->getBookById(s.str()).getTitle() : kainjow::mustache::data(false));
```
veloman-yunkan, line 98: This fix simply undoes the functionality introduced by #705. If this fix is correct (but it obviously isn't) it can be simplified to

```
result.set("bookTitle", kainjow::mustache::data(false));
```
since `archive` is always `nullptr`¹. Then, given how the `bookTitle` parameter is used during mustache template rendering, you could instead have deleted the entire `try/catch` statement, with the same effect.

Yet, once again, that's not how the crash should be fixed.

---
¹ The `archive` variable was introduced in the initial version of PR #705 and I overlooked that it became unused after requested changes; please remove that pointless variable

veloman-yunkan (changes requested):
Eliminating a crash is a necessary but not a sufficient condition for a bugfix. The crash must be eliminated while preserving the expected functionality.

kelson42:
@veloman-yunkan I'm concerned about the kiwix-desktop crash that this PR attempts to fix and by #705 amyway because:
* Cutting edge (with cutting edge libkiwix) Kiwix Desktop fulltext search if 100% broken
* #705 breaks the API/ABI of `SearchRenderer` and as such should lead to a new major release of libkiwix... just after the release of libkiwix `10.0.0` (so unfortunate IMO). I know that @mgautierfr does not have reacted to your invitation about this problem in #705).
* We are a few days before releasing Kiwix Desktop 2.2.0 (so not a good time to deal with such problems)
*  The way of fixing the problem in a satisfying way seems not straight forward (at least not for me and I guess not for @juuz0 as well).

For all these reasons, I would like  you to handle it please in priority.

With my little understanding here are my (hopefully pertinent) thoughts:
* I don't think we should break the API now, so I would recommend (even if this is only temporary - so put a deprecated warning) to keep the old `SearchRenderer` constructor and build a new one with the new feature).
* I have no problem to have the rendering not including the feature brought by #705 if we still use the old constructor (or anyway not delivering the necessary information to make this feature possible).
* No problem to build new minor version of libkiwix if necessary (so with additional methods in the API)

kelson42:
Closed in favou of #705. @juuz0 Thx for your first PR.

PR #718, opened 2022-02-25
Fix search rendered for book title

mgautierfr:
This PR fix issues introduced by #705

- Do not break the API
- Fix crash when user code doesn't provide a library to the rendered.

This supersedes #717

### DIFF
--- a/include/search_renderer.h
+++ b/include/search_renderer.h
@@ -49,10 +49,24 @@ class SearchRenderer
   /**
    * Construct a SearchRenderer from a SearchResultSet.
    *
+   * The constructed version of the SearchRendered will not introduce
+   * the book name for each result. It is better to use the another constructor
+   * with a Library pointer to have a better html page.
+   *
    * @param srs The `SearchResultSet` to render.
    * @param mapper The `NameMapper` to use to do the rendering.
-   * @param library The `Library` to use to look up book details for search
-   *        results
+   * @param start The start offset used for the srs.
+   * @param estimatedResultCount The estimatedResultCount of the whole search
+   */
+  SearchRenderer(zim::SearchResultSet srs, NameMapper* mapper,
+                 unsigned int start, unsigned int estimatedResultCount);
+
+  /**
+   * Construct a SearchRenderer from a SearchResultSet.
+   *
+   * @param srs The `SearchResultSet` to render.
+   * @param mapper The `NameMapper` to use to do the rendering.
+   * @param library The `Library` to use to look up book details for search results.
    * @param start The start offset used for the srs.
    * @param estimatedResultCount The estimatedResultCount of the whole search
    */
--- a/include/search_renderer.h
+++ b/include/search_renderer.h
@@ -49,8 +49,8 @@ class SearchRenderer
   /**
    * Construct a SearchRenderer from a SearchResultSet.
    *
-   * The constructed version of the SearchRendered will not introduce
-   * the book name for each result. It is better to use the another constructor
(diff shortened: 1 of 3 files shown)


kelson42 (approved):
Thanks for the new PR, LGTM

include/search_renderer.h
```diff
@@ -46,13 +46,27 @@ class SearchRenderer
    */
   DEPRECATED SearchRenderer(Searcher* searcher, NameMapper* mapper);
 
+  /**
+   * Construct a SearchRenderer from a SearchResultSet.
+   *
+   * The constructed version of the SearchRendered will not introduce
```
veloman-yunkan, line 52: SearchRendered → SearchRenderer

include/search_renderer.h
```diff
@@ -46,13 +46,27 @@ class SearchRenderer
...
+   * the book name for each result. It is better to use the another constructor
```
veloman-yunkan, line 53: another → other

src/search_renderer.cpp
```diff
@@ -98,9 +98,11 @@ std::string SearchRenderer::getHtml()
     std::string zim_id(it.getZimId());
     result.set("resultContentId", mp_nameMapper->getNameForId(zim_id));
     std::shared_ptr<zim::Archive> archive;
-    try {
+    if (!mp_library) {
+      result.set("bookTitle", kainjow::mustache::data(false));
```
veloman-yunkan, line 101: Wouldn't the same effect be achieved by *not* setting "bookTitle" at all?

juuz0 (changes requested):
I have some inconsistencies with crashing on kiwix-desktop:
These are the files where I experience crashes - 
https://drive.google.com/file/d/1UHx0GiN6c359UVk0aaYzT2wsEV6Z7VxZ/view?usp=sharing
https://drive.google.com/file/d/1ix8fzCywWZRWvPTlsj960dyHQlExkSZk/view?usp=sharing

(If you can't use google drive for any reasons, let me know where it is possible to upload. File names are 
* Arch Wiki [36 MB]
* ZDoom Wiki [7.28 MB]


Steps to reproduce crash:
* Do a full-text search with words `F`, `No`, `Nice` in Arch Wiki, kiwix desktop. It crashes

These crashes don't happen for a libkiwix version from : ec94d9bfd90b463240716d7a39d20e97fc15cdd4

juuz0:
Changing contructor to:
```
SearchRenderer::SearchRenderer(Searcher* searcher, NameMapper* mapper)
    : m_srs(searcher->getSearchResultSet()),
      mp_nameMapper(mapper),
      mp_library(nullptr), <---- this line is added
      protocolPrefix("zim://"),
      searchProtocolPrefix("search://?"),
      estimatedResultCount(searcher->getEstimatedResultCount()),
      resultStart(searcher->getResultStart())
{}
```
fixes it for me

kelson42:
@mgautierfr Can we please quickly move on this PR? This is really annoying to have a broken search engine in Kiwix-Desktop dev version.

src/search_renderer.cpp
mgautierfr: I don't know how mustache behave in case of absence of variable.
So I prefer to add to false. And it make more sense as we use it if the template (at least for testing it)

mgautierfr:
> Changing contructor to: [...]

Fixed. Thanks.

juuz0 (approved):
Works for me now :)

veloman-yunkan (approved):

PR #721, opened 2022-03-01
Use encoded URLs for searchSuggestionHtml 

juuz0:
Previously, the seachURL was not encoded.
This resulted in an XSS vulnerability, a concept of proof is:

start kiwix-serve
visit - `http://192.168.18.1:8081/"><svg onload=alert(1)>`
This would display an alert message.

This encodes the searchURL before passing it to searchSuggestionHtml

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -798,7 +798,7 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
   } catch (const std::out_of_range& e) {}
 
   if (archive == nullptr) {
-    std::string searchURL = m_root+"/search?pattern="+pattern; // Make a full search on the entire library.
+    std::string searchURL = m_root + "/search?pattern=" + kiwix::urlEncode(pattern, true); // Make a full search on the entire library.
     const std::string details = searchSuggestionHTML(searchURL, kiwix::urlDecode(pattern));
 
     return Response::build_404(*this, request.get_full_url(), bookName, "", details);
@@ -831,7 +831,7 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     if (m_verbose.load())
       printf("Failed to find %s\n", urlStr.c_str());
 
-    std::string searchURL = m_root+"/search?content="+bookName+"&pattern="+pattern; // Make a search on this specific book only.
+    std::string searchURL = m_root + "/search?content=" + bookName + "&pattern=" + kiwix::urlEncode(pattern, true); // Make a search on this specific book only.
     const std::string details = searchSuggestionHTML(searchURL, kiwix::urlDecode(pattern));
 
     return Response::build_404(*this, request.get_full_url(), bookName, getArchiveTitle(*archive), details);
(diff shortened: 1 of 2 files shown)


kelson42 (changes requested):
Unit tests are broken and it seems you PR encode `&` URL sepearator and it should not.

src/server/internalServer.cpp
```diff
@@ -741,8 +741,10 @@ std::string get_book_name(const RequestContext& request)
   }
 }
 
-std::string searchSuggestionHTML(const std::string& searchURL, const std::string& pattern)
+std::string searchSuggestionHTML(std::string& searchURL, const std::string& pattern)
 {
+  searchURL = replaceRegex(searchURL, "&lt;", "<");
+  searchURL = replaceRegex(searchURL, "&gt;", ">");
```
mgautierfr, line 747: We should replace `<` and `>` with the url encoding equivalent, not the html encoding.
This is a url and we want it to work for articles with `<`/`>` in it (at least be able to search for them)

mgautierfr (changes requested):
I wonder if this is the right place to secure the url. At least we already have a function to url encode a string (`urlEncode`), we should use it.

src/server/internalServer.cpp
```diff
@@ -778,7 +778,7 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
   } catch (const std::out_of_range& e) {}
 
   if (archive == nullptr) {
-    std::string searchURL = m_root+"/search?pattern="+pattern; // Make a full search on the entire library.
+    std::string searchURL = kiwix::urlEncode(m_root + "/search?pattern=" + pattern); // Make a full search on the entire library.
```
mgautierfr, line 781: There is another flaw here, it is possible to forge a specific url with:
`tld/foo&content=bar`, we will generate a link to `tld/search?pattern=foo&content=bar`.
It is less important that inserting a js in the page but we should fix it anyway.ç
The solution is probably to urlEncode only the pattern with `encodeReserved` at true.

mgautierfr (changes requested):

mgautierfr (approved):

kelson42 (approved):

legoktm:
@juuz0 @kelson42 do we know what version the XSS vulnerability was introduced in?

kelson42:
@legoktm I don't know, but I would guess for an old weakness

juuz0:
https://github.com/kiwix/libkiwix/pull/465
Probably this
the method added into this explicitly defines no escaping `{{{searchURL}}}`

legoktm:
Thanks, so https://github.com/kiwix/libkiwix/commit/a59787002510a52171684487a10562d8d096c421 was only included in 10.0.0 (no released Debian versions are affected, just unstable).

Could we do a 10.0.2 release with just this cherry-picked? I note that even library.kiwix.org is vulnerable to this. Or if 10.1.0 is coming pretty soon then waiting wouldn't be too bad.

And we should also get a CVE ID assigned for this vulnerability, @kelson42 if you haven't gone through this [process](https://cveform.mitre.org/) before I'm happy to help out.

kelson42:
@legoktm Thx for your last comment. For the moment I don’t have a strong opinion on this but understand the rationals. It seems as well your have a clearer opinion on the next release than I do. But, considering this PR to be closed, could you please open a new ticket with this arguments and why you woukd like to request early release of 10.1.0?

PR #724, opened 2022-03-09
Small search improvement.

mgautierfr:
This PR fixes small issues found on search handling on the server.
It is somehow a preparatory work for openSearch feature coming soon.

Fixes #723

PR #727, opened 2022-03-18
Testing of "Fulltext search unavailable" page

veloman-yunkan:
This is the next batch of groundwork for kiwix/kiwix-tools#59. All commits of this PR have been extracted out of #679.


veloman-yunkan:
@mgautierfr Note that this PR as well as some refactoring changes still remaining in #679 interfere with #724. If we can defer the latter a little then it can be rewritten in a nicer way, using an eDSL based on operator-overloading for building error responses that are automatically translated. For example:

```c++
  if (archive == nullptr) {
    const std::string searchURL = m_root+"/search?pattern="+kiwix::urlEncode(pattern, true);
    return HTTP404HtmlResponse(*this, request)
           + urlNotFoundMsg
           + suggestSearchMsg(searchURL, kiwix::urlDecode(pattern))
           + TaskbarInfo(bookName);
  }
```
If #724 is merged before the proposed new way of building error responses, then I will address it in the context of breaking #679 up into smaller PRs.

mgautierfr (approved):

mgautierfr:
Nothing to say on this PR.
It will not be enough to fix complex merge conflict but we we can merge it right now.

PR #729, opened 2022-03-23
Better handling of multi search

mgautierfr:
Based on #724 

- Better cache system
- Be able to specify a list of zim files to search in.
- Correctly protect search from multithread race condition (may be the root cause of #760)
- Maximal number of books on which we do a multizim search can be configured

The filtering of books to use in the search is made using new querystring parameter:
- `books.id` to specify book's id to use (may be provided several times to select several books)
- `books.name` to specify book's name to use (may be provided several times to select several books).
- `content`, same as `books.name`. Keep for compatibility. `content` can be provided only once
- `books.filter.foo` to do a search on the books using the `foo` criteria. Available criterias are the same as to search books in the opds stream

This PR now integrate #730 as both PR must be merge together to have something coherent.

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -70,7 +70,6 @@ struct Library::Impl
 
   Library::Revision m_revision;
   std::map<std::string, Entry> m_books;
-  std::map<std::string, std::shared_ptr<Reader>> m_readers;
   std::map<std::string, std::shared_ptr<zim::Archive>> m_archives;
   std::vector<kiwix::Bookmark> m_bookmarks;
   Xapian::WritableDatabase m_bookDB;
@@ -139,7 +138,7 @@ bool Library::addBook(const Book& book)
   try {
     auto& oldbook = mp_impl->m_books.at(book.getId());
     if ( ! booksReferToTheSameArchive(oldbook, book) ) {
-      dropReader(book.getId());
+      dropCache(book.getId());
     }
     oldbook.update(book); // XXX: This may have no effect if oldbook is readonly
                           // XXX: Then m_bookDB will become out-of-sync with
@@ -173,9 +172,8 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
 }
 
 
-void Library::dropReader(const std::string& id)
+void Library::dropCache(const std::string& id)
 {
-  mp_impl->m_readers.erase(id);
   mp_impl->m_archives.erase(id);
 }
 
@@ -183,7 +181,7 @@ bool Library::removeBookById(const std::string& id)
 {
   std::lock_guard<std::mutex> lock(m_mutex);
   mp_impl->m_bookDB.delete_document("Q" + id);
-  dropReader(id);
+  dropCache(id);
   return mp_impl->m_books.erase(id) == 1;
 }
 
@@ -242,19 +240,12 @@ const Book& Library::getBookByPath(const std::string& path) const
 
 std::shared_ptr<Reader> Library::getReaderById(const std::string& id)
 {
-  try {
-    std::lock_guard<std::mutex> lock(m_mutex);
-    return mp_impl->m_readers.at(id);
-  } catch (std::out_of_range& e) {}
-
-  const auto archive = getArchiveById(id);
-  if ( !archive )
+  auto archive = getArchiveById(id);
+  if(archive) {
+    return std::make_shared<Reader>(archive);
+  } else {
     return nullptr;
-
-  const shared_ptr<Reader> reader(new Reader(archive, true));
-  std::lock_guard<std::mutex> lock(m_mutex);
-  mp_impl->m_readers[id] = reader;
-  return reader;
+  }
 }
 
 std::shared_ptr<zim::Archive> Library::getArchiveById(const std::string& id)
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -66,13 +66,14 @@ struct Library::Impl
   struct Entry : Book
   {
     Library::Revision lastUpdatedRevision = 0;
-
   };
 
   Library::Revision m_revision;
   std::map<std::string, Entry> m_books;
   using ArchiveCache = ConcurrentCache<std::string, std::shared_ptr<zim::Archive>>;
   std::unique_ptr<ArchiveCache> mp_archiveCache;
+  using SearcherCache = ConcurrentCache<std::string, std::shared_ptr<zim::Searcher>>;
+  std::unique_ptr<SearcherCache> mp_searcherCache;
   std::vector<kiwix::Bookmark> m_bookmarks;
   Xapian::WritableDatabase m_bookDB;
 
@@ -87,6 +88,7 @@ struct Library::Impl
 
 Library::Impl::Impl()
   : mp_archiveCache(new ArchiveCache(std::max(getEnvVar<int>("ARCHIVE_CACHE_SIZE", 1), 1))),
+    mp_searcherCache(new SearcherCache(std::max(getEnvVar<int>("SEARCHER_CACHE_SIZE", 1), 1))),
     m_bookDB("", Xapian::DB_BACKEND_INMEMORY)
 {
 }
@@ -156,6 +158,9 @@ bool Library::addBook(const Book& book)
     if (getEnvVar<int>("ARCHIVE_CACHE_SIZE", -1) <= 0) {
       mp_impl->mp_archiveCache->setMaxSize(new_cache_size);
     }
+    if (getEnvVar<int>("SEARCHER_CACHE_SIZE", -1) <= 0) {
+      mp_impl->mp_searcherCache->setMaxSize(new_cache_size);
+    }
     return true;
   }
 }
@@ -182,6 +187,7 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
 void Library::dropCache(const std::string& id)
 {
   mp_impl->mp_archiveCache->drop(id);
+  mp_impl->mp_searcherCache->drop(id);
 }
 
 bool Library::removeBookById(const std::string& id)
@@ -275,7 +281,22 @@ std::shared_ptr<zim::Archive> Library::getArchiveById(const std::string& id)
   } catch (std::invalid_argument&) {
     return nullptr;
   }
+}
 
+std::shared_ptr<zim::Searcher> Library::getSearcherById(const std::string& id)
+{
+  try {
+    return mp_impl->mp_searcherCache->getOrPut(id,
+    [&](){
+      auto archive = getArchiveById(id);
+      if(!archive) {
+        throw std::invalid_argument("");
+      }
+      return std::make_shared<zim::Searcher>(*archive);
+    });
+  } catch (std::invalid_argument&) {
+    return nullptr;
+  }
 }
 
 unsigned int Library::getBookCount(const bool localBooks,
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -108,8 +108,8 @@ struct Library::Impl
 };
 
 Library::Impl::Impl()
-  : mp_archiveCache(new ArchiveCache(std::max(getEnvVar<int>("ARCHIVE_CACHE_SIZE", 1), 1))),
-    mp_searcherCache(new SearcherCache(std::max(getEnvVar<int>("SEARCHER_CACHE_SIZE", 1), 1))),
+  : mp_archiveCache(new ArchiveCache(std::max(getEnvVar<int>("KIWIX_ARCHIVE_CACHE_SIZE", 1), 1))),
+    mp_searcherCache(new SearcherCache(std::max(getEnvVar<int>("KIWIX_SEARCHER_CACHE_SIZE", 1), 1))),
     m_bookDB("", Xapian::DB_BACKEND_INMEMORY)
 {
 }
@@ -176,10 +176,10 @@ bool Library::addBook(const Book& book)
     static_cast<Book&>(newEntry) = book;
     newEntry.lastUpdatedRevision = mp_impl->m_revision;
     size_t new_cache_size = std::ceil(mp_impl->getBookCount(true, true)*0.1);
-    if (getEnvVar<int>("ARCHIVE_CACHE_SIZE", -1) <= 0) {
+    if (getEnvVar<int>("KIWIX_ARCHIVE_CACHE_SIZE", -1) <= 0) {
       mp_impl->mp_archiveCache->setMaxSize(new_cache_size);
     }
-    if (getEnvVar<int>("SEARCHER_CACHE_SIZE", -1) <= 0) {
+    if (getEnvVar<int>("KIWIX_SEARCHER_CACHE_SIZE", -1) <= 0) {
       mp_impl->mp_searcherCache->setMaxSize(new_cache_size);
     }
     return true;
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -59,6 +59,27 @@ bool booksReferToTheSameArchive(const Book& book1, const Book& book2)
       && book1.getPath() == book2.getPath();
 }
 
+template<typename Key, typename Value>
+class MultiKeyCache: public ConcurrentCache<std::set<Key>, Value>
+{
+  public:
+    explicit MultiKeyCache(size_t maxEntries)
+      : ConcurrentCache<std::set<Key>, Value>(maxEntries)
+    {}
+
+    bool drop(const Key& key)
+    {
+      std::unique_lock<std::mutex> l(this->lock_);
+      bool removed = false;
+      for(auto& cache_key: this->impl_.keys()) {
+        if(cache_key.find(key)!=cache_key.end()) {
+          removed |= this->impl_.drop(cache_key);
+        }
+      }
+      return removed;
+    }
+};
+
 } // unnamed namespace
 
 struct Library::Impl
@@ -72,7 +93,7 @@ struct Library::Impl
   std::map<std::string, Entry> m_books;
   using ArchiveCache = ConcurrentCache<std::string, std::shared_ptr<zim::Archive>>;
   std::unique_ptr<ArchiveCache> mp_archiveCache;
-  using SearcherCache = ConcurrentCache<std::string, std::shared_ptr<zim::Searcher>>;
+  using SearcherCache = MultiKeyCache<std::string, std::shared_ptr<zim::Searcher>>;
   std::unique_ptr<SearcherCache> mp_searcherCache;
   std::vector<kiwix::Bookmark> m_bookmarks;
   Xapian::WritableDatabase m_bookDB;
@@ -285,8 +306,9 @@ std::shared_ptr<zim::Archive> Library::getArchiveById(const std::string& id)
 
 std::shared_ptr<zim::Searcher> Library::getSearcherById(const std::string& id)
 {
+  std::set<std::string> ids {id};
   try {
-    return mp_impl->mp_searcherCache->getOrPut(id,
+    return mp_impl->mp_searcherCache->getOrPut(ids,
     [&](){
       auto archive = getArchiveById(id);
       if(!archive) {
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -33,6 +33,7 @@
 #include <pugixml.hpp>
 #include <algorithm>
 #include <set>
+#include <cmath>
 #include <unicode/locid.h>
 #include <xapian.h>
 
@@ -151,6 +152,10 @@ bool Library::addBook(const Book& book)
     auto& newEntry = mp_impl->m_books[book.getId()];
     static_cast<Book&>(newEntry) = book;
     newEntry.lastUpdatedRevision = mp_impl->m_revision;
+    size_t new_cache_size = std::ceil(mp_impl->getBookCount(true, true)*0.1);
+    if (getEnvVar<int>("ARCHIVE_CACHE_SIZE", -1) <= 0) {
+      mp_impl->mp_archiveCache->setMaxSize(new_cache_size);
+    }
     return true;
   }
 }
@@ -184,6 +189,12 @@ bool Library::removeBookById(const std::string& id)
   std::lock_guard<std::mutex> lock(m_mutex);
   mp_impl->m_bookDB.delete_document("Q" + id);
   dropCache(id);
+  // We do not change the cache size here
+  // Most of the time, the book is remove in case of library refresh, it is
+  // often associated with addBook calls (which will properly set the cache size)
+  // Having a too big cache is not a problem here (or it would have been before)
+  // (And setMaxSize doesn't actually reduce the cache size, extra cached items
+  //  will be removed in put or getOrPut).
   return mp_impl->m_books.erase(id) == 1;
 }
 
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -304,17 +304,21 @@ std::shared_ptr<zim::Archive> Library::getArchiveById(const std::string& id)
   }
 }
 
-std::shared_ptr<zim::Searcher> Library::getSearcherById(const std::string& id)
+std::shared_ptr<zim::Searcher> Library::getSearcherByIds(const BookIdSet& ids)
 {
-  std::set<std::string> ids {id};
+  assert(!ids.empty());
   try {
     return mp_impl->mp_searcherCache->getOrPut(ids,
     [&](){
-      auto archive = getArchiveById(id);
-      if(!archive) {
-        throw std::invalid_argument("");
+      std::vector<zim::Archive> archives;
+      for(auto& id:ids) {
+        auto archive = getArchiveById(id);
+        if(!archive) {
+          throw std::invalid_argument("");
+        }
+        archives.push_back(*archive);
       }
-      return std::make_shared<zim::Searcher>(*archive);
+      return std::make_shared<zim::Searcher>(archives);
     });
   } catch (std::invalid_argument&) {
     return nullptr;
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -93,7 +93,7 @@ struct Library::Impl
   std::map<std::string, Entry> m_books;
   using ArchiveCache = ConcurrentCache<std::string, std::shared_ptr<zim::Archive>>;
   std::unique_ptr<ArchiveCache> mp_archiveCache;
-  using SearcherCache = MultiKeyCache<std::string, std::shared_ptr<zim::Searcher>>;
+  using SearcherCache = MultiKeyCache<std::string, std::shared_ptr<ZimSearcher>>;
   std::unique_ptr<SearcherCache> mp_searcherCache;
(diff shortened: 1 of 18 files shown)


veloman-yunkan (commented):
Since @kelson42 requested my review on this WIP PR I started looking at the changes but I soon figured out that I was missing some context. The outcome of my first iteration are a few low value comments for the first couple of commits. It will be much helpful if a high level description of  the use-model and functional enhancement sought by this PR is provided.

src/library.cpp
```diff
@@ -27,10 +27,13 @@
 #include "tools/regexTools.h"
 #include "tools/pathTools.h"
 #include "tools/stringTools.h"
+#include "tools/otherTools.h"
```
veloman-yunkan, line 30: This comment applies to the commit "Use a concurrent_cache to store Archive in library. "

I recommend to split it into three:
- Get rid of `m_readers` in `Library::Impl`
- Switch `Library::Impl::m_archives` to a concurrent cache
- Link the cache size to the book count (unless explicitly configured by the user via the env var).

src/library.cpp
```diff
@@ -181,17 +186,16 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
 }
 
 bool Library::removeBookById(const std::string& id)
 {
   std::lock_guard<std::mutex> lock(m_mutex);
   mp_impl->m_bookDB->delete_document("Q" + id);
-  dropReader(id);
+  dropCache(id);
   return mp_impl->m_books.erase(id) == 1;
 }
```
veloman-yunkan, line 228: Shouldn't the cache size be updated here too? If not, a comment must be added that it is not done intentionally

src/tools/lrucache.h
```diff
@@ -138,12 +138,18 @@ class lru_cache {
     return _cache_items_map.size();
   }
 
+  size_t set_max_size(size_t new_size) {
```
veloman-yunkan, line 141: I know that this class is a hybrid of a snake with a camel, but I wonder if your choice of the style here was deliberate :)

Now getting serious. If the cache's current size exceeds the new value of max size, shouldn't it be truncated immediately? Though, a deeper question is - do we really need a dynamic cache size at all (i.e. is linking the cache size to the actual amount of data a good idea)?

veloman-yunkan (changes requested):
This is only part of the review (I only had time to skim through the first several commits).

src/library.cpp
```diff
@@ -93,7 +94,7 @@ class Library::Impl::BookDB : public Xapian::WritableDatabase
 };
 
 Library::Impl::Impl()
-  : m_bookDB(new BookDB)
```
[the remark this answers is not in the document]
veloman-yunkan: I see that this accidental deletion was undone in a later commit. Yet, the history of the PR better not contain such mistakes.

src/server/internalServer.cpp
```diff
@@ -571,25 +575,22 @@ std::unique_ptr<Response> InternalServer::handle_search(const RequestContext& re
...
+              try {
+                auto bookId = mp_nameMapper->getIdForName(bookName);
+                bookIds.insert(bookId);
+              } catch(const std::out_of_range&) {
+                throw std::invalid_argument("The requested book doesn't exist.");
+              }
+            }
           } else {
             for (auto& bookId: mp_library->filter(kiwix::Filter().local(true).valid(true))) {
               bookIds.insert(bookId);
```
veloman-yunkan: Nesting level has grown too high here. This is a sign for introducing a helper function. IMHO, a private method `std::shared_ptr<zim::Search> Library::createSearch(const SearchInfo&)` is justified.

src/server/internalServer.cpp
```diff
@@ -602,12 +603,20 @@ std::unique_ptr<Response> InternalServer::handle_search(const RequestContext& re
...
+                                m_root + "/skin/search_results.css")
+        + nonParameterizedMessage("no-search-results")
+      );
+      if(searchInfo.bookNames.size() == 1) {
+        auto bookName = *searchInfo.bookNames.begin();
+        auto bookId = mp_nameMapper->getIdForName(bookName);
+        return response + TaskbarInfo(bookName, mp_library->getArchiveById(bookId).get());
+      } else {
+        return response;
+      }
```
veloman-yunkan, line 702: I don't like this way of building responses conditionally. I believe a better way is to introduce `operator +=()`, so that the code looks like below:

```c++
      HTTPErrorHtmlResponse response(*this, request, MHD_HTTP_NOT_FOUND,
                                    "fulltext-search-unavailable",
                                    "404-page-heading",
                                    m_root + "/skin/search_results.css");
      response += nonParameterizedMessage("no-search-results");
      if(searchInfo.bookNames.size() == 1) {
        auto bookName = *searchInfo.bookNames.begin();
        auto bookId = mp_nameMapper->getIdForName(bookName);
        response += TaskbarInfo(bookName, mp_library->getArchiveById(bookId).get());
      }
      return response;
```

src/server/internalServer.cpp
```diff
@@ -132,6 +132,56 @@ std::vector<T> subrange(const std::vector<T>& v, size_t s, size_t n)
...
+    auto name_vec = request.get_arguments(prefix+"name");
+    if (name_vec.empty()) {
+      throw std::invalid_argument("You must provide a value for the name.");
+    }
+    std::set<std::string> bookIds;
+    for(const auto& bookName: name_vec) {
+      try {
+        bookIds.insert(mp_nameMapper->getIdForName(bookName));
+      } catch(const std::out_of_range&) {
+        throw std::invalid_argument("The requested book doesn't exist.");
```
veloman-yunkan: Which book doesn't exist?

src/server/internalServer.cpp
```diff
@@ -132,6 +132,56 @@ std::vector<T> subrange(const std::vector<T>& v, size_t s, size_t n)
 
 } // unnamed namespace
 
+std::set<std::string> InternalServer::selectBooks(const RequestContext& request, const std::string& prefix) const
```
veloman-yunkan, line 190: The meaning of `prefix` (which first appeared in the previous commit) is still mysterious here. Please describe its purpose in the commit message or in a code comment.

src/server/internalServer.cpp
```diff
@@ -192,9 +192,10 @@ SearchInfo::SearchInfo(const std::string& pattern, GeoQuery geoQuery)
     geoQuery(geoQuery)
 {}
 
-SearchInfo::SearchInfo(const RequestContext& request)
+SearchInfo::SearchInfo(const RequestContext& request, const std::set<std::string>& bookIds)
```
veloman-yunkan, line 250: This constructor has became ugly. Parsing of the request better be consolidated in one place. The constructor must be

```c++
SearchInfo(const std::string& pattern, const GeoQuery&, const BookIdSet&)
```

and there must be a method

```c++
SearchInfo InternalServer::getSearchInfo(const RequestContext& ) const;
```

veloman-yunkan (changes requested):
A second set of review comments covering the next few commits.

src/server/internalServer.cpp
veloman-yunkan: I see that a later commit eliminates this issue.

src/server/request_context.h
```diff
@@ -72,6 +72,9 @@ class RequestContext {
         std::istringstream stream(get_argument(name));
         T v;
         stream >> v;
+        if(stream.fail()) {
+          throw std::runtime_error("Invalid argument");
+        }
         return v;
     }
```
veloman-yunkan, line 79: Better reuse `kiwix::extractFromString<T>()` and implement all error handling in the latter.

static/templates/search_result.html
```diff
@@ -137,32 +137,32 @@
...
-              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start={{resultLastPageStart}}&pageLength={{pageLength}}">
+              <a href="{{searchProtocolPrefix}}pattern={{searchPatternEncoded}}{{#contentId}}&content={{.}}{{/contentId}}&start={{start}}&pageLength={{pagination.itemsPerPage}}">
                 ▶
               </a>
             </li>
-          {{/resultLastPageStart}}
+          {{/pagination.lastPage}}
         </ul>
-      {{/hasPages}}
+      {{/pagination.hasPages}}
```
veloman-yunkan, line 165: Doesn't it make sense to simplify this template by removing the special handling of the goto first and last page triangle markers (`◀`, `▶`) but instead including them (conditionally) as labels (with corresponding `start` values) in the regular list of `pagination.pages`?

src/search_renderer.cpp
```diff
@@ -86,6 +86,22 @@ void SearchRenderer::setSearchProtocolPrefix(const std::string& prefix)
...
+  const std::set<std::string>& bookIds
+) {
+  kainjow::mustache::data query;
+  query.set("pattern", kiwix::encodeDiples(pattern));
+  std::ostringstream ss;
+  ss << "?pattern=" << urlEncode(pattern);
+  for (auto& bookId: bookIds) {
+    ss << "&books.id="<<urlEncode(bookId);
+  }
+  query.set("path", ss.str());
```
veloman-yunkan, line 101: `path` isn't a good name for this attribute of the query. What about getting rid of `searchProtocolPrefix` in the mustache template and replacing the stable combination `{{searchProtocolPrefix}}{{query.path}}` with `{{unpaginatedQuery}}`?

src/server/internalServer.cpp
```diff
@@ -167,6 +172,21 @@ ParameterizedMessage nonParameterizedMessage(const std::string& msgId)
 
 } // unnamed namespace
 
+struct Error : public std::runtime_error {
+  Error(ParameterizedMessage message)
+    : std::runtime_error("Invalid_argument"),
+      _message(message)
+  {}
```
veloman-yunkan, line 179: 1. Make `explicit`
2. Pass parameter by const reference
3. Fix the message passed to `std::runtime_error`

src/server/internalServer.cpp
```diff
@@ -167,6 +172,21 @@ ParameterizedMessage nonParameterizedMessage(const std::string& msgId)
 
 } // unnamed namespace
 
+struct Error : public std::runtime_error {
```
veloman-yunkan, line 175: Move the declaration of `struct Error` into the unnamed namespace

static/i18n/en.json
```diff
@@ -4,12 +4,15 @@
 		]
 	},
 	"name":"English",
-	"suggest-full-text-search": "containing '{{{SEARCH_TERMS}}}'..."
-	, "no-such-book": "No such book: {{BOOK_NAME}}"
+	"suggest-full-text-search" : "containing '{{{SEARCH_TERMS}}}'..."
+	, "no-such-book" : "No such book: {{BOOK_NAME}}"
+	, "no-book-found" : "No book found"
```
veloman-yunkan, line 9: When new messages are added, `static/i18n/qqq.json` must be updated too.

veloman-yunkan (changes requested):
This completes the first pass over the entire PR. But I think that with the big picture of the PR now in my head I will make another pass even before any of my comments is addressed.

src/tools/otherTools.h
```diff
@@ -53,6 +56,23 @@ namespace kiwix
...
 
   std::string render_template(const std::string& template_str, kainjow::mustache::data data);
+
+  template<typename T>
+  T getEnvVar(const char* name, const T& defaultValue)
+  {
+    try {
+      const char* envString = std::getenv(name);
+      if (envString == nullptr) {
+        throw std::runtime_error("Environment variable not set");
```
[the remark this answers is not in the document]
mgautierfr: This execption is never thrown to the caller as it is catched by following `catch(...)`
We have to provide a string, so we use it as "documentation", but it is useless to generate one.

src/library.cpp
mgautierfr: Yes, I think it is not needed to update the cache size here. I've added a comment.

static/templates/search_result.html
mgautierfr: Done

src/search_renderer.cpp
mgautierfr: Done

src/library.cpp
```diff
@@ -93,7 +94,7 @@ class Library::Impl::BookDB : public Xapian::WritableDatabase
...
+  : mp_archiveCache(new ArchiveCache(getEnvVar("ARCHIVE_CACHE_SIZE", 1))),
```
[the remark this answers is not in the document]
mgautierfr: Yes, they are now prefixed by `KIWIX_`

src/server/internalServer.cpp
mgautierfr: This is fixed in later commit with the `Error` class

src/server/internalServer.cpp
mgautierfr: Done

src/server/internalServer.cpp
mgautierfr: Addi in commit message

> The prefix will be used to parse a "query to select book" in different context.
    For now we have only one context : selecting books for the catalog search.
    But we will want to select books to do fulltext search on them
    (will be done in later commit)

mgautierfr:
@veloman-yunkan I should have handle all your numerous remarks. Ready for another review pass.

src/server/internalServer.h
```diff
@@ -148,7 +148,9 @@ class InternalServer {
 
     bool etag_not_needed(const RequestContext& r) const;
     ETag get_matching_if_none_match_etag(const RequestContext& request) const;
-    Library::BookIdSet selectBooks(const RequestContext& r, const std::string& prefix) const;
+    Library::
+
+    BookIdSet selectBooks(const RequestContext& r, const std::string& prefix) const;
```
veloman-yunkan, line 153: I guess that this is an accidental change

veloman-yunkan (commented):
Please rebase and fixup this PR so that I can review its clean version. While doing so, please also move the commit "Introduce operator+= to simplify response creation" earlier in the history so that the change depending on it is nicer at birth.

src/server/internalServer.h
mgautierfr: Indeed

veloman-yunkan (commented):
The chosen attempt to reconcile the taken approach to multizim searching with caching contains the following problem.

A multizim searcher object can be arbitrarily large - at the extreme it can include all books in a library. Consider a library with 100 books. For the archive and searcher cache sizes both set to 10, the proposed caching scheme allows  to have 10 multizim searchers, each containing 99 books (with book \#k not included in the set of searcher \#k). As a result

- All 100 `zim::Archive`'s will be opened. During the creation of the first searcher, 99 `zim::Archive`'s will come to life but only 10 of them will reside in the archive cache (the remaining ones being kept alive by the searcher object). When the second searcher is created, 9 or 10 of those cached archives will be reused but the other 90 or 89 will be opened again. The end result will be that each ZIM file is opened \~9 times as independent `zim::Archive` objects.

- Similarly, 990 Xapian database objects will reside in memory, most of them duplicating each other while being independent objects (unless the Xapian architecture uses caching internally and detects attempts to open again an already opened database and reuses the pre-existing object).

Of course the above example is far from being realistic and is intended to illustrate a potential issue via an edge case.

To prevent such a situation we must let the archive cache grow as large as needed to support the searcher cache. Or, on the contrary, creation of a searcher must be rejected if that would result in exceeding the upper limit on the count of simultaneously open archives.

---

I interrupted my second pass as soon the understanding of the above conceptual problem struck me. I believe that we must have an answer to the raised concern before we move on with this PR. Yet, I propose to extract the general purpose code improvement commits as an independent PR and merge it separately. This will facilitate subsequent iterations on the functionality sought by this PR.

kelson42:
@veloman-yunkan Considering that the two blocker PRs are ready and will be merged soon. Would you please make a (final) review od this one!

veloman-yunkan:
> With multizim searcher, we indeed face a problem as the number of archive per searcher can be pretty important.
> 
> I see several solutions:
> 
>   * Do not limit archive cache as you suggest (and as it is done in master)
> 
>   * Estimate the `means of archives per searcher` and set the cache size ratio accordingly
> 
>   * Do not interpret the searcher cache size as the `number of searcher` but as `total number of archives in stored searchers`. I still don't know if the limits is the number of archives (and so we must count n duplication of the same archive as 1) or if it is the open xapian database (and count n duplication as n). It would allow us to have a "lot" a monosearcher but "few" multisearcher in the same time.

I think that the main requirement must be ensuring that the same ZIM file is never opened more than once at the same time, otherwise it will result in additional pressure on RAM by the cluster caches (since the searcher isn't limited to reading the xapian database via a direct reference but also accesses article content during snippet generation). To this end the archive cache system must keep track of all `zim::Archives` that are currently alive. Then we can think how to build the searcher cache on top of the archive cache.

mgautierfr:
On top of rebasing on new master, the last push add four commits:
- The two first limit the number of Archive in a multizimSearch (in the Server only). The limit is configurable (and set to no limit by default). `kiwix-serve` need to be updated to allow such configuration to the user.
- The two last commit allow the cache to keep a weak ref to each shared_ptr added to the cache. So, if a `shared_ptr<Archive>` is alive somewhere (in a Searcher) we will reuse it instead of reopening the `Archive` (even if the `Archive` is dropped from the cache)

mgautierfr:
And last commit, protect search from multithreading race condition.
Without it, we may throw a `Xapian::Error` class because of invalid db format.
As `Xapian::Error` is NOT inheriting from `std::exception` this is probably a fix for https://github.com/kiwix/libkiwix/issues/760
This may also be the cause of `ERR_CONNECTION_REFUSED` in https://github.com/kiwix/libkiwix/issues/769#issuecomment-1124468042

kelson42:
@mgautierfr Sounds all good, is that ready for a new review?

src/tools/concurrent_cache.h
```diff
@@ -95,6 +95,113 @@ class ConcurrentCache
...
+          m_weakStore.add(key, value);
+        } catch (std::exception& e) {
+          drop(key);
+          throw;
+        }
+      }
+    }
+
+    return x.value().get();
+  }
```
veloman-yunkan, line 186: My feeling is that a similar result can be achieved through composition (without needing to specialize `ConcurrentCache` for `shared_ptr`).

1. `ConcurrentCache<Key, std::shared_ptr<Value>> cachedRefs;`
2. `WeakStore<Key, Value> allRefs;`
4. `cachedRefs.getOrPut(key, allRefs.wrap(key, f));`

where `WeakStore<Key, Value>::wrap(const Key& key, F f)` returns an augmented functor that first looks for `key` in `WeakStore`  and calls `f()` only if the key was not found.

src/server/internalServer.cpp
```diff
@@ -195,6 +195,17 @@ struct Error : public std::runtime_error {
...
+Library::BookIdSet limitBooksNumber(Library::BookIdSet&& bookSet, unsigned int limit) {
+  if (bookSet.size() <= limit) {
+    return std::move(bookSet);
+  }
+  auto it = bookSet.begin();
+  while(limit--) {
+    it++;
+  }
+  bookSet.erase(it, bookSet.cend());
+  return std::move(bookSet);
```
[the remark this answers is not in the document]
mgautierfr: It depends of the user:
- If the user is the maintainer of the service which explicitly set a limit in the number of books per search. We are respecting they.
- If the user is the one using such service, yes we are neglecting they, but I would say that it is the responsibility of the maintainer.

We could ask what would be the best `limit` books to keep but it is another question and I don't want to handle in this PR (we don't really have criteria for book relevance order)

src/tools/concurrent_cache.h
mgautierfr: Indeed.

The advantage of specialization is that it is a facto replacement for all ConcurrentCache storing shared_ptr.
The composition means it is the responsibility of the caller to do the composition.

I don't see why we may have a cache system that would not reuse a existing value.
I even think we should backport this in libzim. It could be use for the cluster and dirent cache. If a item (which keep a shared_ptr to dirent) exists somewhere. We should reuse the dirent even if it is not "in" the cache.

src/server/internalServer.cpp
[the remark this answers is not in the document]
mgautierfr: After discussion with @kelson42, it is indeed better to return a 40x error instead of silently remove books from user request.

veloman-yunkan (commented):
Reminding that this PR still lacks unit-tests for multizim search.

src/server/internalServer.cpp
```diff
@@ -95,35 +95,174 @@ inline std::string normalizeRootUrl(std::string rootUrl)
...
+        bookIds.insert(mp_nameMapper->getIdForName(bookName));
+      } catch(const std::out_of_range&) {
+        throw Error(noSuchBookErrorMsg(bookName));
+      }
+    }
+    return limitBooksNumber(std::move(bookIds), m_multizimSearchLimit);
+  } catch(const std::out_of_range&) {}
+
+  // Check for filtering
+  Filter filter = get_search_filter(request, prefix+"query.");
```
veloman-yunkan, line 252: I think that  the word "filter" (instead of "query") is a better fit for the URL query parameter (so that the URL reads `.../search?books.filter.lang=eng&pattern=...`)

src/tools/concurrent_cache.h
veloman-yunkan: > I even think we should backport this in libzim. It could be use for the cluster and dirent cache. If a item (which keep a shared_ptr to dirent) exists somewhere. We should reuse the dirent even if it is not "in" the cache.

It definitely makes sense for the cluster cache, but I don't think that it is a good idea for the dirent cache, since the count of dirents can be quite large and in a (quite realistic) worst case scenario all of them can end up being "tracked" by the `WeakStore`.

On the other hand, the dirent cache is based on the lower level `lru_cache`. If this optimization for `std::shared_ptr` can be justified to be always applicable then it probably must be pushed down to `lru_cache`. Otherwise, it can still be implemented in (or on top of) `lru_cache` as an optional feature.

src/server/internalServer.cpp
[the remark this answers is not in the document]
mgautierfr: Returning a error is made with a fixup commit in the last push

src/server/internalServer.cpp
```diff
@@ -288,13 +293,14 @@ SearchInfo InternalServer::getSearchInfo(const RequestContext& request, const st
     throw Error(nonParameterizedMessage("no-query"));
   }
 
-  return SearchInfo(pattern, geoQuery, bookIds);
+  return SearchInfo(pattern, geoQuery, bookIds.second, bookIds.first);
 }
 
-SearchInfo::SearchInfo(const std::string& pattern, GeoQuery geoQuery, const Library::BookIdSet& bookIds)
+SearchInfo::SearchInfo(const std::string& pattern, GeoQuery geoQuery, const Library::BookIdSet& bookIds, const std::string& searchQuery)
```
veloman-yunkan, line 307: `searchQuery` is misleading in this context. `bookFilter` or `bookFilterQuery` would describe the meaning of this variable better.

mgautierfr: Indeed. Done in last fixup commit

src/server/internalServer.cpp
```diff
@@ -95,35 +95,200 @@ inline std::string normalizeRootUrl(std::string rootUrl)
...
+  }
+}
+
 } // unnamed namespace
 
-SearchInfo::SearchInfo(const std::string& pattern)
-  : pattern(pattern),
-    geoQuery()
-{}
+std::pair<std::string, Library::BookIdSet> InternalServer::selectBooks(const RequestContext& request, const std::string& prefix) const
```
veloman-yunkan, line 217: It looks like we don't need `prefix` as a parameter for `selectBooks()` and `getSearchInfo()` (since the latter is called only with a prefix of "books" while being the only caller of the former). If `prefix` is justified later, let's add it when we really need it.

static/i18n/en.json
```diff
@@ -4,12 +4,16 @@
 		]
 	},
 	"name":"English",
-	"suggest-full-text-search": "containing '{{{SEARCH_TERMS}}}'..."
-	, "no-such-book": "No such book: {{BOOK_NAME}}"
+	"suggest-full-text-search" : "containing '{{{SEARCH_TERMS}}}'..."
+	, "no-such-book" : "No such book: {{BOOK_NAME}}"
+	, "too-many-books" : "Too many books requested ({{NB_BOOKS}}) where limit is {{LIMIT}}"
+	, "no-book-found" : "No book found"
```
veloman-yunkan, line 10: 1. Isn't "No book matches selection criteria" a better message?
2. This message is not documented in qqq.json

src/server/internalServer.cpp
```diff
@@ -95,35 +95,200 @@ inline std::string normalizeRootUrl(std::string rootUrl)
...
+    for(const auto& bookId: id_vec) {
+      try {
+        // This is a silly way to check that bookId exists
+        mp_nameMapper->getNameForId(bookId);
+      } catch (const std::out_of_range&) {
+        throw Error(noSuchBookErrorMsg(bookId));
+      }
+    }
+    const auto bookIds = Library::BookIdSet(id_vec.begin(), id_vec.end());
+    checkBookNumber(bookIds, m_multizimSearchLimit);
```
veloman-yunkan, line 249: 1. `checkBookNumber()` is called at too many locations. It is better to have this check at a single location (right after the call to `selectBooks()`)
2. Please add a unit test covering the failure of this check

veloman-yunkan (changes requested):
After fixing the new comments, please rebase and fixup for the final review.

static/i18n/qqq.json
```diff
@@ -2,16 +2,21 @@
...
 	"suggest-full-text-search": "Text appearing in the suggestion list that, when selected, runs a full text search instead of the title search",
 	"no-such-book": "Error text when the requested book is not found in the library",
+	"too-many-books":"Error text when user request more books than the limit set by the administrator",
 	"url-not-found": "Error text about wrong URL for an HTTP 404 error",
+	"no-book-found": "Error text when no book matches the selection criteria",
 	"suggest-search": "Suggest a search when the URL points to a non existing article",
 	"random-article-failure": "Failure of the random article selection procedure",
 	"invalid-raw-data-type": "Invalid DATATYPE was used with the /raw endpoint (/raw/<book>/DATATYPE/...); allowed values are 'meta' and 'content'",
+	"no-value-for-arg" : "Error text when no value has been provided for ARGUMENT in the request's query string",
+	"no-query" : "Error text when no query has been provided to select one or several books",
```
veloman-yunkan, line 19: The explanation of this message seems to be incorrect

veloman-yunkan (changes requested):
A couple more small fixes that don't deserve separate fixup commit and can be done in place while squashing the current fixup commits.

veloman-yunkan (approved):
LGBT

PR #732, opened 2022-03-24
New way of building 404 error HTML responses

veloman-yunkan:
This is the next step on our path to internationalizing kiwix-serve (and making #679 a smaller PR).

The main goal of this PR was to introduce a new eDSL like way of building error responses. Here it is done only for 404 HTML error responses. A couple of further generalizations still remain in #679 because those enhancements were a little too dependent (factually but not conceptually) on the changes related to the internationalization of kiwix-serve.


mgautierfr:
I like the combination concept you introduce with this PR. This create a more cleaner way to create responses.

kelson42:
@veloman-yunkan @mgautierfr Would be great to fix the `docs/readthedocs.org:libkiwix` task before merging.

mgautierfr:
I'm on it https://github.com/kiwix/libkiwix/issues/735

PR #738, opened 2022-03-29
Shared template for error HTML pages

veloman-yunkan:
The exploratory work on the internationalization of kiwix-serve (performed in a WIP PR #679) has lead to the generalization of the template for producing error HTML responses. This PR packages that finding in a relatively small change that should be easy to review.


mgautierfr:
I agree with the changes. I was even thinking doing the same after I've rebased my PR.
But could we go a bit further and replace `build_500` also ?

veloman-yunkan:
> I agree with the changes. I was even thinking doing the same after I've rebased my PR. But could we go a bit further and replace `build_500` also ?

Done.

In my work on the internationalization of kiwix-serve I circumvented the internal server error because of having no way to test it (other than adding special debugging code for simulating it). The requested change was still added without any testing.

kelson42:
@mgautierfr Can we merge this PR now?

src/server/response.cpp
```diff
@@ -195,26 +191,6 @@ std::unique_ptr<Response> Response::build_416(const InternalServer& server, size
...
-        server.m_root, //root
-        true, //verbose
-        true, //raw
-        false, //withTaskbar
-        false, //withLibraryButton
-        false, //blockExternalLinks
-        content, //content
-        "text/html" //mimetype
-  ));
-  response->set_code(MHD_HTTP_INTERNAL_SERVER_ERROR);
```
mgautierfr, line 214: We now use `ContentResponse::build` to build the ContentResponse for error500.
But it will not construct a minimal response (`raw=true`, `with*=false`, ...).

We should override `generateResponseObject` for the specific case of 500.

mgautierfr (changes requested):

src/server/response.cpp
veloman-yunkan: Done

veloman-yunkan:
I also figured out a way to trigger an internal server error. So, now the HTTP 500 error page is unit-tested.

src/server/response.cpp
```diff
@@ -169,12 +160,29 @@ HTTP400HtmlResponse& HTTP400HtmlResponse::operator+(InvalidUrlMsg /*unused*/)
...
+  // operator+() is a state-modifying operator (akin to operator+=)
+  *this + "An internal server error occured. We are sorry about that :/";
 }
 
+std::unique_ptr<ContentResponse> HTTP500HtmlResponse::generateResponseObject() const
+{
+  // We want a 500 response to be a minimalistic one (so that the server doesn't
+  // have to provide additional resources required for its proper rendering)
+  // ";raw=true" in the MIME-type below disables response decoration
+  // (see ContentResponse::contentDecorationAllowed())
```
mgautierfr, line 180: This is in the same time nice and a bit disturbing.
The "raw=true" has been added to handle something totally different (and on which I disagreed, https://github.com/kiwix/libkiwix/pull/403#issuecomment-671277426) and in the same time, it fits perfectly here ..

mgautierfr (approved):
LGTM

mgautierfr:
(Rebased/fixup on master)

PR #743, opened 2022-04-04
Correctly detect the number of article for zim version <= 6

mgautierfr:
Fix kiwix/kiwix-tools#536

### DIFF
--- a/src/tools/archiveTools.cpp
+++ b/src/tools/archiveTools.cpp
@@ -125,6 +125,30 @@ unsigned int getArchiveMediaCount(const zim::Archive& archive) {
   return counter;
 }
 
+unsigned int getArchiveArticleCount(const zim::Archive& archive) {
+  // [HACK]
+  // getArticleCount() returns different things depending of the "version" of the zim.
+  // On old zim (<=6), it returns the number of entry in `A` namespace
+  // On recent zim (>=7), it returns:
+  //  - the number of entry in `C` namespace (==getEntryCount) if no frontArticleIndex is present
+  //  - the number of front article if a frontArticleIndex is present
+  // The use case >=7 without frontArticleIndex is pretty rare so we don't care
+  // We can detect if we are reading a zim <= 6 by checking if we have a newNamespaceScheme.
+  if (archive.hasNewNamespaceScheme()) {
+    //The articleCount is "good"
+    return archive.getArticleCount();
+  } else {
+    // We have to parse the `M/Counter` metadata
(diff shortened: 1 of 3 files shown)


kelson42:
@mgautierfr I don't think this is proper. We need to reintroduce the media and article count based on M/Counter... this is what was done earlier.

mgautierfr:
This is what is done with (old) zim files (<=6). For new zim files, we use the list of front articles (the same on which we do random or title suggestion).
Using the front articles list is probably more accurate as we don't count `text/html` items with are html redirects (warc2zim).

src/tools/archiveTools.cpp
```diff
@@ -125,6 +125,30 @@ unsigned int getArchiveMediaCount(const zim::Archive& archive) {
...
 }
 
+unsigned int getArchiveArticleCount(const zim::Archive& archive) {
+  // [HACK]
+  // getArticleCount() returns different things depending of the "version" of the zim.
+  // On old zim (<=6), it returns the number of entry in `A` namespace
+  // On recent zim (>=7), it returns:
+  //  - the number of entry in `C` namespace (==getEntryCount) if no frontArticleIndex is present
+  //  - the number of front article if a frontArticleIndex is present
+  // The use case >=7 without frontArticleIndex is pretty rare so we don't care
```
veloman-yunkan, line 135: Nice poetry!

On a more serious note, presence of the `frontArticleIndex` doesn't guarantee that the return value of `zim::Archive::getArticleCount()` is meaningful. It will be equal to zero for ZIM files created by tools linked against pre-7.2.0 libzim7 which weren't upgraded to supply the `FRONT_ARTICLE` hint (this was fixed by https://github.com/openzim/libzim/commit/682e38b0 in openzim/libzim#647). Maybe `kiwix::getArchiveArticleCount()` should return the number of entries in `C` namespace in that case?

mgautierfr: Those zim files are pretty rare I think. mwoffliner is still using the old libzim version as [node-libzim](https://github.com/openzim/node-libzim) is not yet ported (openzim/node-libzim#72) and we've created very few zim with the python wrapper (if we publish them). @kelson42 and @rgaudin may confirm this.

`zim::Archive::getArticleCount()` already return the number of entries in `C` namespace if there is no front article list. (https://github.com/openzim/libzim/blob/master/src/archive.cpp#L69-L78)

veloman-yunkan (approved):

PR #744, opened 2022-04-05
Got rid of static/templates/no_search_result.html

veloman-yunkan:
This follow-up to #738 is probably the last step before removing the WIP status from #679.


src/server/internalServer.cpp
```diff
@@ -447,6 +447,11 @@ std::string noSuchBookErrorMsg(const std::string& bookName)
   return "No such book: " + bookName;
 }
 
+std::string noSearchResultsMsg(const std::string& pattern)
+{
+  return "There is no article with the title <b>\"" + encodeDiples(pattern) + "\"</b> and the fulltext search engine is not available for this content.";
```
mgautierfr, line 452: With #497, we don't redirect to article titled with `pattern`.
We should rewrite this message with "The fulltext search engine is not available for this content" only.

mgautierfr (changes requested):
We should just change the details message but else we are good.

The question about the md5id in cpp is interesting but this PR is working as it is, and we should discuss about this in #712. So this is not a blocker for this PR.

src/server/internalServer.cpp
veloman-yunkan: Done

mgautierfr (approved):

mgautierfr:
Rebased/fixup on master

PR #757, opened 2022-04-22
[SERVER] Support gzip encoding instead of deflate.

### LINKED ISSUE #753: [URGENT] Use gzip instead of deflate
The server currently supports Content-Encoding for the `deflate` flavor only.

There are 3 flavors in the wild:

- `deflate` which uses the DEFLATE algorithm. It used to have a lot of issues in browser implementations in the past but that's all forgotten with modern browsers I believe. It's slightly faster than gzip (simpler digest).
- `gzip` which uses the DEFLATE algorithm also but its format is different (headers and CRC). Most supported everywhere in browsers but also in reverse proxies.
- `br` Brotli. Most recent and fastest I believe but support is missing in some browser and only through modules in reverse proxies.

Has we use kiwix-serve extensively behind a reverse proxy, it would greatly help and simplify deployments if it supports gzip as gzip is the lingua-franca for compression in reverse-proxies (nginx, varnish).

[this post](https://stackoverflow.com/questions/1838699/how-can-i-decompress-a-gzip-stream-with-zlib) and [zlib](https://zlib.net/manual.html#Gzip) describes the differences and indicates switching should be fairly simple.

I'm in favor of simply replacing deflate by gzip for maintenance's sake. I am unaware of any use case of deflate-only support in a client.

mgautierfr:
The `compress` function is copied from httplib

Fix #753

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -33,8 +33,15 @@
 
 #include <array>
 
-
-#define KIWIX_MIN_CONTENT_SIZE_TO_COMPRESS 100
+// This is somehow a magic value.
+// If this value is too small, we will compress (and lost cpu time) too much
+// content.
+// If this value is too big, we will not compress enough content and send too
+// much data.
+// If we assume that MTU is 1500 Bytes it is useless to compress
+// content smaller as the content will be sent in one packet anyway.
+// 1400 Bytes seems to be a common accepted limit.
+#define KIWIX_MIN_CONTENT_SIZE_TO_COMPRESS 1400
 
 namespace kiwix {
 
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -31,8 +31,10 @@
 #include <mustache.hpp>
 #include <zlib.h>
 
+#include <array>
 
-#define KIWIX_MIN_CONTENT_SIZE_TO_DEFLATE 100
+
+#define KIWIX_MIN_CONTENT_SIZE_TO_COMPRESS 100
 
 namespace kiwix {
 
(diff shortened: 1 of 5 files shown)


src/server/response.cpp
```diff
@@ -58,6 +60,41 @@ bool is_compressible_mime_type(const std::string& mimeType)
       || mimeType.find("application/json") != string::npos;
 }
 
+inline bool compress(std::string &content) {
```
veloman-yunkan, line 63: `inline` is not justified here and only adds some noise.

src/server/response.cpp
```diff
@@ -365,35 +402,17 @@ ContentResponse::create_mhd_response(const RequestContext& request)
...
-         We need to chunk them off (move the content 2bytes)
-         It has no incidence on other browsers
-         See http://www.subbu.org/blog/2008/03/ie7-deflate-or-not and comments */
-      m_content = string((char*)&compr_buffer[2], comprLen - 2);
-      m_etag.set_option(ETag::COMPRESSED_CONTENT);
-    } else {
-      shouldCompress = false;
-    }
-  }
+  bool isCompressed = can_compress(request) && compress(m_content);
```
veloman-yunkan, line 405: I would add a `const` here

src/server/response.cpp
```diff
@@ -31,8 +31,10 @@
 #include <mustache.hpp>
 #include <zlib.h>
 
+#include <array>
 
-#define KIWIX_MIN_CONTENT_SIZE_TO_DEFLATE 100
+
+#define KIWIX_MIN_CONTENT_SIZE_TO_COMPRESS 100
```
veloman-yunkan, line 37: I wonder if this threshold is too low - why trying to compress payload that is smaller than the header section?

veloman-yunkan (commented):
Since our unit tests don't check the correctness of compressed content, did you verify that the code actually works with the browsers?

src/server/response.cpp
mgautierfr: Probably yes. It is a pretty old magic value and it seems it has been introduce as some kind of workaround for a segfault https://github.com/kiwix/kiwix-tools/commit/4c79cfcf06e75ef4c38fd45649a36ed426e80445

The question is what would be the good value ? I've seen different values but it seems that MTU is a good limit (https://www.computerworld.com/article/2693941/why-it-doesn-t-make-sense-to-gzip-all-content-from-your-web-server.html). So we can move this to 1400 Bytes.

veloman-yunkan (approved):
I am fine with the code but I didn't build and manually test it (since the unit tests don't exhaustively validate this change). Before merging, this functionality must be tested via kiwix-serve and the browser development console (in order to verify that kiwix-serve sends gzip-compressed content and the browser is able to decompress it).

mgautierfr:
> Before merging, this functionality must be tested via kiwix-serve and the browser development console (in order to verify that kiwix-serve sends gzip-compressed content and the browser is able to decompress it).

Done

Capture d’écran du 2022-04-27 11-40-15


Everything works correctly, as far as I can see on my browser

veloman-yunkan:
> > Before merging, this functionality must be tested via kiwix-serve and the browser development console (in order to verify that kiwix-serve sends gzip-compressed content and the browser is able to decompress it).
> 
> Done
> 
> Capture d’écran du 2022-04-27 11-40-15
> 
> Everything works correctly, as far as I can see on my browser

Perfect! Why don't you merge this PR then?

mgautierfr:
> Perfect! Why don't you merge this PR then?

Because I was waiting for CI to pass after my rebase-fixup and switch to something else.

PR #761, opened 2022-04-28

translatewiki:
Translation updates


kelson42:
@veloman-yunkan Our strings are now on translatewiki.net. Once or twice a week, a bot from TW will (1) resync with our master strings in English (2) make a PR with the latest changes/additions. If a PR exists, then the current PR will be updated/erased. It is expected that this PR is reviewed and merged time to time (does not have to be done each week, but maybe once a month in the worth case). Would be great to secure that this is done before a release as well.

veloman-yunkan:
> BTW, there is not documentation in the `README.md` about the localisation process and what should be run to apply new translations. A bit of documentation would be really welcome.

@kelson42 I enhanced the build process so that new translations are picked up automatically - see #763. Do we need anything else in the documentation other than a link to the Translatewiki project for Kiwix Serve?

veloman-yunkan:
> Our strings are now on translatewiki.net. Once or twice a week, a bot from TW will (1) resync with our master strings in English (2) make a PR with the latest changes/additions. If a PR exists, then the current PR will be updated/erased. It is expected that this PR is reviewed and merged time to time (does not have to be done each week, but maybe once a month in the worth case). Would be great to secure that this is done before a release as well.

@kelson42 This is the first time when I review a Translatewiki PR. What is the expected procedure for feedback? If I write a review comment who is going to address it.

I noticed an issue for a language that I know and I fixed it in a [new commit](https://github.com/kiwix/libkiwix/pull/761/commits/26a0d92588e5f4472880e634abf50e6e10bac52a) added to this PR. It is possible that a similar confusion is present for other languages (but should be prevented in subsequent translations due to a clarification added in qqq.json).

kelson42:
@veloman-yunkan No, a simple explanation about translatewiki and how it gets integrated is enough indeed.

static/i18n/qqq.json
```diff
@@ -1,27 +1,28 @@
...
-	, "404-page-heading" : "Heading of the 404 error page"
-	, "500-page-title" : "Title of the 500 error page"
-	, "500-page-heading" : "Heading of the 500 error page"
-	, "fulltext-search-unavailable" : "Title of the error page returned when search is attempted in a book without fulltext search database"
-	, "no-search-results": "Text of the error page returned when search is attempted in a book without fulltext search database"
-	, "library-button-text": "Tooltip of the button leading to the welcome page"
-	, "home-button-text": "Tooltip of the button leading to the main page of a book"
-	, "random-page-button-text": "Tooltip of the button opening a randomly selected page"
-	, "searchbox-tooltip": "Tooltip displayed for the search box"
+	"name": "{{doc-important|Don't write \"English\" in your language!}}\n\n'''Write the name of ''your'' language in its native script.'''\n\nCurrent language to which the string is being translated to.\n\n'''Important:''' Do not use your language’s word for “English”. Use the word that your language uses to refer to itself. If you translate this message to mean “English” in your language, your change will be reverted.",
```
veloman-yunkan, line 8: Good that this clarification was added!

The Russian translation got this wrong. I fixed it in a new commit added to this PR, but I suspect that the same issue may be present for other translations included in this PR.

kelson42: We should have a look to TW doc. There for sire a way to request reassment of a string translation without modifying the string in English.

mgautierfr: We may even remove the name. At which moment we want to display to the user the language we are using to speak to him ?

veloman-yunkan: I copied `name` from kiwix-desktop. I suspect that we will need it in our language selection UI. We can get rid of `name` if we either

1. switch to 3-letter language codes in our i18n resource filenames or 
2. enhance the language code-to-name translation function `getLanguageSelfName()`
 
https://github.com/kiwix/libkiwix/blob/f90cc39a522b97e8df4acb5692d6c6b56beaf100/src/opds_dumper.cpp#L177-L183

The first option won't work if we want to support language variations (e.g. en_US) in our translations.

kelson42: Names from languages (in their respective languages) should come to 99 percent from libicu (the other one ate hardcoded). Pretty sure it works already that way.

For the monent I'm not sure this is a good to handle local variants of languages... in particular if this bring complexity and new problem to handle.

mgautierfr:
I see no real reason to not merge this.
We have this [discussion](https://github.com/kiwix/libkiwix/pull/761#discussion_r862509646) about the name to remove or not but it will not be done in this PR 
We also have this discussion on the PR https://github.com/kiwix/libkiwix/pull/763 but yet again, it will not be fixed here and it is not a blocker.

PR #764, opened 2022-05-05
Preparatory work on multizim

mgautierfr:
This PR is an extract of improvement made in #729 but not especially related to multizim search.

### DIFF
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -58,7 +58,7 @@ SearchRenderer::SearchRenderer(zim::SearchResultSet srs, NameMapper* mapper, Lib
       mp_nameMapper(mapper),
       mp_library(library),
       protocolPrefix("zim://"),
-      searchProtocolPrefix("search://?"),
+      searchProtocolPrefix("search://"),
       estimatedResultCount(estimatedResultCount),
       resultStart(start)
 {}
@@ -68,12 +68,12 @@ SearchRenderer::~SearchRenderer() = default;
 
 void SearchRenderer::setSearchPattern(const std::string& pattern)
 {
-  this->searchPattern = pattern;
+  searchPattern = pattern;
 }
 
-void SearchRenderer::setSearchContent(const std::string& name)
+void SearchRenderer::setSearchContent(const std::string& content)
 {
-  this->searchContent = name;
+  searchContent = content;
 }
 
 void SearchRenderer::setProtocolPrefix(const std::string& prefix)
@@ -84,6 +84,21 @@ void SearchRenderer::setProtocolPrefix(const std::string& prefix)
 void SearchRenderer::setSearchProtocolPrefix(const std::string& prefix)
 {
   this->searchProtocolPrefix = prefix;
+}
+
+kainjow::mustache::data buildQueryData
+(
+  const std::string& searchProtocolPrefix,
+  const std::string& pattern,
+  const std::string& searchContent
+) {
+  kainjow::mustache::data query;
+  query.set("pattern", kiwix::encodeDiples(pattern));
+  std::ostringstream ss;
+  ss << searchProtocolPrefix << "?pattern=" << urlEncode(pattern, true);
+  ss << "&content=" << urlEncode(searchContent, true);
+  query.set("unpaginatedQuery", ss.str());
+  return query;
 }
 
 kainjow::mustache::data buildPagination(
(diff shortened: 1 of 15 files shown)


src/search_renderer.cpp
```diff
@@ -162,28 +162,28 @@ kainjow::mustache::data buildPagination(
...
-  kainjow::mustache::data results{kainjow::mustache::data::type::list};
-
+  kainjow::mustache::data results;
+  kainjow::mustache::data items{kainjow::mustache::data::type::list};
   for (auto it = m_srs.begin(); it != m_srs.end(); it++) {
     kainjow::mustache::data result;
+    std::string zim_id(it.getZimId());
     result.set("title", it.getTitle());
-    result.set("url", it.getPath());
+    result.set("absolutePath", urlEncode(protocolPrefix+mp_nameMapper->getNameForId(zim_id)+"/"+it.getPath()));
```
veloman-yunkan, line 171: At first, this line seemed dubious to me because of `protocolPrefix` being inside the call to `urlEncode()`. While checking how `urlEncode()` would handle the colon character, I discovered to my surprise, that by default, `kiwix::urlEncode()` doesn't encode reserved URL symbols (including `&`, `+`, and `?`). 

https://github.com/kiwix/libkiwix/blob/26eccb5a5fad04e2c44f2bb204491750fb0cf328/src/tools/stringTools.h#L43
https://github.com/kiwix/libkiwix/blob/26eccb5a5fad04e2c44f2bb204491750fb0cf328/src/tools/stringTools.cpp#L215-L230
https://github.com/kiwix/libkiwix/blob/26eccb5a5fad04e2c44f2bb204491750fb0cf328/src/tools/stringTools.cpp#L149-L191

I guess we have to review all call sites of `urlEncode()` and, maybe, its default behaviour, too!

veloman-yunkan (changes requested):
While reviewing this PR I again felt the need for a unit test for the "/search" endpoint that checks the contents of the response (the need for such a unit test started suggesting itself during the work on #679 and #712, however a straightforward implementation didn't fit my standards for a C++ unit test format that's why I didn't pursue that goal enthusiastically). But we need to do it (not claiming that it has to be done in this PR).

veloman-yunkan:
> While reviewing this PR I again felt the need for a unit test for the "/search" endpoint that checks the contents of the response. The need for such a unit test started suggesting itself during the work on #679 and #712, however a straightforward implementation didn't fit my standards for a C++ unit test format that's why I didn't pursue that goal enthusiastically. But we need to do it (not claiming that it has to be done in this PR).

I am now working on this in #765. Planning to complete it in couple of days.

src/search_renderer.cpp
```diff
@@ -162,28 +162,28 @@ kainjow::mustache::data buildPagination(
 std::string SearchRenderer::getHtml()
 {
   // Build the results list
-  kainjow::mustache::data results{kainjow::mustache::data::type::list};
-
+  kainjow::mustache::data results;
```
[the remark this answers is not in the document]
mgautierfr: Done

src/search_renderer.cpp
mgautierfr: Yes, I was surprised also the first time I see it, but it is usefull.
Maybe we should duplicate the function (with `urlEncodeAll` and `urlEncodeSoft` ?) to be explicit about what is done and avoid such surprise in the future.

mgautierfr: Probably that none of the two options are good:
- We don't want to url encode `/` chars as it is part of the normal path
- As there is no query string here, we want to url encode `&`/`?` chars (potentially present in the article path)

From this page https://www.tutorialspoint.com/html/html_url_encoding.htm, we should probably have a more complete choice on what encode depending of which context.

kelson42:
@mgautierfr What is the status here? Does it need a new review? Îm slightly concerned that waiting to #765 might significantly pospone the release date. What kind of delay should we deal with?

veloman-yunkan:
@kelson42 I will complete #765 by Monday morning and will immediately address any review comments so that we can merge it the same day.

veloman-yunkan (commented):
LGTM. Now that #765 is ready, let's merge it and update this PR accordingly (some tests will need to be changed because of the fixes contained by this PR).

veloman-yunkan:
> Rebased on new master after the merge of #765
> 
> Ready for a new (last?) review.

@mgautierfr Please rebase&fixup.

test/server.cpp
```diff
@@ -1158,7 +1158,7 @@ R"SEARCHRESULT(
 )SEARCHRESULT",
 
 R"SEARCHRESULT(
-            <a href="/ROOT/zimfile/A/Catchin&apos;_Some_Rays:_The_Music_of_Ray_Charles">
+            <a href="/ROOT/zimfile/A/Catchin'_Some_Rays:_The_Music_of_Ray_Charles">
```
veloman-yunkan, line 1161: The explanation of  this change disappeared when the fixup commit was merged

veloman-yunkan (approved):
It would be nice if the commit message written for the fixup commit is restored.

PR #765, opened 2022-05-08
Unittests for search results page

veloman-yunkan:


mgautierfr:
Maybe it is a determinism based on the start page ?

The snippet is generated by passing the article content to the `MSet::snippet` method (https://xapian.org/docs/apidoc/html/classXapian_1_1MSet.html#ab3af7b20654dcc6e3335cc21be74efda)
And the MSet is the Match Set and it is dependent of the actual document in the set. Change the start page/page length change the set (and so the term frequency and so)
Reading xapian source code, it seems that the method use so stats on the term frequency in the results  to score the different parts in the snippet.

veloman-yunkan:
> Maybe it is a determinism based on the start page ?
> 
> The snippet is generated by passing the article content to the `MSet::snippet` method (https://xapian.org/docs/apidoc/html/classXapian_1_1MSet.html#ab3af7b20654dcc6e3335cc21be74efda) And the MSet is the Match Set and it is dependent of the actual document in the set. Change the start page/page length change the set (and so the term frequency and so) Reading xapian source code, it seems that the method use so stats on the term frequency in the results to score the different parts in the snippet.

@mgautierfr Yes, that explanation sounds plausible. So we have another obstacle to creating a manageable unit test.

veloman-yunkan:
> > Maybe it is a determinism based on the start page ?
> > The snippet is generated by passing the article content to the `MSet::snippet` method (https://xapian.org/docs/apidoc/html/classXapian_1_1MSet.html#ab3af7b20654dcc6e3335cc21be74efda) And the MSet is the Match Set and it is dependent of the actual document in the set. Change the start page/page length change the set (and so the term frequency and so) Reading xapian source code, it seems that the method use so stats on the term frequency in the results to score the different parts in the snippet.
> 
> @mgautierfr Yes, that explanation sounds plausible. So we have another obstacle to creating a manageable unit test.

@mgautierfr What do you think of [my solution](https://github.com/kiwix/libkiwix/pull/765/commits/c94652acce2e1f677117b37894abe9721e92f719) to the said problem? Does it seem viable in the long run?

veloman-yunkan:
The next problem with snippets is the difference across different platforms as demonstrated by the failures of **Packages / build-deb (ubuntu-focal)** and **Packages / build-deb (ubuntu-bionic)** CI runs. In those runs the snippets are cut at the edges slightly differently:
- Example 1
  >  `Japan (1975......</cite>`
  > vs
  >  `Japan (1975)...</cite>`
- Example 2
  >  `"Don't Let the Sun Catch You Cryin......</cite>`
  > vs
  >  `"Don't Let the Sun Catch You Cryin'"......</cite>`
- Example 3
  > `<cite>...n' Nasty"`
  > vs
  > `<cite>...'n' Nasty"`
- Example 4
  > `<cite>...60s and '70s ABC-Paramount`
  > vs
  > `<cite>...'60s and '70s ABC-Paramount`

I wonder if this has to do with the code or the environment (e.g. locale).

kelson42:
> I wonder if this has to do with the code or the environment (e.g. locale).

To me this looks like a bug upstream in Xapian. If nothing obvious is on our side, please report it.

mgautierfr:
> @mgautierfr What do you think of [my solution](https://github.com/kiwix/libkiwix/pull/765/commits/c94652acce2e1f677117b37894abe9721e92f719) to the said problem? Does it seem viable in the long run?

At the end, it looks like we will have some kind of "core" text that we want to be present in the snippet and peripheral text that can be present or not.
Why not checking only the presence of the core text ?
With some kind of regex:
```
<cite>...(.*)blind African American musician. Like his debut, this album failed to generate hit singles as Motown struggled to find a sound to fit Wonder, who was just 12 when this album was released. Tribute to Uncle Ray Studio album by Little Stevie Wonder Released October 1962 Recorded 1962 Studio Studio A, Hitsville USA, Detroit Genre Soul, <b>jazz</b> Label Tamla Producer Henry Cosby, Clarence Paul Stevie Wonder chronology The <b>Jazz</b> Soul(.*)......</cite>
```
The `(.*)` regex may have to be a bit more complex to not catch things link `<cite>...</cite><cite>blind African` 

Then test could be that the expected regex match the actual result. It may also be used to be tolerant to variation in the different platform (and xapian version) which are out of our control.

---

Another possibility (but with more work and adding feature) would be to have a json/xml results instead of a html rendering and then:
- Test the json result (with may be simple to test)
- Test the html rendering of controlled search result

veloman-yunkan:
> > @mgautierfr What do you think of [my solution](https://github.com/kiwix/libkiwix/pull/765/commits/c94652acce2e1f677117b37894abe9721e92f719) to the said problem? Does it seem viable in the long run?
> 
> At the end, it looks like we will have some kind of "core" text that we want to be present in the snippet and peripheral text that can be present or not. Why not checking only the presence of the core text ? With some kind of regex:
> 
> ```
> <cite>...(.*)blind African American musician. Like his debut, this album failed to generate hit singles as Motown struggled to find a sound to fit Wonder, who was just 12 when this album was released. Tribute to Uncle Ray Studio album by Little Stevie Wonder Released October 1962 Recorded 1962 Studio Studio A, Hitsville USA, Detroit Genre Soul, <b>jazz</b> Label Tamla Producer Henry Cosby, Clarence Paul Stevie Wonder chronology The <b>Jazz</b> Soul(.*)......</cite>
> ```
> 
> The `(.*)` regex may have to be a bit more complex to not catch things link `<cite>...</cite><cite>blind African`
> 
> Then test could be that the expected regex match the actual result. It may also be used to be tolerant to variation in the different platform (and xapian version) which are out of our control.

@mgautierfr Thank you for the input.

A regex-based solution has the following problems:

1. The contents of the snippet will have to be properly escaped.
2. Currently, the entire response body is compared via a single `EXPECT_EQ()` to the expected text generated from the test-case data. In case of its failure, googletest automatically reports the diff (though for such large multiline strings it works with some quirks). A regex-based approach will require either parsing the search results and comparing individual entries against the expected test data via `std::regex_match()`, or building one large regex out of the test data and give up the convenience of the easy insight into the reason of the failure. 

> 
> Another possibility (but with more work and adding feature) would be to have a json/xml results instead of a html rendering and then:
> 
>   * Test the json result (with may be simple to test)
> 
>   * Test the html rendering of controlled search result

This is an interesting design idea, but it is orthogonal to the problem of fluctuating snippets which will still need to be solved, e.g. using the regex method (though most likely less effort will be required).

---

I realized that my current solution only works for the problematic snippet and doesn't account for the different ways of if and where the Xapian's `...` omission designator text may be inserted in the snippet.

I will try to re-think it.

veloman-yunkan:
@mgautierfr Thanks again for your feedback. In addition to the preceding discussion it helped me to come up with an approach and implementation that, IMHO, can be improved incrementally/as needed. The count of [previously observed failures](https://github.com/kiwix/libkiwix/pull/765#issuecomment-1122897588) in Packages CI runs has been significantly reduced, the only remaining one looking like a bug in Xapian (the search term in the snippet is not highlighted).

I haven't yet removed the first attempt at handling of pagination-induced variation in snippets, but it should definitely be dropped in the final version of this PR.

veloman-yunkan:
> The count of [previously observed failures](https://github.com/kiwix/libkiwix/pull/765#issuecomment-1122897588) in Packages CI runs has been significantly reduced, the only remaining one looking like a bug in Xapian (the search term in the snippet is not highlighted).

The problem in this case seems to be caused by incorrect handling of stemming during snippet generation. As a result the search pattern "beatles" is transformed to a search term "beatl" which then doesn't match the word "beatles" in the text from which a snippet must be extracted.

veloman-yunkan:
@mgautierfr Even though I have removed the WIP status from this PR I still deliberately left in its history two commits with the implementation of an approach that was then overriden with a simpler idea requiring to keep in the test data an oversized snippet for some results. The more complex idea is a little more stable to potential variations in snippet generation, yet with the simple implementation any such new failure should be easy to fix by increasing the extent of the snippet in the test data. I think it makes sense to preserve the more complex implementation in the PR history so that we can take advantage of it if needed.

mgautierfr:
This PR is surely impacted by openzim/libzim#697

veloman-yunkan:
> This PR is surely impacted by [openzim/libzim#697](https://github.com/openzim/libzim/pull/697)
> 
> This change (in libzim) makes me wonder if it is a good idea to check for snippet content (more that testing than requested term are highlighted). Changing the scoring in xapian/libzim may make xapian select a totally different region of the article (and we still have valid snippet). I'm afraid that if we test too aggressively for the snippet content, we will simply update the test data because the new returned snippet seems valid (requested terms highlighted), in this case, it is better to simply test if it is valid.

I don't have a strong opinion on this. I only know that anything that is not tested is far more probable to break without being noticed on time.

veloman-yunkan:
@mgautierfr Should I update this PR with respect to the changes brought by openzim/libzim#697? Or it is better that we first converge on the approach to snippet validation?

mgautierfr:
As you are both agree with the idea to test the snippet text. Yes please update this PR o openzim/libzim#697. If in the future we found it is too complex to maintain this test we will see how/if it should be simplified.

mgautierfr:
Agree with the final code.
I wonder about the git history. There is few commits with way of testing which are replaced by other way of testing just after. It is understandable in regards of how you made the PR and the different problem you faced, but it may be better to clean the git history now we agree on how to test snippet.

veloman-yunkan:
> Agree with the final code. I wonder about the git history. There is few commits with way of testing which are replaced by other way of testing just after. It is understandable in regards of how you made the PR and the different problem you faced, but it may be better to clean the git history now we agree on how to test snippet.

@mgautierfr Should the commit https://github.com/kiwix/libkiwix/pull/765/commits/aa5498e5e5a4904ff77439a74acc32f4298949d6 be preserved?

veloman-yunkan:
I believe that this is PR is ready to be merged. The commit titled "[Changed a test in order to avoid a bug in Xapian](https://github.com/kiwix/libkiwix/pull/765/commits/8f2f93371b98c759c66290b2b6372021fbd05cac)" is preserved and serves as documentation of a known issue (of course, we also have to file it in the Xapian bug-tracker).

kelson42:
@mgautierfr Neeing approval and merging then.

mgautierfr (approved):

PR #770, opened 2022-05-11
Use real magnet link in download modal

### LINKED ISSUE #767: Magnet link makes no sense
kiwix-serve, when offering ZIM file downloads offers both a torrent file and a *magnet file*.

`magnet:` is an URL scheme that is understood by browsers and torrent applications but it is not a file type. The magnet link should have the content of that file as its `href` value. Should this be too complicated to implement, I believe (but that's just my personal opinion) that the magnet link should be removed altogether.

Serving a text file with a long link as content looks broken and decreases the overall impression of the tool.

juuz0:
Fixes #767

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -142,10 +142,21 @@
         }
     }
 
+    async function getMagnetLink(downloadLink) {
+        const magnetUrl = downloadLink + '.magnet';
+        const controller = new AbortController();
+        setTimeout(() => controller.abort(), 5000);
+        const magnetLink = await fetch(magnetUrl, { signal: controller.signal }).then(response => {
+            return response.ok ? response.text() : '';
+        }).catch((_error) => '');
+        return magnetLink;
+    }
+
     function insertModal(button) {
         const downloadLink = button.getAttribute('data-link');
-        button.addEventListener('click', (event) => {
+        button.addEventListener('click', async (event) => {
             event.preventDefault();
+            const magnetLink = await getMagnetLink(downloadLink);
             document.body.insertAdjacentHTML('beforeend', `<div class="modal-wrapper">
                 <div class="modal">
                     <div class="modal-heading">
@@ -181,12 +192,13 @@
                                 <div>Sha256 hash</div>
                             </a>
                         </div>
-                        <div class="modal-regular-download">
-                            <a href="${downloadLink}.magnet" target="_blank">
+                        ${magnetLink ? 
+                        `<div class="modal-regular-download">
+                            <a href="${magnetLink}" target="_blank">
                                 <img src="../skin/magnet.png?KIWIXCACHEID" alt="download magnet" />
                                 <div>Magnet link</div>
                             </a>
-                        </div>
+                        </div>` : ``}
                         <div class="modal-regular-download">
                             <a href="${downloadLink}.torrent" download>
                                 <img src="../skin/bittorrent.png?KIWIXCACHEID" alt="download torrent" />
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -319,7 +319,7 @@ R"EXPECTEDRESULT(      src="/ROOT/skin/jquery-ui/external/jquery/jquery.js?cache
           src: url("/ROOT/skin/fonts/Roboto.ttf?cacheid=84d10248") format("truetype");
     <script src="/ROOT/skin/isotope.pkgd.min.js?cacheid=2e48d392" defer></script>
     <script src="/ROOT/skin/iso6391To3.js?cacheid=ecde2bb3"></script>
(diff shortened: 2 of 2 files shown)


juuz0:
I can't really test this because I don't have any idea on how to get the `Download button` to appear when working on localhost. Any help there would be nice :)

kelson42:
@juuz0 You need to put a "url" attribute in library.xml

juuz0:
That works, thanks!
Hold on the review till my next update then please, thank you

kelson42:
@juuz0 Great, don't forget to rebase on `master` please.

juuz0:
> With the recent merge of https://github.com/kiwix/libkiwix/pull/712 the link to resource are now prefixed with the first 8 chars of sha1sum of the resource. And it is tested.
As you've changed index.js, please update the test with the new prefix.

Done.

kelson42 (approved):
From a user perspective, seems to work fine.

kelson42:
@veloman-yunkan Can you please make the code review?

veloman-yunkan (commented):
The implemented solution downloads the `.magnet` files for all books displayed on the welcome/library page. Isn't it better to resolve the magnet link (and initiate a download) *after* the user clicks it?

kelson42:
> The implemented solution downloads the `.magnet` files for all books displayed on the welcome/library page. Isn't it better to resolve the magnet link (and initiate a download) _after_ the user clicks it?

My request was formulated to request this at the time the dialog overlay box is shown, so it should be only after the user has clicked on the "Download" button and the modal overlay is displayed. By reading the code, it seems to work that way! So IMO it should not be for all tiles (because this would generated too many requests) and not at the time there is a click on the link (because then the user can not copy the link).

veloman-yunkan:
@kelson42 My fault. Never used the download functionality, and overlooked a tiny detail that should be obvious if I knew how it worked.

static/skin/index.js
```diff
@@ -142,10 +142,19 @@
...
-    function insertModal(button) {
+    async function getMagnetLink(downloadLink) {
+        const magnetUrl = downloadLink + '.magnet';
+        const magnetLink = await fetch(magnetUrl).then(response => {
+            return response.ok ? response.text() : '';
+        }).catch((error) => console.log(error));
+        return magnetLink;
+    }
+
+    async function insertModal(button) {
```
veloman-yunkan, line 153: Why do you need the `async` here? The fact that this function registers an `async` event listener doesn't make it `async` itself.

juuz0: Forgot to remove because an earlier version had `await getMagnetLink()` outside the listener, fixed now.

static/skin/index.js
```diff
@@ -325,11 +335,11 @@
         });
         books = [...books].filter((book) => {return !booksToFilter.has(getInnerHtml(book, 'id'))});
         booksToDelete.forEach(book => {iso.remove(book);});
-        books.forEach((book) => {
+        books.forEach(async (book) => {
             iso.insert(generateBookHtml(book, sort))
             const downloadButton = document.querySelector(`[data-id="${getInnerHtml(book, 'id')}"] .book__download span`);
             if (downloadButton) {
-                insertModal(downloadButton);
+                await insertModal(downloadButton);
```
veloman-yunkan, line 342: These `async` and `await` are not needed either

static/skin/index.js
```diff
@@ -142,10 +142,19 @@
         }
     }
 
+    async function getMagnetLink(downloadLink) {
+        const magnetUrl = downloadLink + '.magnet';
+        const magnetLink = await fetch(magnetUrl).then(response => {
+            return response.ok ? response.text() : '';
+        }).catch((error) => console.log(error));
```
veloman-yunkan, line 149: Please [add a short timeout](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_operation_with_a_timeout) to the `fetch()`. An alternative would be to display an animated progress placeholder for the magnet link in the immediately opened download box, but that is probably too much to ask.

kelson42:
Sorry for the bad merge as well, will fix this,

kelson42:
> Sorry for the bad merge as well, will fix this,

Fixed

static/skin/index.js
juuz0: What will be the appropriate timeout then?
I feel 3 seconds should be fine

Ping: @kelson42

kelson42: @juuz0 This is maybe a be short, put 10s.

juuz0:
[AbortSignal.timeout](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout) is not supported by most browsers, used a setTimeout to abort controller instead.

veloman-yunkan (approved):

PR #779, opened 2022-06-04
Support for serving customized resources

veloman-yunkan:
During work on the kiwix-serve front-end, the edit-save-test cycle is a multistep procedure:

1. build and install libkiwix
2. build kiwix-tools
3. run kiwix-serve
4. reload the web-page in the browser

When making changes in static resources that are served by kiwix-serve unmodified, the steps 1-3 can be eliminated if kiwix-serve is capable of serving resources from the file-system. This commit adds such a functionality to kiwix-serve. Now, if during startup of kiwix-serve the environment variable `KIWIX_SERVE_CUSTOMIZED_RESOURCES` is defined it is assumed to point to a file here every line has the following format:

URL MIMETYPE RESOURCE_FILE_PATH

When a request is received by kiwix-serve and its URL matches any of the URLs read from the customized resource file, then the resource data is read from the respective file RESOURCE_FILE_PATH and served with mime-type MIMETYPE.

Though this feature was introduced in order to facilitate the development of the iframe-based content viewer, it can also be useful to users who would like to customize the kiwix-serve front-end on their own (without re-building all of kiwix-serve).

There is some overlap with a feature of the kiwix-compile-resources script that also allows to override resources. The differences are:

1. The new way of customizing front-end resources has all such resources listed in a text file and there is a single environment variable from which the path of that file is read. kiwix-compile-resources associates a separate environment variable with each resource.

2. The new way uses regular paths to identify a resource. The kiwix-compile-resources method encodes the resource path by replacing any non-alphanumeric characters (including the path separator) with underscores (so that the resulting resource identifier can be used to construct the name of the environment variable controlling that resource).

3. The new method allows adding new front-end resources. The old method only allows to modify existing resources.

4. The new method allows (actually requires) to specify the URL at which the overriden resource should be served (similarly, the MIME-type can/must be specified, too). The old method only allows to override the contents of a resource.

6. The new method only allows to override front-end resources that are served without any preprocessing by kiwix-serve at runtime. The old method allows to override template resources as well (note that internationalization/translation resources cannot be overriden using the old method, either).


veloman-yunkan:
This PR originates from https://github.com/kiwix/libkiwix/pull/716#discussion_r834364917. The new functionality in its current form doesn't fully take over the old solution that's why I didn't remove the old solution from `kiwix-compile-resources`. If we decide to enhance this PR and simplify `kiwix-compile-resources` instead, then we should also try to do it in a way that eliminates the need for #607 & kiwix/kiwix-tools#477 (`--customIndexTemplate` becoming a special case of a more general solution).

kelson42:
@mgautierfr Can you please review this Pr?

mgautierfr (approved):
LGTM.

It would be good to have a homogenized code for this. I like the idea of having a file describing the resources (and the mimetype). We may move this on a lower level, in the "resource engine".
But for a next PR.

mgautierfr (changes requested):
Tests are missing

veloman-yunkan:
> Tests are missing

Added a couple of tests that provide only basic coverage of the new code. The last commit outlines an aspect of the desired functionality that is not tested.

mgautierfr (approved):

PR #731, opened 2022-03-23
Render xml result - opensearch

mgautierfr:
Fixes #586

PR #780, opened 2022-06-04
Deduplication of server search results unittests

veloman-yunkan:
A follow-up of https://github.com/kiwix/libkiwix/pull/731#discussion_r888832663


test/server_search.cpp
```diff
@@ -736,11 +806,70 @@ struct TestData
...
+    std::string expectedXmlResultsString() const
+    {
+      if ( results.empty() ) {
+        return "\n    ";
+      }
+
+      std::string s;
+      for ( const auto& r : results ) {
+        s += "\n    <item>\n";
+        s += maskSnippetsInXmlSearchResults(r.getXml());
```
mgautierfr, line 849: I wonder if we could generate xml (and html) without snippets instead of generated a full xml and mask the snippet.

test/server_search.cpp
```diff
@@ -1319,5 +1456,13 @@ TEST_F(ServerTest, searchResults)
     const auto htmlRes = taskbarlessZimFileServer().GET(htmlSearchUrl.c_str());
     EXPECT_EQ(htmlRes->status, 200);
     t.checkHtml(htmlRes->body);
+
+    const std::string xmlSearchUrl = t.xmlSearchUrl();
+    const auto xmlRes1 = zfs1_->GET(xmlSearchUrl.c_str());
+    const auto xmlRes2 = taskbarlessZimFileServer().GET(xmlSearchUrl.c_str());
```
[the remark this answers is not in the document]
veloman-yunkan: I think that I don't understand this question.

test/server_search.cpp
veloman-yunkan: That would mean having a special mode for the /search endpoint (e.g. `/search?snippets=no`), however I don't think that adding it for testing alone makes sense. Besides, we ignore snippets only at this point. Later snippets are validated too.

mgautierfr: I was speaking about the generation of the expected result. We would still have to mask the snippet in the actual results of the tested server.

test/server_search.cpp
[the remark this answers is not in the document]
veloman-yunkan: Here the same request is made for both modes (enabled and disabled) of taskbar injection and next we check that the responses are identical. Then we proceed to validating only one of the responses. What's wrong with it?

test/server_search.cpp
veloman-yunkan: That makes sense. But not to an extent to pursue it if this is going to be the only thing to fix about this PR.

veloman-yunkan: Frankly I would rather change `TestData::expectedXmlResultsString()` to return the real expected response body and perform the masking out of snippets at the check site. That will eliminate some confusion implanted in the current code.

test/server_search.cpp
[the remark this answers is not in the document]
mgautierfr: > What's wrong with it?

Absolutely nothing. I've just totally miss the line 1461. My bad.

test/server_search.cpp
mgautierfr: > But not to an extent to pursue it if this is going to be the only thing to fix about this PR.

Agree with that. 

> Frankly I would rather change TestData::expectedXmlResultsString() to return the real expected response body and perform the masking out of snippets at the check site.

This is the other way around. Feel free to do it (but not in this PR, I approve and merge it)

mgautierfr (approved):

PR #781, opened 2022-06-09
Remove the java wrapper.

mgautierfr:
The meson's `wrapper` option is removed.
New meson's option `static-linkage` is added to tell meson to link with static library.


eli-schwartz:
> New meson's option `static-linkage` is added to tell meson to link with static library.

Meson in git master has a global builtin option for this, incidentally: https://mesonbuild.com/Release-notes-for-0-63-0.html#new-prefer_static-builtin-option

eli-schwartz:
Yeah, my comment was just a heads-up. Like I said, even the last version of Meson doesn't have it, it's only present in git master. :p

PR #783, opened 2022-06-10
Do not use deprecated constructor for Reader.

mgautierfr:
We have a specific private non deprecated constructor especially for that,
let's use it.


mgautierfr:
Fixes are easy and already tested on kiwix-build side (https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/43819342)
I'm merging without review

PR #786, opened 2022-06-23
More tiles on kiwix-serve welcome page (optimised)

juuz0:
Fixes https://github.com/kiwix/kiwix-tools/issues/538

Before:
image
image

Now:
image
image

### DIFF
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -31,7 +31,7 @@ body {
     font-size: 1.6rem;
     color: #333333;
     font-family: poppins;
-    margin: 10px 0 -5px;
+    margin: -10px 0 -5px;
     position: relative;
     top: -10px;
     left: 13px;
@@ -129,9 +129,12 @@ body {
 }
 
 .kiwixHomeBody {
-    padding: 20px 15vw;
+    padding: 20px 0;
     min-height: calc(100vh - 170px);
     position: relative;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
 }
 
 .tagFilterLabel {
@@ -157,6 +160,8 @@ body {
 
 .book__list {
     position: relative;
+    margin: 0 auto;
+    padding-top: 20px;
 }
 
 .book {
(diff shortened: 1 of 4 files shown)


juuz0:
Though this can be tested from a user-perspective, I have to check something in code hence the WIP status :+1:

kelson42:
@juuz0 Thank you for your PR, but I can’t believe this is the right approach. Far too complex and no padding change should be necessary.

mgautierfr:
In https://svelte.dev/repl/2b6fd14cbd454b339dc8c55ea32d67e0?version=3.48.0 I've simply use a flexbox. The flexbox "compute" the best number of columns and adjust the padding/margin to have a proper layout.
There is no complex computation on our side.
You can try it by changing the width of the result preview.

juuz0:
@kelson42 ping for re-review

kelson42 (changes requested):
@juuz0 Thank you for this new PR which seems really more reasonable. I few remarks:
* My personal main problem was the problem visiblem on my iPad. It seems to be fixed now.
* I have added a commit to better align the filters on the left. Hope this is OK for you.
* I want the "number of books" title align left to the tile at the very left
* On my computer I think the margin left and right are still too big (see screenshot). Why there is not one additional tile in the row? 
image

juuz0:
I have removed the centering layout, this is back to normal with a reduced left and right space.
Centring the tiles caused the problem with label, if there were 1 tile, 2 tiles, ....X tiles, it had to be changed all the time. Not easy to calculate correctly.

>  have added a commit to better align the filters on the left. Hope this is OK for you.

Yes, looks better.

image
image

kelson42:
@juuz0 This is what I get:
image

* Margin right bigger than margin left
* There is still place for one tile

I'm not against center the tiles, but secure the "X books" title is aligned to the most left tile

juuz0:
@kelson42 Review ping

kelson42:
@juuz0 title is ok, but second problem is not fixed,
image

There is still plently of place for an additional tile in a row. Why you don't reduce the margins left and right?!

juuz0:
@kelson42 One more review ping :'D

kelson42 (approved):
From a user perspective, it finally works fine!

PR #789, opened 2022-06-24
Remove libzim's wrapper.

### LINKED ISSUE #430: Drop libzim wrapper

mgautierfr:
It is time to remove them. They are deprecated since 10.0.0

Fixes #430


test/server_search.cpp
```diff
@@ -692,11 +692,11 @@ struct TestData
     std::string url = "/ROOT/search?" + query;
 
     if ( start >= 0 ) {
-      url += "&start=" + to_string(start);
+      url += "&start=" + std::to_string(start);
```
veloman-yunkan, line 695: I think all `to_string()`s in this file are supposed to be `kiwix::to_string()` from `src/tools/stringTools.h`. https://github.com/kiwix/libkiwix/blob/b9e03d27726617c8c3305eede44f71df7f28e031/src/tools/stringTools.h#L56-L62

Though if the need for `kiwix::to_string()` has disappeared (since the platforms that didn't provide `std::to_string()` are no longer supported) we better drop it.

mgautierfr: It is supposed to be `std::to_string()` has it is this namespace that was used by default here.
But indeed, we should simply remove `kiwix::to_string()` (will do it in this PR)

kelson42:
@veloman-yunkan I guess this needs a feedback from you? Good to merge?

veloman-yunkan (approved):

PR #791, opened 2022-06-26
CI triggered on pull_request event

### LINKED ISSUE #410: CI should be run on pull_request event

kelson42:
New attempt to fix #410 and maybe superseed #409

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   Macos:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: macos-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
       - name: Setup python 3.10
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin Yes the doc is really unclear about new commits... I worried about the same. But, it works like expected.

kelson42:
> I agree with @mgautierfr in #409 and #410 that “CI” may not be the best name for this workflow.

"Continuous integration (CI) is the practice of automating the integration of code changes from multiple contributors into a single software project. It’s a primary [DevOps best practice](https://www.atlassian.com/devops/what-is-devops/devops-best-practices), allowing developers to frequently merge code changes into a central repository where builds and tests then run. Automated tools are used to assert the new code’s correctness before integration." https://www.atlassian.com/continuous-delivery/continuous-integration

This is exactly what this workflow does: checking code to secure it can be merged (and checking master to secure it is merged correctly and still run). I hardly see what could be a better name?

Issue #792, opened 2022-06-28
Tests are failing on AlpineLinux


kmmndr:
I'm trying to upgrade to [11.0.0](https://github.com/kiwix/libkiwix/releases/tag/11.0.0) on AlpineLinux (Musl-libc) but there is one test failing:

```
...
[==========] 52 tests from 3 test suites ran. (2159 ms total)
[  PASSED  ] 51 tests.       
[  FAILED  ] 1 test, listed below:
[  FAILED  ] LibraryServerTest.catalog_v2_languages
                                                                                       
 1 FAILED TEST                                                                         
stderr:                                                                                
===== Unhandled error : Entry redirect_loop.html is a redirect entry.
```

Full error log : [testlog.txt](https://github.com/kiwix/libkiwix/files/9003869/testlog.txt)

This issue wasn't on last release [10.1.1](https://github.com/kiwix/libkiwix/releases/tag/10.1.1)

## How to reproduce

Create the following Dockerfile

```Dockerfile
FROM alpine:edge

ENV pkgver 11.0.0

WORKDIR /buildir
RUN wget https://github.com/kiwix/libkiwix/archive/$pkgver.tar.gz \
 && tar xvfz $pkgver.tar.gz

WORKDIR /buildir/libkiwix-$pkgver

RUN apk add --no-cache build-base

RUN apk add --no-cache \
   curl-dev \
   gtest-dev \
   libmicrohttpd-dev \
   libzim-dev \
   meson \
   mustache \
   pugixml-dev \
   python3 \
   samurai \
   xapian-core-dev


# build
RUN meson . build --prefix /usr \
 && ninja -C build
```

Run `docker build -t kiwix-lib-test . && docker run --rm -it kiwix-lib-test ninja -C build meson-test`

veloman-yunkan:
The test `LibraryServerTest.catalog_v2_languages` fails with the following diff suggesting that under Alpine the ICU data misses language self-names:

```
@@ -25,5 +25,5 @@
   </entry>
   <entry>
-    <title>French</title>
+    <title>fran\xC3\xA7" "ais</title>
     <dc:language>fra</dc:language>
     <thr:count>1</thr:count>
@@ -35,5 +35,5 @@
   </entry>
   <entry>
-    <title>Russian</title>
+    <title>\xD1\x80\xD1\x83\xD1\x81\xD1\x81\xD0\xBA\xD0\xB8\xD0\xB9</title>
     <dc:language>rus</dc:language>
[...]

PR #795, opened 2022-07-07
Unit test for checking ICU data

veloman-yunkan:
This PR adds a unit test requested in https://github.com/kiwix/libkiwix/issues/792#issuecomment-1169955908.

A known issue with this PR is that `kiwix::ICULanguageInfo` introduced in it started as a free function but was later converted to a class, whereupon `stringTools.h` stopped being a good home for it.

### DIFF
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -22,7 +22,6 @@
 
 #include "kiwixlib-resources.h"
 #include <mustache.hpp>
-#include <unicode/locid.h>
 
 #include "tools/stringTools.h"
 #include "tools/otherTools.h"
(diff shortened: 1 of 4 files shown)


kelson42:
@mgautierfr Can you please review this PR?

mgautierfr (approved):
> A known issue with this PR is that kiwix::ICULanguageInfo introduced in it started as a free function but was later converted to a class, whereupon stringTools.h stopped being a good home for it.

`stringTools.h` is used internally only. It seems overkill to create another specific header for `ICULanguageInfo`. If in the future we think it would be time to move this, we will. But for now it is ok.

PR #796, opened 2022-07-13
No jquery

### LINKED ISSUE #369: Jquery code should not be part of the code
It is a dependence and should be downloaded.

### LINKED ISSUE #610: jQuery error logs on suggestion hover
Using latest nightly, when using the search bar, I can type in and get a list of suggestions.

When I hover over any of them, I get `Uncaught TypeError: jQuery is undefined` errors in the console (triggered in jquery-ui).

Screen Shot 2021-08-27 at 12 26 27

I can select/click and I am redirected properly to the suggestion though.

### LINKED ISSUE #719: Vulnerable Version of JQuery (v1.10.2) Used
Hello, I host a Kiwix-serve instance on my local network. I discovered while running vulnerability scans that Kiwix-serve uses an outdated version of JQuery (v1.10.2) that has multiple XSS vulnerabilities (CVE-2020-11022 and CVE-2020-11023 are the ones my scanner stated). Upon further researched I realized libkiwix makes use of the same JQuery version, so I decided it'd be best to only open an issue here to avoid a duplicate issue, since I believe Kiwix-tools pulls from here (let me know if I'm supposed to make an issue on the Kiwix-tools repo as well). I am not sure if there is actually a way to exploit the vulnerability but I didn't see any other issues mentioning this so I thought it's worth reporting. Apologies if I made a mistake somewhere reporting this, I'm new to all of this.

Edit: Forgot to mention in the original message, the outdated JQuery is located at libkiwix/static/skin/jquery-ui/external/jquery/jquery.js ([Link](https://github.com/kiwix/libkiwix/blob/master/static/skin/jquery-ui/external/jquery/jquery.js))

juuz0:
Fixes #610
Fixes #719
Fixes #369

Supersedes https://github.com/kiwix/libkiwix/pull/639


juuz0:
Didn't push in previous PR to not ruin it with my experiments :P
Rebased on master
I haven't rebased it on https://github.com/kiwix/libkiwix/pull/716 yet, as @mgautierfr told to do so. Will see that.

As a change to previous taskbar suggestions, I went with a hyperlink in suggestions, this gives us the ability to copy/open link address in new tab just how wikipedia does. If this is not needed, will pull back.
Fixed suggestions, submit button

kelson42:
@juuz0 Please rebase and secure CI is OK. There a problem currently with the packages, but the rest should be green.

test/server.cpp
```diff
@@ -442,13 +436,10 @@ std::string TestContentIn404HtmlResponse::expectedResponse() const
...
 )FRAG",
 
-    R"FRAG(
-  <link type="root" href="/ROOT"><link type="text/css" href="/ROOT/skin/jquery-ui/jquery-ui.min.css?cacheid=e1de77b3" rel="Stylesheet" />
-<link type="text/css" href="/ROOT/skin/jquery-ui/jquery-ui.theme.min.css?cacheid=2a5841f9" rel="Stylesheet" />
-<link type="text/css" href="/ROOT/skin/taskbar.css?cacheid=49365e9c" rel="Stylesheet" />
-<script type="text/javascript" src="/ROOT/skin/jquery-ui/external/jquery/jquery.js?cacheid=1d85f0f3" defer></script>
-<script type="text/javascript" src="/ROOT/skin/jquery-ui/jquery-ui.min.js?cacheid=d927c2ff" defer></script>
-<script type="text/javascript" src="/ROOT/skin/taskbar.js?cacheid=5982280c" defer></script>
+    R"FRAG(  <link type="root" href="/ROOT"><link type="text/css" href="/ROOT/skin/taskbar.css" rel="Stylesheet" />
```
mgautierfr, line 439: This remove a `\n` as the content starts just after the `R"FRAG(`. This is maybe why the tests are failing.

test/server.cpp
```diff
@@ -442,13 +428,11 @@ std::string TestContentIn404HtmlResponse::expectedResponse() const
...
 )FRAG",
 
-    R"FRAG(
-  <link type="root" href="/ROOT"><link type="text/css" href="/ROOT/skin/jquery-ui/jquery-ui.min.css?cacheid=e1de77b3" rel="Stylesheet" />
-<link type="text/css" href="/ROOT/skin/jquery-ui/jquery-ui.theme.min.css?cacheid=2a5841f9" rel="Stylesheet" />
-<link type="text/css" href="/ROOT/skin/taskbar.css?cacheid=49365e9c" rel="Stylesheet" />
-<script type="text/javascript" src="/ROOT/skin/jquery-ui/external/jquery/jquery.js?cacheid=1d85f0f3" defer></script>
-<script type="text/javascript" src="/ROOT/skin/jquery-ui/jquery-ui.min.js?cacheid=d927c2ff" defer></script>
-<script type="text/javascript" src="/ROOT/skin/taskbar.js?cacheid=5982280c" defer></script>
+    R"FRAG(
```
mgautierfr, line 431: You have two space after `R"FRAG(`

kelson42:
@mgautierfr Can you please reivew this PR ?

test/server.cpp
```diff
@@ -520,15 +521,11 @@ std::string TestContentIn404HtmlResponse::hiddenBookNameInput() const
...
-  return R"(          <input autocomplete="off" class="ui-autocomplete-input" id="kiwixsearchbox" name="pattern" type="text" title=")"
-       + searchboxTooltip
-       + R"(" aria-label=")"
-       + searchboxTooltip
-       + R"(">
+  return R"(          <input autocomplete="off" id="kiwixsearchbox" name="pattern" type="text" title="Search ')"
+       + bookTitle
+       + R"('" aria-label="Search ')"
+       + bookTitle
+       + R"('">
```
mgautierfr, line 521: Here we test the translation feature.
We may change a bit the test as we modify the class of the input, be we should still test the
title and aria-label is translated.

static/templates/taskbar_part.html
```diff
@@ -5,7 +5,7 @@
         <form class="kiwixsearch" method="GET" action="{{root}}/search" id="kiwixsearchform">
           {{#hascontent}}<input type="hidden" name="content" value="{{content}}" />{{/hascontent}}
           <label for="kiwixsearchbox">&#x1f50d;</label>
-          <input autocomplete="off" class="ui-autocomplete-input" id="kiwixsearchbox" name="pattern" type="text" title="{{{SEARCHBOX_TOOLTIP}}}" aria-label="{{{SEARCHBOX_TOOLTIP}}}">
+          <input autocomplete="off" id="kiwixsearchbox" name="pattern" type="text" title="Search '{{title}}'" aria-label="Search '{{title}}'">
```
mgautierfr: The original PR was based on a version without translation.
As we rebase the PR, we must not revert the translation feature.

static/skin/taskbar.css
```diff
@@ -123,22 +124,16 @@ label[for=kiwixsearchbox] {
     vertical-align: middle;
 }
 
+a.suggest, a.suggest:visited, a.suggest:hover, a.suggest:active {
+    display: block;
+    text-decoration: none;
+    color: inherit;
+}
```
mgautierfr, line 131: This probably change the color of all the link in the page and we don't want that

static/skin/taskbar.js
```diff
@@ -40,65 +41,17 @@ function on_ready() {
...
-            focus: function () {
-                $('.kiwix_searchform').addClass('full_width');
-                $('label[for="kiwix_button_show_toggle"], .kiwix_button_cont').addClass('searching');
-            },
-            blur: function () {
-                $('.kiwix_searchform').removeClass('full_width');
-                $('label[for="kiwix_button_show_toggle"], .kiwix_button_cont').removeClass('searching');
-            }
-        });
-    };*/
```
mgautierfr: I have commented this code because I don't know what to do with it (and I was not sure of the purpose of it).
We may remove the cybook hack if it is not relevant anymore (but we have to state that explicitly)
But the code with the scrool and focus/blur is probably useful at a moment.

mgautierfr (changes requested):

juuz0:
@kelson42 :white_check_mark: 

Question:
For the codefactor thing, it shows it can autofix, should we try that?
image

kelson42:
@juuz0 Honestly I prefer a manual fix.

kelson42 (approved):
From a user perspective, seems to work fine :)

kelson42:
@veloman-yunkan Considering that @mgautierfr is away for a while, could you please review this PR?

static/resources_list.txt
```diff
@@ -53,3 +34,8 @@ templates/url_of_search_results_css
 opensearchdescription.xml
 ft_opensearchdescription.xml
 catalog_v2_searchdescription.xml
+skin/css/autoComplete.01.css
+skin/css/autoComplete.02.css
+skin/css/autoComplete.css
+skin/css/images/magnifier.svg
+skin/css/images/search.svg
```
veloman-yunkan, line 42: `autoComplete.min.js` is missing

static/skin/index.js
```diff
@@ -320,7 +320,7 @@
         const booksToFilter = new Set();
         const booksToDelete = new Set();
         iso.arrange({
-            filter: function (idx, elem) {
+            filter: function (elem) {
```
veloman-yunkan, line 323: Why is this change needed?

veloman-yunkan (commented):
Since I haven't reviewed this PR before, I would prefer to see its clean version (with fixup commits squashed). But I would also like to know the intended plan of dealing with the conflicting PR #716 - is this PR intended to be merged into master before #716 or it must be rebased on top of the latter?

juuz0:
> But I would also like to know the intended plan of dealing with the conflicting PR https://github.com/kiwix/libkiwix/pull/716 - is this PR intended to be merged into master before https://github.com/kiwix/libkiwix/pull/716 or it must be rebased on top of the latter?

I believe it is intended to be rebased on top of it later (https://github.com/kiwix/libkiwix/pull/639#issuecomment-1168376186)

kelson42:
There PR is old and simple in comparison to the iframe one, I would like to merge it ASAP. I see no reason to make the other way around.

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
     return doc.documentElement.textContent;
 }
 
-const jq = jQuery.noConflict(true);
-jq(document).ready(() => {
-    (function ($) {
-        const root = $( `link[type='root']` ).attr("href");
+function on_ready() {
```
veloman-yunkan, line 26: This named functions asks for a conflict with javascript code coming from the ZIM file. Please use an unnamed function.

static/skin/taskbar.js
```diff
@@ -3,6 +3,26 @@ function htmlDecode(input) {
...
+        if (st > lastScrollTop) {
+            kiwixToolBar.style.top = '-100%';
+        } else {
+            kiwixToolBar.style.top = '0';
+        }
+
+        lastScrollTop = st;
+    }
+}
+
```
veloman-yunkan, line 25: This and related changes better be extracted into a separate commit.

veloman-yunkan (changes requested):
Selecting the suggestion using the keyboard (with arrow keys) and pressing ENTER runs a search for the text found in the searchbox instead of going to the suggested page.

Also I ran against an issue mostly unrelated to this PR but the previous version seemed to work fine under those circumstances. In my browser suggestions are not displayed nicely on the main page of the Ray Charles ZIM:

Screenshot from 2022-07-23 16-36-24

Also the styling of the taskbar slightly changes between the said page and other pages. Firefox reports the following warning in the console which probably explains the issue:

>  This page is in Quirks Mode. Page layout may be impacted. For Standards Mode use “<!DOCTYPE html>”

Regarding the history of this PR - I would say that it is somewhat messy. The need for the new unit test in the context of this PR is not clear at all. Next, Jquery is removed first and the problems are then fixed. Such style of work can lead to problems if a bigger change is undertaken. A safer approach of dropping something is to first eliminate the dependency on it and only then removing it. So a better sequence would be:

1. Migrate the suggestions UI to autocomplete.js
2. Say thank you and good-bye to jquery.

juuz0:
> Selecting the suggestion using the keyboard (with arrow keys) and pressing ENTER runs a search for the text found in the searchbox instead of going to the suggested page.

Done, thank you.

> Also the styling of the taskbar slightly changes between the said page and other pages. 

:white_check_mark: 

> The need for the new unit test in the context of this PR is not clear at all.

I agree, let's ping @mgautierfr for clarification :D

> Regarding the history of this PR - I would say that it is somewhat messy.

Here's a try to make it less messy ^^

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
...
+            } catch (error) {
+              return error;
+            }
+          }
+        },
+        searchEngine: (query, record) => {
+          // We accept all records
+          return true;
+        },
+        keys: ['label'],
```
veloman-yunkan: This seems to be a wrong configuration option and is simply ignored by `autoComplete()`

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
...
+        resultsList: {
+          noResults: true,
+          /* We must display 10 results (requested) + 1 potential link to do a full text search. */
+          maxResults: 11,
```
veloman-yunkan: What if you just omit this option? Won't all suggestions sent by the server be displayed?

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
...
+          highlight: true
```
veloman-yunkan: According to [documentation](https://tarekraafat.github.io/autoComplete.js/#/configuration) this should belong to `resultItem`

veloman-yunkan (changes requested):

kelson42:
@juuz0 Please close conversations with a comment if you are so far.

static/skin/taskbar.js
veloman-yunkan: I saw that this comment was marked resolved without being addressed. After checking the documentation I found out that the default value for `maxResults` is 5, which perfectly justifies specifying the desired value here.

But please don't resolve comments like that.

static/skin/taskbar.js
```diff
@@ -76,9 +76,9 @@ document.addEventListener('DOMContentLoaded', function () {
         element: (item, data) => {
           let searchLink;
           if (data.value.kind == "path") {
-            searchLink = `${root}/${bookName}/${encodeURI(data.value.path)}`;
+            searchLink = `${root}/${bookName}/${encodeURIComponent(data.value.path)}`;
           } else {
-            searchLink = `${root}/search?content=${bookName}&pattern=${data.value.value}`;
+            searchLink = `${root}/search?content=${bookName}&pattern=${encodeURIComponent(data.value.value)}`;
```
veloman-yunkan, line 81: `data.value.path` and `data.value.value` must be `htmlDecode()`d before being URI-encoded.

Check with R&B on the Ray Charles ZIM.

1. The suggestion "Let's Go Get Stoned (R&B song)" points to http://localhost:8080/ray_charles/A%2FLet%26apos%3Bs_Go_Get_Stoned_(R%26amp%3BB_song) with a **Not Found** result

2. The search suggestion "containing 'R&B'..." points to http://localhost:8080/search?content=ray_charles&pattern=R%26amp%3BB%20 leading to No results were found for **"R&B "**

veloman-yunkan (changes requested):
Please don't just fix review comments but also test the functionality.

static/skin/taskbar.js
juuz0: @veloman-yunkan GitHub bug?

I replied to every comment except one (suggestion to use title of kiwixsearchbox one)
image

juuz0:
> Please close conversations with a comment if you are so far.

@kelson42 Eh, I reply to every comment usually. So weird now that I don't see them in incognito mode but see when logged in (shows Pending though for days)
image

juuz0:
So weird :facepalm:
Found the solution, I never had this problem earlier. So sorry for the trouble
https://github.com/community/community/discussions/10369

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
...
+              return data;
+            } catch (error) {
+              return error;
+            }
+          }
+        },
+        searchEngine: (query, record) => {
+          // We accept all records
+          return true;
+        },
```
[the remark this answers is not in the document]
veloman-yunkan: The documentation is very poor with regards to the explanation of the `searchEngine` option. Did you make your conclusion by reading the source code of autocomplete.js?

static/skin/taskbar.js
veloman-yunkan: I meant the `highlight` option. `resultList` doesn't have one

static/skin/taskbar.js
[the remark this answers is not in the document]
juuz0: No, that was added by @mgautierfr . I'm in favour of choosing the `loose` option though, much more clear.

static/skin/taskbar.js
juuz0: Done.

static/skin/taskbar.js
juuz0: This is fixed, but I would like to bring up another similar problem which is also present in old taskbar.
Searching for R$B, R#B doesn't bring up correct results
Old taskbar results (same with new) - 
image

veloman-yunkan: This issue (unlike the previous one) belongs to the search engine (Xapian). We shouldn't do anything in the front-end to handle it. Any workarounds must be implemented in the back-end.

static/skin/taskbar.js
```diff
@@ -3,6 +3,39 @@ function htmlDecode(input) {
     return doc.documentElement.textContent;
 }
 
+function handleScrollFor520Width() {
```
veloman-yunkan, line 6: Including the width value in the function name is too much detail. `handleScrollForNarrowClient()` is better but still incorrect since this function doesn't *handle* scrolling but rather *sets up* handling of scrolling. A good name is `setupAutoHidingOfTheToolbar()`. The relation to the small client isn't reflected in this proposed name since nothing in the implementation depends on the size of the window and it would do exactly the same in a large window too. The conditions under which we decide to enable this feature matter only at the call site of this function.

static/skin/taskbar.js
```diff
@@ -3,6 +3,39 @@ function htmlDecode(input) {
...
+  setInterval(function() {
+    if (didScroll) {
+      hasScrolled();
+      didScroll = false;
+    }
+  }, 250);
+
+  function hasScrolled() {
+    let st = document.documentElement.scrollTop || document.body.scrollTop;
+    const kiwixToolBar = document.querySelector('#kiwixtoolbar');
```
veloman-yunkan, line 24: Move `kiwixToolBar` into the outer function

static/resources_list.txt
```diff
@@ -53,3 +35,7 @@ templates/url_of_search_results_css
 opensearchdescription.xml
 ft_opensearchdescription.xml
 catalog_v2_searchdescription.xml
+skin/css/autoComplete.01.css
+skin/css/autoComplete.02.css
+skin/css/autoComplete.css
```
veloman-yunkan, line 40: These two  (`autoComplete.css` and `autoComplete.02.css`) are not used. Why not keeping only `autoComplete.01.css` but renaming it to `autoComplete.css`?

veloman-yunkan (changes requested):
We've approached the final review. Please rebase-fixup and fix the history of the PR as follow:

1. I don't think we need the first two commits (the new unit test and the .gitignore) in this PR.
2. The commit titled "Update `index.js` to not use jquery anymore" should be taught good manners. Squeezing oneself between two logically related changes is impolite.
3. Having a separate change for fixing codefactor warnings is not justified.

static/skin/taskbar.js
kelson42: @juuz0 I woukd recommend to open a dedicated ticket, but not sure something can be done.

static/resources_list.txt
juuz0: Yes that's right, were bothering me since I took over this PR. Done, thanks.

static/skin/taskbar.js
juuz0: :white_check_mark:

static/skin/taskbar.js
juuz0: Thanks, much better.

static/skin/taskbar.js
```diff
@@ -3,46 +3,51 @@ function htmlDecode(input) {
...
+    {
+      selector: "#kiwixsearchbox",
+      placeHolder: document.querySelector("#kiwixsearchbox").title,
+      threshold: 1,
+      debounce: 300,
+      data : {
+        src: async (query) => {
+          try {
+            // Fetch Data from external Source
+            const source = await fetch(`${root}/suggest?content=${bookName}&term=${encodeURIComponent(query)}`);
```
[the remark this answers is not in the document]
juuz0: Done, thanks, never knew about `test/data`

test/server.cpp
```diff
@@ -189,7 +189,7 @@ R"EXPECTEDRESULT(                                <img src="../skin/download.png?
       /* url */ "/ROOT/zimfile/A/index",
 R"EXPECTEDRESULT(<link type="root" href="/ROOT"><link type="text/css" href="/ROOT/skin/taskbar.css?cacheid=26082885" rel="Stylesheet" />
 <link type="text/css" href="/ROOT/skin/css/autoComplete.01.css?cacheid=08951e06" rel="Stylesheet" />
-<script type="text/javascript" src="/ROOT/skin/taskbar.js?cacheid=68384d57" defer></script>
+<script type="text/javascript" src="/ROOT/skin/taskbar.js?cacheid=a95561f5" defer></script>
```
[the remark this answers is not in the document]
juuz0: Going with the second option

static/skin/taskbar.js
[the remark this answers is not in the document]
veloman-yunkan: I briefly looked at the source code but couldn't fully figure out all the subtleties. Let's keep @mgautierfr's version

static/skin/taskbar.js
```diff
@@ -3,6 +3,39 @@ function htmlDecode(input) {
     return doc.documentElement.textContent;
 }
 
+function setupAutoHidingOfTheToolbar() {
+  let lastScrollTop = 0;
+  let delta = 5;
```
veloman-yunkan, line 8: `let` → `const`

static/skin/taskbar.js
```diff
@@ -3,6 +3,39 @@ function htmlDecode(input) {
...
+  let didScroll = false;
+  const kiwixToolBar = document.querySelector('#kiwixtoolbar');
+
+  window.onscroll = function (e) {
+    didScroll = true;
+  }
```
veloman-yunkan, line 14: This cancels the previous `onscroll` handler (if any)

static/skin/taskbar.js
```diff
@@ -3,6 +3,39 @@ function htmlDecode(input) {
...
+
+  setInterval(function() {
+    if (didScroll) {
+      hasScrolled();
+      didScroll = false;
+    }
+  }, 250);
+
+  function hasScrolled() {
+    let st = document.documentElement.scrollTop || document.body.scrollTop;
```
veloman-yunkan, line 24: `let` → `const`

veloman-yunkan (changes requested):
A few final touches

static/skin/taskbar.js
juuz0: used an eventHandler instead

static/skin/taskbar.js
```diff
@@ -3,46 +3,52 @@ function htmlDecode(input) {
...
-const jq = jQuery.noConflict(true);
-jq(document).ready(() => {
-    (function ($) {
-        const root = $( `link[type='root']` ).attr("href");
+document.addEventListener('DOMContentLoaded', function () {
+  const root = document.querySelector(`link[type='root']`).getAttribute("href");
+  let bookName = (window.location.pathname == `${root}/search`)
+    ? (new URLSearchParams(window.location.search)).get('content')
+    : window.location.pathname.split(`${root}/`)[1].split('/')[0];
+  bookName = encodeURIComponent(htmlDecode(bookName));
```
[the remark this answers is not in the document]
juuz0: Yeah, I shouldn't. I have also only url-encoded it in 
` const source = await fetch(${root}/suggest?content=${encodeURIComponent(bookName)}&term=${encodeURIComponent(query)});`
  now. If whole `bookName` was url encoded it caused problems with the `zimfile&other.zim` file.

static/skin/taskbar.js
```diff
@@ -36,15 +37,32 @@ document.addEventListener('DOMContentLoaded', function () {
...
-        highlight: true
       },
       resultItem: {
         element: (item, data) => {
-          item.innerHTML = `${htmlDecode(data.value.label)}`;
+          let searchLink;
+          if (data.value.kind == "path") {
+            searchLink = `${root}/${bookName}/${htmlDecode(data.value.path)}`;
+          } else {
+            searchLink = `${root}/search?content=${bookName}&pattern=${encodeURIComponent(htmlDecode(data.value.value))}`;
```
veloman-yunkan, line 80: Now you've undone the fix of https://github.com/kiwix/libkiwix/pull/796#discussion_r933761318

veloman-yunkan (changes requested):
In a PR with multiple commits and changes amounting to more than a couple dozen lines, please don't fix comments in place unless asked to do so. Fixup commits help the reviewer to quickly analyze all changes since the last review instead of trying to be more careful catching modifications in excess of what was requested.

static/skin/taskbar.js
juuz0: Yes, for reason (https://github.com/kiwix/libkiwix/pull/796#discussion_r933978052)
URL encoding `bookName` here will lead to a URL like 
`http://192.168.18.8:8080/zimfile%26other/A/Heaven_Knows_(Taylor_Hicks_song)`
After clicking that suggestion if we search again, the `bookName` that is sent will be:
`zimfile%26other`
This is URL encoded to
`zimfile%2526other`
(`%` turns into `%25`)
Fetch fails, no suggestions.

veloman-yunkan: Both in the original comment and now too I was only referring to the usage of `bookName` in `/search?content=${bookName}...`

veloman-yunkan: It definitely has to be done. If the `/search` endpoint then puts a URL-encoded book name in the links in the search results, that has to be fixed in the back-end.

veloman-yunkan: Alternatively, we may always URL-decode `bookName` in the front-end

juuz0: Ok all the time, I kept thinking it was said for both.
Sending a fixup

veloman-yunkan (commented):
Great. Time to rebase, fixup and put the final cacheids!

juuz0:
Done.
Now time for a doubt I had since forever,
what's the correct way to find cache IDs?
Currently, I compile, run the tests, let them fail. Go to testlog, find the correct cacheid -> replace. Is the only way?

veloman-yunkan:
> Now time for a doubt I had since forever, what's the correct way to find cache IDs? Currently, I compile, run the tests, let them fail. Go to testlog, find the correct cacheid -> replace. Is the only way?

That's how I do it too. The actual values of cacheids don't matter much. The important behaviour that is being checked is that whenever certain files change links to those resources get updated.

veloman-yunkan (approved):

veloman-yunkan:
> From a user perspective, seems to work fine :)

@kelson42 Please validate the PR from user perspective once again before merging.

kelson42:
@veloman-yunkan I will, glad to see ir ready. Thx a lot for all the review work.

static/skin/taskbar.js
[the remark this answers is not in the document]
mgautierfr: Documentation is poor yes. But from the code source https://github.com/TarekRaafat/autoComplete.js/blob/d300e47f776de0d1662d5d9b8ba8a406a6ebc22e/src/controllers/dataController.js#L39-L68, the function is used to filter the data returned by `getData` with the query of the user. As we want to accept everything, there is no real "match" string, but we must return something "true" to pass the `if (!match) return;`

mgautierfr:
Thanks to both of you to finish and cleanup this PR !

PR #797, opened 2022-07-16
add widget support

juuz0:
Widget endpoint: /widget
Testing endpoint (temporary): /widgettest (has an iframe to see contents)
Fixes https://github.com/kiwix/libkiwix/issues/585


juuz0:
Will fix the code coverage :+1: 
Obviously not completed but I think we can review it from a user perspective with its current form and if this the way to move forward

kelson42:
@juuz0 I don't see any changes folowing @mgautierfr! Any problem?

juuz0:
@kelson42 https://libkiwix--797.org.readthedocs.build/en/797/widget.html

kelson42:
@juuz0 Can you please rebase (and resolve the conflicts) now that we have merged the "no jquery" PR?

juuz0:
@kelson42 Done.

juuz0:
@kelson42 Any updates here? :)

kelson42 (changes requested):
Thank you @juuz0 For this first version. It start to be like I expected. Here a few question/remarks:
* The CSS example does not work for me (i see nothing in red). Are you sure it is OK?
* We should have a way to filter normaly (via URL) like on library.kiwix.org, it seems to not be the case
* The ability to show one or many ZIM files based on their name (`M/Name`) should work as well on welcome page (and should probably be done via a dedicated PR).

juuz0:
> The CSS example does not work for me (i see nothing in red). Are you sure it is OK?

I've checked again, it works on my side with the example in the docs. What browser do you use? I tried on Vivaldi (Chromium based) and Firefox. Exact code of html file?

> We should have a way to filter normaly (via URL) like on library.kiwix.org, it seems to not be the case

These work for me again!

> http://localhost:8181/widget?book=gutenberg_fa_all_2022-01 does not deliver only one book with the M/Name value = gutenberg_fa_all_2022-01

Yes I confirm this mistake, happened during a recent rebase.

> The ability to show one or many ZIM files based on their name (M/Name) should work as well on welcome page (and should probably be done via a dedicated PR).

Looking through the C++ code, I have found there's indeed a way to filer by book name already. Though with two problems:
1. It doesn't work (`http://192.168.18.8:8080/?name=gutenberg_fa_all_2022-04` should give the book but it doesn't, I have found the fix and will be fixing it in the dedicated PR)
2. It isn't made to accept multiple values which we want here, doing that too.

kelson42:
> > The CSS example does not work for me (i see nothing in red). Are you sure it is OK?
> 
> I've checked again, it works on my side with the example in the docs. What browser do you use? I tried on Vivaldi (Chromium based) and Firefox. Exact code of html file?

I test with Firefox, the code comes from the your documentation:
```html
$ cat widget.html 
<!DOCTYPE html>
<html lang="en">
<head>
<title>Widget Test</title>
</head>
<body>
  <iframe src="http://localhost:8181/widget" width=1000 height=1000></iframe>

  <script>
window.onload = function() {
var receiver = document.getElementById('receiver').contentWindow;
function sendMessage() {
    let msg = {
    css: `
    .book__header {
        color:red;
    }`,
    js: `
    function widgetTest() {
    console.log("Testing widget");
    }
    widgetTest();
    `
    }
    receiver.postMessage(msg, 'http://192.168.18.8:8080/widget');
}
sendMessage();
}
</script>

</body>
</html>
```
and the result looks like

image

I see nothing which is different than without custom CSS. Considering that there is nothing which has the DOM `id="receiver"`, I suspect the documentation to be simply wrong or incomplete.

> > We should have a way to filter normaly (via URL) like on library.kiwix.org, it seems to not be the case
> 
> These work for me again!

Indeed, seems to work. Sorry I can not reproduce the problem.

> > http://localhost:8181/widget?book=gutenberg_fa_all_2022-01 does not deliver only one book with the M/Name value = gutenberg_fa_all_2022-01
> 
> Yes I confirm this mistake, happened during a recent rebase.
> 
> > The ability to show one or many ZIM files based on their name (M/Name) should work as well on welcome page (and should probably be done via a dedicated PR).
> 
> Looking through the C++ code, I have found there's indeed a way to filer by book name already. Though with two problems:
> 
>     1. It doesn't work (`http://192.168.18.8:8080/?name=gutenberg_fa_all_2022-04` should give the book but it doesn't, I have found the fix and will be fixing it in the dedicated PR)
> 
>     2. It isn't made to accept multiple values which we want here, doing that too.

All of this seems to be the same problem and actually is 100% independent of the widget system IMO. Can you please open a ticket dedicated to this (and also work on the solution).

An other detail, "Powered by Kiwix" should not be part of the widget IMO. I continue with my testing....

kelson42:
Modulo the remarks done earlier, this first version of the new `/widget` feature looks good to me. Adding @veloman-yunkan for the code review

juuz0:
> I see nothing which is different than without custom CSS. Considering that there is nothing which has the DOM id="receiver", I suspect the documentation to be simply wrong or incomplete.

The documentation does have the `id=receiver` in iframe, only in CSS/JS section :)
There are two mistakes in the code you sent:
1. `id=receiver` in iframe
2. `receiver.postMessage(msg, <URL here>);` and `<iframe src="URL here" width=1000 height=1000></iframe>` should have the same URL (I should probably mention it clearly in docs)
Here's the fixed version:
```
<!DOCTYPE html>
<html lang="en">
<head>
<title>Widget Test</title>
</head>
<body>
  <iframe id="receiver" src="http://localhost:8181/widget" width=1000 height=1000></iframe>

  <script>
window.onload = function() {
var receiver = document.getElementById('receiver').contentWindow;
function sendMessage() {
    let msg = {
    css: `
    .book__header {
        color:red;
    }`,
    js: `
    function widgetTest() {
    console.log("Testing widget");
    }
    widgetTest();
    `
    }
    receiver.postMessage(msg, 'http://localhost:8181/widget');
}
sendMessage();
}
</script>

</body>
</html>
```

juuz0:
> An other detail, "Powered by Kiwix" should not be part of the widget IMO

Done.
Also the commit which provided a way to filter books by name (should be done in C++ side in other PR)

juuz0:
@kelson42 #807

static/skin/index.js
```diff
@@ -265,16 +271,20 @@ const kiwixServe = (function() {
...
-                const hfTitle = humanFriendlyTitle(title);
-                if (valueEntryNode == 'language') {
-                    languages[value] = hfTitle;
-                }
-                optionStr += (hfTitle != '') ? `<option value="${value}">${hfTitle}</option>` : '';
-            });
-            document.querySelector(nodeQuery).innerHTML += optionStr;
+            const entryList = data.querySelectorAll('entry');
+            const nodeQueryElem = document.querySelector(nodeQuery);
+            if (entryList && nodeQueryElem) {
```
veloman-yunkan, line 276: When `nodeQueryElem == null` doesn't the `loadAndDisplayOptions()` function become a no-op? If so why run `fetch(query)` at all?

docs/widget.rst
```diff
@@ -0,0 +1,82 @@
...
+Currently, the following arguments are supported.
+
+disabledesc (value = N/A)
+    Disables the description part of a tile.
+
+disablefilter (value = N/A)
+    Disables the search filters: language, category, tag and search function.
+
+disableclick (value = N/A)
+    Disables clicking the book to open it for reading.
```
veloman-yunkan, line 41: For me (Firefox 103.0) a blank tab is opened

static/skin/widget.js
```diff
@@ -1 +1,87 @@
...
+    const hideResultsLabelRule = `
+    .kiwixHomeBody__results {
+        display: none;
+    }`;
+    const hideTagFilterRule = `
+    .book__tags {
+        pointer-events: none;
+    }`;
+    insertNewCssRules(widgetStyles, [hideNavRule, hideResultsLabelRule, hideTagFilterRule]);
+}
```
veloman-yunkan, line 15: I think that if you introduce custom CSS handling first, then `disableSearchFilters()`, `disableDownload()`, `disableDescription()` and `hideFooter()` can be rewritten via `addCustomCss()` in a simpler way.

kelson42:
@juuz0 Any feedback?

PR #800, opened 2022-07-20
Exit download modal on pressing escape key.

### LINKED ISSUE #799: ESC keyboard key stroke should escape from Download modal view
If I have something like that:
image

Then typing on ESC on the keyboard should exit the modal view (so like clicking on the modal box X button).

juuz0:
Adds an event listener to call closeModal() when Escape key is pressed.
Fixes #799

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -439,6 +439,12 @@
 
     window.addEventListener('scroll', loadSubset);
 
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
@juuz0 Thx, works fine.

PR #801, opened 2022-07-25
PPA: Add kinetic

legoktm:
Not sure why this is broken, debugging...

PR #804, opened 2022-07-31
Illustration URL uses the book UUID

### LINKED ISSUE #755: Why use human ID in OPDS illustration endpoint?
Illustration URLs from the OPDS are of the following format

```
/catalog/v2/illustration/ted_en_playlist-open-source-open-world_2021-12/?size=48
```

What is the reasoning behind using the *dated-human ID* instead of the Book ID in this endpoint?

veloman-yunkan:
Fixes #755

If the server is initialized with a library.xml file, then the id specified in the XML file is used (rather than the UUID recorded in the
ZIM file).

Note that in test/data/library.xml the book ids are fake and different from the real ZIM IDs; that file was created for testing of the /catalog endpoint which doesn't access ZIM content, so the the same ZIM file zimfile.zim was added to library.xml three times as
three different books (with unique human-friendly ids). This explains the diff in test/library_server.cpp.

### DIFF
--- a/src/server/internalServer_catalog_v2.cpp
+++ b/src/server/internalServer_catalog_v2.cpp
@@ -154,8 +154,7 @@ std::unique_ptr<Response> InternalServer::handle_catalog_v2_languages(const Requ
 std::unique_ptr<Response> InternalServer::handle_catalog_v2_illustration(const RequestContext& request)
 {
   try {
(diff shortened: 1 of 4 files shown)


mgautierfr (approved):

mgautierfr:
Ci is broken for impish packages but it will be fixed with #801

PR #806, opened 2022-08-05
/content endpoint

veloman-yunkan:
This PR was extracted from #716 so that the latter can be rewritten on top of it in a cleaner way.

Introducing `/content` endpoint makes sense on its own, so this PR is justified even if an unexpected finding makes us discard the idea of switching to an iframe-based viewer.

User observable changes are as follows:

* A new `/content` endpoint is added.
* Book content is primarily served under `/content/<book>/<path>/<to>/<entry>`.
* `/<book>/<path>/<to>/<entry>` URLs are redirected to `/content/<book>/<path>/<to>/<entry>`.


test/server.cpp
```diff
@@ -268,18 +251,12 @@ TEST_F(ServerTest, 400)
 const char* urls404[] = {
   "/",
   "/zimfile",
-  "/ROOT/non-existent-item",
```
mgautierfr, line 251: Shouldn't we check that all those removed urls404 are now redirections ?
(Maybe not all as they may now fall in a common scheme, but at least few of them)

mgautierfr (changes requested):

test/server.cpp
veloman-yunkan: Good suggestion. I took advantage of it to improve the `ServerTest.NonEndpointUrlsAreRedirectedToContentUrls` unit-test even more.

mgautierfr:
We are good. Please rebase-fixup before final approval

veloman-yunkan:
Done

mgautierfr (approved):

PR #808, opened 2022-08-15
Filter ZIMs by filename in catalog/search

### LINKED ISSUE #807: Catalog filter using ZIM filename
Currently, if I visit:
`http://library.kiwix.org/catalog/search?name=archlinux_en_all_nopic_2022-05`
I get no books.
The `name` is not actually the filename so to get the desired book I need to visit:
`http://library.kiwix.org/catalog/search?name=archlinux_en_all`
and then search through the X number of books with same name that it gives. Hardly flexible. This is needed in `/widget` for now but may be needed in something else later on.
Apologies if this already exists, I couldn't find please guide if it does exist. :)
If it doesn't, I propose we add a new filter `book` for this filename.
`http://library.kiwix.org/catalog/search?book=archlinux_en_all_nopic_2022-05`  when filename is `archlinux_en_all_nopic_2022-05`

juuz0:
Fixes #807


juuz0:
Example test link:
`http://192.168.18.8:8080/?book=termux_en_all_maxi_2022-05&book=installgentoo_en_all_nopic_2019-09`
image

kelson42:
Pretty sure that installgentoo_en_all_nopic_2019-09 is not the zim name.

juuz0:
It's the filename (whatever it is on system, you can try changing a file name and see).
ZIM's name field is not unique as per my observation (for example archwiki_en_all__nopic and archwiki_en_all__maxi have the same name: archwiki_en_all)

juuz0:
@kelson42 Updated to use alias name.
image

kelson42:
@juuz0 Thx, it's great if can now work with both filename and aliases. I will test it. Reading the source code, I have too remarks:
* supporting of aliases shoud probably works only if `--nodatealiases` is used (just to be coherent)
* I would reuse the same primitive like there to compute the alias name

kelson42:
@juuz0 Any feedback?

src/server/internalServer.cpp
```diff
@@ -119,6 +119,9 @@ Filter get_search_filter(const RequestContext& request, const std::string& prefi
     try {
       filter.rejectTags(kiwix::split(request.get_argument(prefix+"notag"), ";"));
     } catch (...) {}
+    try {
+      filter.aliasName(request.get_argument(prefix + "book"));
```
veloman-yunkan: My guess is that selection of books by alias/filename should not be combined with other filtering criteria. If so, then it rather must be done in `InternalServer::selectBooks()` (similar to selection by id or name).

veloman-yunkan (changes requested):
No unit tests have been added!

src/server/internalServer.cpp
juuz0: Sorry, I didn't understand this fully. We need a similar way to filter using query string (just like lang/category/tag)

src/library.cpp
```diff
@@ -461,6 +461,9 @@ void Library::updateBookDB(const Book& book)
   indexer.index_text(normalizeText(book.getPublisher()), 1, "XP");
   indexer.index_text(normalizeText(book.getName()),      1, "XN");
   indexer.index_text(normalizeText(book.getCategory()),  1, "XC");
+  const auto bookName = book.getHumanReadableIdFromPath();
+  const auto aliasName = replaceRegex(bookName, "", "_[[:digit:]]{4}-[[:digit:]]{2}$");
```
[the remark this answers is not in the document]
juuz0: :+1:

src/library.cpp
```diff
@@ -541,6 +545,11 @@ Xapian::Query categoryQuery(const std::string& category)
   return Xapian::Query("XC" + normalizeText(category));
 }
 
+Xapian::Query aliasNameQuery(const std::string& fileName)
+{
+  return parseQuery(fileName, "XF");
```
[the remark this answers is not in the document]
juuz0: Yes, correct. Spaces, underscores, etc.

src/library.cpp
```diff
@@ -521,6 +521,16 @@ Xapian::Query buildXapianQueryFromFilterQuery(const Filter& filter)
   return queryParser.parse_query(normalizeText(filter.getQuery()), flags);
 }
 
+Xapian::Query parseQuery(const std::string& query, const std::string& prefix)
```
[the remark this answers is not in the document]
juuz0: :white_check_mark:

src/name_mapper.cpp
```diff
@@ -51,6 +51,10 @@ HumanReadableNameMapper::HumanReadableNameMapper(kiwix::Library& library, bool w
   }
 }
 
+std::string HumanReadableNameMapper::removeDateFromBookId(const std::string& bookId) {
+  return replaceRegex(bookId, "", "_[[:digit:]]{4}-[[:digit:]]{2}$");
+}
+
```
veloman-yunkan, line 57: Please extract this function in a separate preparatory commit.

src/server/internalServer.cpp
veloman-yunkan: That was not the point of my comment. Does the following queries make sense?

- `catalog/v2/entries?book=<bookid>&lang=eng`
- `catalog/v2/entries?book=<bookid1>&book=<bookid2>&category=wikipedia`
- `catalog/v2/entries?book=<bookid>&tag=nopic`

test/library.cpp
```diff
@@ -500,6 +500,24 @@ TEST_F(LibraryTest, filterByTags)
...
+  EXPECT_FILTER_RESULTS(kiwix::Filter().aliasNames({"zimfile", "example"}),
+  "An example ZIM archive",
+  "Ray Charles"
+  );
+
+  // filtering by alias name requires full text match
+  EXPECT_FILTER_RESULTS(kiwix::Filter().aliasNames({"wrong_name"}),
+  /* no results */
+  );
+}
```
veloman-yunkan, line 520: I was rather expecting new tests in `test/library_server.cpp`, that in particular would demonstrate [how the new filter is expected to interact with other filters](https://github.com/kiwix/libkiwix/pull/808#discussion_r954639462)?

veloman-yunkan (changes requested):
Please get rid of the fixup commits for the next review.

src/server/internalServer.cpp
kelson42: @veloman-yunkan Not sure to understand your remark either (is that a principle remark about filtering by humanid or a just an implementation remark). Your example don't include any "human_id" argument? A request of the type `catalog/v2/entries?book=<bookid>&lang=eng` can not fullfill the feature request because it is not stable over time. The goal here is to have a way to selection an arbitrary list of ZIM file over time. AFAIK, using the human ids and there corresponding aliases is the only way for the moment... and so far it has works fine for the Web URLS (for example library.kiwix.org/wikipedia_en_all_maxi/).

veloman-yunkan: @kelson42 In my examples `<bookid>` is supposed to be the "human_id". The examples illustrate the following cases:

- selection by "human_id" is combined with filtering by language
- selection by "human_id" is combined with filtering by category
- selection by "human_id" is combined with filtering by tag

In my understanding, selection by human_id should be mutually exclusive with other ways of content filtering.

kelson42: @veloman-yunkan I understand you now. I'm not in favour of hardwiring exclusivity but yes, your examples don't really make sense. For example, if the `humanid` is for a book in Armenian, then if `lang=eng` then not result, but if `lang=arm` then if will give one result. Etc...

veloman-yunkan: @kelson42 My vision is that human_id must not be combined with other selection criteria. My [original proposal](https://github.com/kiwix/libkiwix/pull/808#discussion_r954639462) applies only to such a use model.

veloman-yunkan: @kelson42 In other words selection by humanid should work similar to selection by name or uuid (in which case other filtering criteria are ignored).

kelson42: Nothing fondamental against this, even if I would prefer to avoid a special treatment. @juuz0 good to you to fix the points noticed by @veloman-yunkan ?

mgautierfr:
I want to mention this new issue #828 
Please remember the server can be run with a namemapper using the book uuid as a name (https://github.com/kiwix/libkiwix/blob/master/include/name_mapper.h#L41-L45) and it is the default name mapper for `kiwix::Server`.

Be sure that you are not assuming too much that we will use a `HumanReadableNameMapper`

kelson42:
@juuz0 Any feedback?

PR #822, opened 2022-09-25
Update format_code.sh script

### LINKED ISSUE #817: Usage of format_code.sh
Few questions:
* Still in use? usefull?
* I guess it should be updated?
* Probably belongs to `scripts` directory?

kelson42:
Fixes #817


scripts/format_code.sh
```diff
@@ -1,38 +1,18 @@
...
+    echo "./format_code.sh MY_DIR_TO_FORMAT"
+    exit 1
+fi
 
-for i in "${files[@]}"
+# Retrieve directory to parse
+DIR=$1
+
+# Apply formating to all *.cpp and *.h files
+for FILE in `find "$DIR" -name '*.h' -o -name '*.cpp'`
```
veloman-yunkan, line 14: I concur with CodeFactor. Let's use $(...) notation instead of legacy backticks \`...\`.

scripts/format_code.sh
```diff
@@ -1,38 +1,18 @@
...
-"src/common/networkTools.cpp"
-"src/common/stringTools.cpp"
-"src/xapianSearcher.cpp"
-"src/manager.cpp"
-"src/reader.cpp"
-)
+# Print usage() if no argument given
+if [ $# -eq 0 ]
+then
+    echo "./format_code.sh MY_DIR_TO_FORMAT"
```
veloman-yunkan, line 6: Isn't it better to assume that the `format_code.sh` script is primarily intended to format the code of its parent repository? Then the `DIR` argument can be made optional, being deduced as follows:

```bash
mydir=$(dirname "$0")
reporootdir=$(readlink -f "$mydir"/..)
# or reporootdir=$(cd "$mydir"/.. && pwd) 
```

scripts/format_code.sh
```diff
@@ -1,38 +1,18 @@
...
+    exit 1
+fi
 
-for i in "${files[@]}"
+# Retrieve directory to parse
+DIR=$1
```
veloman-yunkan, line 11: Handling multiple directories in a single command line is simple: `DIRS=("$@")`
And then `find "${DIRS[@]}" ...`.
Or simply using `"$@"` as an argument to `find` directly without introducing the variable `DIRS`.

Usage example: `scripts/format_code.sh ./include ./src`

veloman-yunkan (changes requested):

scripts/format_code.sh
kelson42: yes

scripts/format_code.sh
kelson42: OK

scripts/format_code.sh
kelson42: I have added `include` to parse. The script does not take any argument anymore (like before).

scripts/format_code.sh
```diff
@@ -1,17 +1,12 @@
...
-# Retrieve directory to parse
-DIR=$1
+# Compute 'src' path
+SCRIPT_DIR=$(dirname "$0")
+REPO_DIR=$(readlink -f "$SCRIPT_DIR"/..)
+DIRS="$REPO_DIR/src $REPO_DIR/include"
 
 # Apply formating to all *.cpp and *.h files
-for FILE in `find "$DIR" -name '*.h' -o -name '*.cpp'`
+for FILE in $(find $DIRS -name '*.h' -o -name '*.cpp')
```
veloman-yunkan, line 9: Such declaration and usage of `DIRS` is vulnerable to whitespace or some special symbols in being present `"$REPO_DIR"`. The same problem applies to `for FILE in $(find ...)` (because of unquoted `$(...)`). The simplest solution is to `cd` to `"$REPO_DIR"`:

```bash
cd "$REPO_DIR"
DIRS="src include"

# Apply formating to all *.cpp and *.h files
for FILE in $(find $DIRS -name '*.h' -o -name '*.cpp')
...
```

veloman-yunkan (changes requested):

scripts/format_code.sh
veloman-yunkan: Did you miss the `test` subdirectory or you excluded it intentionally?

scripts/format_code.sh
kelson42: yes, fixing

veloman-yunkan (approved):

PR #827, opened 2022-10-04
New logic for generating HTTP-redirects

### LINKED ISSUE #826: Redirection error in latest (2022-09) ifixit zim (English and Russian version)
From ifixit created by [laggykiller](https://github.com/laggykiller): openzim/ifixit#86  

The latest (2022-09) ifixit zim (English and Russian version) has redirection error ('redirected you too many times')

The zim in question is here:
https://download.kiwix.org/zim/ifixit/ifixit_en_all_2022-09.zim
https://download.kiwix.org/zim/ifixit/ifixit_ru_all_2022-09.zim

You can test it out now from here:
https://library.kiwix.org/ifixit_en_all_2022-09
https://library.kiwix.org/ifixit_ru_all_2022-09

The previous versions (2022-06) are normal:
https://download.kiwix.org/zim/ifixit/ifixit_en_all_2022-06.zim
https://download.kiwix.org/zim/ifixit/ifixit_ru_all_2022-06.zim

Other languages are not affected

When viewed from kiwix-serve, it produces redirection error:
image

When viewed from desktop application (e.g. Windows), the homepage is broken:
image

veloman-yunkan:
Fixes #826 (with some side-effects).

Before this fix the root URL for a book was assumed to resolve to the main page. This was not true for ZIM files containing an entry at an empty path or with a path equal to "/", leading to #826. The logic behind this behaviour is found in `kiwix::getEntryFromPath()`.

The fix to that issue by this PR is a little more general and will result in an HTTP redirect in any case where `kiwix::getEntryFromPath(zim, path)` returns an entry with a real path different from the requested one. In particular, this will affect the behaviour on ZIM files with the old namespace scheme, where the requested resource - if not found - is also looked up in the 'A', 'I', 'J', and/or '-' namespaces. Now instead of returning the contents of that other resource an HTTP redirect response will be sent.

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -1043,9 +1043,14 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
 
   try {
     auto entry = getEntryFromPath(*archive, urlStr);
-    if (entry.isRedirect() || urlStr.empty()) {
-      // If urlStr is empty, we want to mainPage.
-      // We must do a redirection to the real page.
+    if (entry.isRedirect() || urlStr != entry.getPath()) {
+      // In the condition above, the second case (an entry with a different
+      // URL was returned) can occur in the following situations:
+      // 1. urlStr is empty or equal to "/" and the ZIM file doesn't contain
+      //    such an entry, in which case the main entry is returned instead.
+      // 2. The ZIM file uses old namespace scheme, and the resource at urlStr
+      //    is not present but can be found under one of the 'A', 'I', 'J' or
+      //    '-' namespaces, in which case that resource is returned instead.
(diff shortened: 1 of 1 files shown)


mgautierfr (approved):
For information, the test on `urlStr.empty()` was made when using "old" namespace schema.
At this time (and still with old zim files), the main entry returned by `getEntryFromPath` was directly the actual main page without a redirection. We had to do the redirection ourselves. And it was the only case where libzim returned a different entry than requested.

Such condition (`urlStr != entry.getPath()`) can only appear with old zim files
(It can appear also with main page in new zim file, but in this case entry is also a redirect).
So this condition is now somehow a compatibility feature with old zim files only. When (if) we stop support old files, we can remove it.

Issue #828, opened 2022-10-03
OPDS catalog must respect the used nameMapper


MohitMaliFtechiz:
**Describe the bug**
On libkiwix 10.1.1 , zim file while opening over hotspot resulting ```The requested URL "/" was not found on this server.```.

**Expected behavior**
Its should open the zim file without any issue

**Screenshots**
![Screenshot from 2022-10-03 18-52-57](https://user-images.githubusercontent.com/34593983/193588946-e4684153-6689-46ec-9f8b-7cfa55dc1bae.png)
![Screenshot from 2022-10-03 18-53-01](https://user-images.githubusercontent.com/34593983/193588952-83e8bc93-4e7d-446c-ace0-b7e469981f57.png)

**Environment**
- Version of Kiwix Android : 3.7.0
- Device : REDMI NOTE 9
- OS version : 11

kelson42:
@mgautierfr @veloman-yunkan Sounds like a regression to me. Isn't it?

mgautierfr:
> However, the opds_stream always uses the "humanReadableIdFromPath". It is the bug (to fix) and it always have been this way.
The "regression" comes from the switch from backend generated home page to dynamic js home page which use the catalog as source.

This is not a regression from the opds_stream point of view as the bug was always here.
This is a regression from the homepage point of view and it dates from libkiwix 10.0.0 (but the bug is on opds_stream side)

kelson42:
@mgautierfr Thx for clarification. IMO, we better get it fixed befpre 12.0.0 release.

veloman-yunkan:
I will work on this.

mgautierfr:
I'm working on this

veloman-yunkan:
I've started with some clean-up of OPDS dumper. I've now shared my work as a self contained PR #829.

veloman-yunkan:
I guess that #829 in its current form can be merged first without causing too much pain for #830. I won't continue working on it.

PR #829, opened 2022-10-06
OPDS dumper cleanup

veloman-yunkan:
OPDS dumper cleanup (as a preparatory step for fixing #828)

This cleanup should also result in a slightly faster OPDS feed generation when only partial entries are requested.


mgautierfr (approved):
LGTM

Issue #740, opened 2022-04-01
Expose (kind of) `InternalServer` in the public API ?


mgautierfr:
The server feature in libkiwix does several things:
- Listen for http request (using libmicrohttpd) and convert it to a `RequestContext`
- Parse the request to understand what user want.
- Find the correct books/Archive/Searcher in the library and "resolve" the user request.
- Build a `Response`
- Convert the `Response` to a libmicrohttpd's response and send back the data to the user.

The first and last step is pretty tied to a http server but the other points are pretty common and it is somehow what is made in all applications.
In kiwix-desktop, we span some kind of internal server and answer to QWebEngine requests (https://github.com/kiwix/kiwix-desktop/blob/master/src/urlschemehandler.cpp)
This is probably what is done in kiwix-android or macos application.

It would be a nice thing to move all the handling in `InternalServer` as a first class citizen (libmicrohttpd neutral) in libkiwix API.
Then we could :
- Unit test the "request handler" without launching a temporary server
- Create the server on top of this "request handler" (and keep the current testing system here)
- Use the "request handler" in kiwix-desktop

Some open questions or points to keep in mind:
- Should we enforce a url schema or be configurable (kiwix-desktop uses a different schema than kiwix-serve)
- I'm convinced (but you can counter-argument) that we should move to architecture where we serve "raw" content in a zim file and we add chrome/gui around using different technologies (Qt Widgets in kiwix-desktop, iframe/js in kiwix-serve, ....) With this idea, the "request handler" should handle "raw" content (I include search and co) and the "http server" should add the Chrome.

At your comments :)

kelson42:
First impression on my side: sounds a good idea, but I have a really superficial understanding of all what this concretly mean. Probably the feedback from @veloman-yunkan here is far more valuable than mine :)

mossroy:
This looks interesting, but would most probably not be usable in kiwix-js.

In kiwix-js, we do such thing in a ServiceWorker (which traps the user HTTP requests). But a ServiceWorker is very isolated, and notably can not access the ZIM file.
So we would not be able to use it as a one-chunk emscripten-compiled C code, that handles the requests from A to Z.

mgautierfr:
> But a ServiceWorker is very isolated, and notably can not access the ZIM file.

How do you respond to the user request then ? You "transfer" the request to something that can access the ZIM file ?

The idea of this issue is to separate the front request handling (parsing the incoming request to know what to do) from the back (doing what is requested).
[...]

PR #830, opened 2022-10-06
OPDS name mapper

### LINKED ISSUE #828: OPDS catalog must respect the used nameMapper

mgautierfr:
Fixes #828

### DIFF
--- a/include/server.h
+++ b/include/server.h
@@ -29,85 +29,84 @@ namespace kiwix
   class NameMapper;
   class InternalServer;
 
-  class ServerConfiguration {
-    public:
-      ServerConfiguration(Library* library, NameMapper* nameMapper=nullptr)
-        : mp_library(library),
-          mp_nameMapper(nameMapper)
-      {}
-
-      ServerConfiguration& setRoot(const std::string& root);
-      ServerConfiguration& setAddress(const std::string& addr) {
-        m_addr = addr;
-        return *this;
-      }
-
-      ServerConfiguration& setPort(int port) {
-        m_port = port;
-        return *this;
-      }
-
-      ServerConfiguration& setNbThreads(int threads) {
-        m_nbThreads = threads;
-        return *this;
-      }
-
-      ServerConfiguration& setMultiZimSearchLimit(unsigned int limit) {
-        m_multizimSearchLimit = limit;
-        return *this;
-      }
-
-      ServerConfiguration& setIpConnectionLimit(int limit) {
-        m_ipConnectionLimit = limit;
-        return *this;
-      }
-
-      ServerConfiguration& setVerbose(bool verbose) {
-        m_verbose = verbose;
-        return *this;
-      }
-
-      ServerConfiguration& setIndexTemplateString(const std::string& indexTemplateString) {
-        m_indexTemplateString = indexTemplateString;
-        return *this;
-      }
-
-      ServerConfiguration& setTaskbar(bool withTaskbar, bool withLibraryButton) {
-        m_withTaskbar = withTaskbar;
-        m_withLibraryButton = withLibraryButton;
-        return *this;
-      }
-
-      ServerConfiguration& setBlockExternalLinks(bool blockExternalLinks) {
-        m_blockExternalLinks = blockExternalLinks;
-        return *this;
-      }
-
-      Library* mp_library;
-      NameMapper* mp_nameMapper;
-      std::string m_root = "";
-      std::string m_addr = "";
-      std::string m_indexTemplateString = "";
-      int m_port = 80;
-      int m_nbThreads = 1;
-      unsigned int m_multizimSearchLimit = 0;
-      bool m_verbose = false;
-      bool m_withTaskbar = true;
-      bool m_withLibraryButton = true;
-      bool m_blockExternalLinks = false;
-      int m_ipConnectionLimit = 0;
-  };
-
-
 
   class Server {
-     public:
+    public:
+      class Configuration {
+        public:
+          explicit Configuration(Library* library, NameMapper* nameMapper=nullptr)
+            : mp_library(library),
+              mp_nameMapper(nameMapper)
+          {}
+
+          Configuration& setRoot(const std::string& root);
+          Configuration& setAddress(const std::string& addr) {
+            m_addr = addr;
+            return *this;
+          }
+
+          Configuration& setPort(int port) {
+            m_port = port;
+            return *this;
+          }
+
+          Configuration& setNbThreads(int threads) {
+            m_nbThreads = threads;
+            return *this;
+          }
+
+          Configuration& setMultiZimSearchLimit(unsigned int limit) {
+            m_multizimSearchLimit = limit;
+            return *this;
+          }
+
+          Configuration& setIpConnectionLimit(int limit) {
+            m_ipConnectionLimit = limit;
+            return *this;
+          }
+
+          Configuration& setVerbose(bool verbose) {
+            m_verbose = verbose;
+            return *this;
+          }
+
+          Configuration& setIndexTemplateString(const std::string& indexTemplateString) {
+            m_indexTemplateString = indexTemplateString;
+            return *this;
+          }
+
+          Configuration& setTaskbar(bool withTaskbar, bool withLibraryButton) {
+            m_withTaskbar = withTaskbar;
+            m_withLibraryButton = withLibraryButton;
+            return *this;
+          }
+
+          Configuration& setBlockExternalLinks(bool blockExternalLinks) {
+            m_blockExternalLinks = blockExternalLinks;
+            return *this;
+          }
+
+          Library* mp_library;
+          NameMapper* mp_nameMapper;
+          std::string m_root = "";
+          std::string m_addr = "";
+          std::string m_indexTemplateString = "";
+          int m_port = 80;
+          int m_nbThreads = 1;
+          unsigned int m_multizimSearchLimit = 0;
+          bool m_verbose = false;
+          bool m_withTaskbar = true;
+          bool m_withLibraryButton = true;
+          bool m_blockExternalLinks = false;
+          int m_ipConnectionLimit = 0;
+       };
+
        /**
         * The default constructor.
         *
         * @param library The library to serve.
         */
-       Server(const ServerConfiguration& configuration);
+       explicit Server(const Configuration& configuration);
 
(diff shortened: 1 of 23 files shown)


src/opds_dumper.cpp
```diff
@@ -94,24 +95,24 @@ kainjow::mustache::object getSingleBookData(const Book& book)
     };
 }
 
-std::string getSingleBookEntryXML(const Book& book, bool withXMLHeader, const std::string& rootLocation, const std::string& endpointRoot, bool partial)
+std::string getSingleBookEntryXML(const Book& book, const NameMapper& nameMapper,  bool withXMLHeader, const std::string& rootLocation, const std::string& endpointRoot, bool partial)
```
veloman-yunkan, line 98: I started #829 in order to avoid ending up with such a function receiving 6 parameters.

mgautierfr: Yes, that's why I've started a factorization after this commit :)

include/server.h
```diff
@@ -29,14 +29,85 @@ namespace kiwix
   class NameMapper;
   class InternalServer;
 
+  class ServerConfiguration {
+    public:
+      ServerConfiguration(Library* library, NameMapper* nameMapper=nullptr)
```
veloman-yunkan: The second parameter being optional, `explicit` is recommended here

include/server.h
```diff
@@ -29,14 +29,85 @@ namespace kiwix
...
+
   class Server {
      public:
        /**
         * The default constructor.
         *
         * @param library The library to serve.
         */
-       Server(Library* library, NameMapper* nameMapper=nullptr);
+       Server(const ServerConfiguration& configuration);
```
veloman-yunkan, line 107: `explicit` was due even before this change

include/server.h
```diff
@@ -29,14 +29,85 @@ namespace kiwix
   class NameMapper;
   class InternalServer;
 
+  class ServerConfiguration {
```
veloman-yunkan, line 32: What about making this a nested class of `Server` (`Server::Configuration`)?

src/server/internalServer.cpp
```diff
@@ -406,13 +385,13 @@ bool InternalServer::start() {
...
 #endif
-  if (m_verbose.load())
+  if (m_configuration.m_verbose)
     flags |= MHD_USE_DEBUG;
 
   struct sockaddr_in sockAddr;
   memset(&sockAddr, 0, sizeof(sockAddr));
   sockAddr.sin_family = AF_INET;
-  sockAddr.sin_port = htons(m_port);
+  sockAddr.sin_port = htons(m_configuration.m_port);
```
veloman-yunkan, line 380: You could have avoided the numerous changes like this one by making `InternalServer` a (private) subclass of `ServerConfiguration`.

test/library.cpp
```diff
@@ -236,14 +236,14 @@ namespace
...
+  auto lib = std::make_shared<kiwix::Library>();
+  kiwix::Manager manager(lib);
   manager.readOpds(sampleOpdsStream, "library-opds-import.unittests.dev");
 
-  EXPECT_EQ(10U, lib.getBookCount(true, true));
+  EXPECT_EQ(10U, lib->getBookCount(true, true));
 
   {
-  const kiwix::Book& book1 = lib.getBookById("0c45160e-f917-760a-9159-dfe3c53cdcdd");
+  const kiwix::Book& book1 = lib->getBookById("0c45160e-f917-760a-9159-dfe3c53cdcdd");
```
veloman-yunkan, line 246: You could have avoided the numerous changes like this one by utilizing a `shared_ptr` with a non-deleting deallocator:

https://github.com/kiwix/libkiwix/blob/a844bc4000de0cce1e0b5a1b0b5356a3160fd99a/src/manager.cpp#L33-L36

https://github.com/kiwix/libkiwix/blob/a844bc4000de0cce1e0b5a1b0b5356a3160fd99a/src/manager.cpp#L92-L102

src/opds_dumper.cpp
```diff
@@ -99,12 +100,12 @@ std::string fullEntryXML(const Book& book, const std::string& rootLocation, cons
     return render_template(RESOURCE::templates::catalog_v2_entry_xml, data);
 }
 
-std::string partialEntryXML(const Book& book, const std::string& rootLocation)
+std::string partialEntryXML(const ServerConfiguration& configuration, const Book& book)
```
veloman-yunkan, line 103: I don't agree with this change.

Cons:
- access to more data than needed by this function

Pros:
- ???

veloman-yunkan (changes requested):

src/opds_dumper.cpp
```diff
@@ -71,7 +72,7 @@ IllustrationInfo getBookIllustrationInfo(const Book& book)
     return illustrations;
 }
 
-std::string fullEntryXML(const Book& book, const std::string& rootLocation)
+std::string fullEntryXML(const Book& book, const std::string& rootLocation, const std::string& bookName)
```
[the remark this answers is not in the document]
veloman-yunkan: Update: I see that this concern was later on addressed in a slightly different way.

src/opds_dumper.cpp
mgautierfr: I made this for consistency.
The idea is that those set of free functions to generate xml are black box function generating the xml in accordance with the ServerConfiguration, what ever is actually used.

I have no strong opinion on this,except that passing the actual needed data may leads to too much parameters.
And if several functions are using a ServerConfiguration I prefer having all functions using them.

mgautierfr:
Ready for a new review.
New commits are marked with `[NEW]` (or `fixup!` as usual)

src/server/internalServer.cpp
```diff
@@ -216,7 +216,7 @@ std::pair<std::string, Library::BookIdSet> InternalServer::selectBooks(const Req
   try {
     auto bookName = request.get_argument("content");
     try {
-      const auto bookIds = Library::BookIdSet{m_configuration.mp_nameMapper->getIdForName(bookName)};
+      const auto bookIds = Library::BookIdSet{mp_nameMapper->getIdForName(bookName)};
```
veloman-yunkan, line 219: This comment applies to the commit "[NEW] Make InternalServer inherite Server::Configuration.". It would be better to follow this approach from the very beginning, instead of making a lot of changes and then undoing them.

veloman-yunkan (commented):
Please rebase&fixup. I think that the clean version has to be reviewed due to significant reversal of some changes.

veloman-yunkan (commented):
While reviewing kiwix/kiwix-tools#580 and kiwix/kiwix-desktop#894 I got the impression that this PR goes too far with passing `Library` and `NameMapper` via shared pointers into some objects.

`OPDSDumper` and `SearchRenderer` are cheap-to-instantiate one-shot-use helper classes with the (current) mere purpose of avoiding free functions `dumpOPDSFeed()`, `renderSearchResults()` with too many parameters. Objects of those classes should be created, used and destroyed within a single function call, and therefore storing the references to `Library` and `NameMapper` as shared pointers is not justified. In case of `OPDSDumper` this concern is somewhat masked by replacing individual references to `Library` and `NameMapper` by an entire `Server::Configuration` object, but the result is that `OPDSDumper` gains non-const access to `Library` which it doesn't really need.

Regarding `Manager`, a scenario where it may turn out to be the last object keeping a reference to `Library` doesn't make much sense (at least, for the case of a manager object instantiated via the `Manager(Library*)` constructor). In such a situation, the modification of the library via the `Manager` should have ended before other clients of the library die.

mgautierfr:
I was more concern by the `Server::Configuration` (and so the server) keeping a reference to a destroyed `Library`.

But I agree that `OPDSDumper` and `SearchRendered` become a bit too complex. If we agree that those objects are one-shoot helper, I would tend to move them as free functions (taking a `Configuration` object if there is too many parameters) and keep the class "private". This way it would not be possible to keep an existing instance having a reference to a `Library`.

There is also the `(Updatable)NameMapper` keeping a reference to the `Library`. We may fix that by having the `Library` keeping a reference to the name mapper instead. It would extend `Library` for something that is related only to the server but is should be ok.

This may me think to https://github.com/kiwix/libkiwix/issues/740. `OPDSDumper` and `SearchRendered` are also used in a context of a "Server" (has they need to know how to generate urls)

---

I'm afraid this PR will become more complex to solve that I was initially thinking. I will move the actual fix in another PR to have it quickly merged and will take time to discuss the best design here.

veloman-yunkan:
@mgautierfr

While reviewing #837, I had a new idea. Why do we need to pass `NameMapper` so deep into various operations having to deal with the book content-id? Why don't we replace `Book::getHumanReadableIdFromPath()` with a new user-settable field `Book::m_human_friendly_id` that by default will be initialized to a value previously returned by `getHumanReadableIdFromPath()`? Then name-mapping can be applied to the library in a single place and all clients of the human-friendly id will access it via the enhanced interface of `Book` without needing an extra mapper object. Of course this gives up some flexibility (for example, you can't have different mappings in different contexts for the same library) but do we really need that?

In short, the proposal is to get rid of the `kiwix::NameMapper` interface altogether, incorporating that info in `kiwix::Book`.

mgautierfr:
> Of course this gives up some flexibility (for example, you can't have different mappings in different contexts for the same library) but do we really need that?

Yes, at least for now.

The namemapper is used for two different (but related) things:
- Routing : To know which zim file to use for a request. And several "request" can lead to same zim file (as date alias).
- Generating content : To create links which must of course be in sync with the routing.

The problem is that routing is not part of the "whole" library (libkiwix). It is part of server only. kiwix-desktop has it own routing and it use the namemapper to generate search result in correspondance to its routing. In the same time, kiwix-desktop can serve the same library (set of books) with kiwix-serve (and so with a different namemapper). And it is the same thing for kiwix-android (but I'm not sure kiwix-android is generating search result with `searchRenderer`).

When we implement #740 with some kind of generic routing system, we may rethink this as we could provide of full set (consistant) of features.

PR #833, opened 2022-10-09
HTTP cache control

### LINKED ISSUE #650: Update Kiwix-serve cache strategy

### LINKED ISSUE #831: Cache control of static resources
This is a sub-ticket of #650

#712 has introduced support for versioning static resources, however the HTTP caching side of their handling has not been addressed.

Currently a build of `kiwix-serve` containing static resource `/skin/RESOURCE` with the cache id value `CACHEID` is supposed only to receive and handle requests of the form `/skin/RESOURCE?cacheid=CACHEID`. #650 proposes responding to such requests with `Cache-Control: max-age=31536000, immutable`. That's easy.

Questions:

1. What should be the `Cache-Control` response to a request *without any* cache-id (`/skin/RESOURCE`)?
    a. #716 has introduced a special user-facing static resource served at `/viewer` which is going to be accessed without any cache id. What should be the caching policy for that one (especially in interaction with kiwix/kiwix-tools#17)?
2. What should be the response to a request *with a wrong* cache-id (`/skin/RESOURCE?cacheid=WRONGCACHEID`)?

veloman-yunkan:
Fixes #650
Fixes #831


kelson42:
Fixes https://github.com/kiwix/libkiwix/issues/831 and partly fixes https://github.com/kiwix/libkiwix/issues/650?! Might be better.

veloman-yunkan:
> Fixes #831 and partly fixes #650?! Might be better.

No. This PR in its current state fixes #831 only partly since it doesn't address the aspects discussed in the questions posed in that ticket. Since the higher level ticket #650 was not concerned with such issues, it can be considered to be fully fixed by this PR.

veloman-yunkan:
Now #831 is fully fixed by this PR

src/server/internalServer.cpp
```diff
@@ -1033,6 +1035,11 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
            + suggestSearchMsg(searchURL, kiwix::urlDecode(pattern));
   }
 
+  const std::string archiveUuid = getArchiveId(*archive);
```
mgautierfr, line 1064: `getArchiveId` (and others) has been introduced by 8b12434 to avoid duplication between `kiwix::Reader::getId` and `kiwix::Book::update` which now use the `zim::Archive`.

Now that `kiwix::Reader` has been removed (and `getArchiveId` implementation simplified) you would tend to directly use `archive.getUuid()`

mgautierfr (changes requested):
> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!      QUESTION: Do we need those resources?         !!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!


For favicon, I not sure but I think we should keep them.
Browsers may do some kind of auto discovery of the favicon by directly request them even if they are not declared in the html. @kelson42 may have a better idea here.

For `search-icon.svg`, it is referenced in `static/skin/index.css`
`search.svg` seems indeed not used. I think we can remove it.

mgautierfr:
Real favicon generator may help us to know what we need to keep : https://realfavicongenerator.net/favicon_checker?protocol=http&site=dev.library.kiwix.org

src/server/internalServer.cpp
veloman-yunkan: Done

static/skin/index.css
```diff
@@ -105,7 +105,7 @@ body {
     border-radius: 10px;
     border: solid 1px #b5b2b2;
     padding: 10px;
-    background-image: url('./search-icon.svg');
+    background-image: url('../skin/search-icon.svg?KIWIXCACHEID');
```
mgautierfr, line 108: Why the change in the path (`../skin`) ?

mgautierfr (commented):
I have one open question but else, it is a approval.

static/skin/index.css
veloman-yunkan: Because relative URLs to static resources are recognized by the `kiwix-resources` script using the `/skin/` anchor

mgautierfr:
Please rebase-fixup on new master and we should be good to merge.

veloman-yunkan:
> Please rebase-fixup on new master and we should be good to merge.

Done. The only change in addition to rebase-fixup is the edited commit message of 415ec4109989ef2d2b7443f3dbcd54081b934115 (the question answered by https://github.com/kiwix/libkiwix/pull/833#pullrequestreview-1139011544 was removed).

mgautierfr (approved):

PR #841, opened 2022-10-31
Catalog filtering by multiple languages

veloman-yunkan:
Fixes kiwix/kiwix-tools#358

The requested functionality is implemented in a slightly different way. The OPDS query was assumed to be a conjunction (AND) of the filtering criteria. Handling multiple `lang` parameters as a disjunction (OR) may lead to confusion. Instead the set of desired languages must be specified as a comma-separated list via a single `lang` parameter (similar to how the `tag` parameter works).

The enhancement applies to 3 endpoints:

- `/catalog/search`
- `/catalog/v2/entries`
- `/catalog/v2/partial_entries`

Note, however, that corresponding unit-tests have been provided only for the first two of them. Filtering of `/catalog/v2/partial_entries` was not tested (when that endpoint was introduced) because it shares the same code with `/catalog/v2/entries`.


veloman-yunkan:
It turned out `/catalog/v2` was not documented at https://wiki.kiwix.org/wiki/OPDS. I updated that page, but after this PR is merged it must be updated again.

kelson42:
@veloman-yunkan Thank your for this PR. Please don't forget to update the wiki (again)

veloman-yunkan:
Done

PR #846, opened 2022-11-17
Internationalization of kiwix-serve's frontend

veloman-yunkan:
Will fix kiwix/kiwix-tools#59

Depends on #871

PR #849, opened 2022-11-28
User language control in the server

### LINKED ISSUE #750: Correctly parse `Accept-Language` header.
The PR #679 add a preliminary support for `Accept-Language` header. It assumes the header contains a single language.
But most browser fill the header with a list of weighted accepted language. See commit message of 927c125

We should correctly parse the header.

veloman-yunkan:
Fixes #750

Depends on #848

kiwix server determines the user language from three sources (in that order, earlier source takes precedence):
- `userlang` URL query parameter
- `userlang` cookie
- `Accept-Language` header

In all cases, the server sets the `userlang` cookie in its response using thus determined language.

If `Accept-Language` is used and multiple language options are specified with weights, then the most suitable language is selected by scoring available translation languages - the score value is computed as the product of weight value and the number of translated strings for that language.

### DIFF
--- a/src/server/request_context.cpp
+++ b/src/server/request_context.cpp
@@ -25,6 +25,7 @@
 #include <sstream>
 #include <cstdio>
 #include <atomic>
+#include <cctype>
 
 #include "tools/stringTools.h"
 
@@ -61,6 +62,22 @@ fullURL2LocalURL(const std::string& full_url, const std::string& rootLocation)
   } else {
     return "";
   }
+}
+
+std::string parseAcceptLanguageHeader(const std::string& s)
+{
+  // TODO: implement properly
+
+  if ( s.empty() )
+    return "en";
+
+  for ( const char c :  s ) {
+    if ( ! std::isalpha(c) ) {
+      return "en";
+    }
+  }
+
+  return s;
 }
 
 } // unnamed namespace
@@ -204,7 +221,7 @@ std::string RequestContext::get_user_language() const
   } catch(const std::out_of_range&) {}
 
   try {
-    return get_header("Accept-Language");
+    return parseAcceptLanguageHeader(get_header("Accept-Language"));
   } catch(const std::out_of_range&) {}
 
   return "en";
--- a/src/server/request_context.cpp
+++ b/src/server/request_context.cpp
@@ -97,6 +97,7 @@ RequestContext::RequestContext(struct MHD_Connection* connection,
 {
   MHD_get_connection_values(connection, MHD_HEADER_KIND, &RequestContext::fill_header, this);
   MHD_get_connection_values(connection, MHD_GET_ARGUMENT_KIND, &RequestContext::fill_argument, this);
+  MHD_get_connection_values(connection, MHD_COOKIE_KIND, &RequestContext::fill_cookie, this);
 
   try {
     acceptEncodingGzip =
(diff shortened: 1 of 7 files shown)


veloman-yunkan:
Note that in this version dialects/language variants (like `en-US`, `fr-XVI`, etc) are treated as regular languages - no attempt to convert them into language codes that are available in our translation database is performed.

src/server/request_context.cpp
```diff
@@ -215,11 +226,20 @@ std::string RequestContext::get_header(const std::string& name) const {
 }
 
 std::string RequestContext::get_user_language() const
+{
+  return userlang;
+}
+
+std::string RequestContext::get_user_language_expensive() const
```
mgautierfr, line 218: Should this method being called `determine_user_language` or something like that ?

`get_user_language_expensive` makes think it it the same thing that `get_user_language` but inefficient but `get_user_language` is just a "cache" for `get_user_language_expensive`

test/server.cpp
```diff
@@ -1042,6 +1042,38 @@ TEST_F(ServerTest, UserLanguageControl)
...
+      /* expected <h1> */ "[I18N TESTING] Content not found, but at least the server is alive"
+    },
+    {
+      "userlang cookie is correctly parsed",
+      /*url*/ "/ROOT/content/zimfile/invalid-article",
+      /*Accept-Language:*/ "",
+      /*Request Cookie:*/       "cookie1=abc; userlang=test; cookie2=xyz",
+      /*Response Set-Cookie:*/  "userlang=test",
+      /* expected <h1> */ "[I18N TESTING] Content not found, but at least the server is alive"
+    },
```
mgautierfr, line 1076: Maybe adding a test with to `userlang=foo|bar` in the cookie ?

mgautierfr (commented):
A small question/update but we are mostly good.

kelson42:
> Note that in this version dialects/language variants (like `en-US`, `fr-XVI`, etc) are treated as regular languages - no attempt to convert them into language codes that are available in our translation database is performed.



> Note that in this version dialects/language variants (like `en-US`, `fr-XVI`, etc) are treated as regular languages - no attempt to convert them into language codes that are available in our translation database is performed.

What is the user impact of that, this seems important that `fr-CH` delivers `fr` in case we don't have any `fr-CH` translations.

kelson42:
It seems here, no javascript file is changed, so I wonder about my comment https://github.com/kiwix/libkiwix/issues/750#issuecomment-1328208784. How should we do, to change the code which handle the langauge selection box from javascript to c++?

src/server/request_context.cpp
veloman-yunkan: Done

veloman-yunkan:
> What is the user impact of that, this seems important that `fr-CH` delivers `fr` in case we don't have any `fr-CH` translations.

@kelson42  I would prefer to address that in a separate PR (after this one and #846 are merged).

veloman-yunkan:
> It seems here, no javascript file is changed, so I wonder about my comment [#750 (comment)](https://github.com/kiwix/libkiwix/issues/750#issuecomment-1328208784). How should we do, to change the code which handle the langauge selection box from javascript to c++?

@kelson42 That is addressed in #846.

test/server.cpp
mgautierfr: It would have been "two `userlang=foo|bar`", not "to".

If a user send a cookie with `userlang=en; uselang=test`.
We have to select one, it would be better to test that, even if it is not a strong specification.

mgautierfr (approved):

mgautierfr:
Rebased on master and fixed-up.
I'm merging.

PR #852, opened 2022-12-03
Add unminified version of autoComplete.js and attribution

### LINKED ISSUE #819: `autoComplete.min.js` is already minified in the code repository
I wonder if this is already minified:
* Why this is minified (is that really necessary)?
* How to read it?
* How to modify it?

Usually, the way of doing is to have a "working" version not minified and then minify the file at release (or even runtime). Why we don't do so?

legoktm:
* Get latest version of autoComplete.min.js
* Provide unminified version to make it simpler to make modifications if upstream diverges or disappears
* Document license and attribution information

Fixes #819.

----
It's unclear to me why the version of autoComplete.min.js doesn't match what I downloaded from https://raw.githubusercontent.com/TarekRaafat/autoComplete.js/v10.2.7/dist/autoComplete.min.js - I'm not sure what, if any, changes were made to the libkiwix version.


veloman-yunkan:
> It's unclear to me why the version of autoComplete.min.js doesn't match what I downloaded from https://raw.githubusercontent.com/TarekRaafat/autoComplete.js/v10.2.7/dist/autoComplete.min.js - I'm not sure what, if any, changes were made to the libkiwix version.

@legoktm `autoComplete.min.js` that was added to libkiwix by PR #796 (which, in turn, overrode an earlier PR #639) corresponds to release [v10.2.6](https://github.com/TarekRaafat/autoComplete.js/releases/tag/v10.2.6) of the [original project](https://github.com/TarekRaafat/autoComplete.js). Maybe instead of upgrading to the most recent version we better just add the unminified variant corresponding to that version?

kelson42:
@legoktm Any update?

kelson42:
@legoktm Should I close thia PR?

PR #856, opened 2022-12-07
Gzip compress HTTP response for Web fonts

### LINKED ISSUE #854: kiwix-serve should compress web fonts
After releasing libkiwix `12.0.0` and kiwix-serve `3.4.0`, this has been deployed on https://library.kiwix.org, and I have renewed a performance test of the welcome page. Overall results looks good, but there is improvement potential.

One of them is around resources size/load times:
image

You can see that we have `318KB` of Web-font loading at first load (I understand the Web-font should be cached properly) which is pretty heavy and I think we could/should improve the situation here without much hassle.

It looks like actually the Web-fonts are not gzip compressed, look:
```bash
$ curl --compressed -I "http://192.168.178.24:8181/skin/fonts/Roboto.ttf?cacheid=84d10248"
HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Length: 168260
Content-Type: application/font-ttf
Access-Control-Allow-Origin: *
Cache-Control: max-age=31536000, immutable
Date: Wed, 07 Dec 2022 11:33:09 GMT

```

I made a small test with `gzip` on the command line and we could get both of them for only `159KB`. We should have this for our skin Web-fonts, but would be great also to do the same on ZIM based Web-fonts.

Therefore I would recommend to compress the Web-fonts (all kind of them). It seems actually that compressing the Web-fonts is recognized like a good practice, see for example:
https://www.fastly.com/blog/new-gzip-settings-and-deciding-what-compress#new-gzip-settings

kelson42:
Fixes #854

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -64,7 +64,13 @@ bool is_compressible_mime_type(const std::string& mimeType)
       || mimeType.find("application/javascript") != std::string::npos
       || mimeType.find("application/atom") != std::string::npos
       || mimeType.find("application/opensearchdescription") != std::string::npos
-      || mimeType.find("application/json") != std::string::npos;
+      || mimeType.find("application/json") != std::string::npos
+
+      // Web fonts
+      || mimeType.find("application/font-") != std::string::npos
+      || mimeType.find("application/x-font-") != std::string::npos
+      || mimeType.find("application/vnd.ms-fontobject") != std::string::npos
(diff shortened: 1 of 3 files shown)


veloman-yunkan (approved):

veloman-yunkan (changes requested):
CI is failing because of a unit test being broken by this change.

veloman-yunkan (approved):

PR #859, opened 2022-12-12
Safe href in suggestion links for the iframe-based viewer

veloman-yunkan:
Addresses kiwix/kiwix-tools#587 with respect to suggestions in the iframe-based viewer.

The value of the `href` attributed is assumed to always be a URI-encoded string that has to be decoded. This is true for URLs with  `javascript:` scheme too.

The following example demonstrates the problem

```HTML
<!DOCTYPE html>
<html>
  <head>
    <script>
      function foo() { alert('A%26B'); }
    </script>
  </head>
  <body>
    <a href="javascript:alert('A%26B)">Inline javascript.</a>
    <a href="javascript:foo()">Out-of-line javascript</a>
  </body>
</html>
```

Invoking inline javascript via the first link displays URL-decoded text ("A&B"). Running out-of-line javascript via the second link display the text as is ("A%26B").

This means that suggestion links in the iframe-based viewer that work via javascript need a second level of URL-encoding.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -342,7 +342,14 @@ function setupSuggestions() {
           } else {
             searchLink = `${root}/search?content=${encodeURIComponent(currentBook)}&pattern=${encodeURIComponent(htmlDecode(data.value.value))}`;
           }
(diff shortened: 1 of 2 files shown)


mgautierfr (approved):
This is a tricky bug.
LGTM

PR #860, opened 2022-12-15
Handling of suggestion links with single quotes (and other special symbols)

veloman-yunkan:
This is a follow-up of #859.

Should fix kiwix/kiwix-tools#589

This PR fixes suggestion links in the presence of special symbols such as `'` (a single quote), `?` (a question mark), and/or `#` (hash)  in the root location (value of the `--urlRootLocation` option of `kiwix-serve`), book name or article name.


mgautierfr (approved):
I now know why people use js framework instead of reimplementing everything themselves :)

Issue #775, opened 2022-05-17
kiwix-serve and url encoding


mgautierfr:
On the PR #764, @veloman-yunkan raised a important question https://github.com/kiwix/libkiwix/pull/764#discussion_r866235011
As the issue title suggests, it is about url encoding

In libkiwix, we have a function to urlEncode a string : https://github.com/kiwix/libkiwix/blob/master/src/tools/stringTools.cpp#L146-L230
This function has a boolean parameter `encodeReserved` and the function does:
- keep alphanumerical char as it is
- keep char in `-_.!~*\()` as it is
- if `encodeReserved` is false : keep char in `;,/?:@&=+$` as it is (else encode it)
- encode every other char

However, it doesn't correspond to the standard specified in [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986) see [section 2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2).
Reserved chars can be (at best) but in two sets:
- gen-delims  = `:/?#[]@`
- sub-delims  = `!$&'()*+,;=`

As the name of the sets suggests, they are use to delimitate components in the url. Readers (url parsers) must separate components first using the delimiters and then url decode each components.
On the opposite side, writer must url encode the components and then composed them using delimiters. This make the delimiter chars sometime encode, sometime not.

From what I understand from the spec, we should always build url this way:
```cpp
std::string url = scheme + "://" + urlEncode(host);
for (auto& part: path_parts) {
 url += "/" + urlEncode(part);
}
char sep = '?';
for (auto& query_part: query) {
  url += sep + urlEncode(query_part.key) + "=" + urlEncode(query_part.value);
  sep = '&';
}
```
`urlEncode`  encoding the reserved chars. Of course, this is the simple version. A complete version should handle different scheme, authentication (`user@host`), anchor (`#title`), ....

However, everything is not so easy.
From [zim spec](https://wiki.openzim.org/wiki/ZIM_file_format#URL_Encoding), the url of the articles (path of the item) must NOT be url encoded. So they can contains things like `/`, `?`/`&` and `#`.
And this become funny:
[...]

PR #857, opened 2022-12-08
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates

PR #866, opened 2023-01-07
URI encoding of redirects

veloman-yunkan:
This is the first on of expected multiple fixes addressing kiwix/libkiwix#775.

This particular fix should at least eliminate @DarkmatterUAE's pain described at kiwix/kiwix-tools#589 :smiley:. I tested it with the smaller and newer [wikipedia_en_all_mini_2023-01.zim](http://download.kiwix.org/zim/wikipedia/wikipedia_en_all_mini_2023-01.zim) instead of [wikipedia_en_all_nopic_2022-01.zim](http://download.kiwix.org/zim/wikipedia/wikipedia_en_all_nopic_2022-01.zim) used in the original ticket, but they shouldn't be different with the respect to the data that triggered the bug.

Fixes kiwix/kiwix-tools#589


kelson42:
@mgautierfr @veloman-yunkan Probably not related to this PR, but worried that the macOS CI fails.

test/data/corner_cases/wtf?.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
mgautierfr, line 1: (Comment related to the commit, not the line)

Adding the 3 files to `corner_cases.zim` make the zim file increasing by 32.3 KB (1300%), where remove the fulltextindex was decreasing the file by 71KB. This seems too much. Can you verify that you have commit the right zim file ?

mgautierfr (changes requested):

test/data/corner_cases/wtf?.html
veloman-yunkan: Everything is correct. So far `corner_cases.zim` contained only empty items which provided an opportunity to omitting the title index. Appearance of the first entry with a title resulted in the title index being generated.

veloman-yunkan:
No changes, only rebased after merging #857.

kelson42:
@mgautierfr Good to merge then?

test/data/corner_cases/wtf?.html
mgautierfr: Arf. This is a pitty. I wonder if we should not have a option to not create xapian title index at all.
(Or maybe use a library compiled without xapian)

Anyway, not for this PR

mgautierfr (approved):

PR #868, opened 2023-01-11
Fix git clone on Windows

### LINKED ISSUE #867: Cannot git clone this repository on Windows
The git checkout for this repository on Windows fails due to a `?` in a directory name. The appveyor error is:

```
appveyor\install_libkiwix.cmd
Cloning into 'libkiwix'...
error: invalid path 'test/data/corner_cases/wtf?'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

I believe the problem was introduced [in this commit](https://github.com/kiwix/libkiwix/commit/78b2c1a273620b0afdd4e251b6feb0e10f6486b0).

adamlamar:
This changes `wtf?` to `wtf#`. The `?` is not allowed in filenames or directories on Windows, but `#` is. If I understand https://github.com/kiwix/libkiwix/issues/775 correctly, the `#` should be a reasonably good substitute since it has the same requirements as `?`.

The corner case zimfile was also updated accordingly and the test should pass.

Successful clone on Windows:

```
$ git clone https://github.com/adamlamar/libkiwix.git
Cloning into 'libkiwix'...
remote: Enumerating objects: 12725, done.
remote: Counting objects: 100% (210/210), done.
remote: Compressing objects: 100% (97/97), done.
remote: Total 12725 (delta 127), reused 168 (delta 111), pack-reused 12515
Receiving objects: 100% (12725/12725), 6.99 MiB | 11.50 MiB/s, done.
Resolving deltas: 100% (8415/8415), done.
```

Fixes #867


veloman-yunkan:
@adamlamar Thank you for the fix!

I have only one very superficial objection. Question mark plays well with wtf (and that is why some mental effort was spent on picking up a good and short question-like meme when creating a new unit test for #866). Anyone looking at the new filenames may experience some cognitive dissonance. They might think:

\- WTF? Why "wtf#"? Shouldn't it be "wtf?"? :smile: 

They will definitely have hard time figuring out the reason behind this counterintuitive usage of the hash symbol.

There are two solutions:

1. Replace the entire "wtf?" string with another one containing a hash symbol (e.g. C#). For consistency I would also update the content of the HTML file.
2. Assuming that the `create_corner_cases_zim_file` is not supposed to be executed on Windows, the "wtf?" files can be generated from inside that script.

And, BTW, nothing prevents us from implementing both options - then both "#" and "?" will be tested!

@mgautierfr What do you think? Should we just merge this PR or it makes sense to pursue consistency that perhaps no one will care about?

adamlamar:
Hey @veloman-yunkan, I agree it would be better if the test case made more sense in semantic terms. `wtf#` is a bit weird :) I was just laser focused on fixing this one issue since its blocking the kiwix-desktop windows build. I like the `C#` suggestion, pushed a change using `C#`.

adamlamar:
Oh and I agree that we could write a test case for all of the characters that need encoding (`?`, `#`, `&` if I understand correctly), but we'd have to avoid committing the file with `?`. I suppose we could have the script generate the necessary files, create the zim file, and then delete them (without committing to git).

test/data/corner_cases/c.html
```diff
@@ -0,0 +1 @@
+c#.html
```
veloman-yunkan, line 1: This comment applies to the name of the file.

Holding on to my previous concern wrt semantic consistency, I have to note that C is not C#. Maybe rename this file to c_sharp.html?

veloman-yunkan (changes requested):

test/data/corner_cases/c.html
adamlamar: Should be fixed now 👍

veloman-yunkan (commented):
Almost good to merge.

However, I don't think that we need all those three commits (each with their own 36KiB version of `corner_cases.zim`) in the history. Please squash all commits into one with a commit message explaining why the change is more extensive than it could have been.

adamlamar:
Sure, I squashed and pushed.

veloman-yunkan (approved):
Thanks for bearing with me and my whims!

PR #870, opened 2023-01-18
Quick-fix of issues caused by URL encoding

veloman-yunkan:
Fixes kiwix/kiwix-tools#594 (hopefully, without introducing another bug).

I started working on a more robust solution to #775 on branch [robust_uri_encoding](https://github.com/kiwix/libkiwix/compare/main...robust_uri_encoding) but then started having a feeling of [overdesigning and overcomplicating things](https://github.com/kiwix/libkiwix/commit/0d5ad498ffeb9678c13b2e159f96403d657d5ccf). So I tried a quick hack (see the last commit in this PR) and it seemed to work, so I decided to package the new unit-tests, fixes to a couple of bugs revealed by those tests and the hack into a PR of its own.

### DIFF
--- a/src/opds_catalog.cpp
+++ b/src/opds_catalog.cpp
@@ -42,19 +42,19 @@ std::string buildSearchString(const Filter& f)
 {
   std::ostringstream oss;
   if ( f.hasQuery() )
-    oss << AMP << "q=" << urlEncode(f.getQuery());
+    oss << AMP << "q=" << urlEncode(f.getQuery(), true);
 
   if ( f.hasCategory() )
-    oss << AMP << "category=" << urlEncode(f.getCategory());
+    oss << AMP << "category=" << urlEncode(f.getCategory(), true);
 
   if ( f.hasLang() )
-    oss << AMP << "lang=" << urlEncode(f.getLang());
+    oss << AMP << "lang=" << urlEncode(f.getLang(), true);
 
   if ( f.hasName() )
-    oss << AMP << "name=" << urlEncode(f.getName());
+    oss << AMP << "name=" << urlEncode(f.getName(), true);
 
   if ( !f.getAcceptTags().empty() )
-    oss << AMP << "tag=" << urlEncode(join(f.getAcceptTags(), ";"));
+    oss << AMP << "tag=" << urlEncode(join(f.getAcceptTags(), ";"), true);
 
   return oss.str();
 }
--- a/src/opds_catalog.cpp
+++ b/src/opds_catalog.cpp
@@ -42,19 +42,19 @@ std::string buildSearchString(const Filter& f)
 {
   std::ostringstream oss;
   if ( f.hasQuery() )
-    oss << AMP << "q=" << urlEncode(f.getQuery(), true);
+    oss << AMP << "q=" << urlEncode(f.getQuery());
 
   if ( f.hasCategory() )
-    oss << AMP << "category=" << urlEncode(f.getCategory(), true);
+    oss << AMP << "category=" << urlEncode(f.getCategory());
 
   if ( f.hasLang() )
-    oss << AMP << "lang=" << urlEncode(f.getLang(), true);
+    oss << AMP << "lang=" << urlEncode(f.getLang());
 
   if ( f.hasName() )
-    oss << AMP << "name=" << urlEncode(f.getName(), true);
+    oss << AMP << "name=" << urlEncode(f.getName());
 
   if ( !f.getAcceptTags().empty() )
-    oss << AMP << "tag=" << urlEncode(join(f.getAcceptTags(), ";"), true);
+    oss << AMP << "tag=" << urlEncode(join(f.getAcceptTags(), ";"));
 
   return oss.str();
 }
--- a/src/opds_dumper.cpp
+++ b/src/opds_dumper.cpp
@@ -216,7 +216,7 @@ string OPDSDumper::dumpOPDSFeedV2(const std::vector<std::string>& bookIds, const
      {"endpoint_root", endpointRoot},
      {"feed_id", gen_uuid(libraryId + endpoint + "?" + query)},
      {"filter", onlyAsNonEmptyMustacheValue(query)},
-     {"query", query.empty() ? "" : "?" + urlEncode(query)},
+     {"query", query.empty() ? "" : "?" + query},
      {"totalResults", to_string(m_totalResults)},
      {"startIndex", to_string(m_startIndex)},
      {"itemsPerPage", to_string(m_count)},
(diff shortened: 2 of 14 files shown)


mgautierfr:
If I resume the change:

<table>
<tr>
    <th>Old
    <th>New
<tr>
	<td colspan="2"> EncodeReserved: false
<tr>
	<td><code>/</code> -> <code>/</code>
	<td><code>/</code> -> <code>/</code>
<tr>
	<td><code>#</code> -> <code>%23</code>
    <td><code>#</code> -> <code>#</code>
<tr>
	<td colspan="2"> EncodeReserved: true
<tr>
	<td><code>/</code> -> <code>%2f</code>
	<td><code>/</code> -> <code>/</code>
<tr>
	<td><code>#</code> -> <code>%23</code>
    <td><code>#</code> -> <code>%23</code>
</table>

So with this change and encoded reserved to true, what changes is that we don't encode `/`
with this change and encoded reserved to false, what changes is that we don't encode `#`

(Please confirm I right)

`urlEncode(..., false)` is used for :
- build the redirect. This is indeed what we want
- encode the item's path when building the absolute path of a search result (we are good here)
- encode the query in the ODPSFeed2. **We are NOT good** as `#` is a begging of the fragment, and the fragment is not send to the server
- encode the category name in the categories opds feed. We should encode the `#` here as category may contain it. But I'm not aware of categories containing a `#`. @kelson42 @rgaudin ?

`urlEncode(..., true)` is used for:
- Build a search pattern in `handle_content` if we cannot find a book or an article. I think we are good here to not encode `/`
- Rebuild the queryString we received (parsed) from libmicrohttpd. I think we are good if we not encode `/`
- Build querystring in search_renderer. We are good also
- Encode the zim name when building the absolute path of search result. As zim name should not contains `/` we should be good also
- Encode the contentId (book name) in ods_dumper `fullEntryXML`. Same has above
- In `urlencoded` filter for mustache template. We probably want to encode `/` here (although I haven't found a use of this filter)

rgaudin:
> * encode the category name in the categories opds feed. We should encode the `#` here as category may contain it. But I'm not aware of categories containing a `#`. @kelson42 @rgaudin ?

As you know, [any value but `;` is allowed in Tags](https://wiki.openzim.org/wiki/Tags) but `_category` is private and we should thus only consider our own use, which doesn't make uses of `#`. 👍

veloman-yunkan:
@mgautierfr Thanks for the meticulous analysis!

> So with this change and encoded reserved to true, what changes is that we don't encode `/` with this change and encoded reserved to false, what changes is that we don't encode `#`
> 
> (Please confirm I right)

Confirming

> 
> `urlEncode(..., false)` is used for :
> 
>   ...
> 
>   * encode the item's path when building the absolute path of a search result (we are good here)

This one is a bug. URI delimiters occurring in the path must be encoded! Fixed.

> 
>   * encode the query in the ODPSFeed2. **We are NOT good** as `#` is a begging of the fragment, and the fragment is not send to the server

Good catch. While working on the robust solution, I changed `RequestContext::get_query()` so that it returns a properly URI-encoded representation of the query, but since that change broke more unit-tests than I was willing to accept, I had to stash it away. Now I see that I will have to deal with the same issue in this quick-fix PR. 

>   * encode the category name in the categories opds feed. We should encode the `#` here as category may contain it. But I'm not aware of categories containing a `#`. @kelson42 @rgaudin ?
> 
Fixed. Category name is now fully URI-encoded. 
> 
> `urlEncode(..., true)` is used for:
> 
>  ...
>
>   * In `urlencoded` filter for mustache template. We probably want to encode `/` here (although I haven't found a use of this filter)

Dropped this one.

veloman-yunkan:
> > * encode the query in the ODPSFeed2. **We are NOT good** as `#` is a begging of the fragment, and the fragment is not send to the server
> 
> Good catch. While working on the robust solution, I changed `RequestContext::get_query()` so that it returns a properly URI-encoded representation of the query, but since that change broke more unit-tests than I was willing to accept, I had to stash it away. Now I see that I will have to deal with the same issue in this quick-fix PR.

Done

veloman-yunkan:
Also fixed usage of `urlEncode()` in `kiwix::getSeachUrl()`, whereupon no instance of `urlEncode(..., false)` is left (other than in the unit-test of `urlEncode()`). Should we drop the extra parameter of `urlEncode()`?

mgautierfr:
> >  encode the item's path when building the absolute path of a search result (we are good here)
>
> This one is a bug. URI delimiters occurring in the path must be encoded! Fixed.

We should not. The search result link is in the form : `http://content/<bookName>/<path_to_article>`.
If we encode `/` in `path_to_article` we will break relative links.

> Also fixed usage of urlEncode() in kiwix::getSeachUrl(), whereupon no instance of urlEncode(..., false) is left (other than in the unit-test of urlEncode()). Should we drop the extra parameter of urlEncode()?

Yes. And we could call it `encodeURIComponent`.
For the encoding of a item path (keeping the `/`) we could have a method with this (pseudo) code :
```
string encodeItemPath(string itemPath) {
    auto components = itemPath.split('/');
    string encoded;
    char separator = '';
    for (component: components) {
        encoded += separator + encodeUrIComponent(component);
        separator = '/'
    }
   return encoded;
}
```
This way we preserve `/` in path and we still encode any `#`/`?` in article path which will break the url.

veloman-yunkan:
> > > encode the item's path when building the absolute path of a search result (we are good here)
> > 
> > 
> > This one is a bug. URI delimiters occurring in the path must be encoded! Fixed.
> 
> We should not. The search result link is in the form : `http://content/<bookName>/<path_to_article>`. If we encode `/` in `path_to_article` we will break relative links.

But that's what is changed by this PR! `/` is now never encoded by `urlEncode()`.

mgautierfr:
> But that's what is changed by this PR! / is now never encoded by urlEncode().

🤦🏽‍♂️ My bad. I've mixed too much things. 

So I think we are good. We just have to remove the `encodeReserved` argument. But it should be functionally good.

---

Just want to mention for potential issue later: We are ok to not encode `/` in zim name because the zim name is inferred from the zim filename and so doesn't contain `/`.  But if it changes, we would have to adapt again and encode `/` in bookname (for exemple, we may want to use the name in the library as @rgaudin thank at first sight in https://github.com/kiwix/kiwix-tools/issues/596)
In fact, we should already encode the `/` as the zim name may be any name if the libkiwix user provide its own namemapper. 
But for a quick-fix we are good. To keep in mind for the "final" solution of #775.

veloman-yunkan:
> So I think we are good. We just have to remove the `encodeReserved` argument. But it should be functionally good.

Done (also rebased).

> we may want to use the name in the library as @rgaudin thank at first sight in [kiwix/kiwix-tools#596](https://github.com/kiwix/kiwix-tools/issues/596)

I stumbled upon "thank" in this sentence, but then figured out that you meant "thought" :)

mgautierfr (approved):

kelson42:
@veloman-yunkan Any chance tomget the final PR to close this ticket based on https://github.com/kiwix/libkiwix/compare/main...robust_uri_encoding ?

PR #882, opened 2023-02-05
Add OPDS/Atom Feed link in HTML head

### LINKED ISSUE #832: Add a feed link to kiwix-serve welcome page
The feed link should have the same filter (URL) as the one currently running.

Read https://medium.com/samsung-internet-dev/add-rss-feeds-to-your-website-to-keep-your-core-readers-engaged-3179dca9c91e form more details.

This should help user to be kept informed about new release of their most wanted ZIM files.

juuz0:
Fixes #832


juuz0:
I think this is what was needed here (using `Want My RSS` extension on firefox - there was no inbuilt way to check which i know of). It links to the atom feed with correct filters. Working!

image

kelson42:
@juuz0 Thank you, I will give a try

kelson42:
@juuz0 What you have done looks good to me from a user perspective. That is what I was expecting.

juuz0:
@kelson42 Awesome! I don't why the CI is failing though, is it a common problem to current-libkiwix or just this PR?

(As a sidenote, how did you check the RSS feed? Is there an inbuilt feature in firefox/other browser?)

kelson42:
> @kelson42 Awesome! I don't why the CI is failing though, is it a common problem to current-libkiwix or just this PR?

Yes, a problem with the package building. Nothing related to your PR.
 
> (As a sidenote, how did you check the RSS feed? Is there an inbuilt feature in firefox/other browser?)

I checked the HTML and know how this feature should work. I have also a proper browser which displays an icon in the searchbar.

juuz0:
Added image @kelson42 
It's hidden in mobile version because I don't know the correct design to put it (might need discussion here if it's desirable) 
image

static/templates/index.html
```diff
@@ -10,6 +10,13 @@
       href="{{root}}/skin/index.css?KIWIXCACHEID"
       rel="Stylesheet"
     />
+    <link 
+      rel="alternate" 
+      type="application/rss+xml" 
+      title="Kiwix RSS Feed"
+      id="rssExtension" 
+      href="/catalog/search?"
```
veloman-yunkan, line 18: We better use the [new OPDS API](https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#catalog-v2-opds-api) (in this case - [`/catalog/v2/entries`](https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#catalog-v2-entries)).

static/skin/index.js
```diff
@@ -18,6 +18,20 @@
...
+        let tempParams = new URLSearchParams(window.location.search);
+        let keysForDel = [];
+        tempParams.forEach((value, key) => {
+            if (value == '') {
+                keysForDel.push(key);
+            }
+        });
+        keysForDel.forEach(key => {
+            tempParams.delete(key);
+        });
```
veloman-yunkan, line 31: You can do the same with a single loop if instead of collecting keys to be deleted, you construct a filtered list of params during the first and only iteration:

```js
const inputParams = new URLSearchParams(window.location.search);
const filteredParams = new URLSearchParams();
for (const [key, value] of inputParams) {
    if ( value != '' ) {
        filteredParams.append(key, value);
    }
}
```

static/templates/index.html
```diff
@@ -55,6 +55,9 @@
           </select>
         </div>
       </div>
+      <a href="/catalog/search?" id="rssFeedLink">
```
veloman-yunkan, line 58: Don't forget to change this to `/catalog/v2/entries`, too.

veloman-yunkan (changes requested):

mgautierfr (changes requested):
Code seems good to me.
(Edit: But I have missed the wrong usage of the api, I agree with @veloman-yunkan here, better using `/catalog/v2/entries`)

However, I think the icon is a bit to lower, I would have put it at the top-right of the top bar.
But I will not block for that if there is a consensus against me.

kelson42:
> However, I think the icon is a bit to lower, I would have put it at the top-right of the top bar.

Yes, top-right corner would be better IMO too.

static/skin/index.js
juuz0: smart!

juuz0:
Unrelated to this issue but https://github.com/kiwix/libkiwix/blob/main/static/skin/index.js#L22 is also using the old API, should that be changed too?

kelson42:
> Unrelated to this issue but https://github.com/kiwix/libkiwix/blob/main/static/skin/index.js#L22 is also using the old API, should that be changed too?

yes please

static/templates/index.html
```diff
@@ -10,6 +10,13 @@
       href="{{root}}/skin/index.css?KIWIXCACHEID"
       rel="Stylesheet"
     />
+    <link 
+      rel="alternate" 
+      type="application/rss+xml" 
+      title="Kiwix RSS Feed"
+      id="rssExtension" 
+      href="/catalog/v2/entries"
```
veloman-yunkan, line 18: Missing `{{root}}`. I know that it doesn't matter since it will be fixed by `updateRssFeedLink()` after the HTML is loaded, but it is better to be consistent.

static/templates/index.html
```diff
@@ -42,6 +42,9 @@
     <script type="text/javascript" src="{{root}}/skin/index.js?KIWIXCACHEID" defer></script>
   </head>
   <body>
+    <a href="/catalog/v2/entries" id="rssFeedLink">
```
veloman-yunkan, line 45: Same here

veloman-yunkan (changes requested):

static/templates/index.html
```diff
@@ -35,6 +42,9 @@
...
+      <img src="{{root}}/skin/rss.png?KIWIXCACHEID" class="rssLogo" alt="Kiwix RSS Feed" aria-label="Kiwix RSS Feed" title="Kiwix RSS Feed">
```
kelson42, line 46: Label, alt, title should be "Library OPDS Feed"
Remove "rss" and replace by "feed" everywhere

kelson42 (changes requested):
It works fine but few changes necessary to be polished

juuz0:
Rebased on main

static/skin/index.js
```diff
@@ -26,7 +26,9 @@
                 filteredParams.append(key, value);
             }
         }
-        document.querySelector('#feed_link').href = `${root}/catalog/v2/entries?${filteredParams.toString()}`;
+        const feedLink = `${root}/catalog/v2/entries?${filteredParams.toString()}`;
+        document.querySelector('#feed_link').href = feedLink;
+        document.querySelector('#feedLink').href = feedLink;
     }
```
veloman-yunkan, line 32: I see in the current naming of the OPDS feed link ids a risk that any future links will be named `feedlink`, `FeedLink`, etc :)

juuz0: I was worried about it too :P but kelson's recommendation so I let it pass.
What do you think is a better name?
@kelson42 @veloman-yunkan

mgautierfr: We should keep only version of `feedLink` or `feed_link` to have the naming consistant.
But we should have a id for the link in the head as `head_feed_link` (or `headFeedLink`) and `feed_link` (or `FeedLink`).

juuz0: `headFeedLink` then :)

mgautierfr (approved):
We are good on my side.

You could have set the head id to `headFeedLink` in the first commit to avoid the change in the second commit. But I will not block for that.

veloman-yunkan (approved):

mgautierfr:
@juuz0 Please rebase on master so we can merge.

juuz0:
@mgautierfr Done.

mgautierfr:
@juuz0 Thanks.

Comment from @kelson42 has been addressed. So I think we are free to merge without its review.

PR #886, opened 2023-02-07
Make Download threadsafe and more thread compliant.

### LINKED ISSUE #873: Make aria2 wrapper and Downloader thread safe/compliant.
See https://github.com/kiwix/kiwix-desktop/issues/87#issuecomment-1403594478

aria2 wrapper use a lock to prevent concurrent access to the curl handle. This create a bottleneck as we can do only one request at the time. If one request takes time (and starting a download does on window) all other requests are delayed (and blocked).

We should make aria2 wrapper thread compliant and also make `Downloader` thread safe to allow a real multi thread usage of `Downloader`

(Not sure we need to make `Download` threadsafe. We can simple state that methods on one `Download` are not threadsafe)

mgautierfr:
Fix #873


include/downloader.h
```diff
@@ -71,14 +71,13 @@ class Download {
...
-   * By passing true to follow, `Dowload` will detect that and will track the
-   * second download instead.
+   * By default, `Download` try to detect that and track the second download when
+   * the first one is finished.
+   * By passing false to follow, `Dowload` will only track the first download.
    *
    * `getFoo` methods are based on the last statusUpdate.
    */
-  void updateStatus(bool follow=false);
+  void updateStatus(bool follow=true);
```
veloman-yunkan, line 80: Since downloader is part of the libkiwix API, this semantic change can in theory lead to unpleasant surprises for clients of libkiwix who update to new version without reading the changelog carefully. A safer way would be introducing a new function with the new semantics and deprecating the old one.

src/downloader.cpp
```diff
@@ -180,23 +180,20 @@ Download* Downloader::startDownload(const std::string& uri, const std::vector<st
...
       }
     }
     for (auto gid : mp_aria->tellWaiting()) {
       if (gid == did) {
         m_knownDownloads[gid] = std::unique_ptr<Download>(new Download(mp_aria, gid));
-        m_knownDownloads.at(gid).get()->updateStatus(true);
         return m_knownDownloads[gid].get();
       }
-  }
+    }
```
veloman-yunkan, line 199: A download may be missed in unlucky scenarios when it moves from the waiting list to the active list after the call to `mp_aria->tellActive()` has completed but before the first loop is done and `mp_aria->tellWaiting()` is called. A similar problem is present in the constructor of `Downloader`.

veloman-yunkan (commented):
This is the first time I look at the `kiwix::Downloader` code and, IMO, it will benefit from some extra refactoring/cleanup. Also it might be a good opportunity to replace libcurl with httplib.

However, since this PR  immediately addresses a user problem it makes sense to merge it quickly (after the concern about the semantic change in the API is resolved) and discuss the proposed improvements without any haste.

include/downloader.h
mgautierfr: I agree with the problem you raise. But I would like to avoid adding a `updateStatus` for this.
Maybe we could  simply remove the default value. For people (other than us ?) using with `true` parameter it will change nothing.
For other using the default value (if there is one at all), it will be a compilation error.

veloman-yunkan: Yes, that's a better option

src/downloader.cpp
```diff
@@ -128,20 +128,20 @@ Downloader::Downloader() :
...
   }
   try {
-    for (auto gid : mp_aria->tellWaiting()) {
+    for (auto gid : mp_aria->tellActive()) {
       m_knownDownloads[gid] = std::unique_ptr<Download>(new Download(mp_aria, gid));
       m_knownDownloads[gid]->updateStatus(false);
     }
   } catch (std::exception& e) {
-    std::cerr << "aria2 tellWaiting failed : " << e.what() << std::endl;
+    std::cerr << "aria2 tellActive failed : " << e.what() << std::endl;
```
veloman-yunkan, line 144: Here reversing the order also inverses the issue - now you can have the same download processed twice. In the second loop you only have to create a new download if it is not already in `m_knownDownloads`.

include/downloader.h
```diff
@@ -71,13 +71,15 @@ class Download {
...
+   * If `follow` is true, updateStatus tries to detect that and tracks
+   * the second download when the first one is finished.
+   * By passing false to `follow`, `Download` will only track the first download.
    *
    * `getFoo` methods are based on the last statusUpdate.
+   *
+   * @param follow: Do we have to follow following downloads.
    */
-  void updateStatus(bool follow=true);
+  void updateStatus(bool follow);
```
veloman-yunkan, line 82: During rebase/fixup the title of the squashed commit must be changed

veloman-yunkan (changes requested):

veloman-yunkan (commented):
Please rebase&fixup, fix the commit and PR titles and we should be good to merge.

veloman-yunkan (approved):

PR #890, opened 2023-02-08
Full URI-encoding of URLs returned by kiwix-serve

### LINKED ISSUE #441: Kiwix-serve random feature returns partly broken URLs
With a ZIM with the article `?_(album_de_Nena)` from WPFR. The random button returns the following URL: http://localhost:8080/wikipedia_fr_articles_2019-07/A/?_(album_de_Nena) and it should return http://localhost:8080/wikipedia_fr_articles_2019-07/A/%3F_(album_de_Nena).

veloman-yunkan:
This PR tries to ensure that the full path component of URLs returned by kiwix-serve is properly  URI-encoded. In particular, all of the following constituents of the path component encoded:
- root location
- book name
- path of the entry within the book

Fixes #441


test/server.cpp
```diff
@@ -631,157 +631,157 @@ TEST_F(ServerTest, Http404HtmlError)
...
 
-    { /* url */ R"(/ROOT/content/zimfile/"><svg onload=alert(1)>)",
+    { /* url */ R"(/ROOT%23%3F/content/zimfile/"><svg onload=alert(1)>)",
       book_name=="zimfile" &&
       book_title=="Ray Charles" &&
       expected_body==R"(
     <h1>Not Found</h1>
     <p>
-      The requested URL "/ROOT/content/zimfile/&quot;&gt;&lt;svg onload=alert(1)&gt;" was not found on this server.
+      The requested URL "/ROOT%23%3F/content/zimfile/%22%3E%3Csvg%20onload%3Dalert(1)%3E" was not found on this server.
```
mgautierfr, line 717: I would say that we need the url decoded version here.
The url entered by the user don't contain % encoding, we may need to encode it when used in url. But this is not really a url, this is text for the user.

test/data/create_corner_cases_zim_file
```diff
@@ -1,7 +1,24 @@
...
+#
+#  ?
+#  =
+#  + (that's a pity, since the + symbol in a ZIM filename is replaced with the
+#     text 'plus' when the ZIM file is added to kiwix-serve's library and it
+#     would be nice to test that functionality)
+#
+# Assuming that tests are NOT run under Windows, above symbols can be included
+# in testing if the file is renamed while copying to the build directory (see
+# test/meson.build), though that would make maintenance slightly more confusing.
```
mgautierfr, line 18: We will compile libkiwix on windows and it would be a good thing to test it also.

One thing will could do is rename the file only on linux and don't rename it on windows.
(And adapt the code accordingly). But in this case, that would make maintenance more than slightly more confusing.

mgautierfr (commented):
Two comments that are more open questions than issue.
I'm not sure at all we should implement what suggested.

test/server.cpp
veloman-yunkan: Fixed, though with a subtle difference from how it worked originally. I `urlDecode()` weakly so that most reserved symbols remain encoded (otherwise the URL may be misinterpreted because of the `?` or `#` symbols appearing in it). In this particular case `=` remains URI-encoded.

veloman-yunkan: I fixed it in place (rather than in a fixup commit) so that the commit message could be updated too.

test/data/create_corner_cases_zim_file
veloman-yunkan: Should we do anything about it now/in this PR?

mgautierfr: No, we can leave it at it is.

mgautierfr (approved):

PR #896, opened 2023-02-11
Stick kiwixNav on top

### LINKED ISSUE #883: kiwix-serve welcome page filter bar should be sticky/frozen
image

If we scroll down the filter bar disappears, but actually it should just stay. Only the tiles should scroll down.

First reported at https://github.com/kiwix/libkiwix/issues/832#issuecomment-1418779825

juuz0:
Fixes #883

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -18,6 +18,7 @@
     params.delete('userlang');
     let timer;
     let languages = {};
+    let previousScrollTop = Infinity;
 
     function updateFeedLink() {
         const inputParams = new URLSearchParams(window.location.search);
@@ -453,6 +454,22 @@
         }
     }
 
+    function updateNavVisibilityState() {
+        const st = window.scrollY;
+        const enableAutoHiding = document.body.clientWidth < 590;
+        if ((Math.abs(previousScrollTop - st) <= 5) || !enableAutoHiding)
+            return;
+        const kiwixNav = document.querySelector('.kiwixNav');
+        if (st > previousScrollTop) {
+            kiwixNav.style.position = 'fixed';
+            kiwixNav.style.top = '-100%';
+        } else {
+            kiwixNav.style.position = 'sticky';
+            kiwixNav.style.top = '0';
+        }
+        previousScrollTop = st;
+    }
+
     window.addEventListener('resize', (event) => {
         if (timer) {clearTimeout(timer)}
         timer = setTimeout(() => {
@@ -531,6 +548,7 @@
         }
         updateFeedLink();
         setCookie(filterCookieName, params.toString(), oneDayDelta);
+        setInterval(updateNavVisibilityState, 250);
     };
 
     // required by i18n.js:setUserLanguage()
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -27,6 +27,7 @@ body {
     position: sticky;
     top: 0;
     z-index: 3;
+    transition: all 0.5s ease;
 }
 
 .kiwixHomeBody__results {
@@ -137,6 +138,7 @@ body {
     display: flex;
     flex-direction: column;
     align-items: center;
+    width: 100%;
 }
 
 .tagFilterLabel {
@@ -178,7 +180,7 @@ body {
 }
 
 #uiLanguageSelectorButton {
-  margin: 16px 12px 0 0;
+  margin: 0 12px 0 0;
   float: right;
 }
 
@@ -467,7 +469,6 @@ body {
 }
 
 .feedLogo {
-    margin: 16px 12px 0 0;
     height: 30px;
     float: right;
     border-radius: 5px;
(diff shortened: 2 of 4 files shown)


kelson42 (approved):
@juuz0 LGTM. TXH!

kelson42:
@juuz0 Don't worry if this is not merged straight. We want to merge #846 first.

veloman-yunkan (commented):
On narrow screens the topbar section's use of the screen space is quite uneconomic:
Screenshot from 2023-02-16 15-40-46
Making the topbar sticky worsens the situation.

I propose that on narrow screens 
1. the topbar's vertical space usage is fixed and
2. the topbar is autohidden while scrolling (similar to how ZIM viewer's toolbar behaves).

kelson42:
@juuz0 Any feedback on @veloman-yunkan request? This PR is part of the next milestone and have only a few tickets left. Therefore we should move forward quickly on this. If you could rebase, that woukd be great as well.

juuz0:
Sorry I totally missed this review comment while working on other PR, doing it ASAP

kelson42:
@juuz0 Thx! Github provides great notification features. I recommend to any Kiwix dev to put attention on this. If you don't this just slows down artificially the whole development process and create additional work load to others (like you know, it is not really the role of the CTO or any other dev to chase a colleague so he communicates transparently on his own work).

juuz0:
> like you know, it is not really the role of the CTO or any other dev to chase a colleague so he communicates transparently about his own work
 
 Understandable. Apologies.

juuz0:
There's a style issue, tiles go to the left while scrolling for no reason..Investigating

veloman-yunkan (changes requested):
The UI language selector button and OPDS feed link button have disappeared.

juuz0:
> The UI language selector button and OPDS feed link button have disappeared.

I put them inside `kiwixNav` element now.

static/skin/index.js
```diff
@@ -453,6 +454,24 @@
         }
     }
 
+    function updateNavVisibilityState() {
+        const st = window.scrollY;
+        const scrollPossible = document.querySelector('.kiwixHomeBody').offsetHeight > window.screen.height;
```
veloman-yunkan, line 459: Isn't it better to use `window.innerHeight` instead of `window.screen.height`? After all what matters is the size of the window rather than the size of the screen. Then this functionality will work correctly also in small non-maximized windows on large screens (the way I tested this PR on my computer).

static/skin/index.js
```diff
@@ -453,6 +454,24 @@
...
+        const documentWidthIsLessThan590 = document.body.clientWidth < 590;
```
veloman-yunkan, line 460: There is too much detail in the name of this variable. What if we decide to make the nav bar autohiding on slightly narrower or wider screens? Then either this variable's name will get out of sync with reality or will have to be updated too. It's better to emphasize the meaning/purpose of this variable by naming it something like `enableAutoHiding`.

static/skin/index.js
veloman-yunkan: On the other hand why we need this `scrollPossible` variable at all?

static/skin/index.js
```diff
@@ -453,6 +454,24 @@
...
+        if ((Math.abs(previousScrollTop - st) <= 5) || !scrollPossible || !documentWidthIsLessThan590)
+            return;
+        const kiwixNav = document.querySelector('.kiwixNav');
+        if (st > previousScrollTop) {
+            kiwixNav.style.position = 'fixed';
+            kiwixNav.style.top = '-100%';
+            document.querySelector('.kiwixHomeBody').style.width = '100%';
```
veloman-yunkan, line 467: Why do you need this line here? Is the width of `kiwixHomeBody` element set dynamically to something else somewhere else?

veloman-yunkan (changes requested):

static/skin/index.js
juuz0: There was a blink effect on trying to scroll when only 1 book was shown (it kept removing and adding kiwixNav). No idea how I can't reproduce it anymore. Removed this variable

static/skin/index.js
juuz0: Done.

static/skin/index.js
juuz0: I don't think (and have checked) it's dynamically set anywhere. But still, when I remove `kiwixNav` all the list, gets shifted to the left on scrolling. Checking through inspector, the width is reduced hence the line.

veloman-yunkan: Can't it be set in the CSS instead?

juuz0: You're right. Thank you. At such moments, I realise I shouldn't haste.

veloman-yunkan (approved):

PR #897, opened 2023-02-12
A gift to javascript naysayers

### LINKED ISSUE #615: No JS fallback for kiwix-serve welcome page
The new kiwix-serve UI is more practical but it's now impossible to use kiwix-serve without JS enabled.

There are many ways to solve this but if we consider no-js to be an edge case, we could, at least have a `<noscript />` tag pointing to a separate endpoint with a static list of contents.

Screen Shot 2021-09-13 at 10 25 23

juuz0:
Fixes #615

### DIFF
--- a/include/html_dumper.h
+++ b/include/html_dumper.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2023 Nikhil Tanwar <2002nikhiltanwar@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#ifndef KIWIX_HTML_DUMPER_H
+#define KIWIX_HTML_DUMPER_H
+
+#include <string>
+
+#include "library_dumper.h"
+
+namespace kiwix
+{
+
+/**
+ * A class to dump Library in HTML format.
+ */
+class HTMLDumper : public LibraryDumper
+{
+ public:
+  HTMLDumper(const Library* library, const NameMapper* NameMapper);
+  ~HTMLDumper();
+
+
+  /**
+   * Dump library in HTML
+   *
+   * @return HTML content
+   */
+  std::string dumpPlainHTML() const;
+};
+
+}
+
+#endif // KIWIX_HTML_DUMPER_H
--- a/include/html_dumper.h
+++ b/include/html_dumper.h
@@ -42,7 +42,7 @@ class HTMLDumper : public LibraryDumper
    *
    * @return HTML content
    */
-  std::string dumpPlainHTML() const;
+  std::string dumpPlainHTML(kiwix::Filter filter) const;
 };
 
 }
--- a/include/library_dumper.h
+++ b/include/library_dumper.h
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2023 Nikhil Tanwar <2002nikhiltanwar@gmail.com>
+ * Copyright 2017 Matthieu Gautier <mgautier@kymeria.fr>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#ifndef KIWIX_LIBRARY_DUMPER_H
+#define KIWIX_LIBRARY_DUMPER_H
+
+#include <string>
+
+
+#include "library.h"
+#include "name_mapper.h"
+#include <mustache.hpp>
+
+namespace kiwix
+{
+
+/**
+ * A base class to dump Library in various formats.
+ *
+ */
+class LibraryDumper
+{
+ public:
+  LibraryDumper(const Library* library, const NameMapper* NameMapper);
+  ~LibraryDumper();
+
+  void setLibraryId(const std::string& id) { this->libraryId = id;}
+
+  /**
+   * Set the root location used when generating url.
+   *
+   * @param rootLocation the root location to use.
+   */
+  void setRootLocation(const std::string& rootLocation) { this->rootLocation = rootLocation; }
+
+  /**
+   * Set some informations about the search results.
+   *
+   * @param totalResult the total number of results of the search.
+   * @param startIndex the start index of the result.
+   * @param count the number of result of the current set (or page).
+   */
+  void setOpenSearchInfo(int totalResult, int startIndex, int count);
+  
+  /**
+   * Sets user default language
+   * 
+   * @param userLang the user language to be set
+   */
+  void setUserLanguage(std::string userLang) { this->m_userLang = userLang; }
+
+  /**
+   * Get the data of categories
+   */
+  kainjow::mustache::list getCategoryData() const;
+
+  /**
+   * Get the data of languages
+   */
+  kainjow::mustache::list getLanguageData() const;
+
+ protected:
+  const kiwix::Library* const library;
+  const kiwix::NameMapper* const nameMapper;
+  std::string libraryId;
+  std::string rootLocation;
+  std::string m_userLang;
+  int m_totalResults;
+  int m_startIndex;
+  int m_count;
+};
+}
+
+#endif // KIWIX_LIBRARY_DUMPER_H
--- a/include/opds_dumper.h
+++ b/include/opds_dumper.h
@@ -28,6 +28,7 @@
 
 #include "library.h"
 #include "name_mapper.h"
+#include "library_dumper.h"
 
 using namespace std;
 
@@ -38,11 +39,10 @@ namespace kiwix
  * A tool to dump a `Library` into a opds stream.
  *
  */
-class OPDSDumper
+class OPDSDumper : public LibraryDumper
 {
  public:
-  OPDSDumper() = default;
-  OPDSDumper(Library* library, NameMapper* NameMapper);
+  OPDSDumper(const Library* library, const NameMapper* NameMapper);
   ~OPDSDumper();
 
   /**
@@ -85,38 +85,6 @@ class OPDSDumper
    * @return The OPDS feed.
    */
   std::string languagesOPDSFeed() const;
-
-  /**
-   * Set the id of the library.
-   *
-   * @param id the id to use.
-   */
-  void setLibraryId(const std::string& id) { this->libraryId = id;}
-
-  /**
-   * Set the root location used when generating url.
-   *
-   * @param rootLocation the root location to use.
-   */
-  void setRootLocation(const std::string& rootLocation) { this->rootLocation = rootLocation; }
-
-  /**
-   * Set some informations about the search results.
-   *
-   * @param totalResult the total number of results of the search.
-   * @param startIndex the start index of the result.
-   * @param count the number of result of the current set (or page).
-   */
-  void setOpenSearchInfo(int totalResult, int startIndex, int count);
-
- protected:
-  kiwix::Library* library;
-  kiwix::NameMapper* nameMapper;
-  std::string libraryId;
-  std::string rootLocation;
-  int m_totalResults;
-  int m_startIndex;
-  int m_count;
 };
 }
 
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -19,16 +19,24 @@ std::string HTMLDumper::dumpPlainHTML() const
 {
   kainjow::mustache::list booksData;
   for ( const auto& bookId : library->getBooksIds() ) {
-    const auto bookEx = library->getBookById(bookId);
-    const auto bookName = bookEx.getName();
+    const auto bookObj = library->getBookById(bookId);
+    const auto bookTitle = bookObj.getTitle();
+    const auto bookDescription = bookObj.getDescription();
+    const auto langCode = bookObj.getCommaSeparatedLanguages();
+    const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
+    std::string faviconAttr = "style=background-image:url(" + bookIconUrl + ")";
     booksData.push_back(kainjow::mustache::object{
-      {"name", bookName}
+      {"title", bookTitle},
+      {"description", bookDescription},
+      {"langCode", langCode},
+      {"faviconAttr", faviconAttr}
     });
   }
 
   return render_template(
              RESOURCE::templates::no_js_library_page_html,
              kainjow::mustache::object{
+               {"root", rootLocation},
                {"books", booksData }
              }
   );
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -39,12 +39,17 @@ std::string HTMLDumper::dumpPlainHTML() const
   for ( const auto& bookId : library->getBooksIds() ) {
     const auto bookObj = library->getBookById(bookId);
     const auto bookTitle = bookObj.getTitle();
+    std::string contentId = "";
+    try {
+      contentId = urlEncode(nameMapper->getNameForId(bookId));
+    } catch (...) {}
     const auto bookDescription = bookObj.getDescription();
     const auto langCode = bookObj.getCommaSeparatedLanguages();
     const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
     const auto tags = bookObj.getTags();
     std::string faviconAttr = "style=background-image:url(" + bookIconUrl + ")";
     booksData.push_back(kainjow::mustache::object{
+      {"id", contentId},
       {"title", bookTitle},
       {"description", bookDescription},
       {"langCode", langCode},
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -99,7 +99,8 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
                     getTranslation("book-filtering-all-languages"),
                     getTranslation("powered-by-kiwix-html"),
                     getTranslation("welcome-to-kiwix-server"),
-                    getTranslation("preview-book")
+                    getTranslation("preview-book"),
+                    getTranslation("welcome-page-overzealous-filter", {{"URL", "?lang="}})
   };
 
   return render_template(
@@ -110,6 +111,7 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
                {"searchQuery", searchQuery},
                {"languages", languages},
                {"categories", categories},
+               {"noResults", filteredBooks.size() == 0},
                {"translations", translations}
              }
   );
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -1,6 +1,7 @@
 #include "html_dumper.h"
 #include "libkiwix-resources.h"
 #include "tools/otherTools.h"
+#include "tools.h"
 
 namespace kiwix
 {
@@ -15,6 +16,23 @@ HTMLDumper::~HTMLDumper()
 {
 }
 
+namespace {
+
+kainjow::mustache::list getTagList(std::string tags)
+{
+  const auto tagsList = kiwix::split(tags, ";", true, false);
+  kainjow::mustache::list finalTagList;
+  for (auto tag : tagsList) {
+  if (tag[0] != '_')
+    finalTagList.push_back(kainjow::mustache::object{
+      {"tag", tag}
+    });
+  }
+  return finalTagList;
+}
+
+} // unnamed namespace
+
 std::string HTMLDumper::dumpPlainHTML() const
 {
   kainjow::mustache::list booksData;
@@ -24,12 +42,14 @@ std::string HTMLDumper::dumpPlainHTML() const
     const auto bookDescription = bookObj.getDescription();
     const auto langCode = bookObj.getCommaSeparatedLanguages();
     const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
+    const auto tags = bookObj.getTags();
     std::string faviconAttr = "style=background-image:url(" + bookIconUrl + ")";
     booksData.push_back(kainjow::mustache::object{
       {"title", bookTitle},
       {"description", bookDescription},
       {"langCode", langCode},
-      {"faviconAttr", faviconAttr}
+      {"faviconAttr", faviconAttr},
+      {"tagList", getTagList(tags)}
     });
   }
 
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -1,0 +1,37 @@
+#include "html_dumper.h"
+#include "libkiwix-resources.h"
+#include "tools/otherTools.h"
+
+namespace kiwix
+{
+
+/* Constructor */
+HTMLDumper::HTMLDumper(const Library* library, const NameMapper* nameMapper)
+  : LibraryDumper(library, nameMapper)
+{
+}
+/* Destructor */
+HTMLDumper::~HTMLDumper()
+{
+}
+
+std::string HTMLDumper::dumpPlainHTML() const
+{
+  kainjow::mustache::list booksData;
+  for ( const auto& bookId : library->getBooksIds() ) {
+    const auto bookEx = library->getBookById(bookId);
+    const auto bookName = bookEx.getName();
+    booksData.push_back(kainjow::mustache::object{
+      {"name", bookName}
+    });
+  }
+
+  return render_template(
+             RESOURCE::templates::no_js_library_page_html,
+             kainjow::mustache::object{
+               {"books", booksData }
+             }
+  );
+}
+
+} // namespace kiwix
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -3,6 +3,7 @@
 #include "tools/otherTools.h"
 #include "tools.h"
 #include "tools/regexTools.h"
+#include "server/i18n.h"
 
 namespace kiwix
 {
@@ -88,15 +89,28 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
     });
   }
 
+  auto getTranslation = i18n::GetTranslatedStringWithMsgId(m_userLang);
+
+  const auto translations = kainjow::mustache::object{
+                    getTranslation("search"),
+                    getTranslation("download"),
+                    getTranslation("count-of-matching-books", {{"COUNT", to_string(filteredBooks.size())}}),
+                    getTranslation("book-filtering-all-categories"),
+                    getTranslation("book-filtering-all-languages"),
+                    getTranslation("powered-by-kiwix-html"),
+                    getTranslation("welcome-to-kiwix-server"),
+                    getTranslation("preview-book")
+  };
+
   return render_template(
              RESOURCE::templates::no_js_library_page_html,
              kainjow::mustache::object{
                {"root", rootLocation},
                {"books", booksData },
                {"searchQuery", searchQuery},
-               {"resultsCount", to_string(filteredBooks.size())},
                {"languages", languages},
-               {"categories", categories}
+               {"categories", categories},
+               {"translations", translations}
              }
   );
 }
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -47,14 +47,17 @@ std::string HTMLDumper::dumpPlainHTML() const
     const auto langCode = bookObj.getCommaSeparatedLanguages();
     const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
     const auto tags = bookObj.getTags();
+    const auto downloadAvailable = (bookObj.getUrl() != "");
     std::string faviconAttr = "style=background-image:url(" + bookIconUrl + ")";
+    
     booksData.push_back(kainjow::mustache::object{
       {"id", contentId},
       {"title", bookTitle},
       {"description", bookDescription},
       {"langCode", langCode},
       {"faviconAttr", faviconAttr},
-      {"tagList", getTagList(tags)}
+      {"tagList", getTagList(tags)},
+      {"downloadAvailable", downloadAvailable}
     });
   }
 
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -33,10 +33,13 @@ kainjow::mustache::list getTagList(std::string tags)
 
 } // unnamed namespace
 
-std::string HTMLDumper::dumpPlainHTML() const
+std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
 {
   kainjow::mustache::list booksData;
-  for ( const auto& bookId : library->getBooksIds() ) {
+  const auto filteredBooks = library->filter(filter);
+  const auto searchQuery = filter.getQuery();
+
+  for ( const auto& bookId : filteredBooks ) {
     const auto bookObj = library->getBookById(bookId);
     const auto bookTitle = bookObj.getTitle();
     std::string contentId = "";
@@ -65,7 +68,8 @@ std::string HTMLDumper::dumpPlainHTML() const
              RESOURCE::templates::no_js_library_page_html,
              kainjow::mustache::object{
                {"root", rootLocation},
-               {"books", booksData }
+               {"books", booksData },
+               {"searchQuery", searchQuery}
              }
   );
 }
--- a/src/html_dumper.cpp
+++ b/src/html_dumper.cpp
@@ -2,6 +2,7 @@
 #include "libkiwix-resources.h"
 #include "tools/otherTools.h"
 #include "tools.h"
+#include "tools/regexTools.h"
 
 namespace kiwix
 {
@@ -17,6 +18,13 @@ HTMLDumper::~HTMLDumper()
 }
 
 namespace {
+
+std::string humanFriendlyTitle(std::string title)
+{
+  std::string humanFriendlyString = replaceRegex(title, "_", " ");
+  humanFriendlyString[0] = toupper(humanFriendlyString[0]);
+  return humanFriendlyString;
+}
 
 kainjow::mustache::list getTagList(std::string tags)
 {
@@ -38,6 +46,22 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
   kainjow::mustache::list booksData;
   const auto filteredBooks = library->filter(filter);
   const auto searchQuery = filter.getQuery();
+  auto languages = getLanguageData();
+  auto categories = getCategoryData();
+
+  for (auto &category : categories) {
+    const auto categoryName = category.get("name")->string_value();
+    if (categoryName == filter.getCategory()) {
+      category["selected"] = true;
+    }
+    category["hf_name"] = humanFriendlyTitle(categoryName);
+  }
+
+  for (auto &language : languages) {
+    if (language.get("lang_code")->string_value() == filter.getLang()) {
+      language["selected"] = true;
+    }
+  }
 
   for ( const auto& bookId : filteredBooks ) {
     const auto bookObj = library->getBookById(bookId);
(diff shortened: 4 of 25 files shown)


juuz0:
@kelson42 I haven't written tests yet (still reading through how to write them, should be done soon) but you can test the main functionality. Testing endpoint is `TLD/nojs`

kelson42:
@juuz0 Thank you very much for your PR. I have put @rgaudin as functional reviewer because he has requested the feature and has probably stronger opninion abuot how this shoudl be fixed. I have as well put @veloman-yunkan for the technical review (and functional as well :)

rgaudin:
OK, I've tested it and it works fine ; congrats. Couple of comments though:

- how is this going to be used? Right now, one has to go to `/nojs` manually to get this rendition. How will this be integrated for effective use? In a separate endpoint linked in a `<noscript />` on the main page?
- I understand the style was copied from the regular version but the magnification effect on a card's hovering only makes sense because it's clickable (and even then…) but in the nojs version it's not clickable and it thus just a distracting confusion.
- where should those link point to? Currently it points to the viewer… which is JS dependent

@kelson42 what's the merge strategy for this?

veloman-yunkan (commented):
I skimmed through the changes very quickly and I think that it is yet premature commenting on the code. At this point I will provide high level input:

1. Shouldn't the output be paginated?
2. What about translation? Once #846 is merged you can reuse the same strings for translating in the backend.

src/server/internalServer.cpp
```diff
@@ -744,6 +747,41 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
...
+  const auto domains = kiwix::split(url, "/", true, false);
+  auto filter = get_search_filter(request);
+  try {
+    if (request.get_argument("category") == "") {
+      filter.clearCategory();
+    }
+    if (request.get_argument("lang") == "") {
+      filter.clearLang();
+    }
+  } catch (...) {}
```
mgautierfr, line 763: I'm surprised about that. In which context `get_search_filter` will set a filter on the category and we need to clear it ?
Shouldn't be better to fix `get_search_filter` if it is buggy ?

src/server/internalServer.cpp
```diff
@@ -744,6 +747,41 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
...
+  std::string content = opdsDumper.noJSData(filter);
+
+  if (domains.size() == 1) {
+    content = opdsDumper.noJSData(filter);
+  } else if ((domains.size() == 3) && (domains[1] == "download") && (mp_nameMapper->getIdForName(domains[2]) != "")) {
+    content = opdsDumper.noJSDownloadLinks(mp_nameMapper->getIdForName(domains[2]));
+  } else {
+    return HTTP404Response(*this, request)
+           + urlNotFoundMsg;
+  }
```
mgautierfr, line 776: This set the content to the result of `opdsDumper.noJSData(filter)` all the time and then overwrite it.
So the first initialization is useless. And we should create the filter only if `domain.size() == 1`, not all the time.

juuz0:
> how is this going to be used? Right now, one has to go to /nojs manually to get this rendition. How will this be integrated for effective use? In a separate endpoint linked in a <noscript /> on the main page?

Hiding nav, body, footer when javascript is not enabled & displaying a text "This page can only be accessed with Javascript enabled, please head over to `<a href="{{root}}/nojs">/nojs endpoint</a>` should be fine I think?

> I understand the style was copied from the regular version but the magnification effect on a card's hovering only makes sense because it's clickable (and even then…) but in the nojs version it's not clickable and it thus just a distracting confusion.

It is clickable though? I had to reduce clickable area (currently it's only the description part) because HTML spec doesn't allow `<a>` nesting (another `<a>` is used for download button)

> where should those link point to? Currently it points to the viewer… which is JS dependent

Will go by mgautierfr's suggestion, thanks! 

> Shouldn't the output be paginated?

normal welcome page (JS one) isn't paginated so why this? Performance reasons?

> What about translation? Once Internationalization of kiwix-serve's frontend #846 is merged you can reuse the same strings for translating in the backend.

Noted, should be done soon - once I understand how to write tests which is taking some effort :)

src/server/internalServer.cpp
juuz0: My reasoning for not touching `get_search_filter` was if the behaviour was intentional to display no books when lang/category == "" and if I change that logic, I would break some other functionality.
In kiwix-serve welcome page filters, "" means all categories while `get_search_filter` seems to consider "" as a valid category and rightly shows no book associated

rgaudin:
> Hiding nav, body, footer when javascript is not enabled & displaying a text "This page can only be accessed with Javascript enabled, please head over to `<a href="{{root}}/nojs">/nojs endpoint</a>` should be fine I think?

Fine by me. What do others think about?

> It is clickable though? I had to reduce clickable area (currently it's only the description part) because HTML spec doesn't allow `<a>` nesting (another `<a>` is used for download button)

No it's not clickable. That's the point, it makes user think it is because it grows

> Will go by mgautierfr's suggestion, thanks!

Way to go. This endpoint also handles the mainPath so that should be easy… although this URL should normally be retrieved from OPDS I suppose…

> normal welcome page (JS one) isn't paginated so why this? Performance reasons?

It is. Just look at https://library.kiwix.org/ there's an automatic loading of new cards on scroll. I'm not in favor or paginating for a first version. It's unlikely this will be really useful in practice and adds some complexity. Let's wait for user requests for this.

juuz0:
> how is this going to be used?......

Done as discussed.

> I understand the style was copied from the regular....

Removed hover effect

> where should those link point to? Currently it points to the viewer....

Done.

> What about translation? .....

Still pending, I'm so scared to rebase this PR on top of main currently. Should be done soon.

Added two tests

juuz0:
dunno why CI is failing :(

juuz0:
> What about translation? Once 846 is merged you can reuse the same strings for translating in the backend.

Done :)

juuz0:
@rgaudin ping because there's no button to re request a review! >:)

juuz0:
For translations, use `userlang` as url param.
Example
`http://192.168.18.8:8080/nojs?lang=eng&userlang=test`

rgaudin (changes requested):
Just tested 8c10936
- Home points to `/nojs` as expected 👍 
- Links points to `/content` 👍 
- i18n translation works fine 👍
- I still get the hover effect on the Tag: text is larger, bold and the pointer changes as well. ❌

juuz0:
> I still get the hover effect on the Tag: text is larger, bold and the pointer changes as well. x

Done now, thanks for noticing.

Additionally, I had forgotten to display "No results" when there were 0 books. That's done too. Along with a new test

src/server/internalServer.cpp
juuz0: I had assumed `category==""` was for all categories filter
Is there a specific category for "all categories" then?

rgaudin:
> > I still get the hover effect on the Tag: text is larger, bold and the pointer changes as well. x
> 
> Done now, thanks for noticing.

👍 OK now

src/opds_dumper.cpp
```diff
@@ -282,12 +282,14 @@ std::string OPDSDumper::noJSData() const
   for ( const auto& bookId : library->getBooksIds() ) {
     const auto bookObj = library->getBookById(bookId);
     const auto bookTitle = bookObj.getTitle();
+    const auto contentId = bookObj.getHumanReadableIdFromPath();
```
veloman-yunkan, line 368: Obtain the user-friendly content id via `nameMapper`

static/templates/no_js_download.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Download links</title>
+</head>
+<body>
```
veloman-yunkan, line 9: Shouldn't this page contain the name of the book?

src/opds_dumper.cpp
```diff
@@ -276,6 +276,17 @@ kainjow::mustache::list getTagList(std::string tags)
...
+std::string OPDSDumper::noJSDownloadLinks(const std::string& bookId) const
+{
+  const auto book = library->getBookById(bookId);
+  return render_template(
+             RESOURCE::templates::no_js_download_html,
+             kainjow::mustache::object{
+               {"url", book.getUrl() }
+             }
+  );
+}
```
veloman-yunkan, line 342: Maybe this method should belong to `InternalServer`?

src/opds_dumper.cpp
```diff
@@ -287,10 +287,13 @@ std::string OPDSDumper::noJSDownloadLinks(const std::string& bookId) const
   );
 }
 
-std::string OPDSDumper::noJSData() const
+std::string OPDSDumper::noJSData(kiwix::Filter filter) const
```
veloman-yunkan, line 344: The API of dumping in plain HTML format diverges from the API of dumping in OPDS format. Why not stick to the signature of the `dumpOPDSFeed2()` and obtain the book ids in `InternalServer` (probably using `InternalServer::search_catalog()`)?

static/templates/no_js.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1" />
     <link type="root" href="{{root}}">
-    <title>Welcome to Kiwix Server</title>
+    <title>{{welcome-to-kiwix-server}}</title>
     <link
       type="text/css"
       href="{{root}}/skin/index.css?KIWIXCACHEID"
```
veloman-yunkan, line 10: Passing translation strings as top-level mustache variables feels somewhat fragile to me. Can't all translations required for this template be passed as one `translations` object (so that they are accessed as `{{translations.welcome-to-kiwix-server}}`? I am not sure that it will work but I believe that it should.

test/library_server.cpp
```diff
@@ -941,6 +941,221 @@ TEST_F(LibraryServerTest, no_name_mapper_catalog_v2_individual_entry_access)
...
+  "                    </a>\n" \
+  "              </div>\n" \
+  "            <div class=\"book__languageTag\" >eng</div>\n" \
+  "            <div class=\"book__tags\"><div class=\"book__tags--wrapper\">\n" \
+  "                <span class=\"tag__link\" aria-label='public_tag_without_a_value' title='public_tag_without_a_value'>public_tag_without_a_value</span>\n" \
+  "                <span class=\"tag__link\" aria-label='wikipedia' title='wikipedia'>wikipedia</span>\n" \
+  "            </div>\n" \
+  "        </div>\n" \
+  "        </div>\n"
+#define RAY_CHARLES_UNCTZ \
```
veloman-yunkan, line 1044: Add `_BOOK_HTML` for consistency

veloman-yunkan (changes requested):
This is a first pass. There are some comments that imply a discussion and a final decision together with @mgautierfr before the proposed changes are made.

src/opds_dumper.cpp
```diff
@@ -262,6 +262,25 @@ std::string OPDSDumper::categoriesOPDSFeed() const
   );
 }
 
+std::string OPDSDumper::noJSData() const
```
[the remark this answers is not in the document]
mgautierfr: This is a interesting point. We could indeed have a `LibraryDumper` (may be only a base class if implementation need it) and have the opds, plain html and libxml dumping using it.
All three dumping are a dumping of the library in different formats. If we can do it only with templates it would be nice, if we cannot ... we cannot. But it would be nice to have the common parts (to be defined) put in common code.

At least, the renaming proposed by @veloman-yunkan seems good.

src/opds_dumper.cpp
juuz0: `InternalServer::search_catalog()` doesnt allow me to change the filter :<< I need to clear the category

static/templates/no_js.html
juuz0: Done. Thanks

test/library_server.cpp
```diff
@@ -941,6 +941,221 @@ TEST_F(LibraryServerTest, no_name_mapper_catalog_v2_individual_entry_access)
...
+  "    <link rel=\"manifest\" href=\"/skin/favicon/site.webmanifest?cacheid=bc396efb\">\n" \
+  "    <link rel=\"mask-icon\" href=\"/skin/favicon/safari-pinned-tab.svg?cacheid=8d487e95\" color=\"#5bbad5\">\n" \
+  "    <link rel=\"shortcut icon\" href=\"/skin/favicon/favicon.ico?cacheid=92663314\">\n" \
+  "    <meta name=\"msapplication-TileColor\" content=\"#da532c\">\n" \
+  "    <meta name=\"msapplication-config\" content=\"/skin/favicon/browserconfig.xml?cacheid=f29a7c4a\">\n" \
+  "    <meta name=\"theme-color\" content=\"#ffffff\">\n" \
+  "    <style>\n" \
+  "      @font-face {\n" \
+  "        font-family: \"poppins\";\n" \
+  "        src: url(\"/skin/fonts/Poppins.ttf?cacheid=af705837\") format(\"truetype\");\n" \
```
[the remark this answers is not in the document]
juuz0: Fixed the bug

src/server/internalServer.cpp
mgautierfr: I'm not sure about what we are speaking about.
But if we are speaking about the request send to the http api, if we don't want to filter on the category, we must not pass a `category==""` query string.

src/opds_dumper.cpp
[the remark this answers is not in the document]
mgautierfr: This seems good to me.

(comment based only on the new organisation of  `LibraryDumper`, `OPDSDumper` & `HTMLDumper`. Some other comments may still apply on the code itself)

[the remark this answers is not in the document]
veloman-yunkan: It seems good to me, too.

Can you please rebase and update the history of this PR by extracting `LibraryDumper` out of `OPDSDumper` as the first step and then adding `HTMLDumper`?

src/opds_dumper.cpp
```diff
@@ -341,34 +341,33 @@ std::string OPDSDumper::noJSDownloadLinks(const std::string& bookId) const
   );
 }
 
-std::string OPDSDumper::noJSData(kiwix::Filter filter) const
+std::string OPDSDumper::dumpPlainHTML(const std::vector<std::string>& bookIds, const std::string& query,
+                                const std::string& category, const std::string& language) const
```
veloman-yunkan, line 349: With `HTMLDumper` now being separate from `OPDSDumper` the difference in the signatures of the dump functions doesn't matter that much. The version taking a single `Filter` parameter was better.

veloman-yunkan (changes requested):
I would like to review a clean version of this PR that starts with `LibraryDumper` being extracted out of `OPDSDumper` before `HTMLDumper` is introduced in the second commit.

src/opds_dumper.cpp
juuz0: Done.

src/opds_dumper.cpp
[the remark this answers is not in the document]
juuz0: Done :>

juuz0:
I haven't added the long copyright text in `library_dumper.*` or `html_dumper.*`, should i just copy paste that text from `opds_dumper`?

src/server/internalServer.cpp
juuz0: There needs to be a way for the server to know that the user requests "ALL CATEGORIES", I dont think a query param can be cleared without javascript on the client side hence I consider `category=""` as ALL categories & clear that on the server side.

veloman-yunkan (changes requested):

src/server/internalServer.cpp
mgautierfr: > There needs to be a way for the server to know that the user requests "ALL CATEGORIES"

The absence of filter on category seems to me a good way to specify all categories.

> I dont think a query param can be cleared without javascript on the client side hence I consider category="" as ALL categories & clear that on the server side.

Why we need to clear it ?
The html code is generated server side, so we can generate the request we want, without a `category` parameter.

veloman-yunkan: > > I dont think a query param can be cleared without javascript on the client side hence I consider category="" as ALL categories & clear that on the server side.
> 
> Why we need to clear it ? The html code is generated server side, so we can generate the request we want, without a `category` parameter.

On the server side we generate the HTML for the search form. The form is filled and submitted from the client side. Values are sent for all inputs of the form (unless it is a request generated via JS).

mgautierfr: Ha yes, I've forget about the search form.
I would say that `get_search_filter` should be the only (and correct) way to create a filter from query string.
If we add a modification on the filter in specific api, it will introduce inconsistency in the API and we can be sure that we will mix things later. 

> while get_search_filter seems to consider "" as a valid category and rightly shows no book associated.

I finally think it is wrong. The only valid use case of searching for no category is to search for book which have no category associated. So it is more something about `undefined` category than <code></code> category.

Maybe we could extend a bit the api to accept to specific values `[all]` and `[empty]` (or `[none]`) for no filtering (all) and filtering for book without category (empty/none).

The `[]` would prevent a potential conflict with a real category `all` or `none`. 
What do you think about that ?

juuz0: Extending the API would make life easier definitely & bring more consistency.
Would've been nice if `all`/`none` be reserved words for filters but I don't really mind the `[all]` or `[empty/none]` (`none` is better word imo).
Only consideration would be if this looks pleasant on the eye in URL: `TLD/?lang=[all]&category=[none]`

include/library_dumper.h
```diff
@@ -0,0 +1,91 @@
...
+{
+
+/**
+ * A base class to dump Library in various formats.
+ *
+ */
+class LibraryDumper
+{
+ public:
+  LibraryDumper() = default;
```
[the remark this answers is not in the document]
juuz0: the default constructor for OPDSDumper comes from https://github.com/kiwix/libkiwix/commit/b48428e4436a66dd28a9c7279eb3e6efbc681348 
Though, the setLibrary function has been removed now and the default constructor is not used anywhere now. Safe to remove I guess?

src/library_dumper.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+
+std::string getLanguageSelfName(const std::string& lang) {
+  const auto itr = iso639_3.find(lang);
+  if (itr != iso639_3.end()) {
+    return itr->second;
+  }
+  return lang;
+};
```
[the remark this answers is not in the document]
juuz0: Out of curiosity, what does this help in?

src/server/internalServer.cpp
```diff
@@ -53,6 +53,7 @@ extern "C" {
 #include "name_mapper.h"
 #include "search_renderer.h"
 #include "opds_dumper.h"
+#include "library_dumper.h"
```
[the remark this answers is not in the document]
juuz0: Not needed, got mixed while i was changing commit history :>

include/library_dumper.h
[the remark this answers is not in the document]
veloman-yunkan: Other places to look for are projects depending on libkiwix. I think we should remove the default constructor of `OPDSDumper` in any case, but at least we must know in advance how many things will break and will need to be fixed.

src/library_dumper.cpp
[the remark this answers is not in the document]
veloman-yunkan: It prevents name conflicts (symbol redefinition errors) at link time - this is especially important for libraries since they should ideally only export symbols representing their public API.

include/html_dumper.h
```diff
@@ -23,6 +23,13 @@ class HTMLDumper : public LibraryDumper
   HTMLDumper(Library* library, NameMapper* NameMapper);
   ~HTMLDumper();
 
+
+  /**
+   * Dump library in HTML
+   *
+   * @return HTML content
+   */
+  std::string dumpPlainHTML() const;
```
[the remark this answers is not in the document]
juuz0: Does this mean to combine the 2 commits (introducing HTMLDumper and /nojs)?
dumpPlainHTML is sort of dependent on having no_js file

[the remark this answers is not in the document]
veloman-yunkan: > Does this mean to combine the 2 commits (introducing HTMLDumper and /nojs)?

No. Only include in the commit introducing `HTMLDumper` what is needed by (the initial version of) its `dumpPlainHTML()` method. The name of the new template `no_js.html` doesn't make it an unseparable part of the change that introduces the `/nojs` endpoint (just suppose that the template is named `nojs_library_page.html`; BTW, isn't that a better filename?)

static/skin/i18n/en.json
```diff
@@ -28,7 +28,7 @@
 	, "random-page-button-text": "Go to a randomly selected page"
 	, "searchbox-tooltip": "Search '{{BOOK_TITLE}}'"
 	, "confusion-of-tongues": "Two or more books in different languages would participate in search, which may lead to confusing results."
-	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"?lang=\">reset filter</a>?"
+	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"?lang=&category=&q=\">reset filter</a>?"
```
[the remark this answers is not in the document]
juuz0: It already handles the absence, I thought it was more consistent to give all params that time - I don't think so anymore. Will reset it.
Atlhough, with #907 changing the `?` to `#`, I think we need a overzealous-filter string for `?`, correct?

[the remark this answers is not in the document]
veloman-yunkan: > with #907 changing the `?` to `#`, I think we need a overzealous-filter string for `?`, correct?

Sorry, but I don't understand your question/proposal.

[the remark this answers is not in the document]
juuz0: https://github.com/kiwix/libkiwix/pull/907/files#diff-c70f0ae73979b109350424912c766a72199a66d69ecfa14c1afbe54dc9672ceeR31

Once this change is merged, we'll be using fragment params in js-enabled welcome page while query params in no-js page. My question was if I should be adding a new `welcome-page-overzealous-filter-query` entry with `<a href=\"?lang=\">` as link instead of `<a href=\"#lang=\">`

src/server/internalServer.cpp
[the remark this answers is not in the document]
juuz0: This change doesn't have a fixup associated, but fixed by editing commit

include/html_dumper.h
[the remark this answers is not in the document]
juuz0: This change doesn't have a fixup associated, but fixed by editing commit

static/skin/i18n/en.json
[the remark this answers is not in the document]
juuz0: Very nice :D Thanks
Added in https://github.com/kiwix/libkiwix/pull/897/commits/eac892e52330b107948082ec31f3da8babdf05d8

P.S. #907 should be merged before this PR

include/library_dumper.h
```diff
@@ -77,8 +94,8 @@ class LibraryDumper
 
 
  protected:
-  kiwix::Library* library;
-  kiwix::NameMapper* nameMapper;
+  kiwix::Library* const library;
+  kiwix::NameMapper* const nameMapper;
```
veloman-yunkan: `const` was added in the wrong place. I intended `const Library* library;` and `const NameMapper* nameMapper;` (we don't want `LibraryDumper` to be able to call non-const operations on library and/or name mapper objects). Though the immutability of those data-members doesn't hurt either. So let's end up with

```
const kiwix::Library* const library;
const kiwix::NameMapper* const nameMapper;
```

src/server/internalServer.cpp
```diff
@@ -759,10 +759,22 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
   return ContentResponse::build(*this, RESOURCE::templates::viewer_settings_js, data, "application/javascript; charset=utf-8");
 }
 
+std::string InternalServer::dumpDownloadLinks(const std::string& bookId) const
```
veloman-yunkan, line 764: In the content of `InternalServer` the verb `dump` sounds a little strange. `getNoJSDownloadPageHTML()` is a better name.

src/server/i18n.h
```diff
@@ -65,6 +65,17 @@ class GetTranslatedString
...
+  {
+    return {key, getTranslatedString(m_lang, key)};
+  }
+
+  std::pair<std::string, kainjow::mustache::basic_data<std::string>> getTranslatedPair(std::string key, kainjow::mustache::object params)
+  {
+    return {key, expandParameterizedString(m_lang, key, params)};
+  }
+
+
```
veloman-yunkan, line 78: 1. Make the signatures consistent with those of preexisting functions.

2. It's better to introduce a separate class `GetTranslatedStringWithMsgId` with a pair of overloaded `operator()`s instead of those functions.

src/server/internalServer.cpp
```diff
@@ -759,14 +759,25 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
   return ContentResponse::build(*this, RESOURCE::templates::viewer_settings_js, data, "application/javascript; charset=utf-8");
 }
 
+std::string userLang = "eng";
```
veloman-yunkan, line 762: A mutable global variable is an issue for concurrent access.

veloman-yunkan (changes requested):
Please include testing of the `/nojs` endpoint in the `ServerTest.CacheControlOfDynamicContent` unit-test.

Also, please rebase&fix-up for the next iteration.

include/library_dumper.h
juuz0: Apologies :)
I always mess up between `pointer to const` and `const pointer` :P

src/server/internalServer.cpp
juuz0: Made it a member of internalServer, i've a feeling there are better ways but if this works then nice!

veloman-yunkan: The more general problem for concurrent access is a shared mutable variable. A global variable merely has more chances of becoming one. Your "fix" doesn't prevent m_userLang from being changed from two or more different threads during concurrent execution of multiple `/nosj` requests. However, it is not clear why you needed to make `userLang` a data member of `InternalServer` at all. A local variable should be just fine.

Since I haven't yet looked at your other changes, please correct this in place (i.e. without introducing a fixup commit).

juuz0: Without having some sort of "shared" access to m_userLang between HTMLDumper & InternalServer - the `/nojs/download` page wouldn't know what userlang to display. Hence the requirement of a member `m_userLang`.

Consider this:

`/nojs?userlang=test`
Click a book's download button & you're redirected to `/nojs/download/xyz` - if we're using a local variable, it would ask for `userlang` argument again - not get it & set the language to default English.


Another solution is probably setting all download links to have userlang as query. Wouldn't require any member in internalServer then, is this better?

veloman-yunkan: You must not store any state *in the server* between requests (in this case, with proper timing `m_userLang` set by one user's access to `/nojs` may affect the language used for serving `/nojs/download/xyz` to another user). 

Note that the `userlang` URL query parameter is only one way of specifying the user language. The user-language (no matter where it comes from) is stored in a cookie:
https://github.com/kiwix/libkiwix/blob/88de978a9c52aa8e64dad883e949be1e9250dfa7/src/server/request_context.cpp#L210-L228
https://github.com/kiwix/libkiwix/blob/88de978a9c52aa8e64dad883e949be1e9250dfa7/src/server/response.cpp#L390-L395

juuz0: Done

src/html_dumper.cpp
```diff
@@ -74,15 +75,27 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
     });
   }
 
+  auto i18nTranslations = i18n::GetTranslatedStringWithMsgId(m_userLang);
```
veloman-yunkan, line 78: `i18nTranslations` -> `getTranslation`

src/server/internalServer.cpp
```diff
@@ -759,14 +759,23 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
   return ContentResponse::build(*this, RESOURCE::templates::viewer_settings_js, data, "application/javascript; charset=utf-8");
 }
 
-std::string InternalServer::getNoJSDownloadPageHTML(const std::string& bookId) const
+std::string InternalServer::getNoJSDownloadPageHTML(const std::string& bookId, const std::string& userLang) const
 {
   const auto book = mp_library->getBookById(bookId);
+  auto i18nTranslations = i18n::GetTranslatedStringWithMsgId(userLang);
```
veloman-yunkan, line 765: `i18nTranslations` -> `getTranslation`

veloman-yunkan (changes requested):
Good. There are a few things to fix to keep the code tidy but else we are almost done.

include/opds_dumper.h
```diff
@@ -38,10 +39,9 @@ namespace kiwix
  * A tool to dump a `Library` into a opds stream.
  *
  */
-class OPDSDumper
+class OPDSDumper : public LibraryDumper
 {
  public:
-  OPDSDumper() = default;
   OPDSDumper(Library* library, NameMapper* NameMapper);
```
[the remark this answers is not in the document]
juuz0: I haven't heard that term before, does this mean using `const_cast` to remove constness?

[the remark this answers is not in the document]
veloman-yunkan: `OPDSDumper` is now more demanding than it has to be - if one has a `const Library` object and/or a `const NameMapper` object they cannot create an `OPDSDumper` (without resorting to a `const_cast`, which is bad manners). There is no conceptual or technical reason why the current signature of the `OPDSDumper` constructor (unlike `HTMLDumper`) prohibits such an invocation. That's why I requested to *relax* the too strict requirements that exits merely for historical reasons.

src/html_dumper.cpp
```diff
@@ -35,12 +35,17 @@ std::string HTMLDumper::dumpPlainHTML() const
...
+    try {
+      contentId = nameMapper->getNameForId(bookId);
+    } catch (...) {}
     const auto bookDescription = bookObj.getDescription();
     const auto langCode = bookObj.getCommaSeparatedLanguages();
     const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
     const auto tags = bookObj.getTags();
     std::string faviconAttr = "style=background-image:url(" + bookIconUrl + ")";
     booksData.push_back(kainjow::mustache::object{
+      {"id", contentId},
```
veloman-yunkan, line 68: `contentId` must be URI-encoded

veloman-yunkan (changes requested):

static/templates/no_js_library_page.html
```diff
@@ -0,0 +1,140 @@
...
+
+      .kiwixHomeBody__results {
+        flex-basis: 100%;
+      }
+
+      #book__title>a, .book__download a {
+        text-decoration: none;
+        all: unset;
+      }
+    </style>
```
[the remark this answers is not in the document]
juuz0: `index.css` is reused in no_js_library_page.html.
These changes in <style> are specific to no-js

include/library_dumper.h
```diff
@@ -0,0 +1,102 @@
...
+   * Get the data of languages
+   */
+  kainjow::mustache::list getLanguageData() const;
+
+  /**
+   * Get the human friendly title
+   * 
+   * @param title 
+   */
+  std::string humanFriendlyTitle(std::string title) const;
```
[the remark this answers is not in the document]
juuz0: > Also does this helper function need to be a member of LibraryDumper? Can't it be a free function in an unnamed namespace in html_dumper.cpp?

I thought it would come to use as  a shared function but eh I guess not now, changed as per suggestion.

static/skin/i18n/en.json
```diff
@@ -46,4 +46,6 @@
 	, "library-opds-feed": "Library OPDS Feed"
 	, "filter-by-tag": "Filter by tag \"{{TAG}}\""
 	, "stop-filtering-by-tag": "Stop filtering by tag \"{{TAG}}\""
+	, "welcome-to-kiwix-server": "Welcome to Kiwix Server"
+	, "download-links-title": "Download links for {{BOOK_TITLE}}"
```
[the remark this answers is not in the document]
juuz0: This title goes into the tab, dont think styling works there/is needed
image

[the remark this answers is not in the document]
veloman-yunkan: Well, this is the first time I ran the new code. I was under the wrong impression that this is shown as the heading of the page which I believe is the right thing to do. The title of the tab is not a good place (as the only place) for such information.

Screenshot from 2023-03-20 17-40-44

veloman-yunkan (changes requested):

static/skin/i18n/en.json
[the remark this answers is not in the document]
juuz0: Done.

test/server.cpp
```diff
@@ -1522,6 +1718,38 @@ TEST_F(ServerTest, CacheControlOfDynamicContent)
...
+
+  // no_js_no_books
+  r = zfs1_->GET("/ROOT%23%3F/nojs?lang=fas");
+  EXPECT_EQ(r->status, 200);
+  EXPECT_EQ(r->body,
+            HTML_PREAMBLE
+            FILTERS_HTML("")
+            HOME_BODY_0_RESULTS
+            FINAL_HTML_TEXT);
 }
```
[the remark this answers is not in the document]
juuz0: I created a new file `test/html_dumper.cpp` for nojs tests (though, not too sure if that name is ok)

> In https://github.com/kiwix/libkiwix/pull/897#pullrequestreview-1334738349 I meant that the ServerTest.CacheControlOfDynamicContent unit-test must be enhanced to cover this new /nojs endpoint too.

I don't know how to do this, but added /nojs in `resources200Compressible`. If that's what was needed, cool! otherwise some guidance please! :>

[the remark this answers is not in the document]
veloman-yunkan: > I created a new file `test/html_dumper.cpp` for nojs tests (though, not too sure if that name is ok)

No, that's an overshoot. A new test-case in `test/server.cpp` is fine (but it must be placed reasonably).

> 
> > In [#897 (review)](https://github.com/kiwix/libkiwix/pull/897#pullrequestreview-1334738349) I meant that the ServerTest.CacheControlOfDynamicContent unit-test must be enhanced to cover this new /nojs endpoint too.
> 
> I don't know how to do this, but added /nojs in `resources200Compressible`. If that's what was needed, cool!

Cool! I wonder if ChatGPT would be able to do that (I hope that not).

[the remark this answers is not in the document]
juuz0: > No, that's an overshoot. A new test-case in test/server.cpp is fine (but it must be placed reasonably).

Okie removed and added them in a new test :> I hope this back and forth stuff hasn't annoyed you (yet!) :P

> Cool! I wonder if ChatGPT would be able to do that (I hope that not).

Hoping the same :P
But also waiting for the day when I tell GPT4 to do welcome page UI from a [paper drawing like this](https://twitter.com/heykahn/status/1635752848398102530)

juuz0:
for the wrong zim, download page was giving 500 instead of 400 - fixed in https://github.com/kiwix/libkiwix/pull/897/commits/de29f7c7f3a1a72bd9f9c34d1752320e09da736a

> at least in ServerTest.noJS and ServerTest.CacheControlOfDynamicContent, maybe also a negative test in ServerTest.404

Done.

test/server.cpp
```diff
@@ -2008,3 +2014,273 @@ R"(const viewerSettings = {
...
+  "        <div>Direct</div>\n" \
+  "    </a>\n" \
+  "    <a href=\".sha256\" download>\n" \
+  "        <div>Sha256 hash</div>\n" \
+  "    </a>\n" \
+  "    <a href=\".magnet\" target=\"_blank\">\n" \
+  "        <div>Magnet link</div>\n" \
+  "    </a>\n" \
+  "    <a href=\".torrent\" download>\n" \
+  "        <div></div>\n" \
```
veloman-yunkan, line 2244: The bug made its way into the unit-tests :)

Another thing that bothers me is the values of the `href` attributes in this HTML. Can we make the test more realistic? Probably it can be done easily by moving the noJS test into `test/library_server.cpp`.

veloman-yunkan (changes requested):
I hope that this was the last complete review of this PR. Please address the comments using fixup commits.

test/server.cpp
juuz0: Ye done! Thanks

test/server.cpp
```diff
@@ -284,7 +288,7 @@ R"EXPECTEDRESULT(      href="/ROOT%23%3F/skin/index.css?cacheid=4974c58b"
     <script type="text/javascript" src="/ROOT%23%3F/skin/languages.js?cacheid=b00b12db" defer></script>
     <script src="/ROOT%23%3F/skin/isotope.pkgd.min.js?cacheid=2e48d392" defer></script>
     <script src="/ROOT%23%3F/skin/iso6391To3.js?cacheid=ecde2bb3"></script>
-    <script type="text/javascript" src="/ROOT%23%3F/skin/index.js?cacheid=8e06e864" defer></script>
+    <script type="text/javascript" src="/ROOT%23%3F/skin/index.js?cacheid=c0c69750" defer></script>
```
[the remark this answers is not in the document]
juuz0: Only this change is done through a rebase

static/templates/no_js_library_page.html
```diff
@@ -51,7 +51,7 @@
         <div class="book__list">
         {{#books}}
         <div class="book__wrapper">
-            <a class="book__link" href="#" data-hover="Preview">
+            <a class="book__link" href="{{root}}/content/{{id}}" data-hover="Preview">
```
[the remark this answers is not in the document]
juuz0: Replaced this with title & aria-label.

> Is "Preview" a good hint text?

Good enough to me to be honest. `Preview book` or `Open book` may also fit.

For the full JS version, I think its better to do in a new ticket? Or should I do that here too

veloman-yunkan:
Please, also update `static/skin/i18n/qqq.json` (i.e. document the new messages there)

src/server/internalServer.cpp
```diff
@@ -759,19 +759,58 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
...
+std::string InternalServer::getNoJSDownloadPageHTML(const std::string& bookId) const
+{
+  const auto book = mp_library->getBookById(bookId);
+  auto bookUrl = book.getUrl();
+  if (bookUrl.size() > 5) {
+    const auto subUrl = bookUrl.substr(bookUrl.size() - 6, bookUrl.size() - 1);
+    if (subUrl == ".meta4") {
+      bookUrl = bookUrl.substr(0, bookUrl.size() - 6);
+    }
+  }
```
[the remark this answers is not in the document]
juuz0: added in stringTools with a new test for it

veloman-yunkan (approved):

PR #901, opened 2023-02-23

translatewiki:
Translation updates


veloman-yunkan:
@kelson42 @mgautierfr I had to modify this PR. As some of the new translations don't contain the self-name of the language (defined via the "name" attribute in the JSON file) which is required by the language selector UI, I exclude such translations from that list. Is it OK? 

While writing this comment I also had the idea that maybe only languages with enough translations (e.g. above 70% of strings available in en.json) should be included in the language selector UI. What do you think?

kelson42:
> While writing this comment I also had the idea that maybe only languages with enough translations (e.g. above 70% of strings available in en.json) should be included in the language selector UI. What do you think?

This is something definitly configuration at Translatewiki, but not sure if this is doable at subproject level.

veloman-yunkan:
> > While writing this comment I also had the idea that maybe only languages with enough translations (e.g. above 70% of strings available in en.json) should be included in the language selector UI. What do you think?
> 
> This is something definitly configuration at Translatewiki, but not sure if this is doable at subproject level. Reading https://translatewiki.net/wiki/Translating:Kiwix, it seems currently the threshold is 25%. Ihave no strong opinion for the moment regarding the idea to raise this value.

@kelson42 I meant to do it in our `generate_i18n_resources_list.py` script

veloman-yunkan:
@mgautierfr @kelson42 I would like to merge this PR before it is updated again by Translatewiki (which will erase my changes).

static/generate_i18n_resources_list.py
```diff
@@ -29,7 +29,8 @@ def get_translation_info(filepath):
     lang_code = Path(filepath).stem
     with open(filepath, 'r', encoding="utf-8") as f:
         content = json.load(f)
-        return lang_code, content["name"]
+        lang_name = content["name"] if "name" in content else None
```
mgautierfr, line 32: `lang_name = content.get("name")`

static/generate_i18n_resources_list.py
```diff
@@ -39,7 +40,11 @@ def get_translation_info(filepath):
             continue
         print("Processing", i18n_file.name)
         if i18n_file.name != "test.json":
-            language_list.append(get_translation_info(i18n_file))
+            lang_code, lang_name = get_translation_info(i18n_file)
+            if lang_name is not None:
```
mgautierfr, line 44: An empty string `lang_name` is also a invalid value.
Better using
```python
if lang_name:
    language_list.append(...)
....
```

mgautierfr (changes requested):
Just small changes on the commented lines. But else we are good (you can merge yourself once you have fixed them to avoid latency and take the risk of translatewiki messing up)

static/generate_i18n_resources_list.py
veloman-yunkan: Fixed

static/generate_i18n_resources_list.py
veloman-yunkan: Fixed

PR #904, opened 2023-02-27
Support for multilang ZIMs

### LINKED ISSUE #903: ZIM with multiple language metadata are not properly supported
At least not supported in kiwix-serve filter

Screenshot 2023-02-22 at 09 50 22

First reported at https://github.com/kiwix/libkiwix/issues/872#issuecomment-1439722930

veloman-yunkan:
Fixes #903

Note that a multilanguage ZIM/book is counted as 1 full book in the results to each of the matching `/catalog/v2/entries?lang=<LANG>` queries.

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -455,12 +455,14 @@ void Library::updateBookDB(const Book& book)
 {
   Xapian::Stem stemmer;
   Xapian::TermGenerator indexer;
-  const std::string lang = book.getLanguage();
-  try {
-    stemmer = Xapian::Stem(iso639_3ToXapian(lang));
-    indexer.set_stemmer(stemmer);
-    indexer.set_stemming_strategy(Xapian::TermGenerator::STEM_SOME);
-  } catch (...) {}
+  const auto langs = book.getLanguages();
+  if ( langs.size() == 1 ) {
+    try {
+      stemmer = Xapian::Stem(iso639_3ToXapian(langs[0]));
+      indexer.set_stemmer(stemmer);
+      indexer.set_stemming_strategy(Xapian::TermGenerator::STEM_SOME);
+    } catch (...) {}
+  }
   Xapian::Document doc;
   indexer.set_document(doc);
 
@@ -475,7 +477,9 @@ void Library::updateBookDB(const Book& book)
   // Index all fields for field-based search
   indexer.index_text(title, 1, "S");
   indexer.index_text(desc,  1, "XD");
-  indexer.index_text(lang,  1, "L");
+  for ( const auto& lang : langs ) {
+    indexer.index_text(lang,  1, "L");
+  }
   indexer.index_text(normalizeText(book.getCreator()),   1, "A");
   indexer.index_text(normalizeText(book.getPublisher()), 1, "XP");
   indexer.index_text(normalizeText(book.getName()),      1, "XN");
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -373,12 +373,28 @@ std::vector<std::string> Library::getBookPropValueSet(BookStrPropMemFn p) const
 
 std::vector<std::string> Library::getBooksLanguages() const
 {
-  return getBookPropValueSet(&Book::getLanguage);
+  std::vector<std::string> langs;
+  for ( const auto& langAndCount : getBooksLanguagesWithCounts() ) {
+    langs.push_back(langAndCount.first);
+  }
+  return langs;
 }
 
 Library::AttributeCounts Library::getBooksLanguagesWithCounts() const
 {
-  return getBookAttributeCounts(&Book::getLanguage);
+  std::lock_guard<std::mutex> lock(m_mutex);
+  AttributeCounts langsWithCounts;
+
+  for (const auto& pair: mp_impl->m_books) {
+    const auto& book = pair.second;
+    if (book.getOrigId().empty()) {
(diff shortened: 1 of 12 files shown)


mgautierfr:
While this PR indeed fix #903, it seems a bit too simple to fully support multilang ZIMs. (And I would be pleasantly surprised if it is enough)

At least, in `Library::updateBookDB`, we use `book.getLanguage()` to set the stemmer. With a multilang book, it will not work.
The same way, we don't split the comma separated list when indexing the book itself. It seems to work (based on the tests you have added), propably because we find the queried language in the book's language list but it would be better to explicitly split it.

And I think we should change `std::string Book::getLanguage()` to `std::vector<std::string> Book::getLanguages()`. This way, we would force all our users to handle correctly multi language book.

veloman-yunkan:
> in `Library::updateBookDB`, we use `book.getLanguage()` to set the stemmer. With a multilang book, it will not work.

What stemmer should be used for a multilang book? Currently, in such cases the stemmer is simply not used since an attempt to create a stemmer for a non-existing language fails with an exception which is intercepted and ignored.

> The same way, we don't split the comma separated list when indexing the book itself. It seems to work (based on the tests you have added), propably because we find the queried language in the book's language list but it would be better to explicitly split it.

Will fix.

> And I think we should change `std::string Book::getLanguage()` to `std::vector<std::string> Book::getLanguages()`

Should we do that in one step? Or add `Book::getLanguages()` and deprecate `Book::getLanguage()`?

veloman-yunkan:
> > And I think we should change `std::string Book::getLanguage()` to `std::vector<std::string> Book::getLanguages()`
> 
> Should we do that in one step? Or add `Book::getLanguages()` and deprecate `Book::getLanguage()`?

Another option is to preserve `Book::getLanguage()` as a wrapper around `Book::getLanguages()` that throws when called on a multilang book.

mgautierfr:
Agree with the current code.

I think it would be better to depreciate `Book::getLanguage`. We can be sure that people will forget to catch the potential exception thrown if we are in multizim. It is better to move the change explicitly in the api.

veloman-yunkan:
> I think it would be better to depreciate `Book::getLanguage`. We can be sure that people will forget to catch the potential exception thrown if we are in multizim. It is better to move the change explicitly in the api.

@mgautierfr OK, I will deprecate it. Should its current implementation remain unchanged (i.e. should it keep returning a comma separated list of languages for multilang ZIMs)? The other alternatives are:

1. return the first element of `Book::getLanguages()` (or an empty string if the latter is empty)
2. return the only element of `Book::getLanguages()` if there is only one language, otherwise raise an exception.

mgautierfr:
> Should its current implementation remain unchanged (i.e. should it keep returning a comma separated list of languages for multilang ZIMs) ?

Yes. We keep the function (as deprecated) to not break the api, so we shouldn't change the behavior (even if it is buggy)

> And what about Bookmark::getLanguage()? Should we make the same change to Bookmark?

Bookmark is about only one article. We can expect that even if zim file is multilanguages, one article is in one language only, so no change needed.

src/library.cpp
```diff
@@ -455,12 +455,14 @@ void Library::updateBookDB(const Book& book)
...
-    indexer.set_stemming_strategy(Xapian::TermGenerator::STEM_SOME);
-  } catch (...) {}
+  const auto langs = book.getLanguages();
+  if ( langs.size() == 1 ) {
+    try {
+      stemmer = Xapian::Stem(iso639_3ToXapian(langs[0]));
+      indexer.set_stemmer(stemmer);
+      indexer.set_stemming_strategy(Xapian::TermGenerator::STEM_SOME);
+    } catch (...) {}
+  }
```
mgautierfr, line 465: Could we try to load languages until we found one (not only checking the first one) ?

src/server/internalServer.cpp
```diff
@@ -223,7 +223,8 @@ typedef std::set<std::string> Languages;
 Languages getLanguages(const Library& lib, const Library::BookIdSet& bookIds) {
   Languages langs;
   for ( const auto& b : bookIds ) {
-    langs.insert(lib.getBookById(b).getLanguage());
+    const auto bookLangs = lib.getBookById(b).getLanguages();
+    langs.insert(bookLangs.begin(), bookLangs.end());
```
mgautierfr, line 227: This seems a functional change, not just a deprecation of `getLanguage` so it should go in another commit.
(Not important in itself, but as we have other changes to do, let's split this change correctly)

mgautierfr (changes requested):

src/library.cpp
veloman-yunkan: I think you missed the `if()` surrounding the `try{}catch()`. A language (and a corresponding stemmer) is associated with the book metadata (title, description, tags, etc) only if the book is known to be in one language. Otherwise, if the book is a multilanguage one how can we figure out in which language, say, its description is written?

src/server/internalServer.cpp
veloman-yunkan: Done

src/library.cpp
mgautierfr: I haven't missed the `if()`. But I have missed the fact it was at indexing time. I was missing with libzim feature of detect the query language based on the database language.

I was totally wrong on this side. So we are good here 🙂

mgautierfr (approved):

PR #907, opened 2023-03-06
Use fragment value instead of search query for filters

### LINKED ISSUE #565: [SERVER] Do not use query to convey the filters of the main page.
The main page of the server now have a filter feature.

Of course we need to "expose" the filters in the url. This allow the user to copy/paste the url and keep the applied filter.

But the querystring (the string after `?`) is part of the url send to the server. It means that urls `library.kiwix.org`, `library.kiwix.org?lang=foo` and `library.kiwix.org?lang=bar` are three different urls, browser must send them to the server. This is a pity as the server don't care about the querystring here and always send back the same content.

We could use the hash value (the string after `#`) to store the active filters. The hash value is not send to the server. It means that the three urls, would be "reduce" to `library.kiwix.org`. The browser could detect they are "the same" and cache the content. In the same time, the browser (and our js) will know that urls are not the same and will correctly handle that (change the url in the address bar, add a history state, ...).

juuz0:
Fixes #565

### DIFF
--- a/static/skin/i18n/en.json
+++ b/static/skin/i18n/en.json
@@ -28,7 +28,7 @@
 	, "random-page-button-text": "Go to a randomly selected page"
 	, "searchbox-tooltip": "Search '{{BOOK_TITLE}}'"
 	, "confusion-of-tongues": "Two or more books in different languages would participate in search, which may lead to confusing results."
-	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"?lang=\">reset filter</a>?"
+	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"#lang=\">reset filter</a>?"
 	, "powered-by-kiwix-html": "Powered by&nbsp;<a href=\"https://kiwix.org\">Kiwix</a>"
 	, "search": "Search"
 	, "book-filtering-all-categories": "All categories"
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -1,4 +1,12 @@
 (function() {
+    class FragmentParams extends URLSearchParams {
+        constructor(fragment = '') {
+            if (fragment[0] == '#')
+                fragment = fragment.substring(1);
+            super(fragment);
+        }
+    }
+
     const root = document.querySelector(`link[type='root']`).getAttribute('href');
     const incrementalLoadingParams = {
         start: 0,
@@ -14,18 +22,18 @@
     let isFetching = false;
     let noResultInjected = false;
     let filters = getCookie(filterCookieName);
-    let params = new URLSearchParams(window.location.search || filters || '');
+    let params = new FragmentParams(window.location.hash || filters || '');
     params.delete('userlang');
     let timer;
     let languages = {};
     let previousScrollTop = Infinity;
 
     function updateFeedLink() {
-        const inputParams = new URLSearchParams(window.location.search);
-        const filteredParams = new URLSearchParams();
+        const inputParams = new FragmentParams(window.location.hash);
+        const filteredParams = new FragmentParams();
         for (const [key, value] of inputParams) {
             if ( value != '' ) {
-                filteredParams.append(key, value);
+                filteredParams.set(key, value);
             }
         }
         const feedLink = `${root}/catalog/v2/entries?${filteredParams.toString()}`;
@@ -377,10 +385,10 @@
         incrementalLoadingParams.count = viewPortToCount();
         fadeOutDiv.style.display = 'none';
         bookOrderMap.clear();
-        params = new URLSearchParams(window.location.search);
+        params = new FragmentParams(window.location.hash);
         if (filterType) {
             params.set(filterType, filterValue);
-            window.history.pushState({}, null, `?${params.toString()}`);
+            window.history.pushState({}, null, `#${params.toString()}`);
             setCookie(filterCookieName, params.toString(), oneDayDelta);
         }
         updateFilterColors();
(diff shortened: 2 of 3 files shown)


juuz0:
I also changed some indentation to match rest of the document

kelson42:
@juuz0 A brother PR should be done in kiwix-tools to upgrade documentation.

kelson42:
@juuz0 Using the term "hash" is wrong (see https://stackoverflow.com/questions/2800187/what-is-it-when-a-link-has-a-pound-sign-in-it#2800195) and actually misleading (because in programming "hash" refers to hashing algoritms).

Current naming is IMO appropriate and we can keep it even if we move the parameters after the pound character.

juuz0:
@kelson42 Thanks, `fragment` seems a much better name (another resource suggesting its use: https://www.rfc-editor.org/rfc/rfc3986#section-4.2)

I am not sure what you mean by
`Current naming is IMO appropriate and we can keep it even if we move the parameters after the pound character.`

Do you mean keeping class name as `URLSearchParams`? It could work but that's a [built-in JS interface](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). We probably shouldn't be naming it same - may use search queries in future where URLSearchParams is needed.

Alternative name could be `FragmentParams`

juuz0:
> A brother PR should be done in kiwix-tools to upgrade documentation

I looked through https://github.com/kiwix/kiwix-tools/blob/main/docs/kiwix-serve.rst 
Seems like it doesn't really need any modifications. It doesn't talk about filtering on welcome page (the only target of this PR)

static/skin/index.js
```diff
@@ -1,4 +1,24 @@
...
+            const nmap = new Map (hash
+                .split("&")
+                .map(v => v.split("=")));
+            super(nmap);
+        }
+
+        toString() {
+            let finalString = '';
+            this.forEach((value, key) => {
+                if (key != '')
```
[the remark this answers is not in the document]
juuz0: Empy `#`.
Eg: first time visiting `192.168.18.8:8080/` (no cookies)

static/skin/i18n/en.json
```diff
@@ -28,7 +28,7 @@
 	, "random-page-button-text": "Go to a randomly selected page"
 	, "searchbox-tooltip": "Search '{{BOOK_TITLE}}'"
 	, "confusion-of-tongues": "Two or more books in different languages would participate in search, which may lead to confusing results."
-	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"?lang=\">reset filter</a>?"
+	, "welcome-page-overzealous-filter": "No result. Would you like to <a href=\"#lang=\">reset filter</a>?"
```
mgautierfr, line 31: This is not really related to this change but why specifying `lang=` ? If we don't want a filter, shouldn't we pass no arguments at all ?

static/skin/index.js
```diff
@@ -1,4 +1,24 @@
 (function() {
+    class FragmentParams extends Map {
+        constructor(fragment = '') {
+            if (fragment[0] == '#')
+                fragment = fragment.substring(1); // remove #
+            const nmap = new Map (fragment
+                .split("&")
+                .map(v => v.split("=")));
+            super(nmap);
+        }
```
mgautierfr, line 10: UrlSearchParam properly handle url encoded parameters (see the note in https://url.spec.whatwg.org/#urlsearchparams). We should do the same for the fragments (constructor and tostring).

mgautierfr (changes requested):

static/skin/index.js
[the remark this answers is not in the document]
veloman-yunkan: Isn't it better to parse the fragment query in such a way that spurious empty keys are not added in the first place?

static/skin/i18n/en.json
veloman-yunkan, line 31: The difference between the search (query) and fragment components in a URL is that when the search/query component changes the browser issues a new request and reloads the page, whereas for a changed fragment component the page isn't (re)loaded (the default action is to scroll the page to the anchor which the fragment component is supposed to stand for). You have to set-up a `window.onhashchange` event handler.

veloman-yunkan (changes requested):

static/skin/i18n/en.json
juuz0: If we do that, the arguments will be taken from cookie values and it'll be on the same no results page everytime

mgautierfr: Ok, let's move on is it is for this PR. 
But it somehow rejoin the discussion in https://github.com/kiwix/libkiwix/pull/897#discussion_r1129196868, let's continue the discussion there.

juuz0: Done.

veloman-yunkan (approved):

static/skin/index.js
```diff
@@ -14,18 +22,18 @@
...
 
     function updateFeedLink() {
-        const inputParams = new URLSearchParams(window.location.search);
-        const filteredParams = new URLSearchParams();
+        const inputParams = new FragmentParams(window.location.hash);
+        const filteredParams = new FragmentParams();
         for (const [key, value] of inputParams) {
             if ( value != '' ) {
-                filteredParams.append(key, value);
+                filteredParams.set(key, value);
```
mgautierfr, line 36: I'm not sure about this change.
Why using set instead of append ?
I've the feeling this is a reminder a the previous version where you implement `FragmentParams`  using a map.

veloman-yunkan: Frankly, during my last review I also wanted to comment on this change, even wrote a comment but then deleted it, since this change has no impact on the meaningful URLs of the welcome page (there are no query params that can be repeated).

juuz0: Should I change this then? :>

mgautierfr: Let's move on and merge this PR !

mgautierfr (approved):

Issue #913, opened 2023-03-15
No cursor on UI language switch button?


rgaudin:
Is there a reason for not having a cursor indicating interactivity on the UI language switch button? Or is it just missing?

![Screen_Shot_2023-03-15_at_11_34_29](https://user-images.githubusercontent.com/57929/225297401-d7b59c2b-a7b9-47b2-bd1a-5c8c56253b1a.png)

kelson42:
@veloman-yunkan Not one of the small UI things you want to fix?

rgaudin:
Maybe a duplicate of #900 but the feed one has the pointer…

veloman-yunkan:
This ticket is duplicate of #900 which was closed by #910

Issue #900, opened 2023-02-22
Add a "pointer" cursor and tool-tip on UI language selector.


mgautierfr:
Small suggestion (from @juuz0)
Maybe the UI language selector icon could have cursor as "pointer" & a tool-tip to tell what it does

_Originally posted by @juuz0 in https://github.com/kiwix/libkiwix/issues/846#issuecomment-1438697702_

mgautierfr:
Answer from @veloman-yunkan 

>  Small suggestion: Maybe the UI language selector icon could have cursor as "pointer" & a tool-tip to tell what it does

The tricky question regarding the tooltip for the change-UI-language button - what language should it be in? In the current language of always in English?

Also, I figured out that the header text of the UI language selector is not translated, so the same (above) question applies to it too.

juuz0:
> The tricky question regarding the tooltip for the change-UI-language button - what language should it be in? In the current language of always in English?

Going by how it's done in almost every app, the tooltip & header text should be in current language imo

veloman-yunkan:
> > The tricky question regarding the tooltip for the change-UI-language button - what language should it be in? In the current language of always in English?
> 
> Going by how it's done in almost every app, the tooltip & header text should be in current language imo

The hint for this particular action is special and has to be in a language that has the highest chance of being understood by the user. I believe that at this instant in the history of humankind that language is English.

Issue #899, opened 2023-02-22
Why are tags transformed?


rgaudin:
Sorry if the title is misleading, I am talking about the use of `humanFriendlyTitle()` for the display of tags in the book card.

https://github.com/kiwix/libkiwix/blob/bf80367b5a9f55758947cb481d61543b47bbeca6/static/skin/index.js#L69-L77

<img width="293" alt="Screenshot_2023-02-22_at_10_17_20" src="https://user-images.githubusercontent.com/57929/220591217-567ca4ca-f78b-4dce-b869-25294a1866bd.png">

I understand it **may** look better in some situation but given [the spec](https://wiki.openzim.org/wiki/Tags) allows ZIM creators anything but the `;` character (that includes spaces and uppercase!), if the tags are not satisfying, we should fix them in the scrapers and not transform them on the reader.

- additional code to maintain
- confusion across readers which don't follow same rule (or not at same pace)
- duplication of code
- arbitrary choices from code on a data that is clearly content

Maybe that should raise a discussion about what Tags we should set on all the ZIM we create. We've been very lazy about this.

mgautierfr:
I can't answer you why we are transforming the tag, there is a lot of people working on it but not me (https://github.com/kiwix/libkiwix/blame/bf80367b5a9f55758947cb481d61543b47bbeca6/static/skin/index.js#L69-L77)

But I can say you why we may want to transform the tag: We may want to store tag in a simple, standardized, searchable form (as lower case, no space, sentence of words separated by `_`) and in the same time displaying it nicely to the user.

I don't know if it is a good thing. We may (or not) indeed have better to state that tags must be stored as they will be displayed and then have something else to do search and filtering.

rgaudin:
> I can't answer you why we are transforming the tag, there is a lot of people working on it but not me (https://github.com/kiwix/libkiwix/blame/bf80367b5a9f55758947cb481d61543b47bbeca6/static/skin/index.js#L69-L77)

Really like that github blame view! 

> But I can say you why we may want to transform the tag: We may want to store tag in a simple, standardized, searchable form (as lower case, no space, sentence of words separated by `_`) and in the same time displaying it nicely to the user.

Kind of what we should do about Categories but categories and tags are semantically different.
[...]

PR #920, opened 2023-03-19
Parameterised feed tool tip 

### LINKED ISSUE #914: Feed Icon tooltip
The feed icon is misleading in that its tooltip doesn't change while its target changes with filters… A user might thus copy and use a filtered-feed thinking he is subscribing to the whole thing.

- Current: `Library OPDS Feed`
- Suggestion (no filter): `Library OPDS Feed - All entries`
- Suggestion (filtered): `Library OPDS Feed - Only for Language ENG, Category gutenberg matching XXX` (only relevant parts to apply of course)

Also, the icon is not proportioned with the one it is sitting next to, making it look bigger.
@siemsie could you take a lot at the two icon on the top right and suggest something?

Screen Shot 2023-03-15 at 11 34 29

juuz0:
Fixes #914


kelson42:
Adding @rgaudin as reviewer as he requested this improvement. On my side I have a concerns about how to translate the new hint.

kelson42 (changes requested):
This PR generate XML which is not valid like:
```
<title>Filtered Entries (lang=eng&amp;category=wikipedia)</title>
```

I guess this is the problem because I had: 
image

rgaudin (approved):
Works as expected 👍  and it does look much better. Thanks.

Hopefully #900 #913 shall follow 😉

veloman-yunkan:
> This PR generate XML which is not valid

@kelson42 That issue predates this PR and has to be fixed in the backend.

juuz0:
> The parameterized message doesn't read well when more than one filtering criteria are used:

@veloman-yunkan would this look better then?
`Library OPDS Feed - Only for language ENG category TED matching "scope"`

veloman-yunkan:
> > The parameterized message doesn't read well when more than one filtering criteria are used:
> 
> @veloman-yunkan would this look better then? `Library OPDS Feed - Only for language ENG category TED matching "scope"`

Better but @rgaudin [has some reservations](https://github.com/kiwix/libkiwix/issues/899) regarding changes in capitalization. Also please make sure that the grammar is correct for any combination of filters (e.g. the current template will produce a deficient result in the presence of the search filter only).

rgaudin:
I would not transform category names (uppercase).

I think better separators like commas between values and maybe hyphens between lang and categories might be better.

That's the kind of things that improves with time and feedback. I doubt we can have have something really satisfying using just the title attr.

juuz0:
attempt 2 at making it look meaningful :>
image

rgaudin:
Looks a lot better. I'd change the text to:

```
Library OPDS Feed – entries matching
Language: eng
Category: gutenberg, wikipedia
Query: test
```

- I think the *mathcing* wording is more explicit than the *only for* one.
- Using colons and titled parameters helps distinguish the key from the value.

rgaudin (changes requested):
See my previous comment

veloman-yunkan (changes requested):
`static/skin/i18n/qqq.json` must be updated too

static/skin/i18n/qqq.json
```diff
@@ -47,5 +47,7 @@
 	"torrent-download-alt-text": "Hint for the icon of torrent download",
 	"library-opds-feed": "Hint for the library OPDS feed link",
```
veloman-yunkan, line 48: library-opds-feed must be removed from qqq.json

juuz0: I remember removing it hah. Probably sneaked in while I was pressing Ctrl Z repeatedly. Thanks, removed.

kelson42:
@juuz0 @veloman-yunkan In this PR we have replaced pngs with svg. Does that have been discussed? Is a contol moved? At a first sight, I don’t see any benefit for that… to the contrary.

rgaudin:
One was png, the other was inline svg. Now both are external svg which reduce html size and enhance caching while being harmonized.
Hasn't been discussed though. Was a consequence of simsie sending svg files.

What is your concern regarding svg?

juuz0:
@kelson42 
Two things:
1. ui language selector button was already a SVG, no replacement of filetypes here. The feed link icon was a png, this is replaced.
2. Images in https://github.com/kiwix/libkiwix/issues/914#issuecomment-1470182142 were SVGs.

I dont have any particular reservation on using any, to be honest.

kelson42:
@rgaudin Its significantly more complex to render (energy consumption) and I worry about the support with older browsers.

juuz0:
> I worry about the support with older browsers.

99.11% as per https://caniuse.com/svg. Pretty good i think

PR #923, opened 2023-03-28
svg files only

Amarjit0511:
Made changes in library.xml and library.cpp file


kelson42:
This PR does not pass the CI and makes no sense.

PR #924, opened 2023-03-29
Enable Chromium to display PDFs in the viewer iframe

### LINKED ISSUE #916: PDF content blocked by Chrome
When opening a PDF in a ZIM in the kiwix-serve viewer, on Chrome, there is an unrecoverable error message. This looks like a regression.

Screenshot 2023-03-15 at 14 18 57

Go to https://dev.library.kiwix.org/viewer#lilote_fr_fo_2023-03/xay-va-%C3%A0-la-p%C3%AAche then click the _Lire l'histoire_ button.

veloman-yunkan:
Fixes #916

The fix doesn't work for Chromium versions under 91. Frankly I haven't verified that it works for Chromium 91+ because I only have 90.

### DIFF
--- a/src/server/response.cpp
+++ b/src/server/response.cpp
@@ -412,6 +412,12 @@ ContentResponse::ContentResponse(const std::string& root, bool verbose, const st
   m_mimeType(mimetype)
 {
   add_header(MHD_HTTP_HEADER_CONTENT_TYPE, m_mimeType);
+  if ( !startsWith(m_mimeType, "application/pdf") ) {
+    add_header("Content-Security-Policy",
+               "default-src 'self' data: blob: about: chrome-extension: 'unsafe-inline' 'unsafe-eval'; "
+               "sandbox allow-scripts allow-same-origin allow-modals allow-popups allow-forms allow-downloads;");
+    add_header("Referrer-Policy", "no-referrer");
+  }
 }
 
 std::unique_ptr<ContentResponse> ContentResponse::build(
--- a/static/viewer.html
+++ b/static/viewer.html
@@ -2,6 +2,10 @@
 <html>
   <head>
     <meta charset="UTF-8">
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'self' data: 'unsafe-inline' 'unsafe-eval';
+                   frame-src 'self' moz-extension: chrome-extension:;
+                   object-src 'none';">
     <title>ZIM Viewer</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link type="text/css" href="./skin/taskbar.css?KIWIXCACHEID" rel="Stylesheet" />
@@ -65,8 +69,7 @@
     </div>
 
     <iframe id="content_iframe"
-            referrerpolicy="same-origin"
-            sandbox="allow-same-origin allow-scripts"
+            referrerpolicy="no-referrer"
             onload="on_content_load()"
             src="./skin/blank.html?KIWIXCACHEID" title="ZIM content" width="100%"
             style="border:0px">


src/server/response.cpp
```diff
@@ -412,6 +412,12 @@ ContentResponse::ContentResponse(const std::string& root, bool verbose, const st
   m_mimeType(mimetype)
 {
   add_header(MHD_HTTP_HEADER_CONTENT_TYPE, m_mimeType);
+  if ( !startsWith(m_mimeType, "application/pdf") ) {
```
Jaifroid, line 415: @veloman-yunkan Have you tested that this actually catches PDF files? E.g., what if the mimeType isn't set? You might want to check also the file name with a case-insensitive regex like `/\.pdf$/i`.

src/server/response.cpp
```diff
@@ -412,6 +412,12 @@ ContentResponse::ContentResponse(const std::string& root, bool verbose, const st
...
+    add_header("Content-Security-Policy",
+               "default-src 'self' data: blob: about: chrome-extension: 'unsafe-inline' 'unsafe-eval'; "
```
Jaifroid, line 417: `chrome-extension:` would only be needed if you were planning to package Kiwix Serve inside an extension or inside an Electron App (the latter, at least, would be possible, but probably using a node-based server). Seems unlikely, so this could be removed.

static/viewer.html
```diff
@@ -2,6 +2,10 @@
 <html>
   <head>
     <meta charset="UTF-8">
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'self' data: 'unsafe-inline' 'unsafe-eval';
+                   frame-src 'self' moz-extension: chrome-extension:;
```
Jaifroid, line 7: Ditto for both `moz-extension:` and `chrome-extension:`. Unlikely that you'd be serving extension URIs. The reason we have these in Kiwix JS is that we package the app inside both types of extension.

Jaifroid:
@veloman-yunkan Just as a general comment on this PR, it will probably need to deal with opening external links, which will be blocked from opening in the iframe (or should be, though I no longer trust the parallels with the case of Kiwix JS). The way I suggested dealing with this was to add an on-click event listener on the iframe's content window (code provided in the relevant issue). Of course it would be worth testing first what happens to external links in this implementation. Another way to deal with it would be to alter the HTML of all external links in the server backend, adding a `target="_blank"` attribute. However, that could conflict with an existing target attribute, so doing it dynamically is cleaner.

src/server/response.cpp
veloman-yunkan: > Have you tested that this actually catches PDF files?

On the [test case](https://master.download.kiwix.org/zim/.hidden/dev/lilote_fr_test_2023-01.zim) that I am using - it does.

veloman-yunkan:
> @veloman-yunkan Just as a general comment on this PR, it will probably need to deal with opening external links, which will be blocked from opening in the iframe (or should be, though I no longer trust the parallels with the case of Kiwix JS). The way I suggested dealing with this was to add an on-click event listener on the iframe's content window (code provided in the relevant issue). Of course it would be worth testing first what happens to external links in this implementation. Another way to deal with it would be to alter the HTML of all external links in the server backend, adding a `target="_blank"` attribute. However, that could conflict with an existing target attribute, so doing it dynamically is cleaner.

@Jaifroid I primarily published this PR as a means of sharing my changes with you. At the moment my goal is to reproduce [your solution](https://github.com/kiwix/libkiwix/issues/916#issuecomment-1488508290) in `kiwix-serve` whereupon I will productize it.

mgautierfr:
LGTM. At least I haven't found blockers.
Let's merge it and see/test it with nightly and `dev.library.kiwix.org`

PR #930, opened 2023-04-10
Really enable Chromium to display PDFs in the viewer iframe

### LINKED ISSUE #916: PDF content blocked by Chrome

veloman-yunkan:
Fixes #916

The previous "fix" (merged PR #92 ) was buggy. It not only didn't work in Chromium v90, but in more recent versions too.

I verified that this fix works in Firefox (v111) and Chromium (v90):

- Attempts by the ZIM content to break out of the viewer iframe are blocked.
- PDFs are displayed in the viewer.

Also (unrelated to the bug being addressed) removed `moz-extension:` and `chrome-extension:` from list of valid sources in content security policy as they are not justified for `kiwix-serve` (as [noted](https://github.com/kiwix/libkiwix/pull/924#discussion_r1152134334) in @Jaifroid's review of #924)

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -1123,6 +1123,29 @@ ParameterizedMessage suggestSearchMsg(const std::string& searchURL, const std::s
                               });
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// The content security policy below is set on responses to the /content
+// endpoint in order to prevent the ZIM content from interfering with the
+// viewer (e.g. breaking out of the viewer iframe by performing top-level
+// navigation).
+const std::string CONTENT_CSP_HEADER =
+    "default-src 'self' "
+                "data: "
+                "blob: "
+                "about: "
+                "'unsafe-inline' "
+                "'unsafe-eval'; "
+
+    "sandbox allow-scripts "
+            "allow-same-origin "
+            "allow-modals "
+            "allow-popups "
+            "allow-forms "
+            "allow-downloads;";
+
+// End of content security policy
+///////////////////////////////////////////////////////////////////////////////
+
 } // unnamed namespace
 
 std::unique_ptr<Response>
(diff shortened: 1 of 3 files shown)


mgautierfr (approved):
LGTM

rgaudin (approved):
LGTM as well but I didn't test. I can't compile libzim on my box (https://github.com/openzim/libzim/issues/778) and on an ubuntu one, it compiles but the test fails…

kelson42:
Mandatory to me is the approval of @Jaifroid

kelson42:
@veloman-yunkan @mgautierfr macos CI fails!

rgaudin:
> @veloman-yunkan @mgautierfr macos CI fails!

It's no one's fault. We install packages via brew and one of the package's dependency is python which can't be linked because it's already linked.

I believe passing `--overwrite` to the `brew install` command could fix it but it's unrelated to this PR

kelson42:
Another issue/PR needs then probably to be open asap.

Jaifroid:
Sorry, I'm still 7 hours behind CET, will review shortly.

src/server/internalServer.cpp
```diff
@@ -1184,6 +1207,13 @@ std::unique_ptr<Response> InternalServer::handle_content(const RequestContext& r
     auto response = ItemResponse::build(*this, request, entry.getItem());
     response->set_etag_body(archiveUuid);
 
+    if ( !startsWith(entry.getItem().getMimetype(), "application/pdf") ) {
```
Jaifroid, line 1210: Are we confident that all PDFs will have a set MIME type, and that it will always be `application/pdf` (and always lowercase)? What about `text/pdf` or `application/x-pdf`, which can be used? I'm thinking of Zimit archives, where we don't really control such things (for Gutenberg and mwOffliner, we can be sure of the MIME type, I think).

Jaifroid (approved):
Thanks, this looks good to me with one query about the PDF MIME type which may just be over-cautious. N.B. I can't test because I can't build this on my Windows machine. I'll happily test it thoroughly once it has been merged and is available in the dev server.

src/server/internalServer.cpp
veloman-yunkan: I think @mgautierfr and/or @rgaudin can answer this question better than me

Jaifroid: That's convincing to me!

mgautierfr:
Seems we are good. I'm merging.

PR #931, opened 2023-04-11
Fix macOS CI python dependency install

rgaudin:
meson and ninja both depends on python3 which received an update.
This python3 update fails to install when linking.

This temp fix removes those files. Hopefully a future update will remove the need for this hack

I couldn't find any cleaner way to get passed this…

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ jobs:
       - name: Install packages
         run: |
           brew update
+          brew unlink python3
+          # upgrade from python@3.11.2_1 to python@3.11.3 fails to overwrite those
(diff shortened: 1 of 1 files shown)


mgautierfr:
@rgaudin Do you have issue open somewhere about that ? Or a discussion suggesting those changes  ?

We can be sure that we (I) will forget about that once CI works.

rgaudin:
Nope. We can open a ticket here to check that every week. Should vanish as soon as the github runner is rebuilt

rgaudin:
Yes, when updating as a dependency, the linking is not done properly. Linking here means the creating of symlinks in `/usr/local/bin`. I've tried different combinations of options without success.

PR #944, opened 2023-05-02
Fixed external links in the viewer iframe

### LINKED ISSUE #943: External content is inaccessible in the viewer iframe

veloman-yunkan:
Fixes #943

Before this fix clicking an external link in the viewer iframe had no effect (other than an error being reported in the browser dev tools console) because the attempt to navigate the top browser context was suppressed due to sandboxing.

One downside with this solution is that it again opens the possibility of losing the viewer toolbar if an internal link in a ZIM file has its `target` attribute set to `_top` (however, such links can be detected and corrected by the viewer). I cannot think of other downsides.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -257,7 +257,7 @@ function isExternalUrl(url) {
       || url.startsWith("https:");
 }
 
-function onClickEvent(e) {
+function handleLinkClick(e) {
   const iframeDocument = contentIframe.contentDocument;
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
@@ -266,6 +266,8 @@ function onClickEvent(e) {
       if ( viewerSettings.linkBlockingEnabled ) {
         return blockLink(target);
       }
+    } else {
+      target.setAttribute("target", "content_iframe");
     }
   }
 }
@@ -301,7 +303,7 @@ this.Element && function(ElementPrototype) {
 }(Element.prototype);
 
 function setup_external_link_blocker() {
-  setupEventHandler(contentIframe.contentDocument, 'a', 'click', onClickEvent);
+  setupEventHandler(contentIframe.contentDocument, 'a', 'click', handleLinkClick);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -1088,6 +1088,7 @@ const std::string CONTENT_CSP_HEADER =
 
     "sandbox allow-scripts "
             "allow-same-origin "
+            "allow-top-navigation-by-user-activation "
             "allow-modals "
             "allow-popups "
             "allow-forms "
--- a/static/viewer.html
+++ b/static/viewer.html
@@ -69,6 +69,7 @@
     </div>
 
     <iframe id="content_iframe"
+            name="content_iframe"
             referrerpolicy="no-referrer"
             onload="on_content_load()"
             src="./skin/blank.html?KIWIXCACHEID" title="ZIM content" width="100%"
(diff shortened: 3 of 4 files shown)


veloman-yunkan:
@rgaudin @kelson42 I've tested this PR but expect you to try it too

kelson42:
@veloman-yunkan Seems to work fine on FF 111, but reopening the problem with _top target is not acceptable IMO

Jaifroid:
See https://caniuse.com/?search=allow-top-navigation-by-user-activation. The one browser you might be concerned by is Safari on iOS (e.g. iPad), where this workaround wouldn't work. It also wouldn't work with IE11, but I guess that's less of a concern nowadays.

static/skin/viewer.js
```diff
@@ -266,6 +266,8 @@ function onClickEvent(e) {
       if ( viewerSettings.linkBlockingEnabled ) {
         return blockLink(target);
       }
+    } else {
+      target.setAttribute("target", "content_iframe");
```
mgautierfr, line 270: This change is nice.
I even wonder if you would have done this at first instead of introducing complex sandboxing.

Jaifroid: Well, the sandboxing was necessary to prevent scripts in the ZIM accessing external resources ("phoning home") -- a real issue with Zimit ZIMs -- and also to prevent top-level navigation. We tried lots of combinations...

mgautierfr: Yeah yeah, but I think is better for us to don't count the time we have spend on this 😉 

---

Maybe (as almost all the time), the best would have been to fix the root cause and fix the generated zims.
After all, the SW embedded in the zim files is able to catch a request to `www.foo.com/bar` and do a request to `/A/www.foo.com/bar` instead. We could (and can) make the SW refuse request to unhanded url.

veloman-yunkan:
> I agree with @kelson42. I have the feeling that this PR negates the initial purpose of sandboxing. It was never to add safety but to avoid zim files breaking navigation by changing `_top`.

I tried to fix the concern with `target="_top"` links (see the fixup commit). Though it worked in Firefox, unfortunately it was a little short of working in Chromium (v90).

The [attached ZIM file](https://github.com/kiwix/libkiwix/files/11383032/naughty.zim.zip) demonstrates the problem.

There are two pages. Either page contains a link to the other page with target set to `_top`. Also the pages contain javascript code that tries to break out of the iframe with some delay (without any user interactions). The delay for the first page is 5 seconds, and for the second page - 1 second.

Neither page can break out of the iframe when loaded in the viewer by specifying its URL in the address bar (e.g. http://localhost:8080/viewer#naughty/naughty.html or http://localhost:8080/viewer#naughty/naughty2.html). Clicking the links also prevents the escape attempt by the links themselves. However the link from the first page to the second page enables the javascript breakout attempt (with short delay) to succeed!

Note again that such behaviour is specific to Chromium (at least v90) and not to Firefox (at least v112).

static/skin/viewer.js
veloman-yunkan: Sandboxing was needed to prevent programmatic escape from the iframe (see kiwix/kiwix-tools#604). Unfortunately this trick [doesn't work perfectly in Chromium](https://github.com/kiwix/libkiwix/pull/944#issuecomment-1533056709)

veloman-yunkan:
> See https://caniuse.com/?search=allow-top-navigation-by-user-activation. The one browser you might be concerned by is Safari on iOS (e.g. iPad), where this workaround wouldn't work. It also wouldn't work with IE11, but I guess that's less of a concern nowadays.

@Jaifroid Thanks for pointing out this shortcoming of this approach. I found [another one](https://github.com/kiwix/libkiwix/pull/944#issuecomment-1533056709) before seeing your comment. It looks like we have to abandon this PR but at least it was worth a try.

veloman-yunkan:
Replaced by #946

PR #946, opened 2023-05-09
Fixed external links in the viewer iframe (second attempt)

### LINKED ISSUE #943: External content is inaccessible in the viewer iframe

veloman-yunkan:
Fixes #943

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -309,14 +309,22 @@ function setup_external_link_blocker() {
   setupEventHandler(contentIframe.contentDocument, 'a', 'click', onClickEvent);
 }
 
+function getBlockedUrl(catchExternalUrl) {
+  const p = new URLSearchParams(catchExternalUrl.split('?')[1]);
+  return p.get('source');
+}
+
 function handleInterceptedExternalLink(catchExternalUrl) {
   // The external link blocking page was loaded in the viewer iframe.
-  // We need to get rid of the viewer taskbar and display the confirmation
-  // page in the top frame.
-  const urlpath = `${root}/` + catchExternalUrl;
+  // We need to get rid of the viewer taskbar and load in the top frame either
+  // the external resource or, if running in --blockexternal mode, the
+  // confirmation page
+  const url = viewerSettings.linkBlockingEnabled
+            ? `${root}/` + catchExternalUrl
+            : getBlockedUrl(catchExternalUrl);
   history.back(); // drop from the browsing history the state where the
                   // external link catcher page is loaded in the iframe ...
-  window.location = urlpath; // ... and load it in the top frame instead
+  window.location = url; // ... and load the target in the top frame instead
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -23,7 +23,7 @@ function userUrl2IframeUrl(url) {
 }
 
 function getBookFromUserUrl(url) {
-  if ( url == '' ) {
+  if ( url == '' || url.startsWith('catch/external?') ) {
     return null;
   }
 
@@ -131,6 +131,10 @@ function iframeUrl2UserUrl(url, query) {
     return '';
   }
 
+  if ( url == `${root}/catch/external` ) {
+    return `catch/external${query}`;
+  }
+
   if ( url == `${root}/search` ) {
     return `search${query}`;
   }
@@ -224,8 +228,12 @@ function handle_content_url_change() {
   const iframeContentUrl = iframeLocation.pathname;
   const iframeContentQuery = iframeLocation.search;
   const newHash = iframeUrl2UserUrl(iframeContentUrl, iframeContentQuery);
-  history.replaceState(viewerState, null, makeURL(location.search, newHash));
-  updateCurrentBookIfNeeded(newHash);
+  if ( newHash.startsWith('catch/external?') ) {
+    handleInterceptedExternalLink(newHash);
+  } else {
+    history.replaceState(viewerState, null, makeURL(location.search, newHash));
+    updateCurrentBookIfNeeded(newHash);
+  }
 };
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
One downside of the used approach is that the external link blocker page may briefly appear even if external blocking mode has not been enabled. That can be fixed by making the `/catch/external` API endpoint always deliver a blank page and generating the HTML in the frontend only when needed.

rgaudin (commented):
I am not sure exactly what this PR attempts to fix but here's what I tested:

- opening PDF files in the viewer ; on Chrome and FF
- clicking an external link in the viewer (there's one at bottom of eleda ZIM)
  - on Chrome and FF
  - both with and without `-b`

It does work as expected: going to the website without `-b` and going to the (unframed) catch page with the option on.
I do see a flash as described in the PR comment.

One issue I noted is:

1. using `-b`
2. Go to ZIM ; click external link
3. Arrive at catch page
4. Hit _Back_ to return to ZIM
5. Focus out of the browser (switch to a different app)
6. Click that same link again

```
Content Security Policy: The page’s settings blocked the loading of a resource at https://eleda.education/ (“frame-src”).
```

If I don't switch to a different app at step 5, re-clicking works.

What's funny (not) is that this happens with both FF and Chrome. I suppose the event is more precise than the whole app but just switching tabs doesn't triggers it.

static/skin/viewer.js
```diff
@@ -262,10 +270,7 @@ function onClickEvent(e) {
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
     if ( isExternalUrl(target.href) ) {
-      target.setAttribute("target", "_top");
-      if ( viewerSettings.linkBlockingEnabled ) {
-        return blockLink(target);
-      }
+      return blockLink(target);
```
Jaifroid, line 273: @veloman-yunkan Instead of loading a blocked page here, why not just do `window.open(url, target)` to open in a new window? What's the advantage of loading a blocked page instead., then opening the new window? (I I've understood correctly.) Also, consider using the `mousedown` event instead of `click`, as it happens before any click event listeners that might otherwise fire inside the iframe.

veloman-yunkan: > Instead of loading a blocked page here, why not just do window.open(url, target) to open in a new window? What's the advantage of loading a blocked page instead., then opening the new window? (I I've understood correctly.)

The external page is not opened in a new window. It is opened in the same browser window however in the top frame rather than in the viewer iframe.

> Also, consider using the mousedown event instead of click, as it happens before any click event listeners that might otherwise fire inside the iframe.

A link can also be activated via keyboard which I believe generates a `click` event but not a `mousedown` event.

veloman-yunkan:
@rgaudin Thanks for your meticulous testing! I will try to resolve the issue that you found.

rgaudin:
> @rgaudin I can't reproduce the above behaviour. Did you refresh (Ctrl-shift-R) the browser after restarting `kiwix-serve`?

Yes, I test on a dedicated browser (FF developer edition) that I clear up completely before each test. Chrome is also a test-dedicated browser here.

I forgot to include the mandatory click to the actual website on the external block page ; my bad. That's probably what confuses the browser

veloman-yunkan:
> I forgot to include the mandatory click to the actual website on the external block page ; my bad. That's probably what confuses the browser

@rgaudin That didn't work for me either. However I was able to reproduce with the following sequence (using `kiwix-server -b`):

1. Open a ZIM page containing an external link
2. Click the external link and arrive at the catch page
3. Navigate two steps back in *recorded* history as follows. Right click (or click and hold) the Back button. In Firefox two topmost entries  (the topmost one representing the current location/page) will read "**External link blocked**", whereas in Chromium the current location/page is not included in the list and there is only one history item titled "**External link blocked**". The next history item (3rd in FF and 2nd in Chromium) is the ZIM page where the external link was clicked. Select that entry and return to the ZIM page.
4. Click the external link again - the action will be blocked by the browser due to Content Security Policy violation.

Note that performing backward navigation of step 3 as two consecutive left clicks on the Back button works differently - the crucial effect seems to be contained in skipping over the previous state recorded in the browsing history.

The recorded browsing history for a blocked external link contains a dubious entry - the "**External link blocked**" entry that shows up in Chromium (and the 2nd such entry in FF). Despite its title it actually also points to the ZIM page where the external link was clicked (I fixed that in a fixup commit, all above description assumes absence of that fix). The dubious entry corresponds to the state where the catch page is first loaded in the viewer iframe. My attempt to eliminate that state from the history was only partly successful - when navigating back the user doesn't see the catch page in the viewer iframe however the history entry for the page containing the clicked external link is duplicated.

rgaudin:
Can you push that commit so I can test again?

veloman-yunkan:
> Can you push that commit so I can test again?

The commit only fixes the wrong title of the history entry. It doesn't eliminate the problem.

rgaudin (approved):
Works fine now

mgautierfr:
The code indeed works, but using the url of the iframe (with `catch/external`) to "discuss" with the viewer seems definitively as a workaround.

Maybe it is better to use https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage to allow the script inserted in the page to discuss with the viewer (and potentially move the current "catch external" feature on the viewer side)

Jaifroid:
Just to say that using postMessage requires having an application injected into the HTML of the iframe to receive the messages and act on them, and then that application needs to manipulate the DOM of the iframe document. However, you have access to that DOM from outside the iframe, unless you FULLY sandbox the iframe, so it would only be a clunkier way of doing what you can already do by accessing the iframe contentWindow / contentDocument.

Regarding the current PR, it's hard to comment as I can't build/test the PR, but I don't really understand the logic of trapping a click on the iframe, inspecting it, loading a document that informs the user that the content is blocked, then detecting the block document and redirecting to external content. Is this working around legacy behaviour in the viewer with regard to blocking external content? It *seems* like it's exploiting a side effect rather than directly loading the external content at the top level (or in a new tab). Apologies if I've misunderstood!

veloman-yunkan:
> The code indeed works, but using the url of the iframe (with `catch/external`) to "discuss" with the viewer seems definitively as a workaround.
> 
> Maybe it is better to use https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage to allow the script inserted in the page to discuss with the viewer (and potentially move the current "catch external" feature on the viewer side)

Did it in a separate PR #948 (we have to decide which one should be merged).

While thinking on the cons and pros of the two approaches before implementing the proposed solution I performed some more testing and identified an issue with this PR - when external links should *not* be blocked (`kiwix-serve` is started without the `--blockexternal` option) ctrl-clicking an external link opens the catch page in the new tab.

PR #948, opened 2023-05-16
Fixed external links in the viewer iframe (via postMessage)

### LINKED ISSUE #943: External content is inaccessible in the viewer iframe

veloman-yunkan:
Fixes #943

This PR is an alternative to #946 (as suggested by https://github.com/kiwix/libkiwix/pull/946#issuecomment-1542414873).

Overall this approach is simpler than #946 however while working on it a few issues common for this PR and #946 have been discovered.

Known issues:

- The solution actually introduces a new message-based API for navigation of the viewer window. A committed antiZIMmer can create a malicious page that breaks out of the viewer iframe (though the same could be done with #946 too).
- Ctrl-clicking an external link loads it (or the catch page) both in the viewer window and in a new tab. This can be fixed.
- Middle-clicking (or selecting the "Open in New Tab/Window" from the context menu) doesn't block external links (but that is a common problem for all solutions based solely on intercepting click events).

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -312,8 +312,8 @@ let viewerSetupComplete = false;
 function on_content_load() {
   if ( viewerSetupComplete ) {
     handle_content_url_change();
-    setup_external_link_blocker();
   }
+  setup_external_link_blocker();
 }
 
 function htmlDecode(input) {
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -243,9 +243,10 @@ function matchingAncestorElement(el, context, selector) {
 
 const block_path = `${root}/catch/external`;
 
-function blockLink(target) {
-  const encodedHref = encodeURIComponent(target.href);
-  target.setAttribute("href", block_path + "?source=" + encodedHref);
+function blockLink(url) {
+  return viewerSettings.linkBlockingEnabled
+       ? block_path + "?source=" + encodeURIComponent(url)
+       : url;
 }
 
 function isExternalUrl(url) {
@@ -262,10 +263,8 @@ function onClickEvent(e) {
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
     if ( isExternalUrl(target.href) ) {
-      target.setAttribute("target", "_top");
-      if ( viewerSettings.linkBlockingEnabled ) {
-        return blockLink(target);
-      }
+      target.setAttribute("href", blockLink(target.href));
+      contentIframe.contentWindow.parent.postMessage({ externalURL : target.href }, "*");
     }
   }
 }
@@ -446,6 +445,13 @@ function changeUILanguage() {
   });
 }
 
+function handleMessage(event) {
+  console.log("handleMessage");
+  if ( event.data.externalURL ) {
+    window.location = event.data.externalURL;
+  }
+}
+
 function setupViewer() {
   // Defer the call of handle_visual_viewport_change() until after the
   // presence or absence of the taskbar as determined by this function
@@ -453,6 +459,7 @@ function setupViewer() {
   setTimeout(handle_visual_viewport_change, 0);
 
   window.onresize = handle_visual_viewport_change;
+  window.addEventListener("message", handleMessage);
 
   const kiwixToolBarWrapper = document.getElementById('kiwixtoolbarwrapper');
   if ( ! viewerSettings.toolbarEnabled ) {
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -73,7 +73,7 @@ const ResourceCollection resources200Compressible{
   { DYNAMIC_CONTENT, "/ROOT%23%3F/skin/taskbar.css" },
   { STATIC_CONTENT,  "/ROOT%23%3F/skin/taskbar.css?cacheid=bbdaf425" },
   { DYNAMIC_CONTENT, "/ROOT%23%3F/skin/viewer.js" },
-  { STATIC_CONTENT,  "/ROOT%23%3F/skin/viewer.js?cacheid=d575e81a" },
+  { STATIC_CONTENT,  "/ROOT%23%3F/skin/viewer.js?cacheid=d407a38a" },
   { DYNAMIC_CONTENT, "/ROOT%23%3F/skin/fonts/Poppins.ttf" },
   { STATIC_CONTENT,  "/ROOT%23%3F/skin/fonts/Poppins.ttf?cacheid=af705837" },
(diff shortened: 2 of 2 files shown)


static/skin/viewer.js
```diff
@@ -262,10 +263,8 @@ function onClickEvent(e) {
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
     if ( isExternalUrl(target.href) ) {
-      target.setAttribute("target", "_top");
-      if ( viewerSettings.linkBlockingEnabled ) {
-        return blockLink(target);
-      }
+      target.setAttribute("href", blockLink(target.href));
+      contentIframe.contentWindow.parent.postMessage({ externalURL : target.href }, "*");
```
Jaifroid, line 267: I don't understand the point of this. You are instructing the iframe window to send a message to the event handler below (`handleMessage`), which will simply give you a variable that you already have access to, i.e. `target.href`). Instead of using this circuitous route, why not just issue the action `window.location = target.href` directly here? What is the purpose of sending this message via the iframe, since nothing in the iframe needs to process it (and can't unless you inject a script into the iframe's document to deal with it)?

Sorry if I've misunderstood, as I haven't studied the full viewer code.

veloman-yunkan: > why not just issue the action `window.location = target.href` directly here?

This is what the previous code did and it was blocked because of sandboxing (since the action was performed in the context of the iframe which didn't have privileges of navigating the top level context). Now that action is initiated via a message - iframe sends a message and the top level receives and handles it in a different context where navigation is perfectly legal.

veloman-yunkan:
> @veloman-yunkan If I click on an external link it works fine, but if I come back in the history (with the back arrow on my browser) and click again on the external link, nothing happens. Tested with FF 113.0.1. Otherwise, seems to do the job properly.

Fixed

static/skin/viewer.js
```diff
@@ -262,10 +263,8 @@ function onClickEvent(e) {
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
     if ( isExternalUrl(target.href) ) {
-      target.setAttribute("target", "_top");
-      if ( viewerSettings.linkBlockingEnabled ) {
-        return blockLink(target);
-      }
+      target.setAttribute("href", blockLink(target.href));
```
mgautierfr, line 266: Why do we need to change the target's href ?

It will be the viewer which will change the window location.

mgautierfr (commented):
I like the way it is going.

But I wonder if we are still stick in the past here.

We have introduced the `/catch/external` at the time when we hadn't a viewer and we were injecting `block_external.js` in the content.

But now, we could use a `postMessage` all the time to tell to the viewer that user has clieked on a external link. The viewer can do all the external link check and displaying a confirmation or changing the window location itself without going to `/catch/external` endpoint.

static/skin/viewer.js
veloman-yunkan: This is required if we also want to block links opened in a new tab/window.

veloman-yunkan:
> I like the way it is going.
> 
> But I wonder if we are still stick in the past here.
> 
> We have introduced the `/catch/external` at the time when we hadn't a viewer and we were injecting `block_external.js` in the content.
> 
> But now, we could use a `postMessage` all the time to tell to the viewer that user has clieked on a external link. The viewer can do all the external link check and displaying a confirmation or changing the window location itself without going to `/catch/external` endpoint.

That makes sense, but if we want to block ctrl-clicked external links then we will need a catch page anyway.

static/skin/viewer.js
Jaifroid: OK. I understand that you can't navigate the iframe's contentWindow, but you don't need to that since you have access right here to `window`, which is the top-level window that is outside the iframe and is not sandboxed. It should be possible to navigate that window freely.

veloman-yunkan: No. Sandboxing applied to the iframe's document prevents it (and any script running in its name, which was the case here) from (directly) navigating the top-level window. Hence the need for indirection via `postMessage()`.

Jaifroid: @veloman-yunkan What you say would be true for a script running inside the iframe. However, `viewer.js` isn't running in the iframe, so it has access to and can navigate the top window, or open new windows/tabs. I did a quick check on this, and confirm it's possible to navigate the top window from `viewer.js`. I'll post the check in a separate comment, so as not to pollute the code view here.

Jaifroid:
FWIW, as seen in screen recording below, it is perfectly possible to navigate the top window from viewer.js. This is using the version currently published to dev.library.kiwix.org, but adding an explicit sandbox on the iframe, and editing viewer.js to open a new window AND navigate the top window, as shown in the code in the screen recording.

To explain what is shown:

1. I add the sandbox to the iframe, and I prove it's working by attempting to load a PDF in the iframe. It is blocked by sandbox;
2. I go back and re-add the sandbox (edits in dev tools are not saved to disk or cache);
3. I add code to viewer.js click handler to open the clicked link in a new tab/window, and code to navigate the top window. I add a breakpoint so we can see it is executing;
4. I click on the PDF. It opens in the new tab (not fully yet because of the breakpoint);
5. I show that all variables are accessible, then let the code finish running: it navigates the top window to the PDF, and finishes loading the PDF in the tab.

FYI I did exactly the same experiment using an external link (https://...) rather than a PDF, with exactly the same result. The top window navigated to the external resource (not inside the iframe of course).

@veloman-yunkan It's possible you were getting a different result in a previous version because the navigation might have been handled in some way from within the iframe, but I don't see why it's necessary to do so. It *would* be necessary if we were ever to have a _fully_ isolated iframe, i.e. one that doesn't `allow-same-origin`. But in that case, the code in this PR wouldn't work either, because it wouldn't be able to manipulate the iframe window at all. You would need to load an application to handle messaging inside the iframe (would have to be part of the ZIM contents).

sandbox_experiments

kelson42:
@veloman-yunkan Any update? Does the simplify approach of Jafroid works for you?

kelson42:
@veloman-yunkan ?

veloman-yunkan:
@Jaifroid Thanks for the proof. I think I never tried the proposed approach because I didn't even think that it would work. Such a solution was inconsistent with my mental model for a reasonable mechanism of imposing restrictions on sandboxed content, and I haven't looked into the spec to check if I even got it right (I am still not convinced that the demonstrated behaviour is correct though I understand that it is a result of my point of view on the problem).

veloman-yunkan:
Replaced by #959

PR #954, opened 2023-06-06
Add a new test, showing accents is not properly handle in search endpoint.

### LINKED ISSUE #953: [server] Search endpoint doesn't remove accents from the query.
Searching for pattern with a accent return no result, even if the same search without accent give results.

This can be see with the branch `search_remove_accent` (https://github.com/kiwix/libkiwix/compare/main...search_remove_accent).

Adding a `removeAccents(pattern)` fixes the issue about the searching part (we have results) but test still failing wrong encoding
when trying to render the search page (or the test is broken ?)

mgautierfr:
This doesn't fix #953  (openzim/libzim#804 does it), but it adds and fixes tests.

Depends of openzim/libzim#804

### DIFF
--- a/test/server_search.cpp
+++ b/test/server_search.cpp
@@ -2,6 +2,7 @@
 #define CPPHTTPLIB_ZLIB_SUPPORT 1
 #include "./httplib.h"
 #include "gtest/gtest.h"
+#include "../src/tools/stringTools.h"
 
 #define SERVER_PORT 8101
 #include "server_testing_tools.h"
@@ -716,7 +717,7 @@ struct TestData
 
   std::string getPattern() const
   {
-    return extractQueryValue("pattern");
+    return kiwix::urlDecode(extractQueryValue("pattern"), true);
   }
 
   std::string getLang() const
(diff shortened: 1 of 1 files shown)


test/server_search.cpp
```diff
@@ -1456,6 +1456,32 @@ TEST(ServerSearchTest, searchResults)
         { "▶", 40, false },
       }
     },
+
+    {
+      /* query */          "pattern=j%C3%A0zz"
```
veloman-yunkan, line 1462: Please add a comment explaining the purpose of this test point (so that maintainers don't have to resort to `git blame` to find it out).

veloman-yunkan (changes requested):

test/server_search.cpp
kelson42: @mgautierfr Can we fix and merge please?

mgautierfr: Done (commit directly modified, no `fixup!`)

veloman-yunkan (approved):

kelson42:
@mgautierfr @rgaudin What's going on with mavOS in CI?

rgaudin:
Answer on Slack

mgautierfr:
What was the answer on Slack ?
We have discussed an issue on dependency installation of the CI, but here is it a bug in the tests.

PR #963, opened 2023-07-01
Quasi-URI-encoded suggestion links

### LINKED ISSUE #958: URL encoding suggestion pathname's unintended side effect
Here's another tricky behavior.

In [this ZIM](https://dev.library.kiwix.org/viewer#freecodecamp.org_en_custom_2023-06), because it's a single page app, we have ZIM front entries that are stub HTML documents simply redirecting to the SPA's URL with appropriate fragment.

For instance, one content is available from:

-  [`index.html#english/python-for-everybody/comparing-and-sorting-tuples`](https://dev.library.kiwix.org/viewer#freecodecamp.org_en_custom_2023-06/index.html#english/python-for-everybody/comparing-and-sorting-tuples)

We thus have a front article at [`redirect/english/python-for-everybody/comparing-and-sorting-tuples`](https://dev.library.kiwix.org/viewer#freecodecamp.org_en_custom_2023-06/redirect/english/python-for-everybody/comparing-and-sorting-tuples).

This works fine with and without the viewer. The random feature works as expected, serving up the _redirects_.

The problem is with the Suggestions. As you can see, the suggestion engines returns correct results and those are displayed fine but click on any leads to a `404`.

Instead of arriving (search for “comp”) at the expected location, one arrives at `https://dev.library.kiwix.org/viewer#index.html`

As you can see, the bookName part is gone and one would expect the redirect to have gone to far. It's not the case.

It's not a problem with how the location is changed (https://github.com/kiwix/libkiwix/blob/main/static/skin/viewer.js#L50).

What happens is that the suggestion code urlencodes the result turning `redirect/english/python-for-everybody/comparing-and-sorting-tuples` into `redirect%2Fenglish%2Fpython-for-everybody%2Fcomparing-and-sorting-tuples`.

Now read carefully:

- Browser is fine with that encoded URL and properly retrieves the entry.
- Redirect is performed but the scope/source location is incorrect and we end-up elsewhere.

This is not linked to the viewer and the iframe. You can reproduce on `/content`:

- Go to [`/content/freecodecamp.org_en_custom_2023-06/`](https://dev.library.kiwix.org/content/freecodecamp.org_en_custom_2023-06/) ✅
- Go to [`/content/freecodecamp.org_en_custom_2023-06/redirect/english/python-for-everybody/comparing-and-sorting-tuples`](https://dev.library.kiwix.org/content/freecodecamp.org_en_custom_2023-06/redirect/english/python-for-everybody/comparing-and-sorting-tuples) ✅
- Go to [`/content/freecodecamp.org_en_custom_2023-06/redirect%2Fenglish%2Fpython-for-everybody%2Fcomparing-and-sorting-tuples`](https://dev.library.kiwix.org/content/freecodecamp.org_en_custom_2023-06/redirect%2Fenglish%2Fpython-for-everybody%2Fcomparing-and-sorting-tuples) ❌

Counter-intuitively, the browser does not translates the `%2F` into `/` and _moves_ there but instead it queries the server with the `%2F`, gets the response and renders it from it's `redirectXXX` location, not accounting those `%2F` as path segments.
That's **exactly** the purpose of URL-encoding and all browsers do the same 👍 

My question at this point would be: why are we URL-encoding this information? From what I understand, this is a _ZIM-Entry path_ that is sent to us via the Suggestion API and we just use it as source of the iframe (prefixed with root if any).
Need for URL-encoding would be to prevent characters from that ZIM entry path to be interpreted as special-meaning for being reserved characters.
While ZIM path are entirely free, ZIM creators already limit themselves and will do as long as ZIM are mainly run in a Web context.

- [Article Format](https://wiki.openzim.org/wiki/Article_Format)
- [URLs](https://wiki.openzim.org/wiki/ZIM_file_format#URLs)
- Reserved characters: `␣` `!` `"` `#` `$` `%` `&` `'` `(` `)` `*` `+` `,` `/` `:` `;` `=` `?` `@` `[` `]`

veloman-yunkan:
Fixes #958

Before this fix suggestion links were built out of fully URI-encoded book name and article path components despite the fact that this measure was taken against only a few dangerous symbols such as '#', '?', '"' and '\'.  However, URI-encoding the slash symbols in the path has some undesirable side-effects.

Henceforth only the problematic symbols are encoded in the article path component. The book name is still fully URI-encoded since I don't see any counter-arguments.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -54,6 +54,22 @@ function gotoRandomPage() {
   gotoUrl(`/random?content=${currentBook}`);
 }
 
+// URI-encodes only the specified special symbols (note, however, that '%' is
+// always considered a special symbol).
+function quasiUriEncode(s, specialSymbols) {
+  if ( specialSymbols.match(/[A-Za-z0-9]/) ) {
+    throw "Alphanumeric symbols cannot be special";
+  }
+
+  // Set's guarantee of iterating in insertion order ensures that
+  // all %s in s will be encoded first.
+  for ( const c of new Set('%' + specialSymbols) ) {
+    s = s.replaceAll(c, encodeURIComponent(c));
+  }
+
+  return s;
+}
+
 function performSearch() {
   const searchbox = document.getElementById('kiwixsearchbox');
   const q = encodeURIComponent(searchbox.value);
@@ -386,8 +402,13 @@ function setupSuggestions() {
           const uriEncodedBookName = encodeURIComponent(currentBook);
           let url;
           if (data.value.kind == "path") {
-            const path = encodeURIComponent(htmlDecode(data.value.path));
-            url = `/content/${uriEncodedBookName}/${path}`;
+            // The double quote and backslash symbols are included in the list
+            // of special symbols to URI-encode so that the resulting URL can
+            // be safely quoted inside a dynamically executed piece of
+            // Javascript code a few lines later.
+            const path = htmlDecode(data.value.path);
+            const quasiUriEncodedPath = quasiUriEncode(path, '#?"\\');
+            url = `/content/${uriEncodedBookName}/${quasiUriEncodedPath}`;
           } else {
             const pattern = encodeURIComponent(htmlDecode(data.value.value));
             url = `/search?content=${uriEncodedBookName}&pattern=${pattern}`;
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
Works as expected ; thanks 👍

mgautierfr (approved):

PR #967, opened 2023-07-11
Functions to read OPDS streams - languages and categories

### LINKED ISSUE #968: libkiwix should be able to parse catalog/v2/languages and catalog/v2/categories
This will give a common interface to get all languages and categories

juuz0:
Fixes #968 
Requirement for: https://github.com/kiwix/kiwix-desktop/pull/960

### DIFF
--- a/include/tools.h
+++ b/include/tools.h
@@ -28,6 +28,7 @@
 namespace kiwix {
 typedef std::pair<std::string, std::string> LangNameCodePair;
 typedef std::vector<LangNameCodePair> FeedLanguages;
+typedef std::vector<std::string> FeedCategories;
 
 /**
  * Return the current directory.
@@ -235,5 +236,13 @@ std::string beautifyFileSize(uint64_t number);
  * @return vector containing pairs of language code and their corresponding full language name. 
  */
 FeedLanguages readLanguagesFromFeed(const std::string& content);
+
+/**
+ * Load categories stored in an OPDS stream .
+ * 
+ * @param content the OPDS stream.
+ * @return vector containing category strings.
+ */
+FeedCategories readCategoriesFromFeed(const std::string& content);
 }
 #endif // KIWIX_TOOLS_H
--- a/include/tools.h
+++ b/include/tools.h
@@ -244,5 +244,13 @@ FeedLanguages readLanguagesFromFeed(const std::string& content);
  * @return vector containing category strings.
  */
 FeedCategories readCategoriesFromFeed(const std::string& content);
+
+/**
+ * Retrieve the full language name associated with a given ISO 639-3 language code.
+ * 
+ * @param lang ISO 639-3 language code.
+ * @return full language name.
+ */
+std::string getLanguageSelfName(const std::string& lang);
 }
 #endif // KIWIX_TOOLS_H
--- a/include/tools.h
+++ b/include/tools.h
@@ -26,6 +26,8 @@
 #include <cstdint>
 
 namespace kiwix {
+typedef std::pair<std::string, std::string> LangNameCodePair;
+typedef std::vector<LangNameCodePair> FeedLanguages;
 
 /**
  * Return the current directory.
@@ -226,5 +228,12 @@ std::string getBestPublicIp();
  */
 std::string beautifyFileSize(uint64_t number);
 
+/**
+ * Load languages stored in an OPDS stream.
+ * 
+ * @param content the OPDS stream.
+ * @return vector containing pairs of language code and their corresponding full language name. 
+ */
+FeedLanguages readLanguagesFromFeed(const std::string& content);
 }
 #endif // KIWIX_TOOLS_H
--- a/src/tools/languageTools.cpp
+++ b/src/tools/languageTools.cpp
@@ -1,0 +1,74 @@
+#include "tools.h"
+#include "stringTools.h"
+#include <mutex>
+
+namespace kiwix
+{
+
+namespace
+{
+
+// These mappings are not provided by the ICU library, any such mappings can be manually added here
+std::map<std::string, std::string> iso639_3 = {
+    {"atj", "atikamekw"},
+    {"azb", "آذربایجان دیلی"},
+    {"bcl", "central bikol"},
+    {"bgs", "tagabawa"},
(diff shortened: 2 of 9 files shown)


kelson42:
@juuz0 Ready to review? Code coverage does not pass, new functions should be tested.

mgautierfr (changes requested):
I'm not sure those functions are put at the right place.
The `Manager` has been created to manage a library. The `Library` being a "simple" container of `Book`s.

But here, `parseLanguages` and `parseCategories` do not manage everything. They are "just" parsing a feed.
This could be simple free function instead of method of Manager. And if we want them to be part of `Manager` (to make it something that "parse content"), it should be static method to allow the user to call them without creating a `Manager` instance first.

---

If fact, they are the equivalent of `Library::getBooksLanguages` or `Library::getBooksCategories`, but for a abstract "remote Library". Maybe there is a new class (potentially sharing the same common interface than `Library`) to create to correctly handle remote library.

kelson42:
@juuz0 Can you please create the corresponding issue to explain what is the problem you face with libkiwix?

juuz0:
@kelson42 https://github.com/kiwix/libkiwix/issues/968 created one

mgautierfr (changes requested):

juuz0:
> If fact, they are the equivalent of Library::getBooksLanguages or Library::getBooksCategories, but for a abstract "remote Library". Maybe there is a new class (potentially sharing the same common interface than Library) to create to correctly handle remote library.

Having a remote library class felt like it should handle the download requests to `/catalog/v2/languages` too.
That's being done in kiwix-desktop right now (similar to how a book catalog is downloaded in kiwix desktop but parsed in libkiwix).

I put these functions in `kiwix/tools.h` for now - since these are tools for just parsing (no download requests etc).
But if we still want to go with the Remote Library class, let me know!

include/library_dumper.h
```diff
@@ -76,6 +76,14 @@ class LibraryDumper
    */
   kainjow::mustache::list getLanguageData() const;
 
+  /**
+   * Get the full language name
+   * 
+   * @param lang 
+   * @return std::string 
+   */
+  static std::string getLanguageSelfName(const std::string& lang);
```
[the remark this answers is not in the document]
juuz0: Did in https://github.com/kiwix/libkiwix/pull/967/commits/c4b2a6d1744b22e762c523f7b2f0777c7d6928e3
Removed the commit making getLanguageSelfName static.
Trying not to do so much history rewriting, instead wanted to use fixup commit here - but this looked necessary.

juuz0:
Rebased

kelson42:
@juuz0 @mgautierfr Please talk together, it's not good to see this PR somehow stuck since a week.

juuz0:
@mgautierfr Review needed!

mgautierfr (changes requested):
Few small fixes.
And tests are failing.

include/tools.h
```diff
@@ -236,5 +236,13 @@ FeedLanguages readLanguagesFromFeed(const std::string& content);
  * @return FeedCategories 
  */
 FeedCategories readCategoriesFromFeed(const std::string& content);
+
+/**
+ * Get the full language name
+ * 
+ * @param lang 
+ * @return std::string 
+ */
```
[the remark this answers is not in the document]
juuz0: Fixed in commit, no fixup

juuz0:
> And tests are failing.

@mgautierfr 
This is a CI issue I think, test `server_search` is failing but you fixed that here https://github.com/kiwix/libkiwix/commit/556b94daae55bdc267664688306ecc67466b9ddf
I assume the CI is not using an updated libzim?

juuz0:
> And tests are failing.

Fixed. On my PC, the value was `"und"` for `getLanguageSelfName("")`. On the CI, it was `""`
I made it compulsory to return `"und"` for empty string.

juuz0:
Rebased after the recent merge

src/tools/languageTools.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+} // unnamed namespace
+
+std::string getLanguageSelfName(const std::string& lang)
+{
+    std::call_once(fillLanguagesFlag, fillLanguagesMap);
+    if (lang == "")
+        return "und";
```
mgautierfr, line 65: It should be better to add this directly in the iso639_3 instead of adding a specific case.

mgautierfr (changes requested):
> Fixed. On my PC, the value was "und" for getLanguageSelfName(""). On the CI, it was ""
I made it compulsory to return "und" for empty string.

I'm a bit disturbed by this.
Why `getLanguageSelfName` return to different things. This means that ICU return different things for `icu::Local::getISOLanguages`. We must understand why.
From where is "und" coming on your PC ?

juuz0:
> Why getLanguageSelfName return to different things. This means that ICU return different things for icu::Local::getISOLanguages. We must understand why.

@mgautierfr 
I want to mention that the CI failed only for ubuntu focal earlier (which uses ICU 66). 

After some investigation, I found that this happens when a `icu::Locale` is created using the `und` languageCode.
Such locale will return 
* `""` for `getISO3Language()`
* `und` for `getDisplayName()`

I suspect the ubuntu-focal behaviour was unwanted (see ticket) and fixed in a later ICU version (ICU 69.1. ubuntu jammy uses 70.1)
https://unicode-org.atlassian.net/browse/ICU-21406

I think we should keep the mapping so it stays consistent across all 3 ubuntu builds

src/tools/languageTools.cpp
juuz0: Do you think Undetermined is a better name than "und"? It's more meaningful and will look better in kiwix-serve
https://iso639-3.sil.org/code/und

mgautierfr:
> I think we should keep the mapping so it stays consistent across all 3 ubuntu builds

ok. I agree. And "Undetermined" is indeed better.

src/tools/languageTools.cpp
juuz0: Done in the commit

mgautierfr (approved):

PR #971, opened 2023-07-19
Make beautifyFileSize public

juuz0:
Requirement for https://github.com/kiwix/kiwix-desktop/pull/946

### DIFF
--- a/include/tools.h
+++ b/include/tools.h
@@ -216,5 +216,14 @@ std::map<std::string, std::string> getNetworkInterfaces();
  */
 std::string getBestPublicIp();
 
+/** Converts file size to human readable format.
+ * 
+ * This function will convert a number to its equivalent size using units.
+ * 
+ * @param number file size in bytes.
+ * @return a human-readable string representation of the size, e.g., "2.3 KB", "1.8 MB", "5.2 GB".
+ */
+std::string beautifyFileSize(uint64_t number);
+
 }
 #endif // KIWIX_TOOLS_H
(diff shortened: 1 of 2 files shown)


kelson42:
@juuz0 Not sure this function is documented properly (in readtjedocs), please add the necessary documentation.

mgautierfr:
> Not sure this function is documented properly (in readtjedocs), please add the necessary documentation.

It will be documented this way : https://libkiwix--971.org.readthedocs.build/en/971/api/function_tools_8h_1a18e272a0e35c09646b78f566e92ac292.html

kelson42:
@mgautierfr actually not bad. LGTM if the number given is not expecting to be in a special unity... because this is not written in the doc.

mgautierfr:
The name of the function is `beautifyFileSize` so yes, we expect a size and so the unit is `B`/`KB`/`MB`/`GB`

kelson42:
@mgautierfr If the result is given with a unity, then the unity of the input expect should be specified in the documentaiton. Here I guess Bytes.

juuz0:
@kelson42 updated

include/tools.h
```diff
@@ -216,5 +216,14 @@ std::map<std::string, std::string> getNetworkInterfaces();
  */
 std::string getBestPublicIp();
 
+/** Converts file size to human readable format
+ * 
+ * This function will convert a number to its equivalent size using units.
+ * 
+ * @param file size in bytes
```
mgautierfr, line 223: The first word after `@param` must be the parameter name. So:
```
@param number file size in bytes.
```

And finish the sentences (even param and return descriptions) with `.`

mgautierfr (approved):

PR #978, opened 2023-07-25
Add missing include.

mgautierfr:
`uint64_t` type (used by `beautifyFileSize`) need to be declared.


mgautierfr:
Should fix broken compilation on armv6 and aarch64 (https://github.com/kiwix/kiwix-build/actions/runs/5651619394/job/15310001245)

Issue #877, opened 2023-01-30
gtest C++14 requirement (move to C++17)


xvitaly:
Google Test since version 1.13.0 requires C++14:

```
[4/69] g++ -Itest/tagParsing.p -Iinclude -I../include -I/usr/include/p11-kit-1 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c++11 -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -DGTEST_HAS_PTHREAD=1 -pthread -MD -MQ test/tagParsing.p/tagParsing.cpp.o -MF test/tagParsing.p/tagParsing.cpp.o.d -o test/tagParsing.p/tagParsing.cpp.o -c ../test/tagParsing.cpp
FAILED: test/tagParsing.p/tagParsing.cpp.o 
[...]

PR #983, opened 2023-07-26
Multiple CI/CD fixes

### LINKED ISSUE #982: Multiple problems around CI/CD
* 12.1.0 as not been published on PPA
* Not using the release trigger
* Debian packages are not tested in the CI (and probably don't pass)
* Unable to compile anyway on Debian unstable because of 
* CI not using latest base image
* ....

kelson42:
Fixes #982

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
       HOME: /home/runner
     runs-on: ubuntu-20.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:37"
+      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:38"
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


kelson42:
CI broken by #877 for "debian-unstable"

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       HOME: /home/runner
     runs-on: ubuntu-20.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:37"
+      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:38"
```
mgautierfr, line 86: Why not using `latest` instead of fixed number.
`kiwix-build_ci` image is something we maintain, it will not change if we don't want. And if we publish a new version, it is because we want CIs to use it.

kelson42: I'm not in favour to use unpinned version of any dependence. This is a general rule.

PR #991, opened 2023-08-23
Use std::shared_ptr instead of raw pointer.

### LINKED ISSUE #988: Shared pointer should be used where appropriate
See https://github.com/kiwix/kiwix-android/pull/3467

mgautierfr:
Using raw pointer is problematic as nothing prevent a user to create
a "consumer" on a temporary object, delete the object and use the consumer.

Using a shared_ptr ensure us to the object is keep alive while a consumer
is using it.
But it prevent us to create plain instance on the stack.

Fix #988


include/html_dumper.h
```diff
@@ -33,7 +33,7 @@ namespace kiwix
 class HTMLDumper : public LibraryDumper
 {
  public:
-  HTMLDumper(const Library* library, const NameMapper* NameMapper);
+  HTMLDumper(ConstLibraryPtr library, std::shared_ptr<const NameMapper> NameMapper);
```
veloman-yunkan, line 36: IMHO, `HTMLDumper` and `OPDSDumper` could be replaced with free functions - they were implemented as classes for reasons unrelated to the goals and principles of object oriented design. Now, having them as classes enables creating the dumper at one place and using it later but do we really need that functionality? Isn't it better to require that they are used as if they were free functions (i.e use and discard immediately after construction while ensuring that the `Library` and `NameMapper` objects stay unchanged during that time)?

veloman-yunkan (commented):
> Using raw pointer is problematic as nothing prevent a user to create
> a "consumer" on a temporary object, delete the object and use the consumer.
> 
> Using a shared_ptr ensure us to the object is keep alive while a consumer
> is using it.

This is not 100% true. An `std::shared_ptr` can be created from a regular pointer using a custom deallocator. Thus changing the API only makes the requirements about the object lifespan more explicit but doesn't actually bring any guarantees with it. What it does is the programmer is relieved from explicit management of object lifetimes, however at the expense of some risk of memory leaks (since, unlike garbage collected languages, C++-with-exclusive-usage-of-std::shared_ptrs doesn't automatically take care of reference cycles).

IMHO, the current justification for this relatively big change is not compelling enough.

mgautierfr:
> This is not 100% true. An std::shared_ptr can be created from a regular pointer using a custom deallocator. Thus changing the API only makes the requirements about the object lifespan more explicit but doesn't actually bring any guarantees with it.

Indeed. But moving the constructor private and allowing the creation of the library only through a shared_ptr limits a lot this possibility. It is still possible to create a shared_ptr with a custom deallocator from `shared_ptr::get()`, but here I would say that it is the inherent risk of using c++ and if user does this, it is on its fault, not us.

> What it does is the programmer is relieved from explicit management of object lifetimes, however of at the expense of some risk of memory leaks (since unlike garbage collected languages C++-with-exclusive-usage-of-std::shared_ptrs doesn't automatically take care of reference cycles).

To have a reference cycles, we need to have object stored in the library which have reference to the library itself. I don't think it is the case and we don't allow user to put custom object in the library.

veloman-yunkan:
> > What it does is the programmer is relieved from explicit management of object lifetimes, however at the expense of some risk of memory leaks (since, unlike garbage collected languages, C++-with-exclusive-usage-of-std::shared_ptrs doesn't automatically take care of reference cycles).
> 
> To have a reference cycles, we need to have object stored in the library which have reference to the library itself. I don't think it is the case and we don't allow user to put custom object in the library.

That's correct and I never implied that it wasn't. My point was that `std::shared_ptr` is not a magic bullet, so it doesn't always make sense to spend effort on switching to it.

mgautierfr:
New version of the PR.

Mostly the same strategy, but as agreed in https://github.com/kiwix/libkiwix/issues/988#issuecomment-1704307671:
- Various dumpers have been moved in private API (None of ours project was actually using it, so no alternative provided)
- Use shared_ptr for the server (as previous version of the PR)
- For `SearchRendered`, we keep the class publicly accessible but we change the API to not store NameMapper nor Library.

include/search_renderer.h
```diff
@@ -90,24 +72,32 @@ class SearchRenderer
...
+   * @param mapper The `NameMapper` to use to do the rendering.
+   * @param library The `Library` to use to look up book details for search results.
+                    May be nullptr. In this case, bookName is not set in the rendered string.
+   * @return The xml string
    */
-  std::string getXml();
+  std::string getXml(const NameMapper* mapper, const Library* library);
 
+ protected: // function
+  std::string renderTemplate(const std::string& tmpl_str, const NameMapper* mapper, const Library *library);
```
veloman-yunkan, line 96: It's better to pass the name mapper and library objects by reference rather than by pointer thus meaning two things:

1. both are mandatory (a pointer can be NULL)
2. the called function doesn't store the address of those objects

mgautierfr: I have changed `const NameMapper*` to `const NameMapper&`.
But as library can be nullptr, I have keep a raw pointer.

mgautierfr:
> The second commit (the one titled "Use std::shared_ptr instead of raw pointer.") is too big. I would have made that change in several smaller steps.

Done. The commit is split in several ones.

One noticeable change is that now `InternalServer` do NOT keep a `const` (shared) pointer on the NameMapper (previous version of the PR was).
I feel bad to keeping a const pointer on something that can change (UpdatableNameMapper) even if we (server) don't change it itself.

kelson42:
@veloman-yunkan any feedback?

veloman-yunkan (commented):
LGTM with a minor comment (mainly intended to serve as a proof that I looked at the changes) -
the description of the commit titled "Make the Server keep a shared_ptr instead of a raw Library pointer." was obviously copy-pasted but not fully updated.

mgautierfr:
Commit message rewritten. Time to merge !

PR #997, opened 2023-09-09
Cookieless user language control

### LINKED ISSUE #995: Can we remove cookie management from the backend?
As far as I understand, libkiwix itself uses two cookies for similar features:

- `userlang` which holds a 2 letter language code to record requested UI language.
- `filters` which holds the latest query string for the filters

Those sole purpose is to support the JS Viewer by recording user preferences.

I would thus like to **suggest to move cookie management to JS/Viewer** because:
- The Viewer being entirely independent of the backend it makes sense to let it handle its user preference storage.
- Cookies are per-user and thus prevent us from caching properly in our varnish frontend (library.kiwix.org)

In order to cache kiwix-serve data, we have to remove the cookie:

- from the requests: kiwix-serve backend thus doesn't see the cookie and sends the initial one from Accept-Language
- from the response: causing https://github.com/kiwix/kiwix-tools/issues/632 because UI is set solely from the cookie (and not URL param/fragment contrary to filters). Can be reproduced by disabling cookies in browser.

I don't know about the implementation but given the description in #849, I don't see how any of this requires the use of the backend. The reader part of the viewer is obviously JS-only and i18n handled in i18n.js and the per-lang JSON files. I'm not sure about search results but since those include english text even on a french UI, I'd say it's not i18n ready. Should it be, it could still be called by the reader passing the language.

So, could we consider removing this cookie code from the backend and handling it all [from JS](https://developer.mozilla.org/en-US/docs/Web/API/document/cookie)?

veloman-yunkan:
Fixes #995

A known issue with this fix is that the relatively expensive handling of the "Accept-Language" header is now being performed for all requests (unless the user language is explicitly specified in the URL query: `?userlang=...`). I will fix it after this PR clears all other hurdles of the review process.

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -754,7 +754,8 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
   const kainjow::mustache::object data{
     {"enable_toolbar", m_withTaskbar ? "true" : "false" },
     {"enable_link_blocking", m_blockExternalLinks ? "true" : "false" },
-    {"enable_library_button", m_withLibraryButton ? "true" : "false" }
+    {"enable_library_button", m_withLibraryButton ? "true" : "false" },
+    {"default_user_language", request.get_user_language() }
   };
   return ContentResponse::build(*this, RESOURCE::templates::viewer_settings_js, data, "application/javascript; charset=utf-8");
 }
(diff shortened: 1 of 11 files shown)


rgaudin (approved):
LGTM ; from a user perspective, the issue is fixed: even behind a varnish cache that removes Cookie from and to user, the language switch works.

mgautierfr (approved):

veloman-yunkan:
I think that this PR introduced a small problem - now the text on error pages will be in the language detected from the value of the "Accept-Language" header (since the information about the user-selected language is no longer sent to the server).

kelson42:
@veloman-yunkan can you please open a new ticket? maybe provide a fix?

Issue #998, opened 2023-09-12
ABI break in 12.1.0


legoktm:
There seems to have been an ABI break between 12.0.0 and 12.1.0, when running `kiwix-desktop --version` (compiled against 12.0.0) against a newly built 12.1.0, seeing:
```
kiwix-desktop: symbol lookup error: kiwix-desktop: undefined symbol: _ZN5kiwix10Downloader14getDownloadIdsB5cxx11Ev
```

which demangles to `kiwix::Downloader::getDownloadIds[abi:cxx11]()`

legoktm:
I reverted the part of https://github.com/kiwix/libkiwix/commit/18b7b5f277a7734c024f7b36b5ad9e0584740b1c that touched getDownloadIds() (i.e. removing const) and it fixed the issue.

legoktm:
With the `const` added, the symbol name becomes `_ZNK5kiwix10Downloader14getDownloadIdsB5cxx11Ev` - specifically it starts with `_ZNK5` instead of `ZN5`.

legoktm:
The patch I've drafted right now (but not uploaded to Debian) is to just remove the `const` from this one method - https://salsa.debian.org/debian/libkiwix/-/blob/master/debian/patches/0002-Revert-ABI-breakage-in-kiwix-Downloader-getDownloadI.patch

Does that seem reasonable? I can also submit it as a PR too.

legoktm:
OK, I've submitted #999. If that can get a thumbs up I'll apply it to Debian.

mgautierfr:
A version 12.1.1 of libkiwix has just been published with the api/abi break removed (equivalent to 12.1.0 - breaks)
A version 13.0.0 of libkiwix has also just been published which is equal to 12.1.0 (equivalent to 12.1.0 + breaks)

You should not use version 12.1.0 for now one. Use either 12.1.1 (without api break) or (13.0.0 with api break)

(Api/break break compared to 12.0.0)

legoktm:
Thank you! I'm updating Debian to 12.1.1 right now and will get 13.0.0 rolling shortly. :)

PR #999, opened 2023-09-14
Revert ABI breakage in kiwix::Downloader::getDownloadIds()

legoktm:
Changing it to be `const` is an ABI change since the symbol changes from `_ZN5...` to `_ZNK5...` (addition of "K").

The other changes made in 18b7b5f27 are probably fine to keep since they don't appear to be used from what I can tell.

Fixes https://github.com/kiwix/libkiwix/issues/998.

### DIFF
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -204,7 +204,7 @@ class Downloader
   /**
    * Get the ids of the managed downloads.
    */
-  std::vector<std::string> getDownloadIds() const;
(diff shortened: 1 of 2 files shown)


legoktm:
The only thing I'm not sure about is if there's a plan to use this function in other const stuff and whether this causes issues with that.

mgautierfr:
LGTM, Thanks

Rebased on branch version_12.1.1

Issue #938, opened 2023-04-20
Ability to toggle magnet support?


rgaudin:
Our download modal offers magnet links if an URL similar to ZIM download URL with `.magnet` suffix exists

https://github.com/kiwix/libkiwix/blob/1a92d4a0b5b6be51fdaff719ff0152a807ba71cc/static/skin/index.js#L200-L214

It is understood that the sole* use case for this is library.kiwix.org but [Our magnet links are broken](https://github.com/kiwix/container-images/issues/242), and we can't really fix them.

Additionally, there aren't much client for those magnets

I didn't look for torrent client but [this list on WP](https://en.wikipedia.org/wiki/Magnet_URI_scheme#Clients) is not reassuring. For our webseed-magnet to work, we'd need the client to support HTTP `xr` or `as` (and `tr` for regular transfers). In that list, it means:
- gtk-gnutella: windows, linux
- limewire: dead
- shareaza: windows only
- tixati: proprietary, windows/linux
- vuze: properietary (and heavy)

To me, it sounds like most users would use another client and get a never-starting download… leading them to think it's a peer issue (they have no idea about the webseed) thus leading them to the direct link. Create a vicious circle with less BT users, reinforcing itself.

mgautierfr:
Another solution would be to remove magnet links from `library.kiwix.org` ?

rgaudin:
Sure but It's automatically provided by mirrorbrain as `.magnet` and `?magnet`. It doesn't look like it can be disabled ([doc](https://mirrorbrain-docs.readthedocs.io/en/latest/configuration.html#magnet-links)).

We could hack a hack on the reverse proxy to return a blank payload on every `.magnet$` request though…

rgaudin:
Let's add a hack in libkiwix's index.js for now to fix the magnet.

kelson42:
@rgaudin I would like to revive this ticket as it seems relatively important as well as relatively easy to fix. I understand that we are mostly "stuck" here because of the deprecation of `poeml/mirrorbrain` and our lack of strategy and resources to deal with this challenge.

I believe that this bug could be fixed both by patching Mirrorbrain OR by patching/hacking our kiwix-serve Javascript. Both are OK to me and I believe we should do it ASAP (this is why I have put this ticket in the `13.0.0` milestone at the last moment).

[...]

PR #1001, opened 2023-10-02
Hack for mirrobrain magnet URIs

rgaudin:
As per #938, in regard to https://github.com/kiwix/container-images/issues/242, this adds some processing to the creation of the `magnet:` link for downloads.

It mostly fixes the parameter names/values of the magnet string but also fetches the list of mirrored URLs (using metalink) to add them as alternative webseeds.

Tested with Vuze/Azureus on Windows as well as Tixati on Windows (Tixati is smart enough to retrieve metalink from the webseed and thus didn't even need the extra `ws=`).

Please note that while Transmission understands it's multiple webseeds, there's no support for fetching the torrent/DHT and so the download never starts. Don't be fooled 😉! I guess it's similar in most other torrent client (most of them dont support webseed magnets).

**Suggestion**: `target="_blank"` for magnet is really annoying (leaves a blank tab open). I think it should be removed.

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -197,6 +197,58 @@
         }
     }
 
+    function makeURLSearchString(params, keysToURIEncode) {
+        let output = '';
+        for (const [key, value] of params.entries()) {
+            let finalValue = (keysToURIEncode.indexOf(key) >= 0) ? encodeURIComponent(value) : value;
+            output += `&${key}=${finalValue}`;
+        }
+        return output;
+    }
+
+    /* hack for library.kiwix.org magnet links (created by MirrorBrain)
+        See https://github.com/kiwix/container-images/issues/242 */
+    async function getFixedMirrorbrainMagnet(magnetLink) {
+        // parse as query parameters
+        const params = new URLSearchParams(
+            magnetLink.replaceAll('&amp;', '&').replace(/^magnet:/, ''));
+
+        const zimUrl = params.get('as'); // as= is fallback URL
+        // download metalink to build list of mirrored URLs
+        let mirrorUrls = [];
+
+        const metalink = await fetch(`${zimUrl}.meta4`).then(response => {
+            return response.ok ? response.text() : '';
+        }).catch((_error) => '');
+        if (metalink) {
+            try {
+                const parser = new DOMParser();
+                const doc = parser.parseFromString(metalink, "application/xml");
+                doc.querySelectorAll("url").forEach((node) => {
+                    if (node.hasAttribute("priority")) { // ensures its a mirror link
+                        mirrorUrls.push(node.innerHTML);
+                    }
+                });
+            } catch (err) {
+                // not a big deal, magnet will only contain primary URL
+                console.debug(`Failed to parse mirror links for ${zimUrl}`);
+            }
+        }
+
+        // set webseed (ws=) URL to primary download URL (redirects to mirror)
+        params.set('ws', zimUrl);
+        // if we got metalink mirror URLs, append them all
+        if (mirrorUrls) {
+            mirrorUrls.forEach((url) => {
+                params.append('ws', url);
+            });
+        }
+
+        params.set('xs', `${zimUrl}.torrent`);  // adding xs= to point to torrent URL
+
+        return 'magnet:?' + makeURLSearchString(params, ['ws', 'as', 'dn', 'xs', 'tr']);
+    }
+
     async function getMagnetLink(downloadLink) {
         const magnetUrl = downloadLink + '.magnet';
         const controller = new AbortController();
@@ -204,6 +256,9 @@
         const magnetLink = await fetch(magnetUrl, { signal: controller.signal }).then(response => {
             return response.ok ? response.text() : '';
         }).catch((_error) => '');
+        if (magnetLink) {
+            return await getFixedMirrorbrainMagnet(magnetLink);
+        }
         return magnetLink;
     }
 
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -63,7 +63,7 @@ const ResourceCollection resources200Compressible{
   { DYNAMIC_CONTENT, "/ROOT%23%3F/skin/index.css" },
   { STATIC_CONTENT,  "/ROOT%23%3F/skin/index.css?cacheid=e4d76d16" },
   { DYNAMIC_CONTENT, "/ROOT%23%3F/skin/index.js" },
-  { STATIC_CONTENT,  "/ROOT%23%3F/skin/index.js?cacheid=e1b1ae55" },
+  { STATIC_CONTENT,  "/ROOT%23%3F/skin/index.js?cacheid=ce19da2a" },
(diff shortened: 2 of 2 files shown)


static/skin/index.js
```diff
@@ -197,13 +197,63 @@
...
+                doc.querySelectorAll("url").forEach((node) => {
+                    if (node.hasAttribute("priority")) { // ensures its a mirror link
+                        mirrorUrls.push(node.innerHTML);
+                    }
+                });
+            }
+        } catch (err) {
+            // not a big deal, magnet will only contain primary URL
+            console.debug(`Failed to retrieve mirror links for ${zimUrl}`);
+        }
```
veloman-yunkan, line 226: Why is the `try/catch` block needed? If it is a protection against malformed metalink XML then its scope better be narrowed down (and the error message changed accordingly).

static/skin/index.js
```diff
@@ -197,13 +197,63 @@
...
+            });
+        }
+
+        params.set('xs', `${zimUrl}.torrent`);  // adding xs= to point to torrent URL
+
+        magnetLink = 'magnet:?';
+        // we don't want params.ToString()'s urlencoding
+        for (const [key, value] of params.entries()) {
+             magnetLink += `&${key}=${value}`;
+        }
```
veloman-yunkan, line 243: 1. This produces a link of the form `magnet:?&...` - note the ampersand symbol immediately following the question mark symbol. Is that OK?

2. In our unit tests we have a ZIM file with an ampersand symbol in it, which suggests that such ZIM files occur in the wild. Won't URLs to such files pose a problem in the absence of URL-encoding?

static/skin/index.js
rgaudin: Exactly, since it's not required per se (magnet will work with a single webseed), I don't want something unexpected to break the magnet link. Could be network-related or parsing related. AFAIK, there's no UI feedback on errors

static/skin/index.js
rgaudin: > This produces a link of the form magnet:?&... - note the ampersand symbol immediately following the question mark symbol. Is that OK?

Yes, it is. We could remove the first `&` but that's extra processing for something that's already a valid URI

> In our unit tests we have a ZIM file with an ampersand symbol in it, which suggests that such ZIM files occur in the wild. Won't URLs to such files pose a problem in the absence of URL-encoding?

Yes, actually the spec says it should be urlencoded. I had it previously and removed it when I encountered an issue (prob. unrelated). Let me fix that back

static/skin/index.js
veloman-yunkan: > Could be network-related

Won't network errors be handled by the `catch()` of the `fetch()`?

rgaudin: Right. I guess it's only about parsing. Do you think we should drop it?

veloman-yunkan: No, the suggestion was to enclose in `try{}` only the code the actually needs it (and adjust the error message).

rgaudin: Got it

static/skin/index.js
rgaudin: Although spec is not clear about URLencoding some of the params, I chose to URL encode all now and tested that al of them are properly decoded in my test clients.

Please note that we dont publish any ZIM file with url-encodable characters ATM.

static/skin/index.js
```diff
@@ -207,34 +207,42 @@
...
+        params.set('dn', encodeURIComponent(params.get('dn')));
+        params.set('xs', encodeURIComponent(`${zimUrl}.torrent`));  // adding xs= to point to torrent URL
+        const trackers = params.getAll('tr');
+        params.delete('tr');
+        trackers.forEach((tracker) => {
+            params.append('tr', encodeURIComponent(tracker));
+        });
 
         magnetLink = 'magnet:?';
         // we don't want params.ToString()'s urlencoding
```
veloman-yunkan, line 248: This comment is now misleading

static/skin/index.js
```diff
@@ -207,34 +207,42 @@
...
 
-        params.set('xs', `${zimUrl}.torrent`);  // adding xs= to point to torrent URL
+        params.set('as', encodeURIComponent(zimUrl));
+        params.set('dn', encodeURIComponent(params.get('dn')));
+        params.set('xs', encodeURIComponent(`${zimUrl}.torrent`));  // adding xs= to point to torrent URL
+        const trackers = params.getAll('tr');
+        params.delete('tr');
+        trackers.forEach((tracker) => {
+            params.append('tr', encodeURIComponent(tracker));
+        });
```
veloman-yunkan, line 245: Isn't it better to let `URLSearchParams.toString()` do the URI-encoding?

static/skin/index.js
rgaudin: No it's not. We still don't want [URLSearchParams.toString](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString) because it URL encodes everything like the `:` in URN and all. Result is thus very different from what we produce here and most importantly it's rejected by torrent clients.

static/skin/index.js
rgaudin: No, see comment above

veloman-yunkan: Then why don't you instead URI-encode inside the loop where you build the final link?

static/skin/index.js
veloman-yunkan: `encodeURIComponent()` encodes everything, too. Did you mean to use `encodeURI()`?

rgaudin: No, precisely because we're applying it to URLs only.

static/skin/index.js
rgaudin: Because only URLs containing parts needs `encodeURIComponent` while others would not support it. We'd have to toggle with it and nothing or `encodeURI`. In current implementation, it's very clear what requires to be encoded IMO

veloman-yunkan: IMHO, it will be clearer if there is a list (array) of keys requiring URI-encoding and a single call to `encodeURIComponent()` in a loop.

veloman-yunkan: I would take that approach by introducing a reusable function `makeURLSearchString(params, keysToURIEncode)`

veloman-yunkan (approved):

rgaudin:
@veloman-yunkan thank you ; is any of the commit history worth keeping? I think it can all be squashed into one. WDYT?

veloman-yunkan:
> @veloman-yunkan thank you ; is any of the commit history worth keeping? I think it can all be squashed into one. WDYT?

@rgaudin I concur with you on this. In fact I wanted to suggest that in my approval but didn't want to appear too finicky. :)

rgaudin:
Done ; it's rebased off `main` as well. Thank you for your time.

kelson42 (approved):

kelson42:
@veloman-yunkan @rgaudin @mgautierfr Unfortunately I can not merge because the CI is red and the error seems worrying.

rgaudin:
@mgautierfr thanks for the explanation. @kelson42 it's now green

PR #1003, opened 2023-10-13
Do not use `[[nodiscard]]` attribute on compiler not supporting it.

mgautierfr:
On aarch64, we use gcc version 6.3.0 which doesn't support the `[[nodiscard]]` attribute
(see https://en.cppreference.com/w/cpp/compiler_support/17)

So don't set the attribute if the attribute is not present.

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -186,6 +186,17 @@ class MultiKeyCache;
 using LibraryPtr = std::shared_ptr<Library>;
 using ConstLibraryPtr = std::shared_ptr<const Library>;
 
+
+// Some compiler we use don't have [[nodiscard]] attribute.
+// We don't want to declare `create` with it in this case.
+#define LIBKIWIX_NODISCARD
+#if defined __has_cpp_attribute
+  #if __has_cpp_attribute (nodiscard)
+    #undef LIBKIWIX_NODISCARD
+    #define LIBKIWIX_NODISCARD [[nodiscard]]
+  #endif
+#endif
+
 /**
  * A Library store several books.
  */
@@ -201,7 +212,7 @@ class Library: public std::enable_shared_from_this<Library>
   Library();
 
  public:
-  [[nodiscard]] static LibraryPtr create() {
+  LIBKIWIX_NODISCARD static LibraryPtr create() {
     return LibraryPtr(new Library());
   }
   ~Library();
@@ -407,6 +418,10 @@ private: //data
   std::vector<kiwix::Bookmark> m_bookmarks;
   std::unique_ptr<Xapian::WritableDatabase> m_bookDB;
 };
+
+// We don't need it anymore and we don't want to polute any other potential usage
+// of `LIBKIWIX_NODISCARD` token.
+#undef LIBKIWIX_NODISCARD
 
 
 }


include/library.h
```diff
@@ -34,6 +34,14 @@
 
 #define KIWIX_LIBRARY_VERSION "20110515"
 
+#define NODISCARD
+#if defined __has_cpp_attribute
+  #if __has_attribute (nodiscard)
```
veloman-yunkan, line 39: Shouldn't this be `__has_cpp_attribute` for better portability (and consistency with the above check)?

veloman-yunkan (commented):
The `NODISCARD` macro is introduced under the implicit assumption that `libkiwix` is the only project having run into the portability issue of direct usage of the `[[nodiscard]]` C++ attribute (I mean what if e.g. xapian defines such a macro too?)

I see two reasonable options out of the current situation:

1. Drop the (only current usage of the) `[[nodiscard]]` attribute.
2. Think about supporting it in a better way and define the appropriate macro (along with other useful general purpose macros) in a header file of its own (which may better belong to `libzim`).

include/library.h
mgautierfr: Yes

mgautierfr:
What about renaming `NODISCARD` to `LIBKIWIX_NODISCARD` ?

> (which may better belong to libzim).

The frontier is a bit blurry but we want to keep the two projects separated.
I don't think it is a good idea to define things in libzim based on compiler features to be used only on libkiwix.

veloman-yunkan:
> What about renaming `NODISCARD` to `LIBKIWIX_NODISCARD` ?

That will do, but the next question is "Why does it have to be defined in `library.h`?"

> 
> > (which may better belong to libzim).
> 
> The frontier is a bit blurry but we want to keep the two projects separated. I don't think it is a good idea to define things in libzim based on compiler features to be used only on libkiwix.

I don't see why the usage of `[[nodiscard]]` should be limited to `libkiwix`. If that new C++ feature is a useful one it should be available across all our C++ projects. Hence `libzim` is the right place for it (until we split out another library for general purpose stuff, something like our own STL/boost/etc).

mgautierfr:
> That will do, but the next question is "Why does it have to be defined in library.h?"

It doesn't have to be defined here, but it is used only in `library.h` so why should we define elsewhere ?

> I don't see why the usage of [[nodiscard]] should be limited to libkiwix. If that new C++ feature is a useful one it should be available across all our C++ projects. Hence libzim is the right place for it (until we split out another library for general purpose stuff, something like our own STL/boost/etc).

The purpose of libzim is not to provide useful things to other projects. It is to provide reading and writing zim files (and even that, we want to split libzim in two libraries https://github.com/openzim/libzim/issues/789)

> Drop the (only current usage of the) [[nodiscard]] attribute.

Droping the `[[nodiscard]]` attribute is also a good solution which avoid a lot of questions.

veloman-yunkan:
> > That will do, but the next question is "Why does it have to be defined in library.h?"
> 
> It doesn't have to be defined here, but it is used only in `library.h` so why should we define elsewhere ?

So that it is easier to reuse if/when needed somewhere else. If you care only about `Library::create()` then consider `#undef`ing the macro after it serves its purpose.

mgautierfr:
New version:
- Use of `LIBKIWIX_NODISCARD`
- Undef after use.
- Fix `__has_cpp_attribute` vs `__has_attribute` (the former won)

veloman-yunkan (approved):

PR #1009, opened 2023-10-21
Kiwix frontend style cleanup

### LINKED ISSUE #917: Look-and-feel mismatch between the kiwix-serve welcome page and the viewer
> > In welcome page, font shows correctly 
> > But it doesn't while browsing a book
> 
> This is a manifestation of a more general discrepancy in the look-and-feel of the welcome page and the viewer page. I think we better address it separately.

Please don't forget to create the ticket then.

_Originally posted by @kelson42 in https://github.com/kiwix/libkiwix/issues/910#issuecomment-1468379699_

veloman-yunkan:
Addresses #917

Unfortunately, I failed to achieve consistent look-and-feel across the library page and the viewer toolbar so that they both look nice.

- large border radius of the library page doesn't look good when applied to the small buttons of the viewer toolbar, whereas small radius used for the viewer buttons looks inferior to what we have now on the library page
- color schemes of the buttons and filters could not be unified too; viewer current button icons don't look nice on dark background, whereas the current design of the toolbar of the library page seems to rely on contrast.

This PR simply reduces the gap in the look-and-feel of the two toolbars. If we want to eliminate it completely, I think that we must come up with a new design first.

### DIFF
--- a/static/resources_list.txt
+++ b/static/resources_list.txt
@@ -10,6 +10,7 @@ skin/iso6391To3.js
 skin/isotope.pkgd.min.js
 skin/index.js
 skin/autoComplete.min.js
+skin/kiwix.css
 skin/taskbar.css
 skin/index.css
 skin/fonts/Poppins.ttf
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -325,58 +325,6 @@ body {
     -webkit-box-orient: vertical;
 }
 
-.modal-wrapper {
-    position: fixed;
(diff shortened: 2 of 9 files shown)


kelson42:
@veloman-yunkan Thank you for this PR and this attempt to fix this annoying issue. I know this is not easy. Will give it a test this WE.

kelson42 (approved):
LGTM, as this is only CSS, I will merge straight.

PR #1017, opened 2023-11-09
Switch to macos-13.

mgautierfr:


kelson42:
@mgautierfr we have here to secure macos12 is support, appropriate compilation command lien have to be used to secure this.

rgaudin:
> @mgautierfr we have here to secure macos12 is support, appropriate compilation command lien have to be used to secure this.

@kelson42 as discussed I believe we want to rely on kiwix-build to produce macos12+ binaries and that should be enough for now.

rgaudin (approved):

rgaudin:
⚠️ last commit to be removed before merge

PR #1018, opened 2023-11-11
Test iOS cross-compile in CI

kelson42:
Test macOS to iOS cross-compilation in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,11 +33,11 @@ jobs:
           rm -f /usr/local/bin/2to3 /usr/local/bin/2to3-3.11 /usr/local/bin/idle3 /usr/local/bin/idle3.11 /usr/local/bin/pydoc3 /usr/local/bin/pydoc3.11 /usr/local/bin/python3 /usr/local/bin/python3-config /usr/local/bin/python3.11 /usr/local/bin/python3.11-config
           brew install pkg-config ninja meson
 
-      - name: Install dependencies
-        env:
-          ARCHIVE_NAME: deps2_macos_native_dyn_libkiwix.tar.xz
+      - name: Install deps
+        shell: bash
         run: |
-          wget -O- https://tmp.kiwix.org/ci/${{env.ARCHIVE_NAME}} | tar -xJ -C ${{env.HOME}}
+          ARCHIVE_NAME=deps2_macos_${{matrix.target}}_libzim.tar.xz
+          wget -O- http://tmp.kiwix.org/ci/${ARCHIVE_NAME} | tar -xJ -C $HOME
 
       - name: Compile
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,17 @@ on:
 
 jobs:
   macOS:
-    runs-on: macos-13
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-13
+        target:
+          - native_dyn
+          - iOS_arm64
+          - iOS_x86_64
+    runs-on: ${{ matrix.os }}
+
     env:
       HOME: /Users/runner
     steps:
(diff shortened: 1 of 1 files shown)


kelson42:
@mgautierfr @rgaudin `meson_cross_file.txt` seems to be missing. An other point is that if we really need `-Db_coverage=true` here?!

rgaudin:
> @mgautierfr @rgaudin `meson_cross_file.txt` seems to be missing. An other point is that if we really need `-Db_coverage=true` here?!

It's downloading https://tmp.kiwix.org/ci/deps2_macos_native_dyn_libkiwix.tar.xz which obviously (so I think) doesn't contain `BUILD_iOS_xxxx/` folder (which would contain the cross file.

kelson42:
@rgaudin Thx, indeed!

Now it seems the `libzim` is not part of the deps! See https://github.com/kiwix/libkiwix/actions/runs/6835979771/job/18590377587?pr=1018#step:5:71

.github/workflows/ci.yml
```diff
@@ -23,21 +33,30 @@ jobs:
...
           brew install pkg-config ninja meson
 
-      - name: Install dependencies
-        env:
-          ARCHIVE_NAME: deps2_macos_native_dyn_libkiwix.tar.xz
+      - name: Install deps
+        shell: bash
         run: |
-          wget -O- https://tmp.kiwix.org/ci/${{env.ARCHIVE_NAME}} | tar -xJ -C ${{env.HOME}}
+          ARCHIVE_NAME=deps2_macos_${{matrix.target}}_libzim.tar.xz
```
mgautierfr, line 39: `deps2_*_libzim.tar.xz` is dependencies for `libzim`. Probably a wrong copy/paste :)

rgaudin:
Changed `libzim` suffix to `libkiwix` but in fact we don't produce those deps for the other archs: macOS arm64 (not listed), iOS x86_64 and iOS arm64

The only macos `deps2` we produce for libkiwix is [`deps2_macos_native_dyn_libkiwix.tar.xz`](https://tmp.kiwix.org/ci/deps2_macos_native_dyn_libkiwix.tar.xz)

.github/workflows/ci.yml
kelson42: OK, sorry for this stupid thing!

kelson42:
@mgautierfr Can we easi

> The only macos `deps2` we produce for libkiwix is [`deps2_macos_native_dyn_libkiwix.tar.xz`](https://tmp.kiwix.org/ci/deps2_macos_native_dyn_libkiwix.tar.xz)

@mgautierfr Can we easily fix this?

.github/workflows/ci.yml
```diff
@@ -23,21 +33,30 @@ jobs:
...
 
-      - name: Compile source code
+      - name: Compile
         env:
           PKG_CONFIG_PATH: ${{env.HOME}}/BUILD_native_dyn/INSTALL/lib/pkgconfig
           CPPFLAGS: -I${{env.HOME}}/BUILD_native_dyn/INSTALL/include
+        shell: bash
         run: |
-          meson . build --default-library=shared -Db_coverage=true
+          MESON_OPTION="--default-library=shared -Db_coverage=true"
```
rgaudin, line 48: I find it more maintainable to not use github contexts in `run`. I prefer to pass all values as env instead:

- easier to test your code locally. You just have to prepare your exports and then copy-paste.
- Github logs prints the env always so it's easy to spot any mistake or missing value
- It looks like *dependencies* for the command which is good for renames/cleanups

.github/workflows/ci.yml
```diff
@@ -23,21 +33,30 @@ jobs:
...
+          MESON_CROSSFILE="$HOME/BUILD_${{matrix.target}}/meson_cross_file.txt"
+          if [[ ! "${{matrix.target}}" =~ native_.* ]]; then
+            MESON_OPTION="$MESON_OPTION --cross-file $MESON_CROSSFILE -Dstatic-linkage=true"
+            cat $MESON_CROSSFILE
```
rgaudin, line 52: Do you want to keep it? I think it's useful but it looks like a debug line 😀

PR #1022, opened 2023-11-15
Viewer toolbar improvements

### LINKED ISSUE #1021: UI language selector seems to not display properly anymore
See:
image

The UI language button should:
* Stick on the right like on the welcome page
* Be the same size like other button

Solution:
* Stick the button on the right (should be easy)
* Make the other buttons a bit bigger (like the taskbar) so they all have the same size (like the UI language switcher)

veloman-yunkan:
Fixes #1021

1. All controls in the viewer toolbar are made the same size
2. The UI language selector button is moved to the right
3. Better layout on screens narrower than 0.42 kilopixels.

### DIFF
--- a/static/skin/kiwix.css
+++ b/static/skin/kiwix.css
@@ -90,7 +90,7 @@ body {
 }
 
 #uiLanguageSelectorButton {
-  margin: 0px 12px 6px 12px;
+  margin: 0px 12px;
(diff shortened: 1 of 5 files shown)


kelson42 (approved):
@veloman-yunkan LGTM. Thx!

PR #1036, opened 2023-12-20
Get correct href value on `onClick` for "warc2zim" files.

mgautierfr:
Next to come warc2zim archive will come with "wombat" embedded. The purpose of wombat is to be an interface with js code to mask that we are in a scrapped/zim context to the js.

So it rewrite the `.href` attributes to the original url (ie, an absolute url to the original website), even if the local relative url is valid.

Let's ask to wombat to not rewrite href in our special case.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -290,12 +290,23 @@ function isExternalUrl(url) {
       || url.startsWith("https:");
 }
 
+function getRealHref(target) {
+  // In case of wombat in the middle, wombat will rewrite the href value to the original url (external link)
+  // This is not what we want. Let's ask wombat to not rewrite href
+  const old_no_rewrite = target._no_rewrite;
+  target._no_rewrite = true;
+  const target_href = target.href;
+  target._no_rewrite = old_no_rewrite;
+  return target_href;
+}
+
 function onClickEvent(e) {
   const iframeDocument = contentIframe.contentDocument;
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
-    if ( isExternalUrl(target.href) ) {
-      const possiblyBlockedLink = blockLink(target.href);
+    const target_href = getRealHref(target);
+    if (isExternalUrl(target_href)) {
+      const possiblyBlockedLink = blockLink(target_href);
       if ( e.ctrlKey || e.shiftKey ) {
         // The link will be loaded in a new tab/window - update the link
         // and let the browser handle the rest.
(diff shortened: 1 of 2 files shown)


Jaifroid:
As I will need to keep up with changes on the reader side, I need to understand the effect of this. I had noticed that getting the `href` of an anchor resulted in the original untransformed URL in code, even though the anchor's href is displayed in the browser as rewritten, and I had worked around that. I didn't know it was possible to disable rewriting (where are these things documented???). So I guess I just need to copy the code here that disables that rewriting when getting the href.

mgautierfr:
>  (where are these things documented???)

At the trustiest place : source code of wombat.
I am not aware[*] of a any documentation (neither comments in source code)

[*] I am NOT saying there isn't. Just I'm not aware of.

> So I guess I just need to copy the code here that disables that rewriting when getting the href.

I think so. Another solution could be to do `target.attributs.href` which seems to be the `href` store in the html (dom ?). As I don't know if there is a semantic difference between both, I prefer use the former and deactivate wombat rewriting (which will return "correct" `target.href`)

Jaifroid:
Thanks. I tried many things, and `event.target.getAttribute('href')` was also subject to rewriting, just like accessing the attribute directly. I'll check the suggested store when I rewrite that code. But thanks for info on how to turn off the rewriting temporarily. I find the lack of comments in the source code (at least the source code I could find) makes it really difficult to work these things out (for me, anyway).

Jaifroid:
Just wanted to say "big thank you" 🙏 for finding the `._no_rewrite = true` trick. It turned out to fix an issue in Kiwix PWA where `window.open()` would refuse to respect the instruction to open a new window rather than a new tab (for Zimit ZIMs of course). Doing `window.open._no_rewrite = true;` before issuing the `window.open()` command (and then setting it back to `false` afterwards) solves that and allows me to use the native API without issue.

I'm going to open an issue in the Wombat source Repo to add at least that bit of documentation!

static/skin/viewer.js
```diff
@@ -294,8 +294,14 @@ function onClickEvent(e) {
...
-    if ( isExternalUrl(target.href) ) {
-      const possiblyBlockedLink = blockLink(target.href);
+    // In case of wombat in the middle, wombat will rewrite the href value to the original url (external link)
+    // This is not what we want. Let's ask wombat to not rewrite href
+    const old_no_rewrite = target._no_rewrite;
+    target._no_rewrite = true;
+    const target_href = target.href;
+    target._no_rewrite = old_no_rewrite;
+    if (isExternalUrl(target_href)) {
+      const possiblyBlockedLink = blockLink(target_href);
```
veloman-yunkan, line 304: Please encapsulate (part of) this change in a helper function `get????Href()`. The hard part is figuring out what to put instead of `????` in the function name.

veloman-yunkan (changes requested):
LGTMBIPTMISB*

---

\* Looks good to me, but I propose to make it slightly better

veloman-yunkan (approved):
Fixup commit must be squashed

PR #1041, opened 2024-01-12
Adding the updated timestamp metadata for zim archives

FredoAF:
Now in the library, the description on the cards is extended to show the timestamp of when the zim archive was updated


kelson42:
@FredoAF Where are these enquiries?

FredoAF:
> @FredoAF Where are these enquiries?
@kelson42  I host a kiwix instance.
Maybe have a think about it and discuss with your team - my users find it very helpful.

Screenshot 2024-01-12 at 15 18 51

PR #1043, opened 2024-01-19
Migrate bookmarks between books

### LINKED ISSUE #1037: Allow bookmarks to apply to new versions of a title
Bookmarks apply to a book: a specific ZIM file identified via its UUID. The problem is:
* When this book is updated (same title, but recenter version), then the user "looses" its bookmarks
* This a problem for Kiwix, but kind of even worse for custom apps

We need a way for the user to easily re-apply its bookmarks to the latest version of the title. This operation will be called the "bookmark reconciliation".

Reconciliation requirements:
* Should be fully transparent for the end-user
* Should need (almost) no code change for the Kiwix developer
* Ideally we should have kind of a system which is not data destructive and allows to track things

mgautierfr:
Fix #1037 

Contrary to suggested in https://github.com/kiwix/libkiwix/issues/1037#issuecomment-1870101456, we don't create new bookmark but update existing one (we don't delete bookmark).

If we create a new one without removing the wrong ones, we would create new bookmarks on each migration.

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -55,6 +55,18 @@ enum supportedListMode {
   NOVALID = 1 << 5
 };
 
+enum MigrationMode {
+  /** When migrating bookmarks, do not allow to migrate to an older book than the currently pointed one.
+   *
+   * This is enforce only if the bookmark point to a valid books. If bookmark point to a invalid book,
+   * it will be migrated to an existing one, even if older.
+   */
+  UPGRADE_ONLY = 0,
+
+   /** Allow to migrate the bookmark to an older book. */
+  ALLOW_DOWNGRADE = 1,
+};
+
 class Filter {
   public: // types
     using Tags = std::vector<std::string>;
@@ -272,18 +284,21 @@ class Library: public std::enable_shared_from_this<Library>
    *
    * @return A tuple<int, int>: <The number of bookmarks updated>, <Number of invalid bookmarks before migration was performed>.
    */
-  std::tuple<int, int> migrateBookmarks();
+  std::tuple<int, int> migrateBookmarks(MigrationMode migrationMode = ALLOW_DOWNGRADE);
 
   /**
    * Migrate all bookmarks associated to a specific book.
    *
    * All bookmarks associated to `sourceBookId` book will be migrated to a better book.
    * "Better book", will be determined using heuristics based on book name, flavour and date.
-
-   * @param source the source bookId of the bookmarks to migrate.
+   * Be carrefull, when using with `migrationModde == ALLOW_DOWGRADE`, the bookmark will be migrated to a different book
+   * even if the bookmark points to an existing book and the other book is older than the current one.
+   *
+   * @param sourceBookId the source bookId of the bookmarks to migrate.
+   * @param migrationMode how we will find the best book.
    * @return The number of bookmarks updated.
    */
-  int migrateBookmarks(const std::string& sourceBookId);
+  int migrateBookmarks(const std::string& sourceBookId, MigrationMode migrationMode = UPGRADE_ONLY);
 
   /**
    * Migrate bookmarks
@@ -441,7 +456,8 @@ private: // functions
   AttributeCounts getBookAttributeCounts(BookStrPropMemFn p) const;
   std::vector<std::string> getBookPropValueSet(BookStrPropMemFn p) const;
   BookIdCollection filterViaBookDB(const Filter& filter) const;
-  std::string getBestTargetBookId(const Bookmark& bookmark) const;
+  void cleanupBookCollection(BookIdCollection& books, const std::string& sourceBookId, MigrationMode migrationMode) const;
+  std::string getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const;
   unsigned int getBookCount_not_protected(const bool localBooks, const bool remoteBooks) const;
   void updateBookDB(const Book& book);
   void dropCache(const std::string& bookId);
--- a/include/library.h
+++ b/include/library.h
@@ -325,6 +325,20 @@ class Library: public std::enable_shared_from_this<Library>
    */
   std::string getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const;
 
+  /**
+   * Get the best bookId for a combination of book's name, flavour and date.
+   *
+   * Given a bookName (mandatory), try to find the best book.
+   * If preferedFlavour is given, will try to find a book with the same flavour. If not found, return a book with a different flavour.
+   * If minDate is given, return a book newer than minDate. If not found, return a empty bookId.
+   *
+   * @param bookName The name of the book
+   * @param preferedFlavour The prefered flavour.
+   * @param minDate the minimal book date acceptable. Must be a string in the format "YYYY-MM-DD".
+   * @return A bookId corresponding to the query, or empty string if not found.
+   */
+  std::string getBestTargetBookId(const std::string& bookName, const std::string& preferedFlavour="", const std::string& minDate="") const;
+
   // XXX: This is a non-thread-safe operation
   const Book& getBookById(const std::string& id) const;
   // XXX: This is a non-thread-safe operation
--- a/include/library.h
+++ b/include/library.h
@@ -280,7 +280,7 @@ class Library: public std::enable_shared_from_this<Library>
    *
    * All invalid bookmarks (ie pointing to unknown books, no check is made on bookmark pointing to
    * invalid articles of valid book) will be migrated (if possible) to a better book.
-   * "Better book", will be determined using heuristics based on book name, flavour and date.
+   * "Better book", will be determined using method `getBestTargetBookId`.
    *
    * @return A tuple<int, int>: <The number of bookmarks updated>, <Number of invalid bookmarks before migration was performed>.
    */
@@ -290,9 +290,7 @@ class Library: public std::enable_shared_from_this<Library>
    * Migrate all bookmarks associated to a specific book.
    *
    * All bookmarks associated to `sourceBookId` book will be migrated to a better book.
-   * "Better book", will be determined using heuristics based on book name, flavour and date.
-   * Be carrefull, when using with `migrationModde == ALLOW_DOWGRADE`, the bookmark will be migrated to a different book
-   * even if the bookmark points to an existing book and the other book is older than the current one.
+   * "Better book", will be determined using method `getBestTargetBookId`.
    *
    * @param sourceBookId the source bookId of the bookmarks to migrate.
    * @param migrationMode how we will find the best book.
@@ -311,6 +309,18 @@ class Library: public std::enable_shared_from_this<Library>
    */
   int migrateBookmarks(const std::string& sourceBookId, const std::string& targetBookId);
  
+  /**
+   * Get the best available bookId for a bookmark.
+   *
+   * Given a bookmark, return the best available bookId.
+   * "best available bookId" is determined using heuristitcs based on book name, flavour and date.
+   *
+   * @param bookmark The bookmark to search the bookId for.
+   * @param migrationMode The migration mode to use.
+   * @return A bookId. Potentially empty string if no suitable book found.
+   */
+  std::string getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const;
+
   // XXX: This is a non-thread-safe operation
   const Book& getBookById(const std::string& id) const;
   // XXX: This is a non-thread-safe operation
@@ -457,7 +467,6 @@ private: // functions
   std::vector<std::string> getBookPropValueSet(BookStrPropMemFn p) const;
   BookIdCollection filterViaBookDB(const Filter& filter) const;
   void cleanupBookCollection(BookIdCollection& books, const std::string& sourceBookId, MigrationMode migrationMode) const;
-  std::string getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const;
   unsigned int getBookCount_not_protected(const bool localBooks, const bool remoteBooks) const;
   void updateBookDB(const Book& book);
   void dropCache(const std::string& bookId);
--- a/include/library.h
+++ b/include/library.h
@@ -56,14 +56,18 @@ enum supportedListMode {
 };
 
 enum MigrationMode {
-  /** When migrating bookmarks, do not allow to migrate to an older book than the currently pointed one.
+  /** When migrating bookmarks, do not allow to migrate to an older book than the currently pointed one
+   * (or date stored in the bookmark if book is invalid)
    *
-   * This is enforce only if the bookmark point to a valid books. If bookmark point to a invalid book,
-   * it will be migrated to an existing one, even if older.
+   * If no newer books are found, no upgrade is made.
    */
   UPGRADE_ONLY = 0,
 
-   /** Allow to migrate the bookmark to an older book. */
+  /** Try hard to do a migration. This mostly does:
+   * - Try to find a newer book.
+   * - If book is invalid: find a best book, potentially older.
+   * Older book will never be returned if current book is a valid one.
+   */
   ALLOW_DOWNGRADE = 1,
 };
 
@@ -466,7 +470,7 @@ private: // functions
   AttributeCounts getBookAttributeCounts(BookStrPropMemFn p) const;
   std::vector<std::string> getBookPropValueSet(BookStrPropMemFn p) const;
   BookIdCollection filterViaBookDB(const Filter& filter) const;
-  void cleanupBookCollection(BookIdCollection& books, const std::string& sourceBookId, MigrationMode migrationMode) const;
+  std::string getBestFromBookCollection(BookIdCollection books, const Bookmark& bookmark, MigrationMode migrationMode) const;
   unsigned int getBookCount_not_protected(const bool localBooks, const bool remoteBooks) const;
   void updateBookDB(const Book& book);
   void dropCache(const std::string& bookId);
--- a/include/library.h
+++ b/include/library.h
@@ -71,6 +71,7 @@ class Filter {
     std::string _query;
     bool _queryIsPartial;
     std::string _name;
+    std::string _flavour;
 
   public: // functions
     Filter();
@@ -130,6 +131,7 @@ class Filter {
     Filter& maxSize(size_t size);
     Filter& query(std::string query, bool partial=true);
     Filter& name(std::string name);
+    Filter& flavour(std::string flavour);
     Filter& clearLang();
     Filter& clearCategory();
 
@@ -151,6 +153,9 @@ class Filter {
 
     bool hasCreator() const;
     const std::string& getCreator() const { return _creator; }
+
+    bool hasFlavour() const;
+    const std::string& getFlavour() const { return _flavour; }
 
     const Tags& getAcceptTags() const { return _acceptTags; }
     const Tags& getRejectTags() const { return _rejectTags; }
--- a/include/library.h
+++ b/include/library.h
@@ -255,7 +255,7 @@ class Library: public std::enable_shared_from_this<Library>
   void addBookmark(const Bookmark& bookmark);
 
   /**
-   * Remove a bookmarkk
+   * Remove a bookmark
    *
    * @param zimId The zimId of the bookmark.
    * @param url The url of the bookmark.
@@ -263,6 +263,39 @@ class Library: public std::enable_shared_from_this<Library>
    */
   bool removeBookmark(const std::string& zimId, const std::string& url);
 
+  /**
+   * Migrate all invalid bookmarks.
+   *
+   * All invalid bookmarks (ie pointing to unknown books, no check is made on bookmark pointing to
+   * invalid articles of valid book) will be migrated (if possible) to a better book.
+   * "Better book", will be determined using heuristics based on book name, flavour and date.
+   *
+   * @return A tuple<int, int>: <The number of bookmarks updated>, <Number of invalid bookmarks before migration was performed>.
+   */
+  std::tuple<int, int> migrateBookmarks();
+
+  /**
+   * Migrate all bookmarks associated to a specific book.
+   *
+   * All bookmarks associated to `sourceBookId` book will be migrated to a better book.
+   * "Better book", will be determined using heuristics based on book name, flavour and date.
+
+   * @param source the source bookId of the bookmarks to migrate.
+   * @return The number of bookmarks updated.
+   */
+  int migrateBookmarks(const std::string& sourceBookId);
+
+  /**
+   * Migrate bookmarks
+   *
+   * Migrate all bookmarks pointing to `source` to `destination`.
+   *
+   * @param sourceBookId the source bookId of the bookmarks to migrate.
+   * @param targetBookId the destination bookId to migrate the bookmarks to.
+   * @return The number of bookmarks updated.
+   */
+  int migrateBookmarks(const std::string& sourceBookId, const std::string& targetBookId);
+ 
   // XXX: This is a non-thread-safe operation
   const Book& getBookById(const std::string& id) const;
   // XXX: This is a non-thread-safe operation
@@ -408,6 +441,7 @@ private: // functions
   AttributeCounts getBookAttributeCounts(BookStrPropMemFn p) const;
   std::vector<std::string> getBookPropValueSet(BookStrPropMemFn p) const;
   BookIdCollection filterViaBookDB(const Filter& filter) const;
+  std::string getBestTargetBookId(const Bookmark& bookmark) const;
   unsigned int getBookCount_not_protected(const bool localBooks, const bool remoteBooks) const;
   void updateBookDB(const Book& book);
   void dropCache(const std::string& bookId);
--- a/include/library.h
+++ b/include/library.h
@@ -408,7 +408,7 @@ private: // functions
   void dropCache(const std::string& bookId);
 
 private: //data
-  mutable std::mutex m_mutex;
+  mutable std::recursive_mutex m_mutex;
   Library::Revision m_revision;
   std::map<std::string, Entry> m_books;
   using ArchiveCache = ConcurrentCache<std::string, std::shared_ptr<zim::Archive>>;
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -192,6 +192,7 @@ void Library::cleanupBookCollection(BookIdCollection& books, const std::string& 
 }
 
 std::string Library::getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const {
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   // Search for a existing book with the same name
   auto book_filter = Filter();
   if (!bookmark.getBookName().empty()) {
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -110,7 +110,7 @@ Library::~Library() = default;
 
 bool Library::addBook(const Book& book)
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   ++m_revision;
   /* Try to find it */
   updateBookDB(book);
@@ -141,13 +141,13 @@ bool Library::addBook(const Book& book)
 
 void Library::addBookmark(const Bookmark& bookmark)
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   m_bookmarks.push_back(bookmark);
 }
 
 bool Library::removeBookmark(const std::string& zimId, const std::string& url)
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   for(auto it=m_bookmarks.begin(); it!=m_bookmarks.end(); it++) {
     if (it->getBookId() == zimId && it->getUrl() == url) {
       m_bookmarks.erase(it);
@@ -166,7 +166,7 @@ void Library::dropCache(const std::string& id)
 
 bool Library::removeBookById(const std::string& id)
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   m_bookDB->delete_document("Q" + id);
   dropCache(id);
   // We do not change the cache size here
@@ -184,7 +184,7 @@ bool Library::removeBookById(const std::string& id)
 
 Library::Revision Library::getRevision() const
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   return m_revision;
 }
 
@@ -192,7 +192,7 @@ uint32_t Library::removeBooksNotUpdatedSince(Revision libraryRevision)
 {
   BookIdCollection booksToRemove;
   {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::lock_guard<std::recursive_mutex> lock(m_mutex);
     for ( const auto& entry : m_books) {
       if ( entry.second.lastUpdatedRevision <= libraryRevision ) {
         booksToRemove.push_back(entry.first);
@@ -217,7 +217,7 @@ const Book& Library::getBookById(const std::string& id) const
 
 Book Library::getBookByIdThreadSafe(const std::string& id) const
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   return getBookById(id);
 }
 
@@ -275,7 +275,7 @@ std::shared_ptr<ZimSearcher> Library::getSearcherByIds(const BookIdSet& ids)
 unsigned int Library::getBookCount(const bool localBooks,
                                    const bool remoteBooks) const
 {
-  std::lock_guard<std::mutex> lock(m_mutex);
+  std::lock_guard<std::recursive_mutex> lock(m_mutex);
   return getBookCount_not_protected(localBooks, remoteBooks);
 }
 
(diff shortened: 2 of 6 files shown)


test/library.cpp
```diff
@@ -347,14 +352,33 @@ class LibraryTest : public ::testing::Test {
...
+
+  auto bookmark = createBookmark(book, "/a/url", "A title");
+
+  EXPECT_EQ(bookmark.getUrl(), "/a/url");
+  EXPECT_EQ(bookmark.getTitle(), "A title");
+  EXPECT_EQ(bookmark.getBookId(), bookId);
+  EXPECT_EQ(bookmark.getBookName(), book.getName());
+  EXPECT_EQ(bookmark.getBookName(), "wikipedia_fr_tunisie");
+  EXPECT_EQ(bookmark.getBookFlavour(), book.getFlavour());
+  EXPECT_EQ(bookmark.getLanguage(), book.getCommaSeparatedLanguages());
```
veloman-yunkan, line 410: What is the reason that `Bookmark::getBookName()` is tested against both return value of `Book::getName()` and an explicit string value, whereas `getBookFlavour()` and `getLanguage()` receive a different treatment?

test/library.cpp
```diff
@@ -376,7 +400,7 @@ TEST_F(LibraryTest, bookmarksSerializationTest)
 
     lib->addBookmark(createBookmark(bookId1, "a/url", "Article title1"));
     lib->addBookmark(createBookmark("invalid-bookmark-id", "another/url", "Unknown title"));
-    lib->addBookmark(createBookmark(bookId2, "a/url/2", "Article title2"));
+    lib->addBookmark(createBookmark(book2, "a/url/2", "Article title2"));
```
veloman-yunkan: If it's intentional that the bookmarks for book1 and book2 are created via book-id and the book object, respectively, it's better to have a comment making such a setup more obvious and explaining why it is important.

include/library.h
```diff
@@ -71,6 +71,7 @@ class Filter {
     std::string _query;
     bool _queryIsPartial;
     std::string _name;
+    std::string _flavour;
```
veloman-yunkan, line 74: This comment applies to the entire commit (titled "Allow to filter books by flavour.")

Please add a unit test `LibraryTest.filterByFlavour` in `test/library.cpp`.

include/library.h
```diff
@@ -263,6 +263,18 @@ class Library: public std::enable_shared_from_this<Library>
...
 
+  /**
+   * Resolve bookmarks
+   *
+   * @param source the source bookId of the bookmarks to resolve.
+                   If not present, we resolve all bookmarks pointing to an unknown book.
+   * @param destination the destination bookId to resolve te bookmark to.
+                        If not present, we will try to determine it using heuristics based on
+                        book name, flavour and date.
+   * @return The number of bookmarks created.
```
veloman-yunkan, line 274: created → updated

include/library.h
```diff
@@ -263,6 +263,18 @@ class Library: public std::enable_shared_from_this<Library>
...
+   */
+  int resolveBookmarks(const std::string& source = "", std::string destination = "");
```
veloman-yunkan, line 276: Isn't "migrate" a better verb for this operation?

veloman-yunkan, line 276: I think it's better to have three separate functions:

```
migrateBookmarks(source, destination);
migrateBookmarks(source);
migrateBookmarks();
```
where `source` and `destination` must be valid book-ids. This eliminates the meaningless combination `migrateBookmarks("", destination)` and decomposes the implementation into a set of smaller and easier to understand non-recursive functions.

test/library.cpp
```diff
@@ -453,6 +453,101 @@ TEST_F(LibraryTest, bookmarksSerializationTest)
...
+    lib->addBookmark(createBookmark(book1));
+    lib->addBookmark(createBookmark("invalid-bookmark-id"));
+    lib->addBookmark(createBookmark(book2));
+
+    auto wrongIdBookmark = createBookmark(book1);
+    wrongIdBookmark.setBookId("wrong-bookmark-id");
+    lib->addBookmark(wrongIdBookmark);
+
+    auto wrongIdBookmarkNoName = createBookmark(book2);
+    wrongIdBookmarkNoName.setBookId("wrong-bookmark-id-noname");
```
veloman-yunkan: IMHO, `bookmark-id` is misleading, better replace it with `book-id`.

src/library.cpp
```diff
@@ -158,6 +158,83 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+    }
+    
+    if (targetBooks.empty()) {
+      // No existing book found for the target, or the same than source.
+      return 0;
+    }
+    if (targetBooks.size() != 1) {
+      sort(targetBooks, DATE, false);
+    }
+    destination = targetBooks[0];
```
veloman-yunkan: This piece of code deserves being packaged into a (private) named function `std::string Library::getMostRecentTargetBookId(const Bookmark&) const` (the name could be different but it must correctly reflect the semantics of that function).

src/library.cpp
```diff
@@ -158,6 +158,83 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+          return 0;
+      }
+      book_filter.query("title:\""+firstBookmarkToChange.getBookTitle() + "\"");
+    }
+    auto targetBooks = filter(book_filter);
+    // Remove source book
+    auto foundIT = std::find(targetBooks.begin(), targetBooks.end(), source);
+    if (foundIT != targetBooks.end()) {
+      targetBooks.erase(foundIT);
+    }
```
veloman-yunkan, line 214: Won't this downgrade the book to an earlier version (provided that it exists) in the absence of a newer version?

veloman-yunkan (changes requested):

include/library.h
kelson42: Definitly

test/library.cpp
mgautierfr: It is intentional, I will add a comment.

include/library.h
mgautierfr: Yes, I follow the @kelson42 proposition at first and made the same thought to myself during development, forget about it.

mgautierfr:
Ready for a new review. New commit are either `fixup!` or marked with `[New]`.

There is an open question about overload resolution between `migrateBookmarks(const std::string& source, bool allowDowngrade)` and `migrateBookmarks(const std::string& source const std::string& target)`.
If used with a `char*` (as a literal string), cpp resolves overload to use the `bool` version.

include/library.h
```diff
@@ -113,7 +113,7 @@ class Filter {
     /**
      *  Set the filter to only accept books in the specified category.
      *
-     *  Multiple categories can be specified as a comma-separated list (in
+     *  Mutiple categories can be specified as a comma-separated list (in
```
veloman-yunkan, line 116: I guess this is an accidental change

include/library.h
```diff
@@ -264,16 +264,16 @@ class Library: public std::enable_shared_from_this<Library>
...
                    If not present, we resolve all bookmarks pointing to an unknown book.
-   * @param destination the destination bookId to resolve te bookmark to.
+   * @param destination the destination bookId to migrate the bookmarks to.
                         If not present, we will try to determine it using heuristics based on
                         book name, flavour and date.
-   * @return The number of bookmarks created.
+   * @return The number of bookmarks updated.
    */
-  int resolveBookmarks(const std::string& source = "", std::string destination = "");
+  int migrateBookmarks(const std::string& source = "", std::string destination = "");
```
veloman-yunkan: Reminding that during rebase&squash the commit message of the original commit must be updated too

src/library.cpp
```diff
@@ -192,37 +189,37 @@ int Library::migrateBookmarks(const std::string& source, std::string destination
...
+    return 0;
   }
+  if (targetBooks.size() != 1) {
+    sort(targetBooks, DATE, false);
+  }
+  return migrateBookmarks(source, targetBooks[0]);
+}
 
+int Library::migrateBookmarks(const std::string& source, const std::string& destination) {
   // We finally a set of bookmarks to change and a destination
```
veloman-yunkan, line 255: This comment is now obsolete

include/library.h
```diff
@@ -439,6 +439,7 @@ class Library: public std::enable_shared_from_this<Library>
   AttributeCounts getBookAttributeCounts(BookStrPropMemFn p) const;
   std::vector<std::string> getBookPropValueSet(BookStrPropMemFn p) const;
   BookIdCollection filterViaBookDB(const Filter& filter) const;
+  std::string getBestTargetBookId(const Bookmark& bookmark) const;
```
veloman-yunkan: During rebase&history-cleanup - it's better if this commit is squashed too (so that the implementation of `migrateBookmarks()` is immediately introduced in the most comprehensible way).

test/library.cpp
```diff
@@ -569,9 +569,12 @@ TEST_F(LibraryTest, MigrateBookmark)
     EXPECT_EQ(allBookmarks[4].getBookId(), bookId2);
     EXPECT_EQ(allBookmarks[5].getBookId(), bookId1+"_date_flavour");
 
-    ASSERT_EQ(lib->migrateBookmarks(bookId1, bookId2), 0); // No more bookId1 bookmark
+    // Suprisingly, `char *` make c++ prefers `migrateBookmarks(..., bool)` instead of `std::string`.
+    // [??] Is is a problem ? It should appear only if user use `char *` (so use C or literal string).
+    // A solution would be to rename one method, but which one ?
```
veloman-yunkan, line 574: Another solution is to use an `enum` instead of `bool` for the parameter type, which will also convert the somewhat cryptic call `migrateBookmarks(..., true)` to a more self-explaining `migrateBookmarks(..., ALLOW_DOWNGRADE)`

veloman-yunkan (commented):
Please rebase&fixup for the next review

kelson42:
@BPerlakiH Looks good to you (in terme of feature and interface)?

BPerlakiH:
> @BPerlakiH Looks good to you (in terme of feature and interface)?

Yes, initially it looks good to me.

test/library.cpp
mgautierfr: Indeed. This is a far better solution.

src/library.cpp
```diff
@@ -157,6 +157,90 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+  if (!bookmark.getBookName().empty()) {
+    book_filter.name(bookmark.getBookName());
+  } else {
+    // We don't have a name stored (older bookmarks)
+    // Fallback on title (All bookmarks should have one, but let's be safe against wrongly filled bookmark)
+    if (bookmark.getBookTitle().empty()) {
+        // No bookName nor bookTitle, no way to find target book.
+        return "";
+    }
+    book_filter.query("title:\""+bookmark.getBookTitle() + "\"");
```
veloman-yunkan, line 189: What if the book title contains a double-quote character? A test case is due.

src/library.cpp
```diff
@@ -157,6 +157,90 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
   return false;
 }
 
+int Library::migrateBookmarks() {
```
veloman-yunkan, line 160: I wonder if this function should return not only the number of migrated bookmarks, but also the count of those bookmarks that failed to migrate

src/library.cpp
```diff
@@ -157,6 +157,90 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+}
+
+int Library::migrateBookmarks(const std::string& source, const std::string& destination) {
+  // We finally a set of bookmarks to change and a destination
+  int changed = 0;
+  for (auto& bookmark:m_bookmarks) {
+    if (bookmark.getBookId() == source) {
+      bookmark.setBookId(destination);
+      changed +=1;
+    }
```
veloman-yunkan, line 260: Shouldn't there be any foolproofness here? I guess a bookmark should be allowed to migrate only if the bookmarked path is present in the destination book.

src/library.cpp
```diff
@@ -200,6 +200,11 @@ std::string Library::getBestTargetBookId(const Bookmark& bookmark) const {
     return "";
   }
   if (targetBooks.size() != 1) {
+    // We have several, reduce to same flavour
+    auto flavouredTargetBooks = filter(book_filter.flavour(bookmark.getBookFlavour()));
+    if (!flavouredTargetBooks.empty()) {
+      targetBooks = flavouredTargetBooks;
+    }
```
veloman-yunkan: `flavouredTargetBooks` may contain `bookmark.getBookId()` (which has been removed from `targetBooks` and will now reappear there)

src/library.cpp
```diff
@@ -202,15 +209,25 @@ std::string Library::getBestTargetBookId(const Bookmark& bookmark) const {
...
+    // Remove source book
+    auto foundIT = std::find(flavouredTargetBooks.begin(), flavouredTargetBooks.end(), bookmark.getBookId());
+    if (foundIT != flavouredTargetBooks.end()) {
+      if (allowDowngrade) {
+        flavouredTargetBooks.erase(foundIT);
+      } else {
+        // targetBooks is sorted by date, remove the source and all oldest books
+        flavouredTargetBooks.erase(foundIT, flavouredTargetBooks.end());
+      }
+    }
```
veloman-yunkan: So this actually also addresses my previous comment on the source book not being dropped from `flavouredTargetBooks`. But it introduces code duplication with the same procedure performed on `targetBooks`. Please extract that common operation into a helper function.

include/library.h
```diff
@@ -55,6 +55,11 @@ enum supportedListMode {
   NOVALID = 1 << 5
 };
 
+enum MigrationMode {
+  UPGRADE_ONLY = 0,
+  ALLOW_DOWNGRADE = 1,
+ };
```
veloman-yunkan, line 61: Unjustified indentation

veloman-yunkan (changes requested):

src/library.cpp
mgautierfr: Yes

src/library.cpp
mgautierfr: We agreed with @kelson42 to not check for the path but only for bookId

src/library.cpp
mgautierfr: Done

mgautierfr:
Ready for new review.
Previous `fixup!` have been already fixup.
New commits start from [7db8e03](https://github.com/kiwix/libkiwix/pull/1043/commits/7db8e036cb55a19b6e25c3266601e14746829463)

src/library.cpp
```diff
@@ -191,6 +191,16 @@ void Library::cleanupBookCollection(BookIdCollection& books, const std::string&
...
 
+std::string remove_quote(std::string input) {
+  size_t lower_bound = 0;
+  size_t found = std::string::npos;
+  while ((found = input.find("\"", lower_bound)) != std::string::npos) {
+    input.erase(found, 1);
+    lower_bound = found;
+  }
+  return input;
+}
```
veloman-yunkan, line 202: It is simpler and more efficient to rely on STL:

```c++
std::string remove_all(std::string s, char c)
{
    const auto it = std::remove(s.begin(), s.end(), c);
    s.erase(it, s.end());
    return s;
}
```

src/library.cpp
```diff
@@ -203,7 +213,7 @@ std::string Library::getBestTargetBookId(const Bookmark& bookmark, MigrationMode
         // No bookName nor bookTitle, no way to find target book.
         return "";
     }
-    book_filter.query("title:\""+bookmark.getBookTitle() + "\"");
+    book_filter.query("title:\"" + remove_quote(bookmark.getBookTitle()) + "\"");
```
veloman-yunkan, line 216: Isn't it safer to replace the quotes with whitespace? The answer depends on how xapian indexes a string like `ABC"XYZ`.

veloman-yunkan (changes requested):

mgautierfr:
> @mgautierfr It‘s important that the piece of code detecting if a new book is available (to amke a bookmark/note update) is public and dedicated, so we can move notes on kiwix android the same way (but in kotlin), see https://github.com/kiwix/kiwix-android/issues/3632

I have tried to make that by making our internal (new) method public, but in fact detecting that a new book is available is not the same thing that we are doing here.

The PR is trying to resolve invalid bookmark, not migrate valid bookmark to new ones.
It seems pretty close at first sight, but it is not the same.
As @veloman-yunkan, it triggers some inconsistency. I've tried to mitigate that with documentation, but it may not be enough.

mgautierfr:
Rebased on main and previous fixup commits have been fixed up.

New commits start at [d41a9ad](https://github.com/kiwix/libkiwix/pull/1043/commits/d41a9ad9de37570fb13c0d5ac4158bc9d867816b)

The questions about surprising behavior of `migrationMode` have been fixed (hopefully)

veloman-yunkan (changes requested):
I only looked at the new commits.
For the next (hopefully final) review please fully rebase & fixup the PR (including any fixes addressing the new comments).

include/library.h
```diff
@@ -250,14 +271,59 @@ class Library: public std::enable_shared_from_this<Library>
...
+  int migrateBookmarks(const std::string& sourceBookId, const std::string& targetBookId);
+
+  /**
+   * Get the best available bookId for a bookmark.
+   *
+   * Given a bookmark, return the best available bookId.
+   *
+   * @param bookmark The bookmark to search the bookId for.
+   * @param migrationMode The migration mode to use.
+   * @return A bookId. Potentially empty string if not suitable book found.
```
veloman-yunkan, line 323: not → no

src/library.cpp
```diff
@@ -157,6 +157,123 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+  if (migrationMode == UPGRADE_ONLY) {
+    auto foundIt = std::find(books.begin(), books.end(), bookmark.getBookId());
+    if (foundIt == books.end()) {
+      foundIt = std::find_if(books.begin(), books.end(), [&](std::string& bookId) {
+        auto& book = getBookById(bookId);
+        return book.getDate() <= bookmark.getDate();
+      });
+    }
+    if (foundIt != books.end()) {
+      // books is sorted by date, remove the source and all oldest books
```
veloman-yunkan, line 192: oldest → older

src/library.cpp
```diff
@@ -157,6 +157,123 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+    book_filter.query("title:\"" + remove_quote(bookmark.getBookTitle()) + "\"");
+  }
+  auto targetBooks = filter(book_filter);
+  pruneBookCollectionFromOlderBooks(targetBooks, bookmark, migrationMode);
+    
+  if (targetBooks.empty()) {
+    // No better book to migrate to.
+    // If bookmark is valid, best is the current one.
+    try {
+      auto book = getBookById(bookmark.getBookId());
```
veloman-yunkan, line 225: `auto` → `const auto&`

src/library.cpp
```diff
@@ -157,6 +157,123 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+    }
+  }
+  if (targetBooks.size() != 1) {
+    // We have several, reduce to same flavour
+    auto flavouredTargetBooks = filter(book_filter.flavour(bookmark.getBookFlavour()));
+    pruneBookCollectionFromOlderBooks(flavouredTargetBooks, bookmark, migrationMode);
+    if (!flavouredTargetBooks.empty()) {
+      targetBooks = flavouredTargetBooks;
+    }
+  }
```
veloman-yunkan, line 238: Have you considered the following scenario?

1. `migrationMode == ALLOW_DOWNGRADE` 
2. `targetBooks` contains a newer book of a different flavour
3. `flavouredTargetBooks` only contains older books of the same flavour

This implementation will favour the older book of the same flavour over a newer one, but is it intentional?

src/library.cpp
```diff
@@ -157,6 +157,123 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+  Bookmark firstBookmarkToChange;
+  for(auto& bookmark:m_bookmarks) {
+    if (bookmark.getBookId() == sourceBookId) {
+      firstBookmarkToChange = bookmark;
+      break;
+    }
+  }
+
+  if (firstBookmarkToChange.getBookId().empty()) {
+    return 0;
```
veloman-yunkan, line 254: Drawing attention to Codecov's discovery that "Added line #L254 was not covered by tests"

test/library.cpp
```diff
@@ -32,7 +32,7 @@ const char * sampleOpdsStream = R"(
     <id>urn:uuid:0c45160e-f917-760a-9159-dfe3c53cdcdd</id>
     <icon>/meta?name=favicon&amp;content=wikipedia_fr_tunisie_novid_2018-10</icon>
     <updated>2018-10-08T00:00::00:Z</updated>
-    <dc:issued>8 Oct 2018</dc:issued>
+    <dc:issued>2018-10-08T00:00:00:Z</dc:issued>
```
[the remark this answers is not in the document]
veloman-yunkan: OK, now I see that it was copied from an existing date string above.

test/library.cpp
```diff
@@ -327,28 +411,292 @@ class LibraryTest : public ::testing::Test {
...
+    auto book = lib->getBookById(bookId);
+
+    auto invalidBookmark = createBookmark(book);
+    invalidBookmark.setBookId("invalid-book-id");
+    invalidBookmark.setDate("2020-10-08");
+    lib->addBookmark(invalidBookmark);
+
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::UPGRADE_ONLY), "");
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::ALLOW_DOWNGRADE), bookId+"_date");
 }
```
veloman-yunkan, line 695: 1. In this test it is not obvious what the date of `book` is. An `ASSERT_EQ(book.getDate(), "...")` would be 
helpful.
2. `migrateBookmarks()` is not called/tested (inconsistency with the test name)

test/library.cpp
```diff
@@ -327,28 +411,292 @@ class LibraryTest : public ::testing::Test {
...
+    validBookmark.setDate("2020-10-08");
+    lib->addBookmark(validBookmark);
+
+    //  The best book for the bookmark is bookId...
+    ASSERT_EQ(lib->getBestTargetBookId(validBookmark, kiwix::UPGRADE_ONLY), bookId);
+    // but there is not migration to do as the bookmark already point to it.
+    ASSERT_EQ(lib->migrateBookmarks(bookId, kiwix::UPGRADE_ONLY), 0);
+
+    ASSERT_EQ(lib->getBestTargetBookId(validBookmark, kiwix::ALLOW_DOWNGRADE), bookId);
+}
```
veloman-yunkan, line 666: In this test it is not obvious what the date of `book` is. An `ASSERT_EQ(book.getDate(), "...")` would be helpful.

test/library.cpp
```diff
@@ -327,28 +411,292 @@ class LibraryTest : public ::testing::Test {
...
+    auto book = lib->getBookById(bookId);
+
+    auto validBookmark = createBookmark(book);
+    lib->addBookmark(validBookmark);
+
+    auto onlyValidBookmarks = lib->getBookmarks();
+
+    ASSERT_EQ(lib->getBestTargetBookId(validBookmark, kiwix::UPGRADE_ONLY), bookId+"_date");
+    ASSERT_EQ(lib->getBestTargetBookId(validBookmark, kiwix::ALLOW_DOWNGRADE), bookId+"_date");
+}
```
veloman-yunkan, line 648: 1. `onlyValidBookmarks` is not used
2. `migrateBookmarks()` is not called/tested (inconsistency with the test name)

test/library.cpp
```diff
@@ -327,28 +411,292 @@ class LibraryTest : public ::testing::Test {
...
+
+    auto book = lib->getBookById(bookId);
+
+    auto invalidBookmark = createBookmark(book);
+    invalidBookmark.setBookId("invalid-book-id");
+    lib->addBookmark(invalidBookmark);
+
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::UPGRADE_ONLY), bookId+"_date");
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::ALLOW_DOWNGRADE), bookId+"_date");
+}
```
veloman-yunkan, line 680: `migrateBookmarks()` is not called/tested (inconsistency with the test name)

include/library.h
```diff
@@ -250,14 +271,59 @@ class Library: public std::enable_shared_from_this<Library>
...
+  /**
+   * Migrate all invalid bookmarks.
+   *
+   * All invalid bookmarks (ie pointing to unknown books, no check is made on bookmark pointing to
+   * invalid articles of valid book) will be migrated (if possible) to a better book.
+   * "Better book", will be determined using heuristics based on book name, flavour and date.
+   *
+   * @return A tuple<int, int>: <The number of bookmarks updated>, <Total number of invalid bookmarks>.
+   */
+  std::tuple<int, int> migrateBookmarks();
```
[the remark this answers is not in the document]
mgautierfr: As we will migrate only invalid bookmarks, I think it is better to migrate to a valid book if possible even if it an older one.
I see no reason users would like to keep invalid bookmarks while they can make them valid (knowing that they will be able to do the migration to newer book later when they have them).

But @kelson42 may have a different point of view.

src/library.cpp
mgautierfr: That's a good question.
What is more important, flavour or date ?

I haven't especially think about that, but I agree with the result.
User is asking ALLOW_DOWNGRADE, we are moving to the "best" book, even if older.

@kelson42, do you agree ?

src/library.cpp
```diff
@@ -176,17 +176,21 @@ std::tuple<int, int> Library::migrateBookmarks() {
   return std::make_tuple(changed, invalidBookmarks);
 }
 
-void Library::cleanupBookCollection(BookIdCollection& books, const std::string& sourceBookId, MigrationMode migrationMode) const {
+void Library::pruneBookCollectionFromOlderBooks(BookIdCollection& books, const Bookmark& bookmark, MigrationMode migrationMode) const {
```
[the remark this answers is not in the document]
mgautierfr: I have follow your advice but now we have a bit more code duplication (sort and if).
I can quickly go back if you which.

src/library.cpp
```diff
@@ -176,17 +176,21 @@ std::tuple<int, int> Library::migrateBookmarks() {
...
   sort(books, DATE, false);
 
   // Remove source book
-  auto foundIT = std::find(books.begin(), books.end(), sourceBookId);
-  if (foundIT != books.end()) {
-    if (migrationMode == ALLOW_DOWNGRADE) {
-      books.erase(foundIT);
-    } else {
+  if (migrationMode == UPGRADE_ONLY) {
```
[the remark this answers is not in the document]
mgautierfr: I have follow your advice but now we have a bit more code duplication (sort and if).
I can quickly go back if you which.

mgautierfr:
Ready for another review. All commits are fixed-up.
I have closed all discussion taken into account.
We have 3 left open discussions. We could go with the current code if we agree or we will need another round depending of our your/@kelson42 feedback.

---

(macOs failing is new and not related to this PR, seems to be a (broken) update in homebrew)

test/library.cpp
```diff
@@ -327,28 +411,294 @@ class LibraryTest : public ::testing::Test {
...
+    EXPECT_EQ(book.getDate(), "2018-10-08");
+
+    auto invalidBookmark = createBookmark(book);
+    invalidBookmark.setBookId("invalid-book-id");
+    invalidBookmark.setDate("2020-10-08");
+    lib->addBookmark(invalidBookmark);
+
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::UPGRADE_ONLY), "");
+    ASSERT_EQ(lib->getBestTargetBookId(invalidBookmark, kiwix::ALLOW_DOWNGRADE), bookId+"_date");
 }
```
veloman-yunkan, line 697: When reading the tests explicitly calling the `getBestTargetBookId()` function, `bookId+"_date"` is not intuitive. `bookId+"_updated_1year_later"` or `"newer_version_of_"+bookId` is more self-explaining.

For the current set of tests this comment applies to the `*_flavour` and `*_date_flavour` variants to a much lesser extent, since they are used only in the `LibraryTest.MigrateBookmark` test case. Shouldn't we have tests for `getBestTargetBookId()` involving those combinations too?

src/library.cpp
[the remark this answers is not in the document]
veloman-yunkan: I didn't take that into account because I was mislead by the function name - it didn't mention sorting.

src/library.cpp
```diff
@@ -157,6 +157,127 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+    // We have several, reduce to same flavour
+    auto flavouredTargetBooks = filter(book_filter.flavour(bookmark.getBookFlavour()));
+    sort(flavouredTargetBooks, DATE, false);
+    if (migrationMode == UPGRADE_ONLY) {
+      removeOlderBooksFromBookCollection(flavouredTargetBooks, bookmark);
+    }
+    if (!flavouredTargetBooks.empty()) {
+      targetBooks = flavouredTargetBooks;
+    }
+  }
```
veloman-yunkan, line 239: I think that the code duplication in this function introduced by following my previous suggestion as well as my other comment on the newer-different-flavour-vs-older-same-flavour situation can be resolved as follows.

It doesn't make sense running the filter query a second time. After it turns out that there is more than a single candidate according to the simple criteria we can add more criteria and filter or reorder the result further. Thus the function `pruneBookCollectionFromOlderBooks()` from the previous version of this PR can be renamed to `getPrioritizedMigrationCandidates()` (possibly returning the result instead of modifying the input collection) and extended with a step that runs an additional sort on the remaining books. I think that such a design will allow for easier addition of extra criteria for selecting the best target book.

veloman-yunkan:
The last two comments (each published as a separate review) were made in response to @mgautierfr's feedback. A full commit-wise review is still due.

include/library.h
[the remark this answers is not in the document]
kelson42: Not strong opinion in term of use case, but I'm more concerned about the whole coherency of the migration methods... and for that, I would tend to agree with @veloman-yunkan.

include/library.h
```diff
@@ -250,14 +271,60 @@ class Library: public std::enable_shared_from_this<Library>
...
+   * Get the best available bookId for a bookmark.
+   *
+   * Given a bookmark, return the best available bookId.
+   * "best available bookId" is determined using heuristitcs based on book name, flavour and date.
+   *
+   * @param bookmark The bookmark to search the bookId for.
+   * @param migrationMode The migration mode to use.
+   * @return A bookId. Potentially empty string if no suitable book found.
+   */
+  std::string getBestTargetBookId(const Bookmark& bookmark, MigrationMode migrationMode) const;
```
kelson42, line 326: We need a version of this function taking the needed arguments (name, flavour, etc..) so the function can be used for the notes in Android. See https://github.com/kiwix/libkiwix/issues/1037#issuecomment-1878484240

src/library.cpp
```diff
@@ -157,6 +157,127 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+  int changed = 0;
+  for(auto& sourceBook:sourceBooks) {
+    changed += migrateBookmarks(sourceBook, ALLOW_DOWNGRADE);
+  }
+  return std::make_tuple(changed, invalidBookmarks);
+}
+
+void Library::removeOlderBooksFromBookCollection(BookIdCollection& books, const Bookmark& bookmark) const {
```
kelson42, line 179: This functions, and all functions, should have a documentation.

src/library.cpp
```diff
@@ -503,6 +626,12 @@ Xapian::Query nameQuery(const std::string& name)
   return Xapian::Query("XN" + normalizeText(name));
 }
 
+Xapian::Query flavourQuery(const std::string& name)
+{
+  return Xapian::Query("XF" + normalizeText(name));
+}
+
+
```
kelson42, line 634: To manu empty lines?

src/library.cpp
mgautierfr: All public methods have a documentation in `include/library.h`.

The particular method has been renamed (and greatly changed) in last version.

include/library.h
mgautierfr: Done

src/library.cpp
mgautierfr: Nice input.
I have rewrite the algorithm with your idea.

test/library.cpp
mgautierfr: I have renamed `_date` to `_udpated1yearlater`.

> For the current set of tests this comment applies to the *_flavour and *_date_flavour variants to a much lesser extent, since they are used only in the LibraryTest.MigrateBookmark test case. Shouldn't we have tests for getBestTargetBookId() involving those combinations too?

Indeed. We could. But `MigrateBookmark` is using `getBestTargetBookId` too, so it is indirectly tested.

include/library.h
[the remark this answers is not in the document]
mgautierfr: Done

test/library.cpp
mgautierfr: Small test added.
Not a full testing as the method is indirectly tested with `MigrateBookmark` and `getBestTargetBookId(bookName, ...)`.

mgautierfr:
Next round of review.

I have rewritten the algorithm using input from @veloman-yunkan. It is indeed simpler.
`fixup!` commits have been inserted in the git history to reduce potential conflict at next `rebase-fixup`.

veloman-yunkan (changes requested):
This is a pre-approval.

Please include the proposed final touches in the rebase&fixup and let there be green light!

src/library.cpp
```diff
@@ -157,6 +157,159 @@ bool Library::removeBookmark(const std::string& zimId, const std::string& url)
...
+    }
+  }
+  int changed = 0;
+  for(auto& sourceBook:sourceBooks) {
+    changed += migrateBookmarks(sourceBook, migrationMode);
+  }
+  return std::make_tuple(changed, invalidBookmarks);
+}
+
+std::string Library::getBestFromBookCollection(BookIdCollection& books, const Bookmark& bookmark, MigrationMode migrationMode) const {
```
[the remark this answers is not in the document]
mgautierfr: Definitely, this is error with too much rewrite/rebase..

mgautierfr:
> Please include the proposed final touches in the rebase&fixup and let there be green light!

Done

veloman-yunkan (approved):

PR #1044, opened 2024-01-28
Default UI language is resolved in the frontend

### LINKED ISSUE #1038: library.kiwix.org seems to use italian as the language for non-language specific UI elements
Hi,

It seems like [library.kiwix.org](https://library.kiwix.org) has the language configured wrongly:
screenshot_kiwix_library

As you can see in the provided screenshot, some text elements like "select category" appear to be written italian (?). Other examples include "select language" and the placeholder text in the search field. I'd assume that said words are intended to be in english or dependent on the currently selected language.

I've tested this with multiple browsers and multiple devices. Kiwix appears to properly detect my prefered language (german), but said texts still continue to be in italian.

### LINKED ISSUE #1039: Kiwix serve UI language is not detected (from Browser language) and always in English
I don't know how the language detection is supposed to work, but if I visit https://library.kiwix.org/ with a browser with `Accept-Language: it` I get redirected to https://library.kiwix.org/#lang=ita and the interface stays in English, while the content is mostly about Italian language wikis with Italian descriptions.

Screenshot_20240101_121518

Reorted first by @nemobis at https://github.com/kiwix/libkiwix/issues/1038#issuecomment-1873260728

veloman-yunkan:
Fixes #1038
Fixes #1039

This change makes the viewer resolve the default UI language in the front end, without any dependence on the backend's handling of the `Accept-Language` header. Yet, the latter has not been removed since it is currently used in the following scenarios:

- `/nojs` endpoint
- error responses (e.g. `/invalid_endpoint`, or any other invalid kiwix-serve URL directly loaded in the browser rather than via the Kiwix JS viewer).


static/generate_i18n_resources_list.py
```diff
@@ -30,7 +30,8 @@ def get_translation_info(filepath):
     with open(filepath, 'r', encoding="utf-8") as f:
         content = json.load(f)
         lang_name = content.get("name")
-        return lang_code, lang_name
+        translation_count = len(content)
+        return lang_code, lang_name, translation_count
```
mgautierfr, line 34: Return directly a dict ?
It would avoid error prone `translation_info[1]` access and no need for `make_language_entry`

static/skin/i18n.js
```diff
@@ -73,11 +73,30 @@ const DEFAULT_UI_LANGUAGE = 'en';
...
+      const pseudoQvalue = Math.pow(0.9, rank);
+      const score = pseudoQvalue * lang.translation_count;
+      if ( score > mostSuitableLang.score ) {
+        mostSuitableLang.code = lang.iso_code;
+        mostSuitableLang.score = score;
+      }
+    }
+  }
+  return mostSuitableLang.code;
+}
```
mgautierfr, line 94: I like the idea of using the translated lines counts to select the best language.

But why with formula ? Do you have a known use case where the formula works (or not) ?

We can be sure that we will have issue opened about "My setting are not respected and kiwix use another language". I prefer to make the first move and prepare an answer 🙂

static/generate_i18n_resources_list.py
```diff
@@ -40,14 +41,21 @@ def get_translation_info(filepath):
...
+            lang_name = translation_info[1]
             if lang_name:
-                language_list.append((lang_code, lang_name))
+                language_list.append(translation_info)
             else:
                 print(f"Warning: missing 'name' in {i18n_file.name}")
         f.write(str(i18n_file.relative_to(script_path.parent)) + '\n')
 
-language_list = [{name: code} for code, name in sorted(language_list)]
+def make_language_entry(lang_info_tuple):
```
rgaudin, line 52:
```suggestion
def make_language_entry(iso_code, self_name, translation_count):
```

and expand it when calling

```py
language_list = [make_language_entry(*lang_info) for lang_info in sorted(language_list)]
```

static/generate_i18n_resources_list.py
rgaudin: I agree, seems like a now unnecessary intermediate step

static/skin/i18n.js
rgaudin: I agree.

rgaudin (changes requested):
+1 on @mgautierfr's feedback

static/generate_i18n_resources_list.py
veloman-yunkan: I used a tuple at this point so that the language list could be sorted (by ISO code) and `language.js` is generated in a well-defined order. However processing of *.json files already takes place in that order, thus the extra sorting doesn't make sense.

static/skin/i18n.js
veloman-yunkan: This reproduces the algorithm used in the backend. The only difference is that qvalues are made up.

Consider the following language list: `ab`, `bc`, `cd`, `de`, and suppose that translation of `kiwix-serve` UI into those languages is at 30%, 90%, 50% and 100%, respectively. If we assume no qvalues (or, equivalently, qvalues equal to 1 for all languages) this actually means that the order of languages never matters and we have to select the language with the highest level of translation (in this case `de` which is at the last position). Therefore, in order to account for the order in which the languages are listed we must assume a decreasing sequence of qvalues which in my implementation is 1, 0.9, 0.81, ... 0.9**k, .... In this example `bc` is the winner with a score of 0.81 vs 0.729 of `de`.

Of course the usage of `0.9` as the common ratio of the geometric progression used for  qvalues is arbitrary and `0.8` or `0.7` may be more appropriate. Note however that exponential functions with a base less than 1 decrease quite rapidly so the utility for our purpose of other values with a single digit after decimal point (0.6, 0.5, etc) drops significantly.

Yet nothing forces us to use a formula. I can happily replace it with any hardcoded sequence like:

```python
assumedQvalues = [1, 0.98, 0.76, 0.54, 0.32, ...]
```

veloman-yunkan: FYI, in Firefox 121, the qvalue sequence implicitly generated by the browser depends on the number of languages in Webpage Language Settings. It is
- `[1, 0.5]` for 2 entries
- `[1, 0.7, 0.3]` for 3 entries
- `[1, 0.8, 0.5, 0.3]` for 4 entries

veloman-yunkan: Their algorithm seems to be
```python
[round(1-i/n, 1) for i in range(n)]
```

under the assumption that values at the middle are rounded upward (i.e. `0.25` is rounded up to `0.3` rather than down to `0.2`)

veloman-yunkan:
Reminder to myself - translatewiki PR #1033 has been merged, so `language.js` must be regenerated during rebase&fixup.

static/skin/i18n.js
mgautierfr: Seems good to me.

mgautierfr (approved):

mgautierfr:
Please rebase-fixup.

veloman-yunkan:
> Reminder to myself - translatewiki PR #1033 has been merged, so `languages.js` must be regenerated during rebase&fixup.

Done

veloman-yunkan:
> Please rebase-fixup.

@mgautierfr Done

PR #1046, opened 2024-01-30
Translation of search results page

### LINKED ISSUE #1028: Search results page is not translated
The HTML produced by the `/search` API endpoint is always in English:

1. Title of the page - "**Search**: *query*"
2. First line - "**Results** *x*-*y* **of** *z* **for** "*query*"" or "**No results were found for** "*query*""
3. Information for each search result - "**from** *book*" and  "*XXX* **words**".

veloman-yunkan:
Fixes #1028

Contrary to https://github.com/kiwix/libkiwix/pull/1032#issuecomment-1896002341, no front-end side translation is performed on the search results page. Instead the `/search` endpoint is called with the `userlang` query parameter explicitly provided. The downside is the reduced opportunity for HTTP-caching search requests coming from users using different UI languages. 

Currently no real translations of messages related to search results are available. To test this change you will have to use the `test` language that is excluded from the UI language selector. It can be set by adding (once) `?userlang=test` after `/viewer` in the browser address bar.

Known issues:

- changing the UI language while a search results page is loaded in the viewer doesn't automatically re-translate it but reloading the viewer does it.

### DIFF
--- a/include/search_renderer.h
+++ b/include/search_renderer.h
@@ -73,6 +73,13 @@ class SearchRenderer
   }
 
   /**
+   * set user language
+   */
+  void setUserLang(const std::string& lang){
+    this->userlang  = lang;
+  }
+
+  /**
    * Generate the html page with the resutls of the search.
    *
    * @param mapper The `NameMapper` to use to do the rendering.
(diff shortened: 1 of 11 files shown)


kelson42:
@veloman-yunkan I guess its time to rebase this PR.

veloman-yunkan:
> @veloman-yunkan I guess its time to rebase this PR.

@kelson You are right. I could sense the same recommendation in my horoscope for today and humbly complied with it.

mgautierfr (approved):
LGTM !

PR #1052, opened 2024-02-13
More predictable Downloader::startDownload()

veloman-yunkan:
Fixes kiwix/kiwix-desktop#1022 (as well as the other case described in #1050 without fixing the latter)

Before this change `Downloader::startDownload()` might avoid starting a new download when a download with the specified URI was already present in its cache. This might be confusing for the following reasons:
    
1. uri is not the only parameter of `Downloader::startDownload()` - a target download directory may also be specified through the second `options` parameter. Thus calling `Downloader::startDownload()` twice with the same URI but different download directories would not save files into the second directory.
    
2. Files of a completed download may be removed, whereupon downloading the same files again won't be a no-op. However in such a situation `Downloader` refuses to actually repeat a previous download.

This change is sufficient to fix the bugs described in #1050 without any changes in `kiwix-desktop`. However I am not sure that it is the right way to do that - maybe adding a `removeDownload()` function to `kiwix::Downloader` (and thus fixing #1050) is justified in any case.

### DIFF
--- a/src/downloader.cpp
+++ b/src/downloader.cpp
@@ -169,12 +169,6 @@ std::vector<std::string> Downloader::getDownloadIds() const {
 std::shared_ptr<Download> Downloader::startDownload(const std::string& uri, const std::vector<std::pair<std::string, std::string>>& options)
 {
   std::unique_lock<std::mutex> lock(m_lock);
-  for (auto& p: m_knownDownloads) {
-    auto& d = p.second;
-    auto& uris = d->getUris();
-    if (std::find(uris.begin(), uris.end(), uri) != uris.end())
-      return d;
-  }
   std::vector<std::string> uris = {uri};
   auto gid = mp_aria->addUri(uris, options);
   m_knownDownloads[gid] = std::make_shared<Download>(mp_aria, gid);


kelson42:
@veloman-yunkan What is the status of this PR? Does last comment of @mgautierfr has triggered a difficulty?

veloman-yunkan:
> @veloman-yunkan What is the status of this PR? Does last comment of @mgautierfr has triggered a difficulty?

@kelson42 No additional technical difficulties. As I wrote in the description I was not sure that the proposed change was a good idea and @mgautierfr's feedback was along the same lines. Yet I will keep it open for a while until I decide how to proceed.

veloman-yunkan:
I started working on an alternative solution based on introducing a new method `Downloader::forgetDownload()` but didn't like it early in the process and ended up with #1066. Now we have to decide which way to go.

mgautierfr:
> I started working on an alternative solution based on introducing a new method Downloader::forgetDownload() but didn't like it early in the process and ended up with https://github.com/kiwix/libkiwix/pull/1066. Now we have to decide which way to go.

I like better #1066. At least we are "sure" that aria2 will not misbehave when we have two downloads for the same file/option.

kelson42:
> > @veloman-yunkan What is the status of this PR? Does last comment of @mgautierfr has triggered a difficulty?
> 
> @kelson42 No additional technical difficulties. As I wrote in the description I was not sure that the proposed change was a good idea and @mgautierfr's feedback was along the same lines. Yet I will keep it open for a while until I decide how to proceed.

OK, it seems an alternative has been chosen, but the two issues this PR was going to fix are still open:
* https://github.com/kiwix/kiwix-desktop/issues/1022
* https://github.com/kiwix/libkiwix/issues/1050 without fixing the latter)

So not sure if this PR should be revamped, a new PR started or just the two issue to be closed...

PR #1054, opened 2024-02-14
Enter polyfills.js

### LINKED ISSUE #1051: Make library work with more browsers by polyfilling `replaceAll()`
In this file https://github.com/kiwix/libkiwix/blob/main/static/skin/viewer.js#L76, a `replaceAll()` method is used that is incompatible with a number of older browsers. We at Kiwix JS have found several cases where the fancy library cannot be used simply because of this method (e.g., any Chrome before 85, see https://caniuse.com/?search=replaceAll), even though the browser would otherwise support the library feature.

We tried to polyfill it ourselves, but due to CORS its' not possible for us to tamper with this script.

For more info see https://github.com/kiwix/kiwix-js/issues/1174 and https://github.com/kiwix/kiwix-js/pull/1190. And for a very small polyfill that could simply be added to the viewer.js without needing to download a package, see https://github.com/kiwix/kiwix-js/pull/1190/files.

veloman-yunkan:
Fixes #1051


Jaifroid (approved):
As far as I'm concerned, this code should work, though I'm not able to test it. I also didn't get to test the polyfill itself over at KJS, because it couldn't be injected into the viewer due to CORS. So, testing would be important!

veloman-yunkan:
I only tested that it didn't break my fresh Firefox (didn't test it on older browsers).

kelson42:
@veloman-yunkan @Jaifroid Can we do anything to secure it works before merging?
@mgautierfr Why the CI fails on macOS?

Jaifroid:
Well once this is merged to dev I could test with an older browser. It wouldn't be a tragedy to have a follow-up, remedial PR if necessary, so long as buggy code doesn't end up on the production server.

mgautierfr:
> @mgautierfr Why the CI fails on macOS?

Fixed with #1056 (This PR is rebased on new master).

mgautierfr (approved):
LGTM.
However, I have not tested it, just reviewed.

PR #1066, opened 2024-02-27
Smarter Downloader::startDownload()

veloman-yunkan:
This PR is an alternative to #1052. Similar to the latter it addresses the user-observable problems behind #1050 without fixing that issue the way it was formulated.

The shared property of #1052 and this PR is that they try to solve the bugs without enhancing the libkiwix API. The difference is that this PR tries to make `Downloader::startDownload()` smarter, whereas #1052 deprived it of the slightest smartness that it tried to exercise.

Now having tried both approaches I think that we better follow the dumb path. Although we can close the seemingly small gap in `downloadCanBeReused()` (marked with an `XXX` comment) my feeling is that trying to optimize for a theoretical situation that can hardly happen in practice is not justified.

### DIFF
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -187,8 +187,6 @@ class Downloader
    * have different values for the download directory or output file name
    * options, after the download is reported to be complete the downloaded file
    * will be present only at the location specified for the first request.
-   * Also, due to the above peculiarity there is no straightforward way to
-   * repeat a completed or cancelled download whose files have been deleted.
    *
    * User should call `update` on the returned `Download` to have an accurate status.
    *
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -168,7 +168,10 @@ class Download {
  */
 class Downloader
 {
- public:
+ public: // types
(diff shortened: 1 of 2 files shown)


mgautierfr (approved):

mgautierfr:
It is a technical approval. Maybe we need a extra step of discussion if this is a better solution than #1052 (or not)

mgautierfr:
@veloman-yunkan Do we agree to merge this solution ?

veloman-yunkan:
> @veloman-yunkan Do we agree to merge this solution ?

I think that it may be the best way to proceed. Although my intuition tells me that we may eventually run into other (more subtle) issues connected with this approach, I don't see why we should try to absolutely avoid them, since we can easily switch to the dumb solution at any time.

mgautierfr:
Let's go with this solution then. If we face some bug with this, we will see then what to do.

mgautierfr:
Fixes https://github.com/kiwix/kiwix-desktop/issues/1022

PR #1070, opened 2024-03-13
Add OpenSearch description in Kiwix Server homepage

kelson42:
Fixes https://github.com/kiwix/kiwix-tools/issues/620


mgautierfr:
Let's merge it, we will see if users complain to have xml responses only.

PR #1074, opened 2024-04-20
add ipv6 support

aryanA101a:
**Summary of changes**:
- Updated the APIs for retrieving network interfaces on posix-compliant systems and Windows to enable ipv6 support
- Added dual stack flag to from libmicrohttp to support ipv6 as well if the user specifies the ipv6 flag

**Compatibility notes**
 iphlpapi.h-getadaptersaddresses() : Minimum supported client  Windows XP


src/server/internalServer.cpp
```diff
@@ -451,28 +453,49 @@ bool InternalServer::start() {
...
+
+  if (m_ipv6) {
+    flags|=MHD_USE_DUAL_STACK;
+    sockAddr6.sin6_family = AF_INET6;
+    sockAddr6.sin6_port = htons(m_port);
+    if (m_addr.empty()) {
+      if (0 != INADDR_ANY) {
+        sockAddr6.sin6_addr = in6addr_any;
+      }
+      m_addr = (kiwix::getBestPublicIp(m_ipv6)).addr6;
```
veloman-yunkan, line 469: Well, it looks like this is why we may need the `ipv6` boolean parameter. But are there realistic situations where the user trusts the application to select the IP address on its own but enforces the type of the IP address?

src/server/internalServer.cpp
```diff
@@ -451,28 +453,49 @@ bool InternalServer::start() {
...
-    if (inet_pton(AF_INET, m_addr.c_str(), &(sockAddr.sin_addr.s_addr)) == 0) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    sockAddr4.sin_family = AF_INET;
+    sockAddr4.sin_port = htons(m_port);
+    if (m_addr.empty()) {
+      if (0 != INADDR_ANY) {
+        sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
+      }
+      m_addr = (kiwix::getBestPublicIp(m_ipv6)).addr;
+    } else{
```
veloman-yunkan, line 484: I would move this piece of code along with its counterpart in the other branch of the enclosing `if` statement outside (before) that statement. Of course it will have to initialize both `sockAddr4` and `sockAddr6` but as a result overall nesting will be reduced.

veloman-yunkan (changes requested):

src/server/internalServer.cpp
aryanA101a: None that I can think of.

One more case where the flag is needed is if a user does not specify an ip address but wants the server to support ipv6.

src/tools/networkTools.cpp
```diff
@@ -75,103 +77,121 @@ std::string kiwix::download(const std::string& url) {
...
+                      if (!IN6_IS_ADDR_LINKLOCAL(&(si->sin6_addr)))
+                        interfaces[temp->AdapterName].addr6=host;
+                  }
+                  pUnicast = pUnicast->Next;
+                }
+            }
     }
+  } else {
+    std::cerr << "Call to GetAdaptersAddresses failed with error: "<< dwRetVal << std::endl;
   }
```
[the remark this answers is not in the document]
aryanA101a: There are no tabs.

src/server/internalServer.h
```diff
@@ -103,6 +103,7 @@ class InternalServer {
                    bool withTaskbar,
                    bool withLibraryButton,
                    bool blockExternalLinks,
+                   bool ipv6,
```
veloman-yunkan, line 106: Get rid of this parameter

src/server/internalServer.h
```diff
@@ -118,6 +119,7 @@ class InternalServer {
     void stop();
     std::string getAddress() { return m_addr; }
     int getPort() { return m_port; }
+    bool isAddressIPv6() { return m_ipv6; }
```
veloman-yunkan, line 122: Make `const`

src/tools/networkTools.cpp
```diff
@@ -75,103 +77,144 @@ std::string kiwix::download(const std::string& url) {
...
+#else
+  return getNetworkInterfacesPosix();
 #endif
-  return interfaces;
+
 }
 
-std::string kiwix::getBestPublicIp() {
-  auto interfaces = getNetworkInterfaces();
+std::string kiwix::getBestPublicIp(bool ipv6) {
```
veloman-yunkan, line 193: What is the meaning of the `ipv6` parameter? Does it enable (allow) returning an IPv6 address or requires it?

veloman-yunkan (changes requested):
I am writing this summary after having added the review comments and having deleted/edited some of them (as I felt that my feedback starts being inconsistent). I read the discussion at kiwix/kiwix-tools#545 a little too late. My feeling is that the option `--ipv6` is not a good way to instruct the kiwix server to support IPv6. Is it supposed to enforce IPv6 or merely enables it?

An alternative may be to use special values for the `-i`/`--address` options. For example,

- `all` - listen on all available IP addresses (IPv4 and IPv6)
- `ipv4` - listen on all available IPv4 addresses
- `ipv6` - listen on all available IPv6 addresses
- `auto`
- network interface name (e.g. `eth0`, `wlan0`) - will be resolved to the IP address of the named interface

I think that we have to close in on the use model of this enhancement before proceeding with the PR.

veloman-yunkan:
> With the latest refactoring, any type of ip can be used with -i flag without the necessity of specifying -6 flag.

What happens if you specify an IPv4 address together with the `-6` flag?

aryanA101a:
> I am writing this summary after having added the review comments and having deleted/edited some of them (as I felt that my feedback starts being inconsistent). I read the discussion at [kiwix/kiwix-tools#545](https://github.com/kiwix/kiwix-tools/issues/545) a little too late. My feeling is that the option `--ipv6` is not a good way to instruct the kiwix server to support IPv6. Is it supposed to enforce IPv6 or merely enables it?

Current model:
-i lets you provide both types of ip addr
-6 just tries to enforce ipv6. In case no ip address is provided it makes sure best public ip is ipv6

> 
> An alternative may be to use special values for the `-i`/`--address` options. For example,
> 
> * `ipv4` - listen on all available IPv4 addresses
> * `ipv6` - listen on all available IPv6 addresses

Seems a good approach.

> * `all` - listen on all available IP addresses (IPv4 and IPv6)
> * `auto`
> * network interface name (e.g. `eth0`, `wlan0`) - will be resolved to the IP address of the named interface

I don't think a normal person will care about the specific network interfaces, so adding this much sugar seems redundant.
I think that ipv6 support is more about ensuring compatibility with ipv6 clients, although it does let us serve on ipv6 addresses, but that's not the sauce.

> I think that we have to close in on the use model of this enhancement before proceeding with the PR.

```
      -f, -family <FAMILY>
              Specifies  the  protocol family to use. The protocol family identifier can be one of inet, inet6, bridge, mpls
              or link.  If this option is not present, the protocol family is guessed from other arguments. If the  rest  of
              the  command line does not give enough information to guess the family, ip falls back to the default one, usu‐
              ally inet or any.  link is a special family identifier meaning that no networking protocol is involved.

       -4     shortcut for -family inet.

       -6     shortcut for -family inet6.
``` 
from the man page of ip command

I'd still argue that the -6 flag is a concise way to enforce IPv6, though your suggestion seems like a good approach as well.

@veloman-yunkan @kelson42 What do you think?

aryanA101a:
> > With the latest refactoring, any type of ip can be used with -i flag without the necessity of specifying -6 flag.
> 
> What happens if you specify an IPv4 address together with the `-6` flag?

```
@aryanA101a ➜ /workspaces/kiwix-tools (main) $ kiwix-serve wikipedia_en_100_mini_2023-08.zim -p 2002 -6 -i 127.0.0.1
Ip address 127.0.0.1  is not a valid ipv6 address
```

kelson42:
> > An alternative may be to use special values for the `-i`/`--address` options. For example,
> > 
> > * `ipv4` - listen on all available IPv4 addresses
> > * `ipv6` - listen on all available IPv6 addresses
> > * `all` - listen on all available IP addresses (IPv4 and IPv6)
> > * `auto`
> 
> Seems a good approach.

@aryanA101a Thank you for your patience. I have discussed the case with @veloman-yunkan and we agreed that this proposal was appropriate and leads to the removal of `-4` and/or `-6`. I also don't see what is the value of `auto`!? Would you be able to change the interface that way?

aryanA101a:
Sure

kelson42:
@aryanA101a Is that ready for a new review?

aryanA101a:
@mgautierfr I have modified the meson.build, but still win32-dyn build is failing.

kelson42:
@aryanA101a @veloman-yunkan Not sure about the status here, ready to review?

veloman-yunkan (changes requested):
Please rebase the PR and beautify its history (in the worst case it should consist of a single commit, but I would appreciate if you can split it into a few meaningful incremental changes).

kelson42:
@aryanA101a We are that far that we are almost ready to merge, any chance we can do that soon?

veloman-yunkan (approved):

kelson42:
@aryanA101a @veloman-yunkan I can not merge with what looks like legit CI errors.

aryanA101a:
> @mgautierfr I have modified the meson.build, but still win32-dyn build is failing.

https://github.com/kiwix/kiwix-build/blob/82500c545b386e102819a901508eeb7fbc59a048/kiwixbuild/configs/win32.py#L12
https://github.com/kiwix/libkiwix/actions/runs/9172701816/job/25267685133?pr=1074#step:5:349

kelson42:
@aryanA101a If this PR requires a change at Kiwix Build, please create an issue explaining what is needed.

aryanA101a:
> @aryanA101a If this PR requires a change at Kiwix Build, please create an issue explaining what is needed.

https://github.com/kiwix/kiwix-build/pull/697

I'll need some input from @mgautierfr

veloman-yunkan (commented):
To the best of my understanding the only change since the last review was in `meson.build`. Then my previous approval still holds. 

However, regarding the commit structure of this PR, I think that all changes to the source code (*.cpp, *.h files) should belong to one commit. The fix for building under windows should be in a separate commit.

veloman-yunkan:
> @aryanA101a @veloman-yunkan Last commit seems to fix the compilation problem to create binary for Windows. But this commit in significant and we have to request new review

That's a false impression. The last commit (which combines two unrelated changes) was amended with a small change that finally fixed the windows build. In a more reviewer friendly flow a fixup commit should be created.

kelson42:
OK, then merging so we can finally move forward.

veloman-yunkan:
Note that this PR is a backward incompatible API change. Which means that the next release of libkiwix must be 14.0 (unless backward compatibility is added).

@mgautierfr BTW, why are the versions of openzim and kiwix projects defined in the kiwix-build repository instead of in every project in a standardized way?

mgautierfr:
> BTW, why are the versions of openzim and kiwix projects defined in the kiwix-build repository instead of in every project in a standardized way?

It is defined in every project as a dependency constraint. See for example 
https://github.com/kiwix/kiwix-tools/blob/main/meson.build#L21-L22
Or each project define it own version like in https://github.com/kiwix/libkiwix/blob/main/meson.build#L2

kiwix-build is defining what we are building in our CI/release (and it must fulfills the constraints define in every project). Other packagers are free to use different version (as long as constraints are respected)

veloman-yunkan:
Indeed. I somehow failed to locate the version in `meson.build`. Sorry for the silly question.

PR #1075, opened 2024-05-04
Add Kiwix server a few accesskeys

### LINKED ISSUE #1072: Suggestions for improving Kiwix-server for blind people
Hello the Kiwix team,

To allow Kiwix-server to add accessibility functions for the blind people, I suggest you add shortcuts in the template files for pages read by the Kiwix server in the Kiwix common functions library.

For this I suggest you use the word "accesskey" to define these shortcuts, see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

Of course, there is a risk of conflict of these shortcuts with Browsers or the screen reader. 
In my case with Orca and Firefox, I used the following keys:
- S: to access the search area so Alt+Shift+S
- D: to access an article at random so Alt+Shift+D
- N: to access the home page of the encyclopedia so Alt+Shift+N
- W: to access the encyclopedias home page so Alt+Shift+W

Here are the patch files used to do this [libkiwix.tar.gz](https://github.com/kiwix/libkiwix/files/14911343/libkiwix.tar.gz)

kelson42:
Fixes #1072

Except for the "Library" button, the accesskey values are the same as on Kiwix Desktop or should be implemented the same (like for input search and switch to library) in Kiwix Desktop, see https://github.com/kiwix/kiwix-desktop/issues/1083. This is the reason why I don't have respected exactly the request like requested by @patrick-emmabuntus.

### DIFF
--- a/static/templates/index.html
+++ b/static/templates/index.html
@@ -93,7 +93,7 @@
         </div>
       </div>
       <form id='kiwixSearchForm' class='kiwixNav__SearchForm'>
-        <input type="text" name="q" placeholder="Search" id="searchFilter" class='kiwixSearch filter'>
+        <input type="text" name="q" accesskey="s" placeholder="Search" id="searchFilter" class='kiwixSearch filter'>
         <span class="kiwixButton tagFilterLabel"></span>
         <input type="submit" class="kiwixButton kiwixButtonHover" id="searchButton" value="Search"/>
       </form>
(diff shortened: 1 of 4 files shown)


veloman-yunkan (approved):

mgautierfr:
This is a approval but waiting feedback from user as requested here : https://github.com/openzim/mwoffliner/pull/2023#issuecomment-2098408189

PR #1079, opened 2024-05-11
Revert humanfriendly tag display in Kiwix Server

### LINKED ISSUE #899: Why are tags transformed?

kelson42:
Fixes #899

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -486,9 +486,8 @@
     function addTagElement(tagValue, resetFilter) {
         const tagElement = document.getElementsByClassName('tagFilterLabel')[0];
         tagElement.style.display = 'inline-block';
-        const humanFriendlyTagValue = humanFriendlyTitle(tagValue);
-        tagElement.innerHTML = `${humanFriendlyTagValue}`;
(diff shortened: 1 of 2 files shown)


veloman-yunkan (changes requested):
Unit tests must be fixed by updating the cacheids

veloman-yunkan (approved):
Please squash the changes into one commit when merging.

PR #1086, opened 2024-05-19
Improve  documentation

kelson42:
Small beautification. The most important change been that the data directory is foreseen to store ZIM files and nothing else (by my experience).

### DIFF
--- a/include/tools.h
+++ b/include/tools.h
@@ -38,22 +38,28 @@ typedef std::vector<std::string> FeedCategories;
 std::string getCurrentDirectory();
 
 /**
- * Return the data directory.
+ * Return the data directory
  *
- * The data directory is a directory where to put data (zim files, ...)
- * It depends of the platform and it may be changed by user using environment variable.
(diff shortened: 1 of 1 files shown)


include/tools.h
```diff
@@ -38,22 +38,28 @@ typedef std::vector<std::string> FeedCategories;
...
- * - current directory
- *
- * @return the path of the data directory (utf8 encoded)
+ * Return the data directory
+ *
+ * The data directory is foreseen to put ZIM files (and temporary
+ * files during download).  Its path can vary and depends on a few
+ * things.
+ *
+ * Resolution order is:
```
veloman-yunkan, line 47: The following is a more accurate description of how the data directory is used:

> The data directory is the default directory where downloaded files should be saved (it can be overriden via the `options` parameter of `kiwix::Downloader::startDownload()`).
> 
> Its path can vary and is determined as follows:

kelson42: Will fix and merge.

PR #1092, opened 2024-06-03
Fix wrong usage of `extra_link_args` variable.

mgautierfr:
Previous PR #1074 remove a bit too much in meson.build around extra_libs, extra_link_args.
So the compilation on Windows was failing. This PR fixes that.

The question about `host_machine`/`build_machine` should probably better handled but it will be once we will have a proper native compilation on Windows (soon™️).
This PR is just making compilation of libkiwix works on Windows.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -48,11 +48,14 @@ if host_machine.system() == 'windows' and static_deps
 endif
 
 if host_machine.system() == 'windows'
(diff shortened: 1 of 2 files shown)


veloman-yunkan (approved):
The nature of the changes is such that I can't approve them with an LGTM assessment (as I don't understand all their implications), yet I think that I can safely issue a DNLBTM (Does not look bad to me) warrant

PR #1093, opened 2024-06-03
Backward compatible support for IPv6

veloman-yunkan:
This PR brings backward compatibility to #1074

### DIFF
--- a/include/tools.h
+++ b/include/tools.h
@@ -25,10 +25,13 @@
 #include <map>
 #include <cstdint>
 
-namespace kiwix {
-struct IpAddress{
-    std::string addr;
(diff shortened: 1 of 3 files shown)


mgautierfr (commented):
Maybe we can also set the "old" methods as deprecated
(with https://github.com/kiwix/libkiwix/blob/main/include/common.h#L7-L14)

Else, agree with this PR

veloman-yunkan:
> Maybe we can also set the "old" methods as deprecated

It may be too soon. The nature of the enhancement is such that I feel fine with the old API staying around for some time. We can deprecate it later.

Issue #1099, opened 2024-07-12
`OPDSParser.parse` is always returning true


BPerlakiH:
**Updated:**

On the Apple reader side, we would like to be rest assured, that when invalid data is passed into the ``OPDSParser.parse(data:)`` function, we can get a `false` value back,
so we can handle it gracefully. [We already had a unit test on the reader side, but so far it was disabled](https://github.com/kiwix/kiwix-apple/pull/875/files).

Currently the libkiwix function never returns false:
https://github.com/kiwix/libkiwix/blob/ece40966f1b568ac6ce06c52d146f9f408435810/src/manager.cpp#L167-L179

Whereas on compiled libkiwix header files, I see this information, which is misleading:

```
  /**
   * Load a library content stored in a OPDS stream.
   *
   * @param content The content of the OPDS stream.
   * @param readOnly Set if the library path could be overwritten later with
   *                 updated content.
   * @param libraryPath The library path (used to resolve relative path)
   * @return True if the content has been properly parsed.
   */
  bool readOpds(const std::string& content, const std::string& urlHost);
```

So we either need to return False in case of invalid content, or if that's not possible, update the header file to reflect this special case.

kelson42:
@BPerlakiH Which exception is thrown? And what exactlz is wrong in the OPDS stream to trigger this exception?

veloman-yunkan:
I think that the bug description is not formulated correctly. I believe that it should rather state that `Manager::readOpds()` is not fool-proof against malformed input with an example of such input provided. The fix should then come with one or more unit-tests covering such scenarios.

mgautierfr:
From the screenshot in https://github.com/kiwix/kiwix-apple/issues/851, it seems this is a allocation error (not enough memory ? to much memory allocated ?) inside icu73 (called by `kiwix::removeAccents`).
This may be us doing strange things when trying to remove accents somehow.

But we need to fix this issue first. This is not related to this method always returning `true` (even if we have to fix this anyway).

mgautierfr:
I have open the issue https://github.com/kiwix/libkiwix/issues/1101 about the memory allocation error in `kiwix::removeAccents`.

I keep this issue open about the method always returning true.

kelson42:
@BPerlakiH Can younplease open the corresponding issue lin this repo to clearly state what does not work?

BPerlakiH:
[...]

PR #1100, opened 2024-07-12
Fix return value for Manager::parseOpdsDom

BPerlakiH:
Fixes: #1099

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -143,9 +143,8 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
     m_totalBooks = strtoull(libraryNode.child("totalResults").child_value(), 0, 0);
     m_startIndex = strtoull(libraryNode.child("startIndex").child_value(), 0, 0);
     m_itemsPerPage = strtoull(libraryNode.child("itemsPerPage").child_value(), 0, 0);
-    m_hasSearchResult = true;
   } catch(...) {
-    m_hasSearchResult = false;
+    return false;
   }
 
   for (pugi::xml_node entryNode = libraryNode.child("entry"); entryNode;
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -143,7 +143,9 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
     m_totalBooks = strtoull(libraryNode.child("totalResults").child_value(), 0, 0);
     m_startIndex = strtoull(libraryNode.child("startIndex").child_value(), 0, 0);
     m_itemsPerPage = strtoull(libraryNode.child("itemsPerPage").child_value(), 0, 0);
+    m_hasSearchResult = true;
   } catch(...) {
+    m_hasSearchResult = false;
     return false;
   }
 


src/manager.cpp
```diff
@@ -143,9 +143,8 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
     m_totalBooks = strtoull(libraryNode.child("totalResults").child_value(), 0, 0);
     m_startIndex = strtoull(libraryNode.child("startIndex").child_value(), 0, 0);
     m_itemsPerPage = strtoull(libraryNode.child("itemsPerPage").child_value(), 0, 0);
-    m_hasSearchResult = true;
```
mgautierfr, line 146: `m_hasSearchResul` is a member of `Manager` and is public, so part of the API.
Even if, indeed, it is not used in any of our project, we cannot simply not set it.

The `totalResults`/`startIndex`/`itemsPerPage` nodes are present when parsing a OPDS which is a search result. If parsing a root opds flux, they are not present.
So we

mgautierfr, line 146: `m_hasSearchResul` is a member of `Manager` and is public, so part of the API.
Even if, indeed, it is not used in any of our project, we cannot simply not set it.

src/manager.cpp
```diff
@@ -143,9 +143,8 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
...
   } catch(...) {
-    m_hasSearchResult = false;
+    return false;
```
mgautierfr, line 147: The `totalResults`/`startIndex`/`itemsPerPage` nodes are present when parsing a OPDS which is a search result. If parsing a root opds stream, they are not present.
So we must not stop the parsing of the opds stream if they are not present.

veloman-yunkan (commented):
I am not sure what `kiwix::Manager::parseOpdsDom()` is supposed to return. I am also not sure why that method is declared `protected`. I don't know why the `m_hasSearchResult`, `m_totalBooks`, `m_startIndex`, and `m_itemsPerPage` data members of `kiwix::Manager` are needed and why they are public.

Therefore I think that we should address all of the above comprehensively.

mgautierfr (changes requested):
In itself, this function cannot really fails.
The `doc` is already parsed and all methods call on nodes return [null node](https://pugixml.org/docs/manual.html#node_null) or empty string. (Never throw exception).
And we parse the string with `strtoull` which will always return a value (and set errno in case of error (we should check).

So either we have to deeply modify this function (and `Book::updateFromOpds`) to correctly[1] handle missing node/attribute or the `bool` return is useless and we should change it to `void`.

[1] The next question is "What is correctly`. The current version is tolerant to missing data and setup book with empty attribute in this case. Should we stop parsing of the whole stream if one attribute is missing ? Should we continue but tell (with a bool) that something (without telling what) goes wrong ?

mgautierfr:
> I am not sure what kiwix::Manager::parseOpdsDom() is supposed to return. I am also not sure why that method is declared protected.

It should probably be declared private and return `void` (at least with the current code).

> I don't know why the m_hasSearchResult, m_totalBooks, m_startIndex, and m_itemsPerPage data members of kiwix::Manager are needed and why they are public.

It is not clear even to me :)
As said, opds stream can be also for search result. There is this information in the stream and we must provide it to user one way or the other.

I suppose it was used somehow at a moment but I cannot found where/when.
Maybe I was just (excessively) careful and simply parse it and put it here to have the value but we never used them.

src/manager.cpp
BPerlakiH: Ok, makes sense, I have reverted that part.

veloman-yunkan (commented):
This PR doesn't fix anything. The return value of `kiwix::Manager::parseOpdsDom()` is not used in any of kiwix projects. In theory it *could* be used by some code that inherits from `kiwix::Manager` but even then the fix should start from first defining the semantics of the return value of `kiwix::Manager::parseOpdsDom()`. It looks that a proper solution would be to change the return type to `void`, but that would mean a change in the API. A roughly equivalent solution is to state that the return value of `kiwix::Manager::parseOpdsDom()` doesn't matter.

kelson42:
@veloman-yunkan How should we know if parsing went well or not?

veloman-yunkan:
> @veloman-yunkan How should we know if parsing went well or not?

A higher level function `kiwix::Manager::readOpds()` is responsible for that. See [my comment](https://github.com/kiwix/libkiwix/issues/1099#issuecomment-2254454847) in the discussion of the issue.

PR #1105, opened 2024-07-27
Add String Slugification for Generating File Name

ShaopengLin:
kiwix/kiwix-desktop#1134

Introduced function to do basic sanitization to strings that we intend to use as file names in an operating system. Currently, it only sanitizes reserved characters. In the future, this can be extended to handle the complex file name requirements by Windows.

### DIFF
--- a/include/tools.h
+++ b/include/tools.h
@@ -282,5 +282,14 @@ FeedCategories readCategoriesFromFeed(const std::string& content);
  * @return full language name.
  */
 std::string getLanguageSelfName(const std::string& lang);
+
+/**
+ * Slugifies the filename by converting any characters reserved by the operating
+ * system to '_'. Note filename is only the file name and not a path.
+ *
+ * @param filename Valid UTF-8 encoded file name string. 
+ * @return slugified string.
+ */
+std::string getSlugifiedFileName(const std::string& filename);
 }
 #endif // KIWIX_TOOLS_H
--- a/src/tools/stringTools.cpp
+++ b/src/tools/stringTools.cpp
@@ -32,6 +32,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <regex>
 
 /* tell ICU where to find its dat file (tables) */
 void kiwix::loadICUExternalTables()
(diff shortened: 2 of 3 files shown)


kelson42:
@ShaopengLin Thank you! please don't forget to create the appropriate unit test.

kelson42:
Rebasing to benefit of deb package control file fix.

kelson42:
@veloman-yunkan Can we merge that one?

src/tools/stringTools.cpp
```diff
@@ -32,6 +32,13 @@
 
 #include <iostream>
 #include <iomanip>
+#include <regex>
+
+#ifdef _WIN32
+    const std::regex reservedCharsReg("[<>:\"/\\\\|?*]");
+#else
+    const std::regex reservedCharsReg("/");
+#endif
```
veloman-yunkan, line 41: Move this inside `getSlugifiedFileName()` - it is not a function about whose performance we should care.

include/tools.h
```diff
@@ -282,5 +282,14 @@ FeedCategories readCategoriesFromFeed(const std::string& content);
  * @return full language name.
  */
 std::string getLanguageSelfName(const std::string& lang);
+
+/**
+ * Slugifies the filename by converting any characters reserved by the operating
+ * system to '_'.
+ *
+ * @param filename file name string.
+ * @return slugified string.
```
veloman-yunkan, line 291: I think it should be emphasized that the input is a file name, not a filepath

veloman-yunkan (changes requested):

include/tools.h
mgautierfr: Please also mention that this should be a valid utf8 encoded string.

Windows use different encoding. It is better to be explicit here that we want utf8.

src/tools/stringTools.cpp
ShaopengLin: Done

src/tools/stringTools.cpp
```diff
@@ -32,6 +32,13 @@
 
 #include <iostream>
 #include <iomanip>
+#include <regex>
+
+#ifdef _WIN32
+    const std::regex reservedCharsReg("[<>:\"/\\\\|?*]");
```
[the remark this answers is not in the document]
ShaopengLin: Done

include/tools.h
ShaopengLin: Done

veloman-yunkan (approved):

Issue #1087, opened 2024-05-19
`kiwix::getDataDirectory()` should be moved to `kiwix/kiwix-desktop` repository


kelson42:
Reasons are that `kiwix::getDataDirectory()`:
* Is only used by Kiwix Desktop
* Is anyway very specific whatever the version of Kiwix (OS)
* Should be decided outside the libkiwix (not the duty of libkiwix)
* Only used by the aria2c method and should be given there as function parameter.

#1085 should probably be implemented at the same time

kelson42:
@sgourdas Could you please proceed quickly with this, this should be very straight forward

Issue #1085, opened 2024-04-12
Remove `KIWIX_DATA_DIR`


kelson42:
To me this `ENV` variable `KIWIX_DATA_DIR`:
* Is not clear which feature it serves (seems useless to me)
* I'm again to instrument Kiwix Desktop with "hidden" variable environnements.
* Is not documented (the best doc I have is [this comment](https://github.com/kiwix/kiwix-desktop/issues/694#issuecomment-2049562023) from @veloman-yunkan )
* As a bad naming because only part of the data are stored in it

For all this reasons, I ask if we should just not remove it?

This is a follow-up of https://github.com/kiwix/kiwix-desktop/pull/1081 and https://github.com/kiwix/kiwix-desktop/issues/694

kelson42:
@mgautierfr Thank you for the clarification, I have moved the issue. The documentation link is also very helpful to understand.

kelson42:
All of this is Kiwix Desktop related and should be handled in Kiwix Desktop. The need is not clear to me why we should need to customize this from outside the app and it we would have then it would have to be done via a command line argument. On the top the whole Kiwix Desktop UI does not take care of this and continue to show the usual path. This "feature" has been transformed in a bug. We should just remove it.

PR #1107, opened 2024-08-04
Removed getDataDirectory()

sgourdas:
This removes `getDataDirectory()` and properly modifies all its dependencies, so as to migrate it to kiwix-desktop.
- In the process, `makeDirectory()` is also removed after being marked as non essential.
- `startDownload()` now directly receives the download directory and ignores all `dir`s in options.

Implements #1087 and #1085

### DIFF
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -172,7 +172,12 @@ class Downloader
   typedef std::vector<std::pair<std::string, std::string>> Options;
 
  public: // functions
-  Downloader();
+  /*
+  * Create a new Downloader object.
+  *
+  * @param sessionFileDir: The directory where aria2 will store its session file.
+  */
+  explicit Downloader(std::string sessionFileDir);
   virtual ~Downloader();
 
   void close();
--- a/include/downloader.h
+++ b/include/downloader.h
@@ -191,10 +191,11 @@ class Downloader
    * User should call `update` on the returned `Download` to have an accurate status.
    *
    * @param uri: The uri of the thing to download.
+   * @param downloadDir: The download directory where the thing should be stored (takes precedence over any "dir" in `options`).
    * @param options: A series of pair <option_name, option_value> to pass to aria.
    * @return: The newly created Download.
    */
-  std::shared_ptr<Download> startDownload(const std::string& uri, const Options& options = {});
+  std::shared_ptr<Download> startDownload(const std::string& uri, const std::string& downloadDir, Options options = {});
 
   /**
    * Get a download corrsponding to a download id (did)
--- a/include/tools.h
+++ b/include/tools.h
@@ -45,32 +45,6 @@ typedef std::vector<std::string> FeedCategories;
  */
 std::string getCurrentDirectory();
 
-/**
- * Return the data directory
- *
- * The data directory is the default directory where downloaded files
- * should be saved (it can be overriden via the options parameter of
- * `kiwix::Downloader::startDownload()`).
- *
- * Its path can vary and is determined as follows:
- *
- * * `$KIWIX_DATA_DIR` if `$KIWIX_DATA_DIR` environment variable set, *otherwise...*
- * * On Windows:
- *
- *   * `$APPDATA/kiwix` if environment variable `$APPDATA` set, *otherwise...*
- *   * `$USERPROFILE/kiwix` if environment variable `$USERPROFILE` set, *otherwise...*
- *
- * * On other Operating Systems:
- *
- *   * `$XDG_DATA_HOME/kiwix` if environment variable `$XDG_DATA_HOME` set, *otherwise...*
- *   * `$HOME/.local/share/kiwx` if environment variable `$HOME` set, *otherwise...*
- *
- * * Current working directory.
- *
- * @return the path of the data directory (UTF-8 encoded)
- */
-std::string getDataDirectory();
-
 /** Return the path of the executable
  *
  * Some application may be packaged in auto extractible archive (Appimage) and the
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -65,7 +65,6 @@ Aria2::Aria2():
   std::vector<const char*> callCmd;
 
   std::string rpc_port = "--rpc-listen-port=" + to_string(m_port);
-  std::string download_dir = "--dir=" + getDataDirectory();
   std::string session_file = appendToDirectory(getDataDirectory(), "kiwix.session");
   pauseAnyActiveDownloads(session_file);
   std::string session = "--save-session=" + session_file;
@@ -94,7 +93,6 @@ Aria2::Aria2():
   callCmd.push_back("--enable-rpc");
   callCmd.push_back(rpc_secret.c_str());
   callCmd.push_back(rpc_port.c_str());
-  callCmd.push_back(download_dir.c_str());
   if (fileReadable(session_file)) {
     callCmd.push_back(inputFile.c_str());
   }
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -55,17 +55,15 @@ void pauseAnyActiveDownloads(const std::string& ariaSessionFilePath)
 
 } // unnamed namespace
 
-Aria2::Aria2():
+Aria2::Aria2(std::string sessionFileDir):
   mp_aria(nullptr),
   m_port(42042),
   m_secret(getNewRpcSecret())
 {
-  m_downloadDir = getDataDirectory();
-  makeDirectory(m_downloadDir);
   std::vector<const char*> callCmd;
 
   std::string rpc_port = "--rpc-listen-port=" + to_string(m_port);
-  std::string session_file = appendToDirectory(getDataDirectory(), "kiwix.session");
+  std::string session_file = appendToDirectory(sessionFileDir, "kiwix.session");
   pauseAnyActiveDownloads(session_file);
   std::string session = "--save-session=" + session_file;
   std::string inputFile = "--input-file=" + session_file;
--- a/src/aria2.h
+++ b/src/aria2.h
@@ -22,11 +22,10 @@ class Aria2
     std::unique_ptr<Subprocess> mp_aria;
     int m_port;
     std::string m_secret;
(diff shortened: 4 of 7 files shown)


include/downloader.h
```diff
@@ -172,7 +172,7 @@ class Downloader
   typedef std::vector<std::pair<std::string, std::string>> Options;
 
  public: // functions
-  Downloader();
+  Downloader(std::string downloadDir);
```
veloman-yunkan, line 175: 1. make `explicit`
2. In addition to serving as the default download directory, it is also used as the directory where the aria2 session file is saved. In `kiwix-desktop` it is mostly (or, rather, exclusively) used in that second role, since the actual path of the download is set via `kiwix::Downloader::startDownload()` (and thus the default download directory is never used). We can make the download directory a mandatory parameter of `kiwix::Downloader::startDownload()`, then this parameter of the `Downloader` constructor will only serve as the location of the download session file.

src/aria2.h
```diff
@@ -26,7 +26,7 @@ class Aria2
     std::string doRequest(const MethodCall& methodCall);
 
   public:
-    Aria2();
+    Aria2(std::string downloadDir);
```
veloman-yunkan, line 29: make `explicit`

include/tools.h
```diff
@@ -214,6 +188,8 @@ bool fileExists(const std::string& path);
  */
 bool fileReadable(const std::string& path);
 
+bool makeDirectory(const std::string& path);
```
veloman-yunkan, line 191: Please expose `makeDirectory` in the `kiwix` namespace in a separate commit

veloman-yunkan (changes requested):

include/downloader.h
sgourdas: Sorry but I am not sure I understand. I assume you are reffering to the call being made from within `DownloadManager::startDownload` to `kiwix::Downloader::startDownload()`. Download dir is passed there inside the options parameter. Do we want to replace that or keep it and add the download dir as a flat 3rd parameter?

veloman-yunkan: I meant adding `downloadDir` as a mandatory 2nd parameter of `kiwix::Downloader::startDownload()`. Its `options` parameter will stay as an optional 3rd parameter. The download directory passed via the `downloadDir` parameter will override the download directory, if any, passed via `options`.

sgourdas: So inside `kiwix::Downloader::startDownload()` will the `downloadDir` be forced inside the `options` parameter to override whatever (if any) download directory there? And from there it will be inputted in the rest of the methods as is, through `options`.

Is this correct, or am I missing the point of this?

veloman-yunkan: Yes, that's what I meant.

sgourdas:
@veloman-yunkan is this ok?

src/downloader.cpp
```diff
@@ -209,9 +209,16 @@ bool downloadCanBeReused(const Download& d,
...
-std::shared_ptr<Download> Downloader::startDownload(const std::string& uri, const Options& options)
+std::shared_ptr<Download> Downloader::startDownload(const std::string& uri, const std::string& downloadDir, Options options)
 {
   std::unique_lock<std::mutex> lock(m_lock);
+  for (auto it = options.begin(); it != options.end(); it++) {
+    if (it->first == "dir") {
+      it = options.erase(it);
+      break;
+    }
+  }
```
veloman-yunkan, line 220: Better remove *all* occurrences of "dir". Take advantage of [`std::remove_if()` STL algorithm](https://en.cppreference.com/w/cpp/algorithm/remove) (note that it doesn't remove the undesired elements, but moves them to the end, so you'll have to call `options.erase()` too).

src/aria2.cpp
```diff
@@ -55,12 +55,12 @@ void pauseAnyActiveDownloads(const std::string& ariaSessionFilePath)
 
 } // unnamed namespace
 
-Aria2::Aria2():
+Aria2::Aria2(std::string downloadDir):
   mp_aria(nullptr),
   m_port(42042),
-  m_secret(getNewRpcSecret())
+  m_secret(getNewRpcSecret()),
+  m_downloadDir(downloadDir)
```
veloman-yunkan, line 62: `m_downloadDir` can be dropped - it is private and is used only in the constructor

src/aria2.cpp
```diff
@@ -55,12 +55,12 @@ void pauseAnyActiveDownloads(const std::string& ariaSessionFilePath)
 
 } // unnamed namespace
 
-Aria2::Aria2():
+Aria2::Aria2(std::string downloadDir):
```
veloman-yunkan, line 58: `downloadDir`  → `sessionFileDir`

src/aria2.h
```diff
@@ -26,7 +26,7 @@ class Aria2
     std::string doRequest(const MethodCall& methodCall);
 
   public:
-    Aria2();
+    explicit Aria2(std::string downloadDir);
```
veloman-yunkan, line 29: `downloadDir`  → `sessionFileDir`

src/downloader.cpp
```diff
@@ -125,8 +125,8 @@ void Download::cancelDownload()
 }
 
 /* Constructor */
-Downloader::Downloader() :
-  mp_aria(new Aria2())
+Downloader::Downloader(std::string downloadDir) :
```
veloman-yunkan, line 128: `downloadDir`  → `sessionFileDir`

include/downloader.h
```diff
@@ -172,7 +172,7 @@ class Downloader
   typedef std::vector<std::pair<std::string, std::string>> Options;
 
  public: // functions
-  Downloader();
+  explicit Downloader(std::string downloadDir);
```
veloman-yunkan, line 175: `downloadDir`  → `sessionFileDir`

include/downloader.h
```diff
@@ -194,7 +194,7 @@ class Downloader
    * @param options: A series of pair <option_name, option_value> to pass to aria.
    * @return: The newly created Download.
    */
-  std::shared_ptr<Download> startDownload(const std::string& uri, const Options& options = {});
+  std::shared_ptr<Download> startDownload(const std::string& uri, const std::string& downloadDir, Options options = {});
```
veloman-yunkan, line 197: This change deserves a commit of its own - the description of the change `Remove getDataDirectory` doesn't cover it at all. In that commit, please also stop passing the `--dir` option to the `aria2c` command line (in the `Aria2` constructor).

veloman-yunkan (changes requested):

include/tools.h
```diff
@@ -214,6 +214,8 @@ bool fileExists(const std::string& path);
...
```
veloman-yunkan, line 191: Now I wonder if we need `makeDirectory()` in `libkiwix` at all. In `kiwix-desktop` we can use [`QDir::mkpath()`](https://doc.qt.io/qt-5/qdir.html#mkpath). So if we require that `sessionFileDir` refers to an existing directory we can drop `makeDirectory()` as well.

include/downloader.h
veloman-yunkan, line 197: Documentation must be updated.

include/downloader.h
```diff
@@ -172,7 +172,7 @@ class Downloader
   typedef std::vector<std::pair<std::string, std::string>> Options;
 
  public: // functions
-  Downloader();
+  explicit Downloader(std::string sessionFileDir);
```
veloman-yunkan, line 175: The parameter must be documented

veloman-yunkan (changes requested):

include/tools.h
sgourdas: > So if we require that sessionFileDir refers to an existing directory we can drop makeDirectory() as well.

What I am thinking of in regards to that is if we are ok on the first run.

From what I see the only call to `makeDirectory()` derives from `createDownloader()` which uses `getDataDirectory()`. So if we use `QDir::mkpath()` in that we should be ok, so I will go on with the removal of this also.

include/downloader.h
```diff
@@ -165,14 +165,16 @@ class Download {
  * A Downloader manages `Download` using aria2 in the background.
  * `Downloader` is threadsafe.
  * However, the returned `Download`s are NOT threadsafe.
+ * 
+ * @param sessionFileDir: The directory where aria2 will store its session file.
```
veloman-yunkan, line 169: This is the documentation of the `Downloader` class. You had to document this parameter in the constructor (which so far had no documentation since there was nothing to document).

src/aria2.cpp
```diff
@@ -55,18 +55,16 @@ void pauseAnyActiveDownloads(const std::string& ariaSessionFilePath)
...
   mp_aria(nullptr),
   m_port(42042),
   m_secret(getNewRpcSecret())
 {
-  m_downloadDir = getDataDirectory();
-  makeDirectory(m_downloadDir);
   std::vector<const char*> callCmd;
 
   std::string rpc_port = "--rpc-listen-port=" + to_string(m_port);
   std::string download_dir = "--dir=" + getDataDirectory();
```
veloman-yunkan: `getDataDirectory()` is removed in this commit but is still used here.

veloman-yunkan (changes requested):
I think that the commits must be reordered - you should start with the enhancement of the `Downloader::startDownload()`

include/downloader.h
```diff
@@ -193,10 +193,11 @@ class Downloader
    * User should call `update` on the returned `Download` to have an accurate status.
    *
    * @param uri: The uri of the thing to download.
+   * @param downloadDir: The download directory where the thing should be stored.
```
veloman-yunkan, line 196: Add a note that it takes precedence over the download directory passed via `options` (if any)

include/downloader.h
sgourdas: Yeah, I was not sure if the documentation of the class meant documentation for the class's constructor as well -- as if we assume one means the other. Fixing...

sgourdas:
@veloman-yunkan how is it now?

mgautierfr:
LGTM too

PR #1112, opened 2024-08-13
Public i18n API

### LINKED ISSUE #1103: Expose support for i18n/translation in the public C++ API
`libkiwix` server component has been internationalized a while ago, but the support for that functionality has not been exposed in the public C++ API.

kiwix/kiwix-desktop#972 has now put enough pressure on `libkiwix` to finally start providing translation services to its C++ clients.

veloman-yunkan:
Fixes #1103

- General purpose i18n/translation API of libkiwix is exposed
- API for category translation is introduced

I tested this PR in `kiwix-desktop` with the following experimental change to `ContentManager::setCategories()`:

```c++
void ContentManager::setCategories()
{
    QStringList categories;
    if (m_local) {
        const auto lang = "test"; //QLocale::languageToString(QLocale().language()).toStdString();
        auto categoryData = mp_library->getKiwixLibrary()->getBooksCategories();
        for (auto category : categoryData) {
            const auto translatedCategory = kiwix::translateBookCategory(lang, category);
            auto categoryName = QString::fromStdString(translatedCategory);
            categories.push_back(categoryName);
        }
        m_categories = categories;
        emit(categoriesLoaded(m_categories));
        return;
    }
    m_remoteLibraryManager.getCategoriesFromOpds();
}
```

I hardcoded the language because I couldn't immediately figure out how to set the locale language - I was getting "`Gtk-WARNING: Locale not supported by C library. Using the fallback 'C' locale.`" when running `kiwix-desktop` with `LC_NAME=test`, `LC_NAME=fr`, `LC_NAME=fr_FR`, `LC_NAME=fr_FR.UTF-8`, etc)

Note that for full fix of kiwix/kiwix-desktop#972 `ContentManager::updateCategories()` will need to be updated too.


ShaopengLin:
> I hardcoded the language because I couldn't immediately figure out how to set the locale language - I was getting "Gtk-WARNING: Locale not supported by C library. Using the fallback 'C' locale." when running kiwix-desktop with LC_NAME=test, LC_NAME=fr, LC_NAME=fr_FR, LC_NAME=fr_FR.UTF-8, etc)

The string returned isn't really ISO 639 if you use language(). QLocale's name() method does this and just needs a small parsing before we pass it to libkiwix.

ShaopengLin (approved):
LGTM. Confirmed that the integration works with Kiwix Desktop.

scripts/kiwix-compile-i18n
```diff
@@ -61,7 +61,7 @@ lang_table_entry_cxx_template = '''
 
 cxxfile_template = '''// This file is automatically generated. Do not modify it.
 
-#include "server/i18n.h"
```
mgautierfr, line 64: This first commit introduce the file `test/data/multilang.zim`. It should not.

mgautierfr (changes requested):
LGTM except the binary file (test zim file) that should be removed before merge.

scripts/kiwix-compile-i18n
veloman-yunkan: Thanks for catching this. It has been fixed.

Issue #1081, opened 2024-05-12
Release 14.0.0


kelson42:
- [x] Check if library dependence version check is updated in Meson file
- [x] Check if there is an ABI break we would have missed
- [ ] Update `subprojects` with appropriate library dependence versions (like in kiwix-build)
- [x] Merge/apply latest translations from Translatewiki
- [x] Secure the CI is green on git `main`
- [x] Kiwix-Build is OK
- [x] Update the Changelog
- [x] Update version
- [x] Close current milestone and create new one incrementaly (_a priori_ a minor version)
- [x] Create a tag on git
- [x] Secure new source/sbinaries are published on http://download.kiwix.org/release/libkiwix
- [x] Update the Github release with the Changelog
- [ ] Create new empty entry in Changelog (placeholder for future entries)
- [ ] Publicize these new versions

kelson42:
@veloman-yunkan I plan to release libkiwix14 tomorrow. Any reason why we shoukd wait?

veloman-yunkan:
I think that the right time to release libkiwix 14.0.0 is on 14th at 00:00. Other than that I am not aware of any reason to wait.

PR #1116, opened 2024-09-01
Bump-up version to 14.0.0

kelson42:
Next version will be `14.0.0` as we have breaking changes. Doing this right now to be able to configure properly dependencies in `kiwix-tools` and `kiwix-desktop`.

Related to #1081

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -35,11 +35,12 @@ Description: library of common code for Kiwix (development)
  .
  This package contains development files.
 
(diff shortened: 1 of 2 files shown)


mgautierfr:
> Doing this right now to be able to configure properly dependencies in kiwix-tools and kiwix-desktop.

Would have been nice to have PR in kiwix-tools and kiwix-desktop in the same time ...

PR #1123, opened 2024-09-05
Handling of external app links in the viewer

### LINKED ISSUE #1090: Relax the Content-Security-Policy to allow `mailto:` links in the Kiwix Serve iframe
This relates to issue https://github.com/kiwix/kiwix-tools/issues/680 on Kiwix Tools. Most links that are not considered secure are blocked by the CSP and/or sandbox, and that includes `mailto:` links, which are blocked in Chromium browsers (whereas they seem not be blocked by Firefox). This also affects Kiwix JS Browser Extension and the PWA.

I have determined (in the PWA) that relaxing the CSP ***for the content in the iframe*** should fix this. All that is needed is to add `mailto:` to the list of exceptions in the `internalServer.cpp` code. I am pretty sure it won't be necessary to alter the CSP of the outer document, but that would need testing. The code is here:

https://github.com/kiwix/libkiwix/blob/main/src/server/internalServer.cpp#L1110

A Zimit2 ZIM with a `maitlo:` link is given in the linked issue for testing.

veloman-yunkan:
Fixes kiwix/kiwix-tools#680 
Also closes #1090, but not as requested in the title & description of that issue because I couldn't make it work and chose the general approach proposed in an [afterthought](https://github.com/kiwix/libkiwix/issues/1090#issuecomment-2147755667) that was initially [posted](https://github.com/kiwix/kiwix-tools/issues/680#issuecomment-2147208056) in #680.

Links that should be handled/opened by external applications - such as email addresses (`mailto:`), phone numbers (`tel:`), etc - are opened by the viewer in a new tab/window, thus avoiding any issues with content security policy.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -335,6 +335,13 @@ function getRealHref(target) {
   return target_href;
 }
 
+function setHrefAvoidingWombatRewriting(target, url) {
+  const old_no_rewrite = target._no_rewrite;
+  target._no_rewrite = true;
+  target.setAttribute("href", url);
+  target._no_rewrite = old_no_rewrite;
+}
+
 function onClickEvent(e) {
   const iframeDocument = contentIframe.contentDocument;
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
@@ -351,7 +358,7 @@ function onClickEvent(e) {
       if ( e.ctrlKey || e.shiftKey ) {
         // The link will be loaded in a new tab/window - update the link
         // and let the browser handle the rest.
-        target.setAttribute("href", possiblyBlockedLink);
+        setHrefAvoidingWombatRewriting(target, possiblyBlockedLink);
       } else {
         // Load the external URL in the viewer window (rather than iframe)
         contentIframe.contentWindow.parent.location = possiblyBlockedLink;
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -310,6 +310,12 @@ function blockLink(url) {
        : url;
 }
 
+function urlMustBeHandledByAnExternalApp(url) {
+  const WHITELISTED_URL_SCHEMATA = ['http:', 'https:', 'about:', 'javascript:'];
+
+  return WHITELISTED_URL_SCHEMATA.indexOf(url.protocol) == -1;
+}
+
 function isExternalUrl(url) {
   if ( url.startsWith(window.location.origin) )
     return false;
@@ -334,7 +340,13 @@ function onClickEvent(e) {
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
   if (target !== null && "href" in target) {
     const target_href = getRealHref(target);
-    if (isExternalUrl(target_href)) {
+    const target_url = new URL(target_href, iframeDocument.location);
+    const isExternalAppUrl = urlMustBeHandledByAnExternalApp(target_url);
+    if ( isExternalAppUrl && !viewerSettings.linkBlockingEnabled ) {
+        target.setAttribute("target", "_blank");
+    }
+
+    if (isExternalAppUrl || isExternalUrl(target_href)) {
       const possiblyBlockedLink = blockLink(target_href);
       if ( e.ctrlKey || e.shiftKey ) {
         // The link will be loaded in a new tab/window - update the link
@@ -343,6 +355,7 @@ function onClickEvent(e) {
       } else {
         // Load the external URL in the viewer window (rather than iframe)
         contentIframe.contentWindow.parent.location = possiblyBlockedLink;
+        e.preventDefault();
       }
     }
   }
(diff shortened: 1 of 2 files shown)


static/skin/viewer.js
```diff
@@ -310,6 +310,17 @@ function blockLink(url) {
        : url;
 }
 
+function isExternalAppUrl(url) {
+  const EXTERNAL_APP_PROTOCOLS = ['mailto', 'tel', 'geo', 'magnet', 'zoomus'];
```
Jaifroid, line 314: I personally wouldn't hardcode a list of URI schemata given that there are an infinite number of possible schemata, since webapps can define them arbitrarily. If we were only dealing with Wikimedia ZIMs, a hardcoded list might be appropriate, but for Zimit we could have literally any protocol, e.g. `skype:`, `bingmaps:`, `comgooglemaps:`, `comgooglemaps-x-callback:`, etc., etc., etc....? I think we need a different approach: have a list of URI schemata ("protocols") that we know can or must be opened in the app (I only found `javascript:` and `about:`), and open all the rest in a new tab or window.

Jaifroid (commented):
As mentioned in comment, I'd suggest having a whitelist rather than a blacklist, because the blacklist could be infinite, whereas there are only a couple of protocols in the whitelist.

One thing to be aware of is that I found several cases of PDFs served with http:// protocols instead of https:// protocols in Zimit archives, so I had to include a normalizing step in the code in Kiwix JS here https://github.com/kiwix/kiwix-js/blob/main/www/js/app.js#L2451. Now, I don't know if that is only relevant for Zimit1, or if Zimit2 might have this issue or not. I don't know if you still want to support Zimit1 ZIMs that are likely to be around for quite some time? @benoit74?

benoit74:
I have exactly the same remark / recommendation as @Jaifroid (who has been quicker than I to review this ^^)

In ZIMs, we can indeed have http:, https:, javascript:, about: (and maybe others, I don't really know tbh) that should be handled, all others should be just left for the browser to proceed. All these links should raised a warning from my PoV since the user is not navigating inside the ZIM but going somewhere else. When navigating inside the ZIM, only relative links are used normally.

Usage of http: and https: links are obviously way more common in Zimit than in other scrapers, but we can definitely have both in any (e.g. in ifixit we have many "external" links posted by users in the guides, and these could be http or https depending on what the target server supports or what the user decided to use ; and we do not try to guess if any http link could be replaced by an https one, this is too daunghting). Regarding links processed dynamically by wombat, when internal to the ZIM they use the same protocol as the currently displayed HTML page.

Regarding Zimit 1, currently openZIM does not have anymore Zimit 1 ZIMs published, they have all been either updated to Zimit 2 or archived. That been said, there is probably many in the wild, and we obviously have no control on when users will drop them for a more recent version.

kelson42:
FYI and AFAIK there is such a list of protocol schemes in zimwriterfs source code as well.

Jaifroid:
I think we'll need to test this thoroughly against any links transformed by Wombat. In @mgautierfr's original workaround, it was necessary to turn off Wombat processing temporarily when the user clicked a link in order to get the "real" URL as opposed to the Wombat-transformed URL. It's not clear to me whether that is relevant to this PR (sorry, I'm still travelling, so don't have access to a test environment right now).

static/skin/viewer.js
veloman-yunkan: Addressed in a fixup commit

veloman-yunkan:
> All these links should raised a warning from my PoV since the user is not navigating inside the ZIM but going somewhere else.

Should this be implemented in this PR?

benoit74:
> Should this be implemented in this PR?
No clue ^^

kelson42:
> > All these links should raised a warning from my PoV since the user is not navigating inside the ZIM but going somewhere else.
> 
> Should this be implemented in this PR?

`-b --blockexternal` should apply to all external URLs if activated, otherwise I see no reason why we should display a warning. This is at least not the behaviour which has been chosen so far.

veloman-yunkan:
> > > All these links should raised a warning from my PoV since the user is not navigating inside the ZIM but going somewhere else.
> > 
> > 
> > Should this be implemented in this PR?
> 
> @veloman-yunkan `-b --blockexternal` should apply to all external URLs if activated, otherwise I see no reason why we should display a warning. This is at least not the behaviour which has been chosen so far. Does it apply?

Done. Any links containing a non-whitelisted URL schema are now considered external with respect to link blocking and routed through the link-blocker page. Otherwise such links are opened in a new window/tab.

Also discovered a bug related to handling of ctrl (or shift) clicked links for Zimit2 ZIMs and fixed it in a separate commit.

kelson42 (approved):
LGTM

veloman-yunkan:
> LGTM

I think that the additional bugfix commit was not included in your review.

PR #1124, opened 2024-09-07
Better check libzim version

kelson42:
We check maximum version of libzim to avoid linking against a version which might have breaking changes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,7 +35,9 @@ else
   error('Cannot found header mustache.hpp')
 endif
 
-libzim_dep = dependency('libzim', version : '>=9.0.0', static:static_deps)
+libzim_dep = dependency('libzim', version:'>=9.0.0', static:static_deps)
+libzim_dep = dependency('libzim', version:'<10.0.0', static:static_deps)
+
 if not compiler.has_header_symbol('zim/zim.h', 'LIBZIM_WITH_XAPIAN', dependencies: libzim_dep)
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -35,7 +35,9 @@ else
   error('Cannot found header mustache.hpp')
 endif
 
-libzim_dep = dependency('libzim', version : '>=9.0.0', static:static_deps)
+libzim_dep = dependency('libzim', version:'>=9.0.0', static:static_deps)
+libzim_dep = dependency('libzim', version:'<10.0.0', static:static_deps)
```
veloman-yunkan, line 39: IMHO, this doesn't add a version requirement. It rather overrides the preceding one (so only `version:'<10.0.0'` stays in effect). Yes, the build will likely fail if `libzim` newer than `9.0.0` is not found, but then meson is allowed to pick up a version older than 10.0.0 (including a version lower than 9.0.0).

veloman-yunkan: The correct way is to [specify multiple restrictions](https://mesonbuild.com/Reference-manual_functions.html#dependency_version):  `version:['>=9.0.0', '<10.0.0']`

I am going to fix it in #1133

kelson42: Hmmm, this approach has been validated on an other repo. by @mgautierfr. I don't remember the results of testing I made at that time... but AFAIK it tests the first one and tests then the second one... which is what we want.

@veloman-yunkan Do you have a better formalism?

veloman-yunkan: Done

PR #1127, opened 2024-09-09
Illustration size on book tile is fixed to 48x48

### LINKED ISSUE #1102: ZIM Illustration is resized in the catalog UI and hence a bit blurry
Current icon on ZIMs in the library catalog are always a bit blurry even if the real image if reasonably sharp.

image

Real image:

image

Problem is that the image is displayed as a div background-image, and the div is not exactly 48x48 pixels, so the image is resized to few more pixels (55x60 on my device / browser). And obviously adding few pixels is just adding blur.

Image if I force background size to 48x48 pixel:

image

image

PS: note that just adding the "background-size: 48px 48px" is just a trick to showcase the issue, it is not deemed to be the best approach, it has not been tested on all screen sizes / browsers and it has a side effect that the image is not centered vertically and horizontal anymore. If we go this way we need at least the "background-position: center;" additional style.

veloman-yunkan:
Fixes #1102

### DIFF
--- a/static/skin/index.css
+++ b/static/skin/index.css
@@ -189,6 +189,8 @@
     align-items: center;
     align-content: center;
     justify-content: center;
+    width:  48px;
+    height: 48px;
     margin: 10px 0 0 10px;
 }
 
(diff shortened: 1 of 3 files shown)


kelson42:
To me it looks good! @benoit74 ?

benoit74:
Looks good, to be tested in nightly obviously

benoit74 (approved):

benoit74:
I confirm this work as intended in nightly, thank you!

PR #1128, opened 2024-09-09
Deprived pseudolang mul of a self-name

### LINKED ISSUE #1098: Remove Multiple Languages filter
As discussed live, we should not have a “Multiple Languages” filter entry. We should not have any ZIM with `Languages: mul` metadata and the content team will fix the ZIM but we should not have any handling for this and thus if such a ZIM makes it into the library, it should use default label (`mul`)

Screenshot 2024-07-03 at 16 40 05

veloman-yunkan:
Fixes #1098

ICU package contains a special code "mul" with a self-name of "multiple languages". libkiwix now suppresses that. As a result the self-name of "mul" (like for any other unknown language code) is the code itself (i.e. "mul").

The most prominent user-visible effect of this change is that the language filter in the library page no longer contains a "Multiple languages" entry if there is a legacy ZIM file with the language set to "mul" - that entry now shows up as "Mul".

### DIFF
--- a/src/tools/languageTools.cpp
+++ b/src/tools/languageTools.cpp
@@ -68,6 +68,7 @@ void fillLanguagesMap()
         const kiwix::ICULanguageInfo lang(*icuLangPtr);
         iso639_3.insert({lang.iso3Code(), lang.selfName()});
     }
+    iso639_3.erase("mul");
 }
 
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
LGTM

Issue #815, opened 2022-09-07
We need to test kiwix-serve UI (with frontend js code)


mgautierfr:
With time, we have created a more and more complex frontend with a lot of javascript.
However, our unit/integrations tests only check the backend code.
We need to find a way to add testing on the js/frontend code.

juuz0:
First name that comes to mind :>
https://jestjs.io

If you could setup meson for it, I can start writing tests

kelson42:
Here is the list of supported browsers https://github.com/kiwix/libkiwix/issues/1029#issuecomment-2352741810

Issue #31, opened 2017-03-15
Rethink how we integrate resources in binaries.


mgautierfr:
We have currently a small solution to embedded resource file in binaries.
The resource file is just store in a C array and the code just read it and interpret it as a file content.

This work well but it could be improved :
- Some files/resources work together and can be seen as a pack. (stop words list for xapian, all html/css/js for the topbar, ...) 
- It is somehow complicated to allow customization (kiwix/kiwix-tools#17) or update of the resources. For now the two solutions are either to recompile the binary or use env variables to overwrite the content.

This is somehow a pity as we are a project who embedded different files in a read efficient archive, handling file hierarchy and so. We can probably use our own code base to store and read those resource files :)

kelson42:
Stop words are meanwhile stored in the ZIM files themselves.

kelson42:
@mgautierfr What would be a better approach?

kelson42:
Why not just using a zip blob? Would be easy to read in c++ and easy to build for anybody whatever the platform. We could then just allow to load an external ZIP fs path tomcustomize things.

mgautierfr:
At that time, I was thinking about using a zim archive. Now I would use a [jubako](https://github.com/jubako/jubako) based container.

kelson42:
We won't use anything overkill. Package should better be done via tools people are familiar with. Why is a simple tar or zip not good enough?

PR #1131, opened 2024-09-11
Ungarbled binary resources

### LINKED ISSUE #1129: Binary resources in libkiwix have been broken by dos2unix conversion
The last commit of #1113 (https://github.com/kiwix/libkiwix/pull/1113/commits/4812fb18f6374eaafb98fca9bae075987c8ab436) has broken the binary resources of libkiwix.

This shows up as PNG icons failing to display in the frontend:

image

veloman-yunkan:
Fixes #1129


veloman-yunkan:
This PR fixes the bug introduced by #1113. More importantly it provides protection (in the form of a unit test) against similar future problems.

There are a few things that I don't like about this PR:

- It works by invoking an external command (a small python script) to perform a HTTP request and compute the hash of the response. I chose that approach in order to avoid having to introduce a dependency on a library providing support for SHA1 computation.
- The list of static resources checked by the new test (as well a few old ones) is hardcoded and must be manually updated when new resources are added.
- The list of file extensions considered to be binary (and thus to be excluded from dos2unix conversion) is hardcoded and must be updated when new types of binary resources are introduced.

veloman-yunkan:
@kelson42 How shall we proceed with this PR without @mgautierfr? I am most interested in feedback on the proposed method of protection against bugs similar to the one being fixed.

kelson42:
@veloman-yunkan We have to deal without his feedback. But I have two remarks:
* I'm concerned by the growing difficulty of updating these static files
* This needs at least an update IMHO of https://github.com/kiwix/libkiwix/blob/main/README.md#static-files-compilation

veloman-yunkan:
> * The list of file extensions considered to be binary (and thus to be excluded from dos2unix conversion) is hardcoded and must be updated when new types of binary resources are introduced.

Now when new types of resources are added that won't go unnoticed - the build will fail unless the list of known extensions is updated.

veloman-yunkan:
> @veloman-yunkan We have to deal without his feedback. But I have two remarks:
> 
>   * I'm concerned by the growing difficulty of updating these static files

With the newly committed change no extra difficulty regarding maintenance of static resource files is being introduced  (as long as we are concerned only with things that have to be memorized). The second point in [my earlier comment](https://github.com/kiwix/libkiwix/pull/1131#issuecomment-2343809865) is just a reiteration of a (slight) problem that existed before this PR. There is a way to add a test that would mostly eliminate that issue but IMHO we shouldn't spend any effort on it until it proves to be a real rather than a hypothetical problem. 

>   * This needs at least an update IMHO of https://github.com/kiwix/libkiwix/blob/main/README.md#static-files-compilation

I reread the README but don't think that it needs to be updated. If the pressure to do so mounts, we better enhance our test suite as mentioned in the previous paragraph.

>   * I wonder if tackling [We need to test frontend js code. #815](https://github.com/kiwix/libkiwix/issues/815) would not be a better approach

That may solve some of the addressed concerns (though at the cost of a significantly larger effort) but should not be considered as a substitute for what can be checked on the unit-test level.

>   * What about [Rethink how we integrate resources in binaries. #31](https://github.com/kiwix/libkiwix/issues/31) ?

Our mechanism for embedding resources has been significantly enhanced since when the original issue was filed. That on the one hand makes that ticket more justified since the system has become more complex and perhaps can be optimized. On the other hand the task has become more difficult too and maybe we shouldn't touch what works reasonably well until we notice that we have to spend excessive hours on solving issues rooted in the current implementation of how we deal with static resources.

PR #1132, opened 2024-09-12
Refactor getBestPublicIp for all valid ips

sgourdas:
This refactors `getBestPublicIp(bool)` to `getBestPublicIps(IpMode)` so it can correctly return all attached ips for both protocols ipv4 and ipv6.

Fixes kiwix/kiwix-tools#703
Fixes https://github.com/kiwix/kiwix-tools/issues/709

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -16,7 +16,7 @@
 
 namespace kiwix {
 
-enum class IpMode { ipv4, ipv6, all };
+enum class IpMode { IPV4, IPV6, ALL };
 typedef zim::size_type size_type;
 typedef zim::offset_type offset_type;
 
--- a/include/common.h
+++ b/include/common.h
@@ -16,7 +16,7 @@
 
 namespace kiwix {
 
-enum class IpMode { IPV4, IPV6, ALL };
+enum class IpMode { IPV4, IPV6, ALL, AUTO }; // AUTO: Server decides the protocol
 typedef zim::size_type size_type;
 typedef zim::offset_type offset_type;
 
--- a/include/server.h
+++ b/include/server.h
@@ -81,7 +81,7 @@ namespace kiwix
        bool m_withTaskbar = true;
        bool m_withLibraryButton = true;
        bool m_blockExternalLinks = false;
-       IpMode m_ipMode = IpMode::ipv4;
+       IpMode m_ipMode = IpMode::IPV4;
        int m_ipConnectionLimit = 0;
        std::unique_ptr<InternalServer> mp_server;
   };
--- a/include/server.h
+++ b/include/server.h
@@ -64,8 +64,8 @@ namespace kiwix
        void setBlockExternalLinks(bool blockExternalLinks)
         { m_blockExternalLinks = blockExternalLinks; }
        void setIpMode(IpMode mode) { m_ipMode = mode; }
-       int getPort();
-       std::string getAddress();
+       int getPort() const;
+       std::string getAddress() const;
        IpMode getIpMode() const;
 
      protected:
--- a/include/server.h
+++ b/include/server.h
@@ -22,7 +22,7 @@
 
 #include <string>
 #include <memory>
-#include "common.h"
+#include "tools.h"
 
 namespace kiwix
 {
@@ -52,7 +52,7 @@ namespace kiwix
        void stop();
 
        void setRoot(const std::string& root);
-       void setAddress(const std::string& addr) { m_addr = addr; }
+       void setAddress(const std::string& addr);
        void setPort(int port) { m_port = port; }
        void setNbThreads(int threads) { m_nbThreads = threads; }
        void setMultiZimSearchLimit(unsigned int limit) { m_multizimSearchLimit = limit; }
@@ -65,14 +65,14 @@ namespace kiwix
         { m_blockExternalLinks = blockExternalLinks; }
        void setIpMode(IpMode mode) { m_ipMode = mode; }
        int getPort() const;
-       std::string getAddress() const;
+       IpAddress getAddress() const;
        IpMode getIpMode() const;
 
      protected:
        std::shared_ptr<Library> mp_library;
        std::shared_ptr<NameMapper> mp_nameMapper;
        std::string m_root = "";
-       std::string m_addr = "";
+       IpAddress m_addr;
        std::string m_indexTemplateString = "";
        int m_port = 80;
        int m_nbThreads = 1;
@@ -81,7 +81,7 @@ namespace kiwix
        bool m_withTaskbar = true;
        bool m_withLibraryButton = true;
        bool m_blockExternalLinks = false;
-       IpMode m_ipMode = IpMode::IPV4;
+       IpMode m_ipMode = IpMode::AUTO;
        int m_ipConnectionLimit = 0;
        std::unique_ptr<InternalServer> mp_server;
   };
--- a/include/tools.h
+++ b/include/tools.h
@@ -24,6 +24,7 @@
 #include <vector>
 #include <map>
 #include <cstdint>
+#include "common.h"
 
 namespace kiwix
 {
@@ -214,9 +215,10 @@ std::map<std::string, IpAddress> getNetworkInterfacesIPv4Or6();
 std::map<std::string, std::string> getNetworkInterfaces();
 
 /** Provides the best IP address
- * This function provides the best IP address from the list given by getNetworkInterfacesIPv4Or6()
+ * This function provides the best IP addresses for both ipv4 and ipv6 protocols,
+ * in an IpAddress struct, based on the list given by getNetworkInterfacesIPv4Or6()
  */
-std::string getBestPublicIp(bool ipv6);
+IpAddress getBestPublicIps();
 
 /** Provides the best IPv4 adddress
  * Equivalent to getBestPublicIp(false). Provided for backward compatibility
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -75,12 +75,12 @@ void Server::setRoot(const std::string& root)
   }
 }
 
-int Server::getPort()
+int Server::getPort() const
 {
   return mp_server->getPort();
 }
 
-std::string Server::getAddress()
+std::string Server::getAddress() const
 {
   return mp_server->getAddress();
 }
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -75,12 +75,24 @@ void Server::setRoot(const std::string& root)
   }
 }
 
+// FIXME: this method is implemented under the assumption that it is invoked only once (per object).
+void Server::setAddress(const std::string& addr)
+{
+  if (addr.empty()) return;
+
+  if (addr.find(':') != std::string::npos) { // IPv6
+    m_addr.addr6 = (addr[0] == '[') ? addr.substr(1, addr.length() - 2) : addr; // Remove brackets if any
+  } else {
+    m_addr.addr = addr;
+  }
+}
+
 int Server::getPort() const
 {
   return mp_server->getPort();
 }
 
-std::string Server::getAddress() const
+IpAddress Server::getAddress() const
 {
   return mp_server->getAddress();
 }
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -407,7 +407,7 @@ public:
 
 InternalServer::InternalServer(LibraryPtr library,
                                std::shared_ptr<NameMapper> nameMapper,
-                               std::string addr,
+                               IpAddress addr,
                                int port,
                                std::string root,
                                int nbThreads,
@@ -461,21 +461,30 @@ bool InternalServer::start() {
   sockAddr6.sin6_family = AF_INET6;
   sockAddr6.sin6_port = htons(m_port);
 
-  if (m_addr.empty()) {
-    if (0 != INADDR_ANY) {
-      sockAddr6.sin6_addr = in6addr_any;
-      sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
-    }
-    m_addr = kiwix::getBestPublicIp(m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL);
+  if (m_addr.addr.empty() && m_addr.addr6.empty()) { // No ip address provided
+    if (m_ipMode == IpMode::AUTO) m_ipMode = IpMode::ALL;
+    sockAddr6.sin6_addr = in6addr_any;
+    sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
+    IpAddress bestIps = kiwix::getBestPublicIps();
+    if (m_ipMode == IpMode::IPV4 || m_ipMode == IpMode::ALL) m_addr.addr = bestIps.addr;
+    if (m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL) m_addr.addr6 = bestIps.addr6;
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::ALL;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    const std::string addr = !m_addr.addr.empty() ? m_addr.addr : m_addr.addr6;
+
+    if (m_ipMode != kiwix::IpMode::AUTO) {
+      std::cerr << "ERROR: When an IP address is provided the IP mode must not be set" << std::endl;
       return false;
     }
+
+    bool validV4 = inet_pton(AF_INET, m_addr.addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    bool validV6 = inet_pton(AF_INET6, m_addr.addr6.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
+
+    if (!validV4 && !validV6) {
+      std::cerr << "ERROR: invalid IP address: " << addr << std::endl;
+      return false;
+    } 
+
+    m_ipMode = !m_addr.addr.empty() ? IpMode::IPV4 : IpMode::IPV6; 
   }
 
   if (m_ipMode == IpMode::ALL) {
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -85,6 +85,19 @@ namespace kiwix {
 namespace
 {
 
+bool ipAvailable(const std::string addr)
+{
+  auto interfaces = kiwix::getNetworkInterfacesIPv4Or6();
+
+  for (const auto& [_, interfaceIps] : interfaces) {
+    if ((interfaceIps.addr == addr) || (interfaceIps.addr6 == addr)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 inline std::string normalizeRootUrl(std::string rootUrl)
 {
   while ( !rootUrl.empty() && rootUrl.back() == '/' )
@@ -483,6 +496,11 @@ bool InternalServer::start() {
       std::cerr << "ERROR: invalid IP address: " << addr << std::endl;
       return false;
     } 
+
+    if (!ipAvailable(addr)) {
+      std::cerr << "ERROR: IP address is not available on this system: " << addr << std::endl;
+      return false;
+    }
 
     m_ipMode = !m_addr.addr.empty() ? IpMode::IPV4 : IpMode::IPV6; 
   }
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -466,25 +466,25 @@ bool InternalServer::start() {
       sockAddr6.sin6_addr = in6addr_any;
       sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
     }
-    m_addr = kiwix::getBestPublicIp(m_ipMode == IpMode::ipv6 || m_ipMode == IpMode::all);
+    m_addr = kiwix::getBestPublicIp(m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL);
   } else {
     bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
     bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
     if (ipv6){
-       m_ipMode = IpMode::all;
+       m_ipMode = IpMode::ALL;
     } else if (!ipv4) {
       std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
       return false;
     }
   }
 
-  if (m_ipMode == IpMode::all) {
+  if (m_ipMode == IpMode::ALL) {
     flags|=MHD_USE_DUAL_STACK;
-  } else if (m_ipMode == IpMode::ipv6) {
+  } else if (m_ipMode == IpMode::IPV6) {
     flags|=MHD_USE_IPv6;
   }
 
-  struct sockaddr* sockaddr = (m_ipMode==IpMode::all || m_ipMode==IpMode::ipv6)
+  struct sockaddr* sockaddr = (m_ipMode==IpMode::ALL || m_ipMode==IpMode::IPV6)
                               ? (struct sockaddr*)&sockAddr6
                               : (struct sockaddr*)&sockAddr4;
 
--- a/src/server/internalServer.h
+++ b/src/server/internalServer.h
@@ -117,8 +117,8 @@ class InternalServer {
                                void** cont_cls);
     bool start();
     void stop();
-    std::string getAddress() { return m_addr; }
-    int getPort() { return m_port; }
+    std::string getAddress() const { return m_addr; }
+    int getPort() const { return m_port; }
     IpMode getIpMode() const { return m_ipMode; }
 
   private: // functions
--- a/src/server/internalServer.h
+++ b/src/server/internalServer.h
@@ -27,6 +27,7 @@ extern "C" {
 
 #include "library.h"
 #include "name_mapper.h"
+#include "tools.h"
 
 #include <zim/search.h>
 #include <zim/suggestion.h>
@@ -94,7 +95,7 @@ class InternalServer {
   public:
     InternalServer(LibraryPtr library,
                    std::shared_ptr<NameMapper> nameMapper,
-                   std::string addr,
+                   IpAddress addr,
                    int port,
                    std::string root,
                    int nbThreads,
@@ -117,7 +118,7 @@ class InternalServer {
                                void** cont_cls);
     bool start();
     void stop();
-    std::string getAddress() const { return m_addr; }
+    IpAddress getAddress() const { return m_addr; }
     int getPort() const { return m_port; }
     IpMode getIpMode() const { return m_ipMode; }
 
@@ -166,7 +167,7 @@ class InternalServer {
     typedef ConcurrentCache<std::string, std::shared_ptr<LockableSuggestionSearcher>> SuggestionSearcherCache;
 
   private: // data
-    std::string m_addr;
+    IpAddress m_addr;
     int m_port;
     std::string m_root;                   // URI-encoded
     std::string m_rootPrefixOfDecodedURL; // URI-decoded
--- a/src/tools/networkTools.cpp
+++ b/src/tools/networkTools.cpp
@@ -217,8 +217,7 @@ std::string getBestPublicIp(bool ipv6) {
   std::map<std::string, IpAddress> interfaces = getNetworkInterfacesIPv4Or6();
 
 #ifndef _WIN32
-  const char* const prioritizedNames[] =
-      { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
+  const char* const prioritizedNames[] = { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
   for(auto name: prioritizedNames) {
     auto it=interfaces.find(name);
     if(it != interfaces.end() && !(ipv6 && (*it).second.addr6.empty())) {
--- a/src/tools/networkTools.cpp
+++ b/src/tools/networkTools.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "tools.h"
+#include "stringTools.h"
 #include <tools/networkTools.h>
 
 #include <stdio.h>
@@ -62,6 +63,12 @@ size_t write_callback_to_iss(char* ptr, size_t size, size_t nmemb, void* userdat
   return nmemb;
 }
 
+void updatePublicIpAddress(IpAddress& publicIpAddr, const IpAddress& interfaceIpAddr) 
+{
+  if (publicIpAddr.addr.empty())  publicIpAddr.addr  = interfaceIpAddr.addr;
+  if (publicIpAddr.addr6.empty()) publicIpAddr.addr6 = interfaceIpAddr.addr6;
+}
+
 } // unnamed namespace
 
 std::string download(const std::string& url) {
@@ -84,7 +91,6 @@ std::string download(const std::string& url) {
   }
   return ss.str();
 }
-
 
 namespace
 {
@@ -211,39 +217,35 @@ std::map<std::string, std::string> getNetworkInterfaces() {
   return result;
 }
 
-
-std::string getBestPublicIp(bool ipv6) {
-  IpAddress bestPublicIp = IpAddress{"127.0.0.1","::1"};
+IpAddress getBestPublicIps() {
+  IpAddress bestPublicIps;
   std::map<std::string, IpAddress> interfaces = getNetworkInterfacesIPv4Or6();
-
 #ifndef _WIN32
   const char* const prioritizedNames[] = { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
-  for(auto name: prioritizedNames) {
-    auto it=interfaces.find(name);
-    if(it != interfaces.end() && !(ipv6 && (*it).second.addr6.empty())) {
-      bestPublicIp = (*it).second;
-      break;
+  for (const auto& name : prioritizedNames) {
+    const auto it = interfaces.find(name);
+    if (it != interfaces.end()) {
+      updatePublicIpAddress(bestPublicIps, it->second);
     }
   }
 #endif
-
-  const char* const prefixes[] = { "192.168", "172.16.", "10.0" };
-  for(auto prefix : prefixes){
-    for(auto& itr : interfaces) {
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
+  const char* const v4prefixes[] = { "192.168", "172.16", "10.0", "169.254" };
+  for (const auto& prefix : v4prefixes) {
+    for (const auto& [_, interfaceIps] : interfaces) {
+      if (kiwix::startsWith(interfaceIps.addr, prefix)) {
+        updatePublicIpAddress(bestPublicIps, interfaceIps);
       }
     }
   }
-  return ipv6 ? bestPublicIp.addr6 : bestPublicIp.addr;
-}
 
+  updatePublicIpAddress(bestPublicIps, {"127.0.0.1", "::1"});
+
+  return bestPublicIps;
+}
 
 std::string getBestPublicIp()
 {
-  return getBestPublicIp(false);
+  return getBestPublicIps().addr;
 }
 
 } // namespace kiwix
--- a/test/otherTools.cpp
+++ b/test/otherTools.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "gtest/gtest.h"
+#include "../include/tools.h"
 #include "../src/tools/otherTools.h"
 #include "zim/suggestion_iterator.h"
 #include "../src/server/i18n_utils.h"
@@ -252,11 +253,8 @@ TEST(networkTools, getNetworkInterfaces)
   }
 }
 
-TEST(networkTools, getBestPublicIp)
+TEST(networkTools, getBestPublicIps)
 {
-  using kiwix::getBestPublicIp;
-
-  std::cout << "getBestPublicIp(true)  " << getBestPublicIp(true)  << std::endl;
-  std::cout << "getBestPublicIp(false) " << getBestPublicIp(false) << std::endl;
-  std::cout << "getBestPublicIp()      " << getBestPublicIp()      << std::endl;
+  std::cout << "getBestPublicIps(): " << "[" << kiwix::getBestPublicIps().addr << ", " << kiwix::getBestPublicIps().addr6 << "]" << std::endl;
+  std::cout << "getBestPublicIp(): " << kiwix::getBestPublicIp() << std::endl;
 }


kelson42:
@sgourdas thx for the PR but can you please fix the code so the CI can pass?

sgourdas:
@kelson42 ready

kelson42 (changes requested):
Almost good, but
```
$ kiwix-serve --port=8080 -i 127.0.0.1 /home/kelson/Downloads/wikipedia_en_ray_charles_maxi_2023-12.zim 
The Kiwix server is running and can be accessed in the local network at: http://192.168.0.168:8080 and http://::1:808
```
Should return only `http://127.0.0.1:8080`  as kiwix-serve should only listen on this IP

include/server.h
```diff
@@ -52,7 +53,7 @@ namespace kiwix
        void stop();
 
        void setRoot(const std::string& root);
-       void setAddress(const std::string& addr) { m_addr = addr; }
+       void setAddress(const std::vector<std::string>& addr) { m_addr = addr; }
```
veloman-yunkan, line 56: I am not sure that we can easily support such an API in general. To the best of my understanding you cannot set-up a single socket listening on (arbitrary) multiple addresses. A socket listening on *all* addresses is a special case. The commitment to such an API will require starting multiple servers (one per each address) which is doable but too much for the issue that we are trying to address.

sgourdas: Sure, I will revert to taking a string as a parameter and initializing the vector on this.

src/server/internalServer.cpp
```diff
@@ -462,18 +462,19 @@ bool InternalServer::start() {
   sockAddr6.sin6_port = htons(m_port);
 
   if (m_addr.empty()) {
-    if (0 != INADDR_ANY) {
+    if (0 != INADDR_ANY) { // what is the purpose of this? INADDR_ANY is always 0. Refer: http://www.castaglia.org/proftpd/doc/devel-guide/src/include/inet.h.html
```
sgourdas, line 465: @veloman-yunkan should I also remove this entire if clause?

kelson42:
Still not working fine:
```
$ kiwix-serve --port=8081 -i ipv4 /home/kelson/Downloads/wikipedia_fr_climate_change_maxi_2022-05.zim 
The Kiwix server is running and can be accessed in the local network at: http://:8081
```

@sgourdas Please test all of this completly.

src/server/internalServer.cpp
veloman-yunkan: I think that we should remove the `if` but not its body, thereby probably fixing a bug. To the best of my understanding, in this branch of the enclosing `if` `sockAddr6.sin6_addr = in6addr_any;` must be executed unconditionally. `sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY)` is likely a no-op for `0 == INADDR_ANY`, but it doesn't hurt to have it in case there are systems where `INADDR_ANY` is non-zero.

include/server.h
```diff
@@ -65,14 +66,14 @@ namespace kiwix
         { m_blockExternalLinks = blockExternalLinks; }
        void setIpMode(IpMode mode) { m_ipMode = mode; }
        int getPort();
-       std::string getAddress();
+       std::vector<std::string> getAddresses();
```
veloman-yunkan, line 69: Make `const` (and also please fix `const`-correctness of other getters in a separate commit)

include/tools.h
```diff
@@ -216,13 +217,13 @@ std::map<std::string, std::string> getNetworkInterfaces();
...
  */
-std::string getBestPublicIp(bool ipv6);
+std::vector<std::string> getBestPublicIps(IpMode mode);
 
 /** Provides the best IPv4 adddress
  * Equivalent to getBestPublicIp(false). Provided for backward compatibility
  * with libkiwix v13.1.0.
  */
-std::string getBestPublicIp();
+std::vector<std::string> getBestPublicIps();
```
veloman-yunkan, line 226: Read the docstring of this function and undo this change.

src/server/internalServer.cpp
```diff
@@ -462,18 +462,19 @@ bool InternalServer::start() {
...
       sockAddr6.sin6_addr = in6addr_any;
       sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
     }
-    m_addr = kiwix::getBestPublicIp(m_ipMode == IpMode::ipv6 || m_ipMode == IpMode::all);
+    m_addr = kiwix::getBestPublicIps(m_ipMode);
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    // TODO: Needs more work
```
veloman-yunkan, line 471: What else needs to be done?

src/server/internalServer.h
```diff
@@ -117,7 +117,7 @@ class InternalServer {
                                void** cont_cls);
     bool start();
     void stop();
-    std::string getAddress() { return m_addr; }
+    std::vector<std::string> getAddresses() { return m_addr; }
```
veloman-yunkan, line 120: Make `const` (and also please fix `const`-correctness of other getters in a separate commit)

test/server_testing_tools.h
```diff
@@ -142,7 +142,7 @@ void ZimFileServer::run(int serverPort, std::string indexTemplateString)
   }
   server.reset(new kiwix::Server(library, nameMapper));
   server->setRoot(cfg.root);
-  server->setAddress(address);
+  server->setAddress({address});
```
veloman-yunkan, line 145: Undo

sgourdas:
> Still not working fine:
> 
> ```
> $ kiwix-serve --port=8081 -i ipv4 /home/kelson/Downloads/wikipedia_fr_climate_change_maxi_2022-05.zim 
> The Kiwix server is running and can be accessed in the local network at: http://:8081
> ```
> 
> @sgourdas Please test all of this completly.

@kelson42 I tested this and it was working fine for me. Tested again and I get:

```
kiwix-serve --port 8081 -l /home/ubunter/.local/share/kiwix-desktop/library.xml -i ipv4
Loading the library from the following files:
        /home/ubunter/.local/share/kiwix-desktop/library.xml
The library was successfully loaded.
The Kiwix server is running and can be accessed in the local network at: http://172.27.70.65:8081
```

sgourdas:
Have you built both new kiwix tools and libkiwix versions?

If yes, can you provide more details like if despite the printed message the server is available in browser?

include/tools.h
```diff
@@ -216,13 +217,13 @@ std::map<std::string, std::string> getNetworkInterfaces();
 /** Provides the best IP address
  * This function provides the best IP address from the list given by getNetworkInterfacesIPv4Or6()
  */
-std::string getBestPublicIp(bool ipv6);
+std::vector<std::string> getBestPublicIps(IpMode mode);
```
veloman-yunkan, line 220: What about changing the return type to `IpAddress` (and reverting the function name change)?

veloman-yunkan (changes requested):

src/server/internalServer.cpp
sgourdas: This was a comment left to check if everything is ok after the change, before the entire scope was finalized. It is ok like this. 

Will remove comment.

include/tools.h
sgourdas: Thanks for this.

sgourdas: @veloman-yunkan what would you say about making it `IpMode getBestPublicIps(void)`?

kelson42:
@sgourdas This will wait tomorrow, I kind of suspect something wrong in my testing because of the renaming of the `.pc` file done earlier today and not fully rolled-out.

include/tools.h
veloman-yunkan: I don't get your proposal. Why would a function named `getBestPublicIps()` return an `IpMode` instead of an IP address?

sgourdas: Excuse my typo I meant `IpAddress getBestPublicIps(void)`

veloman-yunkan: Makes sense.

include/server.h
```diff
@@ -52,7 +52,7 @@ namespace kiwix
        void stop();
 
        void setRoot(const std::string& root);
-       void setAddress(const std::string& addr) { m_addr = addr; }
+       void setAddress(const std::string& addr) { if(!addr.empty()) addr.find('[') != std::string::npos ? m_addr.addr6 = addr : m_addr.addr = addr; }
```
veloman-yunkan: 1. Move the definition to .cpp and implement as a multiliner using an `if` statement instead of the conditional expression.
2. Is `[` a mandatory part of an IPv6 address syntax? Isn't it better to test for the presence of a colon (`:`)?

src/server/internalServer.cpp
```diff
@@ -461,19 +461,17 @@ bool InternalServer::start() {
...
-    }
-    m_addr = kiwix::getBestPublicIp(m_ipMode == IpMode::ipv6 || m_ipMode == IpMode::all);
+  if (m_addr.addr.empty() && m_addr.addr6.empty()) { // No specific ip address was provided
+    sockAddr6.sin6_addr = in6addr_any;
+    sockAddr4.sin_addr.s_addr = htonl(INADDR_ANY);
+    m_addr = kiwix::getBestPublicIps();
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    bool ipv6 = inet_pton(AF_INET6, m_addr.addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
```
veloman-yunkan, line 469: `m_addr.addr` → `m_addr.addr6`

I have a feeling that the rest of the code in this block has to be revised too.

src/tools/networkTools.cpp
```diff
@@ -212,39 +212,38 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
       { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
   for(auto name: prioritizedNames) {
     auto it=interfaces.find(name);
-    if(it != interfaces.end() && !(ipv6 && (*it).second.addr6.empty())) {
-      bestPublicIp = (*it).second;
-      break;
+    if (it != interfaces.end()) {
+      if (it->second.addr.find(':') == std::string::npos) bestIp.addr = it->second.addr;
+      else bestIp.addr6 = it->second.addr6;
     }
```
veloman-yunkan, line 227: Since the `break` statement has disappeared. Does that mean that what was previously considered the best public IP address may now be overwritten by continued execution of the loop?

src/tools/networkTools.cpp
```diff
@@ -212,39 +212,38 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
 
   const char* const prefixes[] = { "192.168", "172.16.", "10.0" };
   for(auto prefix : prefixes){
     for(auto& itr : interfaces) {
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
-      }
+      if (itr.second.addr.find(':') == std::string::npos && itr.second.addr.find(prefix) == 0)
```
veloman-yunkan, line 234: Why would `itr.second.addr` (which is supposed to be an IPv4 address) contain a `:`?

src/tools/networkTools.cpp
```diff
@@ -212,39 +212,38 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
+          bestIp.addr = itr.second.addr;
+      if (itr.second.addr6.find(':') != std::string::npos)
+          bestIp.addr6 = itr.second.addr6;
```
sgourdas, line 237: Should use a helper variable for `itr.second.addr`

src/tools/networkTools.cpp
```diff
@@ -212,39 +212,38 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
   for(auto prefix : prefixes){
     for(auto& itr : interfaces) {
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
-      }
+      if (itr.second.addr.find(':') == std::string::npos && itr.second.addr.find(prefix) == 0)
+          bestIp.addr = itr.second.addr;
+      if (itr.second.addr6.find(':') != std::string::npos)
```
veloman-yunkan, line 236: `itr.second.addr6` - unless empty - *should be* an IPv6 address. Therefore the condition should be `if (!itr.second.addr6.empty())`

include/server.h
```diff
@@ -52,7 +52,16 @@ namespace kiwix
...
+       void setAddress(const std::string& addr) {
+          if (!addr.empty()) {
+              if (addr.find('[') != std::string::npos) {
+                  std::string cleanAddr = addr.substr(1, addr.length() - 2);
+                  m_addr.addr6 = cleanAddr;
+              } else {
+                  m_addr.addr = addr;
+              }
+          }
+      }
```
veloman-yunkan, line 64: 1. Move the definition to .cpp
2. Is `[` a mandatory part of an IPv6 address syntax? Isn't it better to test for the presence of a colon (`:`)? Of course we can remove optional square brackets, but we should also accept IPv6 addresses without brackets (like we did before).

veloman-yunkan (changes requested):
I think I started reviewing while the PR was being updated. My feeling is that my feedback doesn't pave the path to this PR being merged on next iteration, it seems like we will have to make at least a couple more back-and-forths.

sgourdas:
> I think I started reviewing while the PR was being updated. My feeling is that my feedback doesn't pave the path to this PR being merged on next iteration, it seems like we will have to make at least a couple more back-and-forths.

Sorry for this. This PR is not up to standard and I have to revise all the changes. When it started, I expected it to be more straight forward and it has become very unstructured. 

Will try to get back ASAP.

include/server.h
sgourdas: Exactly what I want do on next iteration. Thanks.

kelson42:
Seems to work now! How do we secure that only valid IP are given/accepted for `-i`?

sgourdas:
> Seems to work now! How do we secure that only valid IP are given/accepted for `-i`?

Perfect. Currently, we are using `inet_pton()` accordingly for v4 or v6 addresses. If the conversion is successful we know that it is valid and of that type.

kelson42:
@sgourdas This IP `192.168.0.0` is not available on my system, but I have no error telling me the IP is not valid.
```
kelson@camber:~/code/kiwix-tools$ kiwix-serve --port=8081 -i 192.168.0.0 /home/kelson/Downloads/wikipedia_en_ray_charles_maxi_2023-12.zim 
Unable to instantiate the HTTP daemon. The port 8081 is maybe already occupied or need more permissions to be open. Please try as root or with a port number higher or equal to 1024.
```

sgourdas:
> @sgourdas This IP `192.168.0.0` is not available on my system, but I have no error telling me the IP is not valid.
> 
> ```
> 
> kelson@camber:~/code/kiwix-tools$ kiwix-serve --port=8081 -i 192.168.0.0 /home/kelson/Downloads/wikipedia_en_ray_charles_maxi_2023-12.zim 
> 
> Unable to instantiate the HTTP daemon. The port 8081 is maybe already occupied or need more permissions to be open. Please try as root or with a port number higher or equal to 1024.
> 
> ```

What should be the proper behaviour for this case in your opinion? I think this should be a topic for another PR if it add features that were not present before.

kelson42:
@sgourdas OK, I have open https://github.com/kiwix/kiwix-tools/issues/709
@sgourdas @veloman-yunkan We should move aheqd with code review IMO now that it works from a user perspective.

kelson42 (approved):
LGTM from a user perspective

veloman-yunkan (changes requested):
I think that we should close #1136 and bring its enhancement into this PR.

include/tools.h
```diff
@@ -32,6 +33,12 @@ struct IpAddress
 {
     std::string addr;  // IPv4 address
     std::string addr6; // IPv6 address
+
+    bool empty() const { return addr.empty() && addr6.empty(); }
+    bool empty4() const { return addr.empty(); }
+    bool empty6() const { return addr6.empty(); }
+    bool valid4(IpMode mode) const { return (mode == IpMode::ipv4 || mode == IpMode::all) && !this->empty4(); }
+    bool valid6(IpMode mode) const { return (mode == IpMode::ipv6 || mode == IpMode::all) && !this->empty6(); }
```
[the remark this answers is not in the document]
sgourdas: My understanding is the same as yours. Whenever `all` is provided, we should serve all the protocols that we can.

`valid[number]()` is a way to understand if an ip should be used (mode compatibility and corresponding ip field not empty). If `mode == IpMode::all` both fields should be used, if the interface provides them.

test/otherTools.cpp
```diff
@@ -252,11 +253,14 @@ TEST(networkTools, getNetworkInterfaces)
...
+  using kiwix::getBestPublicIps;
   using kiwix::getBestPublicIp;
+  using kiwix::IpMode;
 
-  std::cout << "getBestPublicIp(true)  " << getBestPublicIp(true)  << std::endl;
-  std::cout << "getBestPublicIp(false) " << getBestPublicIp(false) << std::endl;
-  std::cout << "getBestPublicIp()      " << getBestPublicIp()      << std::endl;
+  std::cout << "getBestPublicIps(true)   : [" << getBestPublicIps(IpMode::ipv4).addr << ", " << getBestPublicIps(IpMode::ipv4).addr6 << "]" << std::endl;
+  std::cout << "getBestPublicIps(false)  : [" << getBestPublicIps(IpMode::ipv6).addr << ", " << getBestPublicIps(IpMode::ipv6).addr6 << "]" << std::endl;
+  std::cout << "getBestPublicIps(false)  : [" << getBestPublicIps(IpMode::all).addr << ", " << getBestPublicIps(IpMode::all).addr6 << "]" << std::endl;
```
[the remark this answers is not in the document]
sgourdas: I am crying seeing pieces like this left behind. These fugitives keep resisting. Thanks a lot for catching another one.

src/server.cpp
```diff
@@ -75,12 +75,25 @@ void Server::setRoot(const std::string& root)
   }
 }
 
+void Server::setAddress(const std::string& addr) 
+{
+  if (addr.empty())
+    return;
+
+  if (addr.find(':') != std::string::npos) { // IPv6
+    for (char ch : addr) if (ch == '[' or ch == ']') ch = '\0'; // Clean possible brackets
```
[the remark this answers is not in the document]
sgourdas: Yeah I do not know why I did this honestly. Will fix on next iteration.

include/tools.h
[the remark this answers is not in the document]
veloman-yunkan: Then it is an issue with the naming of the method.

I don't currently feel that it is justified enhancing the `IpAddress` with member functions. Please get rid of them and let's see how it looks.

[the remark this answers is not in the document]
veloman-yunkan: Note that I mean *all* member functions that you added to `IpAddress`.

src/tools/networkTools.cpp
```diff
@@ -211,39 +211,39 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
-    if(it != interfaces.end() && !(ipv6 && (*it).second.addr6.empty())) {
-      bestPublicIp = (*it).second;
-      break;
-    }
+  for(const auto& name: prioritizedNames) {
+    const auto it = interfaces.find(name);
+    if(it == interfaces.end()) continue;
+    const IpAddress& interfaceIps = it->second;
+    if(!bestPublicIps.addr.empty() && validMode4 && !interfaceIps.addr.empty()) 
+      bestPublicIps.addr = interfaceIps.addr;
```
veloman-yunkan, line 226: Shouldn't the condition be `if ( bestPublicIps.addr.empty() ... )` so that `bestPublicIps.addr` is set only once?

The same bug is present for the rest of the similar `if`s in this function.

src/tools/networkTools.cpp
```diff
@@ -211,39 +211,39 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
 
-
-std::string getBestPublicIp(bool ipv6) {
-  IpAddress bestPublicIp = IpAddress{"127.0.0.1","::1"};
+IpAddress getBestPublicIps(IpMode mode) {
+  IpAddress bestPublicIps = IpAddress{"127.0.0.1", "::1"};
   std::map<std::string, IpAddress> interfaces = getNetworkInterfacesIPv4Or6();
-
+  bool validMode4 = (mode == IpMode::ipv4 || mode == IpMode::all);  // Mode is valid to support ipv4
+  bool validMode6 = (mode == IpMode::ipv6 || mode == IpMode::all);  // Mode is valid to support ipv6
```
veloman-yunkan, line 218: I think that these can be moved to the bottom of the function. The function will consist of two parts
- select the best public IPv4 and IPv6 addresses regardless of the mode requested (the logic of this will be slightly simpler compared to the current code)
- compose the result from those IPv4 and IPv6 addresses based on the requested mode

src/tools/networkTools.cpp
```diff
@@ -211,39 +211,39 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
-      }
+  for (const auto& prefix : prefixes) {
+    for (const auto& [_, interfaceIps] : interfaces) {
+        if (!bestPublicIps.addr.empty() && validMode4 && !interfaceIps.addr.empty() && interfaceIps.addr.find(prefix) == 0)
+            bestPublicIps.addr = interfaceIps.addr;
+        if (!bestPublicIps.addr6.empty() && validMode6 && !interfaceIps.addr6.empty())
```
veloman-yunkan, line 236: Is the `interfaceIps.addr.find(prefix) == 0` check omitted intentionally?

test/otherTools.cpp
```diff
@@ -252,11 +253,14 @@ TEST(networkTools, getNetworkInterfaces)
...
+  using kiwix::getBestPublicIps;
   using kiwix::getBestPublicIp;
+  using kiwix::IpMode;
 
-  std::cout << "getBestPublicIp(true)  " << getBestPublicIp(true)  << std::endl;
-  std::cout << "getBestPublicIp(false) " << getBestPublicIp(false) << std::endl;
-  std::cout << "getBestPublicIp()      " << getBestPublicIp()      << std::endl;
+  std::cout << "getBestPublicIps(IpMode::ipv4) : [" << getBestPublicIps(IpMode::ipv4).addr << ", " << getBestPublicIps(IpMode::ipv4).addr6 << "]" << std::endl;
+  std::cout << "getBestPublicIps(IpMode::ipv6) : [" << getBestPublicIps(IpMode::ipv6).addr << ", " << getBestPublicIps(IpMode::ipv6).addr6 << "]" << std::endl;
+  std::cout << "getBestPublicIps(IpMode::all)  : [" << getBestPublicIps(IpMode::all).addr << ", " << getBestPublicIps(IpMode::all).addr6 << "]" << std::endl;
```
veloman-yunkan, line 264: Please overload `operator<<(std::ostream&, kiwix::IpAddress&)` above this test and make this code more readable.

include/tools.h
```diff
@@ -225,6 +225,15 @@ IpAddress getBestPublicIps(IpMode mode);
...
 std::string getBestPublicIp();
 
+/** Checks if IP address is available
+ *
+ * Check if the given IP address is configured on any network interface of the system
+ *
+ * @param addr the IP address to check.
+ * @return true if the IP address is available on the system.
+ */
+bool ipAvailable(const IpAddress& addr);
```
veloman-yunkan, line 235: 1. This function need to be a part of the public API?
2. `IpAddress` can be a pair of addresses (one IPv4 and the other IPv6). How is this function supposed to work if both components are non-empty?

veloman-yunkan (changes requested):

src/server.cpp
```diff
@@ -75,12 +75,23 @@ void Server::setRoot(const std::string& root)
   }
 }
 
+void Server::setAddress(const std::string& addr)
+{
+  if (addr.empty()) return;
+
+  if (addr.find(':') != std::string::npos) { // IPv6
+    m_addr.addr6 = (addr[0] == '[') ? addr.substr(1, addr.length() - 2) : addr; // Remove brackets if any
```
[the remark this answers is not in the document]
sgourdas: I wanted to make it perform a full validation but was not sure if we want to do all that. Is it ok to refactor `Server::setAddress` to return a bool signifying validity of the address?

include/tools.h
sgourdas: > This function need to be a part of the public API?

I think it can be a useful function to have available

> `IpAddress` can be a pair of addresses (one IPv4 and the other IPv6). How is this function supposed to work if both components are non-empty?

Current use case was to check only one ip (in the case of the user giving one). I will refactor it to take as a parameter the mode as well and check accordingly.

src/tools/networkTools.cpp
sgourdas: You are totally correct. Thanks for catching this...

src/tools/networkTools.cpp
sgourdas: Yes, the reason it is omitted is because we only provide v4 prefixes.

sgourdas: Would it be beneficial to provide some?

include/common.h
```diff
@@ -16,7 +16,7 @@
 
 namespace kiwix {
 
-enum class IpMode { ipv4, ipv6, all };
+enum class IpMode { ipv4, ipv6, all, flex };
```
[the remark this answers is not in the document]
kelson42: I have the feeling I have missed something... what is his "glex" option!? Where does it come from?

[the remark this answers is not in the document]
veloman-yunkan: @kelson42 It started from [here](https://github.com/kiwix/kiwix-tools/pull/704#discussion_r1771266061)

src/server.cpp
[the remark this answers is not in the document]
veloman-yunkan: Don't mind about this comment until we fix the more important shortcomings. In fact I am not even sure that we should spend any effort on making it 100% foolproof. Achieving perfect foolproofness rarely pays off and should be pursued only if stakes are high. Getting a wrong error message is no way near threatening someone's life, business or dignity.

src/tools/networkTools.cpp
veloman-yunkan: > Yes, the reason it is omitted is because we only provide v4 prefixes.

The purpose of this piece of code is to select an IP address from a local network, though I am not sure why the `169.254` prefix (see https://en.wikipedia.org/wiki/Link-local_address) has not been included. Your selection of an IPv6 address fails to meet that criteria. Why don't you filter based on the same IPv4 check (`interfaceIps.addr.find(prefix) == 0`) but use the IPv6 address (`interfaceIps.addr6`)?

> Would it be beneficial to provide some for v6?

According to https://en.wikipedia.org/wiki/Link-local_address#IPv6

* In [IPv6](https://en.wikipedia.org/wiki/IPv6), unicast link-local addresses are assigned from the block *`fe80::/10`*

however, unfortunately, it seems like `fe80` cannot be used as a prefix because only its higher 10 bits matter. If we decide to support it a slightly more complex check will be required. But let's not go after it now.

src/tools/networkTools.cpp
sgourdas: I am debating whether we should do a mode check inside this function at all. The name `getBestPublicIps` I think better suits a behavior where the whole `IpAdress` struct is filled. IMO mode checking should be moved outside this function.

include/tools.h
veloman-yunkan: > > This function need to be a part of the public API?
> 
> I think it can be a useful function to have available

When dealing with public API that's not a good argument. Public API is a commitment. Exposing something just because it's useful may lead to unjustified maintenance overhead (like having to deal with backward compatibility issues). Let's not bloat our public API without necessity.

src/tools/networkTools.cpp
veloman-yunkan: Agree. In fact, when writing my previous comment I considered proposing to package the first part (covered by the first bullet) into a separate function `IpAddress getBestPublicIps(void)`. I even think that you proposed something like that earlier.

veloman-yunkan: > I even think that you proposed something like that earlier.

Indeed, [you did](https://github.com/kiwix/libkiwix/pull/1132#discussion_r1765135280).

sgourdas:
Sorry for the lack of feedback here. Getting back tomorrow morning.

src/tools/networkTools.cpp
sgourdas: Ok so if I understand correctly, for this point, I should just add `169.254` for the v4 prefixes, right?

src/tools/networkTools.cpp
sgourdas: Yeah, this has been all over the place, but let me correct it to close this up.

include/common.h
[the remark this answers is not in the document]
sgourdas: I assume we are ok to proceed right?

src/tools/networkTools.cpp
```diff
@@ -211,40 +211,51 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
-      break;
-    }
+  const char* const prioritizedNames[] = { "eth0", "eth1", "wlan0", "wlan1", "en0", "en1" };
+  for(const auto& name: prioritizedNames) {
+    const auto it = interfaces.find(name);
+    if(it == interfaces.end()) continue;
+    const IpAddress& interfaceIps = it->second;
+    if(!bestPublicIps.addr.empty() && validMode4 && !interfaceIps.addr.empty()) 
+      bestPublicIps.addr = interfaceIps.addr;
+    if(!bestPublicIps.addr6.empty() && validMode6 && !interfaceIps.addr6.empty())
```
sgourdas, line 227: @veloman-yunkan IMO these `empty()` checks on the interfaceIPs are not needed and they do not really provide any help. I suggest to remove them to clean up the code.

kelson42:
FYI, this is the last PR to merge blocking release of libkiwix14. This is urgent we can complete this ASAP as many projects, for many different wait to the libkiwix14 to move forward with their own milestones.

sgourdas:
@kelson42 please do one check again on usability since a lot of things have been changed.

src/server/internalServer.cpp
```diff
@@ -1302,3 +1310,23 @@ std::unique_ptr<Response> InternalServer::handle_locally_customized_resource(con
...
 
 }
+
+bool ipAvailable(const kiwix::IpAddress& addr, const kiwix::IpMode& mode)
+{
+  bool validV4 = false, validV6 = false;
+  auto interfaces = kiwix::getNetworkInterfacesIPv4Or6();
+
+  if (addr.addr == "127.0.0.1") validV4 = true;
+  if (addr.addr6 == "::1") validV6 = true;
```
sgourdas, line 1320: These 2 should be removed.

kelson42 (approved):
LGTM from a user perspective

kelson42:
@sgourdas Please prepare the PR at `kiwix/kiwix-desktop` level to support this new API (at least compilation)

sgourdas:
> @sgourdas Please prepare the PR at `kiwix/kiwix-desktop` level to support this new API (at least compilation), see [kiwix/kiwix-desktop#1119](https://github.com/kiwix/kiwix-desktop/issues/1119)

Opened kiwix/kiwix-desktop#1220

sgourdas:
@kelson42 I skept past the support for ipv6 from  kiwix/kiwix-desktop#1119. The PR I opened just fixes compilation. Going to implement the rest also.

kelson42:
@sgourdas AFAIK this PR is causing crashing (segfault) of Kiwix Desktop - this should be checked (fixed?) properly.

sgourdas:
> @sgourdas AFAIK this PR is causing crashing (segfault) of Kiwix Desktop - this should be checked (fixed?) properly.

@kelson42 which exact process causes the app to seg fault? I cannot seem to be able to reproduce it.

src/server.cpp
```diff
@@ -75,12 +75,23 @@ void Server::setRoot(const std::string& root)
...
+  } else {
+    m_addr.addr = addr;
+  }
+}
```
veloman-yunkan, line 87: One more foolproofness issue with this function - it is written under the (reasonable) assumption that it is called only once. This peculiarity of the implementation should be at least noted in a comment.

src/server/internalServer.cpp
```diff
@@ -461,19 +461,27 @@ bool InternalServer::start() {
...
+    if (m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL) m_addr.addr6 = bestIps.addr6;
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::ALL;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    bool providedV4 = !m_addr.addr.empty();
+    m_ipMode = providedV4 ? IpMode::IPV4 : IpMode::IPV6;
```
veloman-yunkan, line 473: This is acceptable only if `m_ipMode` was set to `IpMode::AUTO` (but that's an issue with foolproofness).

src/server/internalServer.cpp
```diff
@@ -461,19 +461,27 @@ bool InternalServer::start() {
...
+    if (m_ipMode == IpMode::IPV4 || m_ipMode == IpMode::ALL) m_addr.addr = bestIps.addr;
+    if (m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL) m_addr.addr6 = bestIps.addr6;
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::ALL;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    bool providedV4 = !m_addr.addr.empty();
```
veloman-yunkan, line 472: I would get rid of this in favour of `const std::string addr = !m_addr.addr.empty() ? m_addr.addr : m_addr.addr6;`

src/tools/networkTools.cpp
```diff
@@ -211,39 +210,36 @@ std::map<std::string, std::string> getNetworkInterfaces() {
   return result;
 }
 
-
-std::string getBestPublicIp(bool ipv6) {
-  IpAddress bestPublicIp = IpAddress{"127.0.0.1","::1"};
+IpAddress getBestPublicIps(IpMode mode) {
```
veloman-yunkan, line 213: `mode` is an unused parameter

src/tools/networkTools.cpp
```diff
@@ -211,39 +210,36 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
-        bestPublicIp = itr.second;
-        break;
-      }
+  const char* const v4prefixes[] = { "192.168", "172.16", "10.0", "169.254" };
+  for (const auto& prefix : v4prefixes) {
+    for (const auto& [_, interfaceIps] : interfaces) {
+      if (bestPublicIps.addr.empty() && interfaceIps.addr.find(prefix) == 0) bestPublicIps.addr = interfaceIps.addr;
+      if (bestPublicIps.addr6.empty()) bestPublicIps.addr6 = interfaceIps.addr6;
     }
   }
```
veloman-yunkan, line 232: I wonder if we should try matching via network address prefix before using looking at the network interface names. @kelson42 What's your opinion?

src/server/internalServer.cpp
```diff
@@ -461,19 +461,27 @@ bool InternalServer::start() {
...
+
+    bool validV4 = inet_pton(AF_INET, m_addr.addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    bool validV6 = inet_pton(AF_INET6, m_addr.addr6.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
+
+    if (!validV4 && !validV6) {
+      std::cerr << "IP address " << (providedV4 ? m_addr.addr : m_addr.addr6) << " is not a valid ip address." << std::endl;
+      return false;
+    } 
+    
+    if (!ipAvailable(m_addr, m_ipMode)) {
```
veloman-yunkan, line 483: There is no such function at this point in the PR history

src/server/internalServer.cpp
```diff
@@ -461,19 +461,27 @@ bool InternalServer::start() {
...
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::ALL;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    bool providedV4 = !m_addr.addr.empty();
+    m_ipMode = providedV4 ? IpMode::IPV4 : IpMode::IPV6; 
+
+    bool validV4 = inet_pton(AF_INET, m_addr.addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    bool validV6 = inet_pton(AF_INET6, m_addr.addr6.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
```
veloman-yunkan, line 476: Are these checks made redundant by the introduction of `ipAvailable()`?

src/server/internalServer.h
```diff
@@ -196,4 +196,5 @@ class InternalServer {
 
 }
 
+bool ipAvailable(const kiwix::IpAddress& addr, const kiwix::IpMode& mode);
```
veloman-yunkan, line 199: Declaration of this function in a header file is not needed

src/server/internalServer.cpp
```diff
@@ -1310,3 +1310,23 @@ std::unique_ptr<Response> InternalServer::handle_locally_customized_resource(con
...
+    kiwix::IpAddress interfaceIps = interface.second;
+    validV4 |= !validV4 && interfaceIps.addr == addr.addr;
+    validV6 |= !validV6 && interfaceIps.addr6 == addr.addr6;
+  }
+
+  if (mode == kiwix::IpMode::IPV4) return validV4;
+  else if (mode == kiwix::IpMode::IPV6) return validV6;
+
+  return validV4 && validV6;
+}
```
veloman-yunkan: 1. Move the definition above the point of usage and hide it in an unnamed namespace
2. Forget about the implementation and describe in a doc comment how you want this function to work, given that `IpAddress` is conceptually a pair of addresses (IPv4 and IPv6). It is intended for the case when only one of `.addr` or `.addr6` is non-empty? If so, then maybe its parameter type should be changed to `std::string`?

veloman-yunkan (changes requested):

src/tools/networkTools.cpp
veloman-yunkan: No, the main concern was about the application of the prefix check when selecting the IPv6 address. The exact instructions what should be done is provided in the new review.

src/server/internalServer.cpp
sgourdas: True. The code assumes `m_ipMode` here to be `kiwix::IpMode::AUTO`, so I will add a return false in case it is not, to increase foolproofness a bit.

src/server/internalServer.cpp
sgourdas: There is some overlap for sure. 

In the sense of checking if an ip can proceed yes, but not in the sense of providing a better error message (sacrificing a bit of compute time).

For example address `256.0.0.1` would be called invalid, which is more accurate than just saying it is not available.

It really depends on what we value more optimization/validation.

What is your opinion?

src/server/internalServer.cpp
sgourdas: What do you mean exactly by this?

veloman-yunkan: `ipAvailable()` is introduced in the next commit

src/server/internalServer.cpp
veloman-yunkan: Here I am concerned about performance least of all. I wanted to find out if we could simplify the code.

But I missed the fact that these are not just checks, they also have a side effect of initializing `sockAddr4.sin_addr.s_addr` and/or `sockAddr6.sin6_addr.s6_addr` so the calls to `inet_pton()` cannot be eliminated. And even if they could - a more adequate error message is a good reason to have those checks in place.

sgourdas: The only thing we could do is initialize based on the mode if we would go the other way, but I agree.

test/otherTools.cpp
```diff
@@ -252,11 +253,12 @@ TEST(networkTools, getNetworkInterfaces)
   }
 }
 
-TEST(networkTools, getBestPublicIp)
-{
-  using kiwix::getBestPublicIp;
+std::ostream& operator<<(std::ostream& os, const kiwix::IpAddress& ip) {
+    return os << "[" << ip.addr << ", " << ip.addr6 << "]";
+}
```
sgourdas, line 258: Questioning whether we should keep this.

src/server/internalServer.cpp
sgourdas: I agree to change this function's scope into just evaluating whether an ip address (std::string) is available, for the same reasons as we did with `getBestPublicIps`.

I am debating though how the check should be done:
- Mode as parameter (like now) and check the corresponding interface fields.
- No mode as parameter but we check inside the function to determine it's type. Then we check on the corresponding interface fields.
- No mode as parameter and we blindly check both interface fields.

kelson42:
CI does not pass anymore! This PR generates a lot of back and forth and I have difficulty to foresee the end of the review process. I'm a bit concerned about the general overhead it generates honestly. I this PR can not be finaly reviewed/merged in the next days, then I would recommend that @veloman-yunkan overtakes the whole polishing process.

sgourdas:
> CI does not pass anymore! This PR generates a lot of back and forth and I have difficulty to foresee the end of the review process. I'm a bit concerned about the general overhead it generates honestly. I this PR can not be finaly reviewed/merged in the next days, then I would recommend that @veloman-yunkan overtakes the whole polishing process.

Sorry, I should have probably turned this to a draft, but this is not ready for review yet as I am waiting for the last points to be addressed first.

src/server/internalServer.cpp
```diff
@@ -461,30 +478,38 @@ bool InternalServer::start() {
...
+    if (m_ipMode == IpMode::IPV4 || m_ipMode == IpMode::ALL) m_addr.addr = bestIps.addr;
+    if (m_ipMode == IpMode::IPV6 || m_ipMode == IpMode::ALL) m_addr.addr6 = bestIps.addr6;
   } else {
-    bool ipv6 = inet_pton(AF_INET6, m_addr.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::all;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
+    if (m_ipMode != kiwix::IpMode::AUTO) return false; // If mode is not AUTO, something must be wrong
```
sgourdas, line 489: If we decide to keep this, kiwix-desktop should set mode to `AUTO` as well

sgourdas: (Of course like the test that is failing)

src/server/internalServer.cpp
sgourdas: I will add a commit implementing the 3rd option which I like most so we can proceed faster later if no disputes.

veloman-yunkan:
> Sorry, I should have probably turned this to a draft, but this is not ready for review yet as I am waiting for the last points to be addressed first.

This comment is somewhat ambiguous. @sgourdas Do you wait for any feedback from my side?

sgourdas:
> > Sorry, I should have probably turned this to a draft, but this is not ready for review yet as I am waiting for the last points to be addressed first.
> 
> This comment is somewhat ambiguous. @sgourdas Do you wait for any feedback from my side?

@veloman-yunkan sorry if it was not clear, but yes. 

Waiting on answers for the below to make a hopefully final squash and push.

https://github.com/kiwix/libkiwix/pull/1132#discussion_r1780735868:
> I wonder if we should try matching via network address prefix before looking at the network interface names. @kelson42 What's your opinion?

https://github.com/kiwix/libkiwix/pull/1132#discussion_r1782409588:
> If we decide to keep this, kiwix-desktop should set mode to AUTO as well

https://github.com/kiwix/libkiwix/pull/1132#discussion_r1781463941:
> Questioning whether we should keep this.

src/server/internalServer.cpp
veloman-yunkan: `AUTO` can be the default mode.

test/otherTools.cpp
veloman-yunkan: That's a minor issue. Proceed however you like.

veloman-yunkan:
@sgourdas

Those are not blockers.

Looks like @kelson42 has scared you too much with his attempt to expedite the release of 2.4.0 libkiwix.
This is *not* your last chance of having this PR merged so an extra iteration or two for final polishing is fine.

sgourdas:
> @sgourdas
> 
> Those are not blockers.
> 
> Looks like @kelson42 has scared you too much with his attempt to expedite the release of 2.4.0 libkiwix. This is _not_ your last chance of having this PR merged so an extra iteration or two for final polishing is fine.

Alright then, lets go for another round 😄

src/tools/networkTools.cpp
```diff
@@ -211,39 +216,34 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
-    for(auto& itr : interfaces) {
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
-      }
+  const char* const v4prefixes[] = { "192.168", "172.16", "10.0", "169.254" };
+  for (const auto& prefix : v4prefixes) {
+    for (const auto& [_, interfaceIps] : interfaces) {
+      if (interfaceIps.addr.find(prefix) == 0) updatePublicIpAddress(bestPublicIps, interfaceIps);
```
veloman-yunkan, line 234: 1. Reformat as a multiline statement with braces
2. Don't upset `kiwix::startsWith()` by making it feel neglected or distrusted.

src/tools/networkTools.cpp
```diff
@@ -211,39 +216,34 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
     }
   }
-  return ipv6 ? bestPublicIp.addr6 : bestPublicIp.addr;
-}
 
+  if (bestPublicIps.addr.empty()) bestPublicIps.addr = "127.0.0.1";
+  if (bestPublicIps.addr6.empty()) bestPublicIps.addr6 = "::1";
```
veloman-yunkan, line 239: Align the bodies (including the assignment operators)

src/server/internalServer.cpp
```diff
@@ -85,6 +85,16 @@ namespace kiwix {
 namespace
 {
 
+bool ipAvailable(const std::string addr)
+{
+  auto interfaces = kiwix::getNetworkInterfacesIPv4Or6();
+
+  for (const auto& [_, interfaceIps] : interfaces)
+    if ((interfaceIps.addr == addr) || (interfaceIps.addr6 == addr)) return true;
```
veloman-yunkan, line 93: Move the return statement onto a line of its own and add braces for bodies of both `for` and `if` statement.

veloman-yunkan (changes requested):
LLIONSPBP

If you wonder what that abbreviation stands for and cannot guess it yourself, than turn for help to an Asian constricting snake and ask it to evaluate the expression `'gnihsilbup erofeb gnihsilop emos sdeen ylno ti ekil skooL'[::-1]`.

src/server/internalServer.cpp
```diff
@@ -461,21 +461,30 @@ bool InternalServer::start() {
...
-    bool ipv4 = inet_pton(AF_INET, m_addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
-    if (ipv6){
-       m_ipMode = IpMode::ALL;
-    } else if (!ipv4) {
-      std::cerr << "Ip address " << m_addr << "  is not a valid ip address" << std::endl;
-      return false;
+    const std::string addr = !m_addr.addr.empty() ? m_addr.addr : m_addr.addr6;
+
+    if (m_ipMode != kiwix::IpMode::AUTO) {
+      std::cerr << "Specific IP address " << addr << " provided, but mode was not set to AUTO." << std::endl;
```
veloman-yunkan, line 485: Change the error message to "ERROR: When an IP address is provided the IP mode must not be set" (yes, don't include the IP address in the message so that one cannot engineer funny error messages by supplying intelligently constructed strings for the `--address` option).

src/server/internalServer.cpp
```diff
@@ -461,21 +461,30 @@ bool InternalServer::start() {
...
+      return false; // If mode is not AUTO, something must be wrong
```
veloman-yunkan, line 486: Delete the comment

src/server/internalServer.cpp
```diff
@@ -461,21 +461,30 @@ bool InternalServer::start() {
...
     }
+
+    bool validV4 = inet_pton(AF_INET, m_addr.addr.c_str(), &(sockAddr4.sin_addr.s_addr)) == 1;
+    bool validV6 = inet_pton(AF_INET6, m_addr.addr6.c_str(), &(sockAddr6.sin6_addr.s6_addr)) == 1;
+
+    if (!validV4 && !validV6) {
+      std::cerr << "IP address " << addr << " is not a valid ip address." << std::endl;
```
veloman-yunkan, line 493: Change the message format to "ERROR: invalid IP address: *HERE_COMES_THE_IP_ADDRESS*" so that it produces a less confusing error for, for example, `kiwix-serve --address "6.6.6.6 is a valid IP address however it asks for trouble, therefore"` ...

src/tools/networkTools.cpp
```diff
@@ -211,39 +217,36 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
+      if (kiwix::startsWith(interfaceIps.addr, prefix)) {
+        updatePublicIpAddress(bestPublicIps, interfaceIps);
       }
     }
   }
-  return ipv6 ? bestPublicIp.addr6 : bestPublicIp.addr;
-}
 
+  if (bestPublicIps.addr.empty())  bestPublicIps.addr  = "127.0.0.1";
+  if (bestPublicIps.addr6.empty()) bestPublicIps.addr6 = "::1";
```
veloman-yunkan, line 242: Why don't we reuse `updatePublicIpAddress()` here?

```c++
updatePublicIpAddress(bestPublicIps, {"127.0.0.1", "::1"});
```

src/server/internalServer.cpp
veloman-yunkan: This was not done.

src/server/internalServer.cpp
```diff
@@ -483,6 +493,10 @@ bool InternalServer::start() {
       std::cerr << "IP address " << addr << " is not a valid ip address." << std::endl;
       return false;
     } 
+    if (!ipAvailable(addr)) {
+      std::cerr << "IP address " << addr << " is not available on this system." << std::endl;
```
veloman-yunkan, line 497: Change the error format to "ERROR: IP address is not available on this system: *HERE_COMES_THE_IP_ADDRESS*"

src/server/internalServer.cpp
```diff
@@ -483,6 +493,10 @@ bool InternalServer::start() {
       std::cerr << "IP address " << addr << " is not a valid ip address." << std::endl;
       return false;
     } 
+    if (!ipAvailable(addr)) {
```
veloman-yunkan, line 496: Add a blank line

veloman-yunkan (changes requested):
There is no limit on how clean code can be. Since you gave me an excuse I will ask for more cleanliness. :wink:

src/server/internalServer.cpp
sgourdas: Lost in de-conflicting

sgourdas:
> There is no limit on how clean code can be. Since you gave me an excuse I will ask for more cleanliness. 😉

I will now stop talking, after saying that it looks very nice!

veloman-yunkan (approved):
@sgourdas Thanks for bearing with the perfectionist side of me! The result of it is that this code can now be safely run in a surgery room.

@kelson Despite my above assessment of the quality of the code of this PR, I am reminding that cleanliness is orthogonal to functionality. A hammer can be super-clean and perfectly disinfected however it is not suitable for performing colonoscopy. Though I paid attention to the semantics of the code during the final review I didn't test it and hence don't have 100% confidence against the infamous treacherousness of C++. Please repeat your testing before merging.

kelson42:
@veloman-yunkan Thx but unable to test, the kiwix-tools PR does not compile!

sgourdas:
> @veloman-yunkan Thx but unable to test, the kiwix-tools PR does not compile!

@kelson42 it should be good now.

kelson42:
LGTM

src/tools/networkTools.cpp
veloman-yunkan: > Ok so if I understand correctly, for this point, I should just add `169.254` for the v4 prefixes, right?

Adding the `169.254` prefix may have been a mistake. While testing the fix for #1144 I ran against a situation when the public API displayed by `kiwix-serve` starts with this prefix but that address doesn't work.

src/tools/networkTools.cpp
```diff
@@ -211,40 +217,35 @@ std::map<std::string, std::string> getNetworkInterfaces() {
...
 #endif
-
-  const char* const prefixes[] = { "192.168", "172.16.", "10.0" };
-  for(auto prefix : prefixes){
-    for(auto& itr : interfaces) {
-      std::string interfaceIp(itr.second.addr);
-      if (interfaceIp.find(prefix) == 0 && !(ipv6 && itr.second.addr6.empty())) {
-        bestPublicIp = itr.second;
-        break;
+  const char* const v4prefixes[] = { "192.168", "172.16", "10.0", "169.254" };
```
veloman-yunkan, line 232: Adding the 169.254 prefix may have been a mistake. While testing the fix for https://github.com/kiwix/libkiwix/issues/1144 I ran against a situation when the public API displayed by kiwix-serve starts with this prefix but that address doesn't work.

kelson42: Not sure what this range actually means?

veloman-yunkan: @sgourdas Can you please try to reproduce the issue? For me it occurred under Windows.

sgourdas: @veloman-yunkan I confirm that this is an issue I run into as well.

veloman-yunkan: > Not sure what this range actually means?

@kelson42 It's a link-local address. Discussed here https://github.com/kiwix/libkiwix/pull/1132#discussion_r1774806791

kelson42: @sgourdas Please open asap an issue

veloman-yunkan: > @veloman-yunkan I confirm that this is an issue I run into as well.

@sgourdas Then please file an issue and let's fix it before libkiwix v14.0 is released.

PR #1136, opened 2024-09-21
Introduce ipAvailable network function

sgourdas:
This introduces `bool ipAvailable(const IpAddress& addr)` allowing us to check if a given IP address is configured on any network interface on the host system. It is used by `InternalServer::start` to check if the given ip is available.

Fixes kiwix/kiwix-tools#709

This is a branch from feature/best-public-ip, so awaiting #1132

### DIFF
--- a/include/server.h
+++ b/include/server.h
@@ -22,7 +22,7 @@
 
 #include <string>
 #include <memory>
-#include "common.h"
+#include "tools.h"
 
 namespace kiwix
 {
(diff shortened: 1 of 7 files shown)


sgourdas:
@kelson42 lets try to fit this in the libkiwix 14. It can be a useful tool

veloman-yunkan:
Let's do this on #1132 directly.

PR #1145, opened 2024-10-03
Fix broken CI on Linux (aarch64, focal)

### LINKED ISSUE #1143: [REGRESSION] Kiwix-Build CI/CD crashes on libkiwix on aarch64
See https://github.com/kiwix/kiwix-build/actions/runs/11154308452/job/31003365353#step:8:552

We should:
* Fix the problem
* Explain why the libkiwix CI does not have caught this compilation problem

veloman-yunkan:
Should fix #1143

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -89,7 +89,8 @@ bool ipAvailable(const std::string addr)
 {
   auto interfaces = kiwix::getNetworkInterfacesIPv4Or6();
 
-  for (const auto& [_, interfaceIps] : interfaces) {
+  for (const auto& kv : interfaces) {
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
The fix is now being validated by kiwix/kiwix-build#712. In that PR I observed the Linux(aarch64+focal) jobs fail, then fixed the compilation and verified the fix by re-running the failed jobs. Then I started a fresh run of all CI jobs to make sure that the fix doesn't break something else (the run is now in progress).

veloman-yunkan:
kiwix-build CI passed - see https://github.com/kiwix/kiwix-build/actions/runs/11162493037?pr=759

PR #1147, opened 2024-10-05
Disable IPV6_ONLY to fix kiwix-serve on Windows

### LINKED ISSUE #1144: kiwix-serve cant instantiate HTTP daemon on windows
Using https://tmp.kiwix.org/ci/dev_preview/zim-tools_windows/kiwix-tools_win-x86_64-2024-09-02.zip on Windows 11 (VM) I can't seem to start kiwix-serve.

Either as regular user or administrator, I always get the following error:

```
C:\Users\reg\test-tools>kiwix-serve.exe -p 8001 ..\python-libzim\rlz\test.zim
Unable to instantiate the HTTP daemon. The port 8001 is maybe already occupied or need more permissions to be open. Please try as root or with a port number higher or equal to 1024.
```

sgourdas:
This fixes support of IPv6 for kiwix-serve on Windows by disabling IPV6_ONLY. By default Linux systems seem to do this internally, so the issue became Windows specific.

Fixes #1144

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -75,9 +75,11 @@ void Server::setRoot(const std::string& root)
   }
 }
 
-// FIXME: this method is implemented under the assumption that it is invoked only once (per object).
 void Server::setAddress(const std::string& addr)
 {
+  m_addr.addr.clear();
+  m_addr.addr6.clear();
+
   if (addr.empty()) return;
 
   if (addr.find(':') != std::string::npos) { // IPv6
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -529,6 +529,12 @@ bool InternalServer::start() {
       closesocket(sock);
       return false;
     }
+
+    if (::bind(sock, (struct sockaddr*)&sockAddr6, sizeof(sockAddr6)) == SOCKET_ERROR) {
+      std::cerr << "ERROR: Failed to bind IPv6 socket" << std::endl;
+      closesocket(sock);
+      return false;
+    }
   }
 #endif
   mp_daemon = MHD_start_daemon(flags,
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -544,9 +544,10 @@ bool InternalServer::start() {
                             MHD_OPTION_THREAD_POOL_SIZE, m_nbThreads,
                             MHD_OPTION_PER_IP_CONNECTION_LIMIT, m_ipConnectionLimit,
                             MHD_OPTION_END);
+
   if (mp_daemon == nullptr) {
-    std::cerr << "Unable to instantiate the HTTP daemon. The port " << m_port
-              << " is maybe already occupied or need more permissions to be open. "
+    std::cerr << "ERROR: Unable to instantiate the HTTP daemon. The port " << m_port
+              << " may already be in use, or more permissions are required to open it. "
                  "Please try as root or with a port number higher or equal to 1024."
               << std::endl;
 #ifdef _WIN32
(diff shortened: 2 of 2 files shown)


sgourdas:
@kelson42 is it possible for you to test how this works on your end? 

This issue is fixed, but for me it seems like there is a different problem. The IPs printed in the terminal do not seem to work for me, which would have to be an issue on a different point in lib.

kelson42:
I can not test on Windows. Whatbdoes this PR means for the end user?

sgourdas:
> I can not test on Windows. Whatbdoes this PR means for the end user?

The daemon now manages to start correctly, but the output urls do not work, so essentially it does not really provide usability. 

In that case I will try to investigate further for other points of failure as well, but they are unclear ATM.

kelson42:
@sgourdas We need the feature to work the same on GNU/Linux and Windows. What is the status? Do you need help?

sgourdas:
> @sgourdas We need the feature to work the same on GNU/Linux and Windows. What is the status? Do you need help?

The initial issue of binding on Windows has been fixed, but there is another one blocking the serving of the content from what it seems. I am still investigating on that.

sgourdas:
Honestly the behavior is very weird/inconsistent on my end. Do you have any ideas @kelson42 @veloman-yunkan?

The daemon start gives no error now, but the handler is not being invoked.

P.S. My last fixup commit does not really provide any improvement it seems like.

kelson42:
@veloman-yunkan Can you please have a look? We really need to release libkiwix14

veloman-yunkan:
> @veloman-yunkan Can you please have a look? We really need to release libkiwix14

@kelson42 I am at it now.

veloman-yunkan:
The problem seems to be that under Windows `libmicrohttpd` is being compiled without support for IPv6. If that is fixed, most likely, we will simply have to drop this PR.

sgourdas:
> The problem seems to be that under Windows `libmicrohttpd` is being compiled without support for IPv6. When that is fixed, we will most likely simply have to drop this PR.

 @veloman-yunkan I think that the disabling of IPV6_ONLY will still needed since before the change the daemon was returning null.  Also, from what I have seen, it is good practice to also use `WSAStartup` while it does not seem to make a difference here.

sgourdas:
@veloman-yunkan I passed the `-Denable_ipv6=true` flag to `libmicrohttpd.py` and rebuilt but it still does not seem to work. Am I missing something?

kelson42:
@veloman-yunkan So we need a patch in kiwix-build?

veloman-yunkan:
> @veloman-yunkan I passed the `-Denable_ipv6=true` flag to `libmicrohttpd.py` and rebuilt but it still does not seem to work. Am I missing something?

The `-D` option is not a magic wand. For `-Denable_ipv6=true` to have some effect an `enable_ipv6` option should be present in `meson_options.txt` and/or `meson.build` of `libmicrohttpd` source tree. Otherwise, why would we bother investigating bugs and fixing them? We could instead do `-Dfix_kiwix_libkiwix_issue1144=true` :smile:

sgourdas:
> > @veloman-yunkan I passed the `-Denable_ipv6=true` flag to `libmicrohttpd.py` and rebuilt but it still does not seem to work. Am I missing something?
> 
> The `-D` option is not a magic wand. For `-Denable_ipv6=true` to have some effect an `enable_ipv6` option should be present in `meson_options.txt` and/or `meson.build` of `libmicrohttpd` source tree. Otherwise, why would we bother investigating bugs and fixing them? We could instead do `-Dfix_kiwix_libkiwix_issue1144=true` 😄

Makes sense. I reached the point where finding no solution after hours makes me want to become a wizard.

kelson42:
Superseeded by https://github.com/kiwix/kiwix-build/pull/761

PR #1151, opened 2024-10-09
Fix Server::setAddress

### LINKED ISSUE #1150: Server::setAddress cannot be used more than once for an object

sgourdas:
This brings a change from the closed #1147 that fixes multiple uses of Server::setAddress.

Now the `m_addr` field is cleared before setting a new ip.

Helps a lot especially in kiwix-desktop local server, when multiple servers may be started.

Fixes #1150

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -75,9 +75,11 @@ void Server::setRoot(const std::string& root)
   }
 }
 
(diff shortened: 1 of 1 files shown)


sgourdas:
@kelson42 would be good to include this in the 14 release as well.

veloman-yunkan (approved):

PR #1168, opened 2024-12-20
Update a few dependencies in the workflows

kelson42:
@rgaudin Removed `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1` and not sure about that move, therefore putting you as reviewer.


rgaudin (approved):
LGTM ; this was part of a workaround

PR #1173, opened 2025-01-07
Two compilation fixes for FreeBSD

OlCe2:
Note that in both cases we are not special-casing FreeBSD, as the first case is in fact Linux-specific, and the second case is just following what POSIX prescribes (the change works on Linux too).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,14 +4,40 @@ project('libkiwix', 'cpp',
   default_options : ['c_std=c11', 'cpp_std=c++17', 'werror=true'])
 
 compiler = meson.get_compiler('cpp')
-
 static_deps = get_option('static-linkage') or get_option('default_library') == 'static'
+extra_libs = []
 
-# See https://github.com/kiwix/libkiwix/issues/371
-if ['arm', 'mips', 'm68k', 'ppc', 'sh4'].contains(host_machine.cpu_family())
-  extra_libs = ['-latomic']
-else
-  extra_libs = []
+# Atomics as compiled by GCC or clang can lead to external references to
+# functions depending on the type size and the platform. LLVM provides them in
+# 'libcompiler_rt', which clang normally automatically links in, while GNU
+# provides them in 'libatomic', which GCC *does not* link in automatically (but
+# this is probably going to change, see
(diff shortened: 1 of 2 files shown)


OlCe2:
Well, wait, it seems that, although it works, the first commit is logically incorrent...

kelson42:
@OlCe2 Thank you. Please ensure commits make sense and are coherently organised. We will review the PR. Unfortunately, we don't have a CI for FreeBSD so it is difficult to keep everything compiling fine.

OlCe2:
Yes, sorry, I figured out after pushing the PR that the first commit is not really correct (I mean, it would be better than the current situation, but still logically incorrect for reasons I will explain later). Working on an alternative, will update this PR when done.

OlCe2:
Ok, this is ready to be reviewed.

meson.build
```diff
@@ -7,11 +7,39 @@ compiler = meson.get_compiler('cpp')
...
+if compiler.links(atomics_program,
+                  name: 'compiler driver readily supports atomics')
+	extra_libs = []
 else
-  extra_libs = []
+  libatomic = compiler.find_library('atomic')
+  compiler.links(atomics_program, name: 'atomics work with libatomic',
+                 dependencies: libatomic, required: true)
+  extra_libs = ['-latomic']
 endif
```
veloman-yunkan, line 43: I know that this is the first step in building the `extra_libs` collection. However since it has become much more verbose now, I would prefer that `extra_libs = []` is moved upward outside the `if` (and probably placed under the definition of `compiler`) and the inclusion of `libatomic` in `extra_libs` is performed via `+=` similar to how that collection is populated with other libs. Then this code will be safer to shuffle around.

veloman-yunkan (changes requested):
LGTM, but let's improve it a little.

veloman-yunkan (approved):
Perfect!

OlCe2:
Thanks!

OlCe2:
I hope to be able to dedicate a bit of time to see how to set up a FreeBSD CI in your project in the future, that would help avoiding breakage (although we do workaround it in our ports tree with custom patches).

PR #1174, opened 2025-01-13
meson.build: Comment on 'threads' dependency required for FreeBSD

OlCe2:
While here, wrap the long test line.


OlCe2:
Forgot to add that there's no functional change here.

PR #1175, opened 2025-01-27
Autocompleter box size matches the content width

veloman-yunkan:
Note however that no upper limit is set on the width of the autocompleter box - it is possible, but I don't see how we could come up with a good value for it.

Fixes [#674](https://github.com/kiwix/kiwix-tools/issues/674)

### DIFF
--- a/static/skin/autoComplete/css/autoComplete.css
+++ b/static/skin/autoComplete/css/autoComplete.css
@@ -23,7 +23,9 @@
 
 .autoComplete_wrapper > ul {
   position: absolute;
+  min-width: 100%;
   width: fit-content;
+  max-width: 600px;
   max-height: 226px;
   overflow-y: scroll;
   top: 100%;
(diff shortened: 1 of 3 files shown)


kelson42:
@veloman-yunkan I propose 300px (it's arbitrary), we should have ellipsis if this happens and HTML hint, so if someone if `hover` its' still possible to see the full title. What do you think?

veloman-yunkan:
Truncating the suggestion and adding a hint (to be displayed on hover) will be a lot more work. And it will be mostly unused on mobile. Should we really pursue it?

kelson42:
@veloman-yunkan All of this should be doable via a few CSS directives... or do I get it wrong!

veloman-yunkan:
> @veloman-yunkan All of this should be doable via a few CSS directives... or do I get it wrong!

Indeed, [`text-overflow: ellipsis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) solves the part I was most concerned with.

kelson42 (changes requested):
LGTM, only a few remarks:
* `300px` is too small. Not sure about the right attitude: removing this or putting something significantly larger. Probably putting `600px` is the best compromise.
* I don't understand why putting the new CSS rules in `kiwix.css` would be better. Looks good to me like it is.

veloman-yunkan:
> * I don't understand why putting the new CSS rules in `kiwix.css` would be better. Looks good to me like it is.

`autoComplete.css` was imported from [another project](https://github.com/TarekRaafat/autoComplete.js).

PR #1176, opened 2025-01-29
Handling of external links with target="_blank"

veloman-yunkan:
Fixes kiwix/kiwix-tools#678

External links with `target="_blank"` attribute are opened in a new tab/window.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -345,6 +345,13 @@ function setHrefAvoidingWombatRewriting(target, url) {
   target._no_rewrite = old_no_rewrite;
 }
 
+function linkShouldBeOpenedInANewWindow(linkElement, mouseEvent) {
+  return linkElement.getAttribute("target") == "_blank"
+      || mouseEvent.ctrlKey
+      || mouseEvent.shiftKey
+      || mouseEvent.metaKey /* on Macs */;
+}
+
 function onClickEvent(e) {
   const iframeDocument = contentIframe.contentDocument;
   const target = matchingAncestorElement(e.target, iframeDocument, "a");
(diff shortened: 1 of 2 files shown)


static/skin/viewer.js
```diff
@@ -355,7 +359,7 @@ function onClickEvent(e) {
 
     if (isExternalAppUrl || isExternalUrl(target_href)) {
       const possiblyBlockedLink = blockLink(target_href);
-      if ( e.ctrlKey || e.shiftKey ) {
+      if ( e.ctrlKey || e.shiftKey || linkShouldBeOpenedInANewWindow(target) ) {
```
benoit74, line 362: That's very interesting, this is why it is also not working with MacOS command (⌘) key. I did not dared to speak about it in the issue. I should. I think we should add at least `e.metaKey` to this list.

veloman-yunkan: On Windows keyboards, `metaKey` is the Windows key (⊞). Clicking a link with that key pressed doesn't open a new tab or window (at least not in Firefox under Ubuntu). Should the `e.metaKey` check be combined with a test that the user agent runs on a Mac?

kelson42: No useful feedback to bring here IMHO. Even if a bit cumbersome, it seems you are on the appropriate path.

veloman-yunkan: @benoit74 This will require browser sniffing which isn't considered a good practice (see [comments to this answer on StackOverflow](https://stackoverflow.com/a/20087506)). I will simply include the `metaKey` in the list. I don't think that it is used as a modifier key for clicks under Windows/Linux.

benoit74: I'm fine with this solution as well, I doubt anyone will complain about strange behavior under Windows/Linux when they Meta+click on a link. While I'm sure Mac users might.

benoit74 (commented):
With my limited understanding of C++, LGTM, thanks

PR #1196, opened 2025-05-19
Nicer Internal Server Error page

veloman-yunkan:
Updated the Internal Server Error page to match the [new 404 (Content Not Found) error page](https://github.com/kiwix/libkiwix/pull/1178).

The page can be tested on https://github.com/kiwix/libkiwix/blob/14.0.0/test/data/poor.zim by trying to load`A/redirect_loop.html` (i.e. navigate to the URL like the following one http://localhost:8080/viewer#poor/A/redirect_loop.html after starting `kiwix-serve` on the said ZIM file).

Note that the title of the page has not been changed (it is still "Internal Server Error").

Since the new text on the page has not gone through translatewiki update process, you may see stale translations if you select any real language. In order to validate that the page translation works properly set the UI language to `test` by running the following command in the Developer tools console: `window.localStorage.setItem('userlang', 'test')`.


veloman-yunkan:
Unlike the old 500 error page, the new one doesn't contain any information about the error (as reported by the backend code). I can add it if needed and the HTML/CSS for that section is provided.

rgaudin (approved):
LGTM ; what kind of additional information could we get?

veloman-yunkan:
> what kind of additional information could we get?

The message of the C++ exception raised while handling the request. It won't be translated.

veloman-yunkan:
PR was updated:
- Used the [illustration](https://github.com/kiwix/kiwix-tools/issues/727#issuecomment-2894018805) provided by @siemsie
- Additional error info (if available) is displayed in a white textbox as is (i.e. untranslated and without any clarifying text like "Unhandled error:", "Server reported the following error:", etc).

PR #1200, opened 2025-06-07
Fix kiwix-serve pagination

kelson42:
Fixes https://github.com/kiwix/kiwix-tools/issues/744

Fix pagination which was (most of the time) wrong by an offset one because of a mixup between index displayed (starting with `1`) and index in the resultset (starting with `0`)

### DIFF
--- a/test/server_search.cpp
+++ b/test/server_search.cpp
@@ -758,8 +758,8 @@ struct TestData
                        ? R"(No results were found for <b>"PATTERN"</b>)"
                        : R"(Results <b>FIRSTRESULT-LASTRESULT</b> of <b>RESULTCOUNT</b> for <b>"PATTERN"</b>)";
 
-    const size_t lastResultIndex = std::min(totalResultCount, firstResultIndex + results.size() - 1);
-    header = replace(header, "FIRSTRESULT", std::to_string(firstResultIndex));
+    const size_t lastResultIndex = std::min(totalResultCount, firstResultIndex + results.size());
+    header = replace(header, "FIRSTRESULT", std::to_string(firstResultIndex+1));
     header = replace(header, "LASTRESULT",  std::to_string(lastResultIndex));
     header = replace(header, "RESULTCOUNT", std::to_string(totalResultCount));
     header = replace(header, "PATTERN",     getPattern());
@@ -940,7 +940,7 @@ TEST(ServerSearchTest, searchResults)
       /* start */            -1,
       /* resultsPerPage */   0,
       /* totalResultCount */ 0,
-      /* firstResultIndex */ 1,
+      /* firstResultIndex */ 0,
       /* results */          {},
       /* pagination */       {}
     },
@@ -951,7 +951,7 @@ TEST(ServerSearchTest, searchResults)
       /* start */            -1,
       /* resultsPerPage */   0,
       /* totalResultCount */ 0,
-      /* firstResultIndex */ 1,
+      /* firstResultIndex */ 0,
       /* results */          {},
       /* pagination */       {}
     },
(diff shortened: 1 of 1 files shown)


kelson42:
@veloman-yunkan I let you confirm the fix is correct before fixing all the tests

src/server/internalServer.cpp
```diff
@@ -1004,11 +1004,11 @@ std::unique_ptr<Response> InternalServer::handle_search_request(const RequestCon
     return response;
   }
 
-  const auto start = max(1u, request.get_optional_param("start", 1u));
+  const auto start = max(0u, request.get_optional_param("start", 0u));
```
veloman-yunkan, line 1024: This changes the semantics of the `start` parameter of the pagination control of the [`/search` API endpoint of `kiwix-serve`](https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#search) from 1-based indexing to 0-based. Either the bug should be fixed in a different way or the documentation in `kiwix/kiwix-tools` has to be updated, too.

kelson42: Indeed. To me correct default value is `0`, therefore I have updated the documentation at https://github.com/kiwix/kiwix-tools/pull/747

veloman-yunkan: With this change there a mismatch between how the results are indexed in the query (0-based) and in the output (1-based). I think that 1-based indexing is more common for non-IT (i.e. majority of) people.

kelson42: Yes, this is on purpose and normal. The query is a technical thing and the labelling is for users. It should always have been like this.

veloman-yunkan: OK, then go ahead with fixing the tests.

kelson42:
@veloman-yunkan Last round?

test/server_search.cpp
```diff
@@ -758,8 +758,8 @@ struct TestData
                        ? R"(No results were found for <b>"PATTERN"</b>)"
                        : R"(Results <b>FIRSTRESULT-LASTRESULT</b> of <b>RESULTCOUNT</b> for <b>"PATTERN"</b>)";
 
-    const size_t lastResultIndex = std::min(totalResultCount, firstResultIndex + results.size() - 1);
-    header = replace(header, "FIRSTRESULT", std::to_string(firstResultIndex));
+    const size_t lastResultIndex = std::min(totalResultCount, firstResultIndex + results.size());
```
veloman-yunkan, line 761: Now there is a subtle confusion here - `firstResultIndex` is a 0-based index, while `lastResultIndex` is 1-based

test/server_search.cpp
```diff
@@ -1028,10 +1028,10 @@ TEST(ServerSearchTest, searchResults)
 
     {
       /* query */          "pattern=yellow&books.id=" RAYCHARLESZIMID,
-      /* start */            1,
+      /* start */            0,
```
veloman-yunkan, line 1031: After this change this test point became identical to the one preceding it.

veloman-yunkan (commented):
There are two minor issues, but the PR can be merged without fixing them.

test/server_search.cpp
kelson42: That sounds quite bad to me. I will have a look.

test/server_search.cpp
kelson42: Not sure what is the best option here... do you?

veloman-yunkan: I think that the better option is to remove the preceding test point, but I don't mind if you remove this test point instead :)

test/server_search.cpp
kelson42: Actually, I don't see the problem, `lastResultIndex` is OK and it is a proper computer index (starting at 0.

PR #1204, opened 2025-06-08
Stop HTML encoding by mustache.js

kelson42:
Fixes https://github.com/kiwix/kiwix-tools/issues/690

Applies [mustache.js recommended approach to avoid automatic HTML encoding](https://github.com/janl/mustache.js?tab=readme-ov-file#variables).


static/skin/i18n/en.json
```diff
@@ -44,12 +44,12 @@
 	, "search-results-page-title": "Search: {{SEARCH_PATTERN}}"
 	, "search-results-page-header": "Results <b>{{START}}-{{END}}</b> of <b>{{COUNT}}</b> for <b>\"{{{SEARCH_PATTERN}}}\"</b>"
 	, "empty-search-results-page-header": "No results were found for <b>\"{{{SEARCH_PATTERN}}}\"</b>"
-	, "search-result-book-info": "from {{BOOK_TITLE}}"
+	, "search-result-book-info": "from {{{BOOK_TITLE}}}"
```
veloman-yunkan: Unlike the other two translations that are used to set DOM attributes via JS, this one is used when rendering an HTML or XML template so it should undergo HTML encoding, otherwise book titles containing text that looks like a HTML tag (esp. `</div>`) will result in invalid HTML on the search results page. You can check that this change isn't needed by selecting the bottommost "containing 'searchterm'..." suggestion and paying attention to the "from BookTitle" line below each search result.

veloman-yunkan (changes requested):

static/skin/i18n/en.json
kelson42: This is with my fix:
Screenshot from 2025-06-09 14-01-15

This is without the fix:
Screenshot from 2025-06-09 14-03-13

Your explanation makes sense, but there is something wrong somewhere else then.

kelson42: The problem is that actually we have here a double encoding, see:
```
            <a href="/content/cest-pas-sorcier_fr_all_2024-11/index/comment_pousse_la_canne_a_sucre_c_est_pas_sorcier-DLE3">
              Comment pousse la canne à sucre ? - C&apos;est pas sorcier
              <div class="book-title">from C&amp;apos;est pas sorcier</div>
```

kelson42: @veloman-yunkan I have made an other fix, because like I suspected we face here a double encoding by Mustache.js.... Not sure what is the best, but pretty sure we have other problem of that kind in other places. What do you think?

veloman-yunkan (approved):

PR #1208, opened 2025-06-15
Popups are allowed to escape the browser sandbox

veloman-yunkan:
Fixes kiwix/kiwix-tools#683

### DIFF
--- a/src/server/internalServer.cpp
+++ b/src/server/internalServer.cpp
@@ -1134,6 +1134,7 @@ const std::string CONTENT_CSP_HEADER =
             "allow-same-origin "
             "allow-modals "
             "allow-popups "
+            "allow-popups-to-escape-sandbox "
             "allow-forms "
             "allow-downloads;";
 


veloman-yunkan:
@Jaifroid Since the implementation of content security policy in `kiwix-serve` was largely (if not exclusively) based on your input in #916, what's your opinion on this change? Can it have any negative effects?

kelson42 (commented):
@veloman-yunkan Testing your PR. To me it does not change anything (visible). I still get 
image

veloman-yunkan:
@kelson42 I had the same problem with non-incognito window of chromium. Please make sure that it doesn't use the cached responses from previous testing.

Jaifroid:
@veloman-yunkan This is a known ~issue~ feature of Chrome's PDF support. PDFs are considered to be untrusted because they can contain active content that can make web calls it is undesirable to allow. So the whole PDF gets blocked by the sandbox. Normally a JS-based viewer gets round this. Are we talking about placing PDFs themselves in a popup, browser-managed, not via an included JS viewer? In that case it's usually better just to force the PDF to open in a new browser tab or new browser window. However, I wasn't aware of this sandbox attribute, so if it solves the issue of an overprotective CSP, then I can't see any real harm in it. The risk is that PDFs may "phone home". If you're happy with the potential risk, then the change doesn't seem bad. There are always trade-offs between security and useability.

The main thing to be sure of is that the content going into the popup is definitely from the ZIM, and not merely a redirect to an external PDF. This was a problem with Zimit-based ZIMs, and one of the reasons we put the sandbox in in the first place: to give users some assurance that they won't suddenly end up navigating the Web, which could put some users in danger. I guess those users would be better off putting their system into offline mode than trusting our implementation 100%.

Jaifroid (approved):
If you're satisfied that the risk is low, then this seems fine to me.

veloman-yunkan:
@Jaifroid Thank you for elaborate feedback!

PR #1212, opened 2025-08-11
Viewer chaperon mode

veloman-yunkan:
Fixes kiwix/kiwix-tools#591

Viewer now rewrites internal links so that opening them in a new tab/window keeps the viewer around. Thus the viewer acts as a chaperon for the users preventing them from finding themselves out of the viewer's supervision. Of course there are ways to circumvent such oversight, however it has always been the case with chaperons in all cultures in all epochs.

I seem to have ended up with a somewhat simpler solution than described in https://github.com/kiwix/kiwix-tools/issues/591#issuecomment-1525787724, however it should have [the same shortcomings](https://github.com/kiwix/kiwix-tools/issues/591#issuecomment-3057492750). In particular, the fix may have the following negative effects:

- link preview functionality (if the page contains such a feature) should be broken
- the fix may interfere with other click event handlers installed on internal links

Also note that this change hasn't been tested on
- ~Zimit 2 ZIM files (where wombat also meddles with the links)~ (it turns out that the potential problem with wombat is behind a barrier - warc2zim uses an iframe of its own, and this fix only goes after links in the top frame).
- the various scenarios with PDFs for which there have been a number of issues reported before
- ~links with non-empty fragment component (e.g. (`./otherpage#subsection1`)~ (any problems with `#anchor` links should be fixed by #1213)

Above list was compiled while considering potential negative effects of the fix and may be incomplete. Even after those scenarios are tested and any discovered problems happen to be easy to fix, there is still a possibility that this change introduces a regression on something else that was omitted from the analysis.

Therefore I propose to make this feature optional (controllable from the command line of `kiwix-serve`).

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -10,6 +10,8 @@ let viewerState = {
   uiLanguage: 'en',
 };
 
+const FULL_ROOT_URL = `${window.location.origin}${root}`;
+
 function dropUserLang(query) {
   const q = new URLSearchParams(query);
   q.delete('userlang');
@@ -347,6 +349,12 @@ function onClickEvent(e) {
   if (target !== null && "href" in target) {
     const target_href = getRealHref(target);
     const target_url = new URL(target_href, iframeDocument.location);
+    if ( target_url.href.startsWith(`${FULL_ROOT_URL}/viewer#`) &&
+         !linkShouldBeOpenedInANewWindow(target, e) ) {
+      contentIframe.contentWindow.parent.location = target_url;
+      e.preventDefault();
+      return;
+    }
     const isExternalAppUrl = urlMustBeHandledByAnExternalApp(target_url);
     if ( (isExternalAppUrl && !viewerSettings.linkBlockingEnabled)
          || goingToOpenALinkToAnUndisplayableResource(target_url) ) {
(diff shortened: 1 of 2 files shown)


static/skin/viewer.js
```diff
@@ -347,6 +349,12 @@ function onClickEvent(e) {
   if (target !== null && "href" in target) {
     const target_href = getRealHref(target);
     const target_url = new URL(target_href, iframeDocument.location);
+    if ( target_url.href.startsWith(`${FULL_ROOT_URL}/viewer#`) &&
+         !linkShouldBeOpenedInANewWindow(target, e) ) {
+      contentIframe.contentWindow.parent.location = target_url;
+      e.preventDefault();
+      return;
```
kelson42, line 356: Just to be sure: JS can stack a list of event handlings to the same event. It is really important that the rest of the stack of events already associated to the mouse click continues to be handled like before. Can you please confirm this?

kelson42 (commented):
Il will make user testing this evening. Thank you again.

static/skin/viewer.js
veloman-yunkan: I didn't think about that. But I think that it is in general impossible to implement this hack without affecting some (possibly exotic) way of custom handling a click event.

kelson42 (approved):
I have tested this PR and to me, at this stage is good enough. We all know this is a kind of a hack, let see *in practive* how viable is this hack. Difficult to me to provide a better way to assess this improvement.

benoit74:
I have no idea how wget has implement the WARC generation, but it looks so ^^

Jaifroid:
> I have no idea how wget has implement the WARC generation, but it looks so ^^

@benoit74 @veloman-yunkan JFYI (because I worked on that a lot), you're both right in historical terms... Original warc2zim archives (the ServiceWorker implementation) were definitely inside their own iframe (with id "replay_iframe"), which is what was controlled by the Service Worker. The outer part of the article only acted as a bootstrapper for content in the WARC archive. So, for implementations that use an iframe to display ZIM articles (KJS, Kiwix Serve), you used to get an iframe inside the iframe. For other frameworks, you just got an iframe inside the webview. Zimit2 (warc2zim2) does away with that, since it relinquished the Service Worker, so all the recent warc-based ZIMs appear as 100% regular ZIMs without a nested iframe.

PR #1218, opened 2025-09-01

translatewiki:
Translation updates


static/skin/i18n/es-formal.json
```diff
@@ -0,0 +1,90 @@
+{
+	"@metadata": {
+		"authors": [
+			"Camachevaz"
+		]
+	},
+	"name": "Inglés",
```
veloman-yunkan: Same problem as in https://github.com/kiwix/kiwix-desktop/pull/1387#discussion_r2309687361

kelson42: @veloman-yunkan This a bug, a dev has pit a string here which represent a language and should not be translated. Can you please just fix this?

kelson42: whole file has been removed

PR #1220, opened 2025-09-06
Introduce Manager::addBooksFromDirectory()

### LINKED ISSUE #739: Allow to serve a full directory of ZIM files
Today `kiwix-serve` can serve:
* A list of ZIM files
* An XML library

There is no way to just give a directory and kiwix-serve will explore the library and load all the ZIM files.

To some extend, we can do that with globbing, but this is a kind of a workaround to the problem.

We should just allow to give a directory as argument and then `kiwix-serve` would visit the whole directory arborescence for ZIM files and load them. Any ZIM file which is not loadable (for example because not readable) should generate an error (like we already do today).

This issue is pretty much connected to #740 

This issue is a follow-up of https://github.com/kiwix/kiwix-tools/issues/735

juuz0:
Fixes https://github.com/kiwix/kiwix-tools/issues/739
Pre-requirement for: https://github.com/kiwix/kiwix-tools/pull/743

Added a function to load books from a directory. Requires rootPath to iterate over.

### DIFF
--- a/include/manager.h
+++ b/include/manager.h
@@ -155,6 +155,15 @@ class Manager
                        const std::string& url = "",
                        const bool checkMetaData = false);
 
+  /**
+   * Add all books from the directory tree into the library.
+   * 
+   * @param path          The path of the directory to scan.
+   * @param verboseFlag   Verbose logs flag.
+   */
+  void addBooksFromDirectory(const std::string& path,
+                             const bool verboseFlag = false);
+
   std::string writableLibraryPath;
 
   bool m_hasSearchResult = false;
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -23,6 +23,14 @@
 #include "tools/pathTools.h"
 
 #include <pugixml.hpp>
+#include <filesystem>
+#include <iostream>
+#include <set>
+#include <queue>
+#include <cctype>
+#include <algorithm>
+
+namespace fs = std::filesystem;
 
 namespace kiwix
 {
@@ -249,6 +257,58 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
   return !(
       this->addBookFromPathAndGetId(pathToOpen, pathToSave, url, checkMetaData)
           .empty());
+}
+
+void Manager::addBooksFromDirectory(const std::string& path,
+                                    const bool verboseFlag)
+{
+  std::set<std::string> iteratedDirs;
+  std::queue<std::string> dirQueue;
+  dirQueue.push(fs::absolute(path).u8string());
+  int totalBooksAdded = 0;
+  if (verboseFlag)
+    std::cout << "Adding books from the directory tree: " << dirQueue.front() << std::endl;
+
+  while (!dirQueue.empty()) {
+    const auto currentPath = dirQueue.front();
+    dirQueue.pop();
+    if (verboseFlag)
+      std::cout << "Visiting directory: " << currentPath << std::endl;
+    for (const auto& dirEntry : fs::directory_iterator(currentPath)) {
+      auto resolvedPath = dirEntry.path();
+      if (fs::is_symlink(dirEntry)) {
+        try {
+          resolvedPath = fs::canonical(dirEntry.path());
+        } catch (const std::exception& e) {
+          std::cerr << "Could not resolve symlink " << resolvedPath.u8string() << " to a valid path. Skipping..." << std::endl;
+          continue;
+        }
+      }
+      const std::string pathString = resolvedPath.u8string();
+      std::string resolvedPathExtension = resolvedPath.extension().u8string();
+      std::transform(resolvedPathExtension.begin(), resolvedPathExtension.end(), resolvedPathExtension.begin(),
+          [](unsigned char c){ return std::tolower(c); });
+      if (fs::is_directory(resolvedPath)) {
+        if (iteratedDirs.find(pathString) == iteratedDirs.end())
+          dirQueue.push(pathString);
+        else if (verboseFlag)
+          std::cout << "Already iterated over " << pathString << ". Skipping..." << std::endl;
+      } else if (resolvedPathExtension == ".zim" || resolvedPathExtension == ".zimaa") {
+        if (!this->addBookFromPath(pathString, pathString, "", false)) {
+          std::cerr << "Could not add " << pathString << " into the library." << std::endl;
+        } else if (verboseFlag) {
+          std::cout << "Added " << pathString << " into the library." << std::endl;
+          totalBooksAdded++;
+        }
+      } else if (verboseFlag) {
+        std::cout << "Skipped " << pathString << " - unsupported file type or permission denied." << std::endl;
+      }
+    }
+    iteratedDirs.insert(currentPath);
+  }
+
+  if (verboseFlag)
+    std::cout << "Traversal completed. Total books added: " << totalBooksAdded << std::endl;
 }
 
 bool Manager::readBookFromPath(const std::string& path, kiwix::Book* book)


include/manager.h
```diff
@@ -155,6 +155,17 @@ class Manager
                        const std::string& url = "",
                        const bool checkMetaData = false);
 
+  /**
+   * Add books from a directory into the libary.
+   * 
+   * @param path          The path of the directory to scan.
+   * @param path          If the function should stop for an invalid file.
```
veloman-yunkan: param name was not updated after copy&paste

include/manager.h
```diff
@@ -155,6 +155,17 @@ class Manager
                        const std::string& url = "",
                        const bool checkMetaData = false);
 
+  /**
+   * Add books from a directory into the libary.
```
veloman-yunkan: 1. libary -> library
2. Please make it explicit that this function adds all books found in the directory tree.

src/manager.cpp
```diff
@@ -251,6 +256,46 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
...
+        resolvedPath = fs::canonical(dirEntry.path());
+      }
+      const std::string pathString = resolvedPath.u8string();
+      if (fs::is_directory(resolvedPath)) {
+        if (iteratedDirs.find(pathString) == iteratedDirs.end())
+          dirQueue.push(pathString);
+        else if (verboseFlag)
+          std::cout << "Already iterated over " << pathString << ". Skipping..." << std::endl;
+      } else {
+        if (!this->addBookFromPath(pathString, pathString, "", false)) {
```
veloman-yunkan, line 297: 1. Is it wise trying to treat all files (regardless of their extension) as ZIM?
2. Shouldn't this operation also be reported in the verbose output?

src/manager.cpp
```diff
@@ -251,6 +256,46 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
...
+          if (skipInvalid)
+            std::cerr << "Skipping invalid file: " << pathString << std::endl;
+          else {
+            throw std::runtime_error("Unable to add file: " + pathString + " into the library.");
```
veloman-yunkan: Will this function ever be called with `skipInvalid=false`? From exception safety perspective its current implementation for that case is a little dubious - if an invalid book is encountered and the exception is caught the library is left in a state where some but not all books from the directory have been added. Does that make sense to the caller?

src/manager.cpp
```diff
@@ -251,6 +256,46 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
...
+          }
+        }
+      }
+    }
+    if (verboseFlag)
+      std::cout << "Completed iterating over directory: " << currentPath << std::endl;
```
veloman-yunkan: This message may be a little misleading for the breadth first traversal of the directory tree that is used.

veloman-yunkan (commented):
The output of the verbose mode may be somewhat confusing in the context of the full log - the purpose of iterating over files and directories is not obvious. Maybe the output should start (and end) with the description of the operation being performed.

include/manager.h
juuz0: Addressed in https://github.com/kiwix/libkiwix/pull/1220/commits/eb959f7cba99d36c3ef55b4dcd8dd29758c54ffc

include/manager.h
juuz0: Done in fixup commit

src/manager.cpp
juuz0: Yeah agreed. For now I removed this and hopefully the overall improved logs look better.

src/manager.cpp
juuz0: This was in the back of my head, I guess my motivation was the skipInvalid in kiwix-serve. But I agree, it makes little sense to have this in the library. Addressed in https://github.com/kiwix/libkiwix/pull/1220/commits/1baa14229b15dbdd79ca6bcdc0db64235d2d6460

src/manager.cpp
juuz0: 1. Probably not, and with the recent changes in https://github.com/kiwix/libkiwix/pull/1220/commits/1baa14229b15dbdd79ca6bcdc0db64235d2d6460. This would report a std:cerrr for each non ZIM file. I will push a commit for it but wanted to ask if you've tackled this before? Should checking the extensions for .zim, .zimaa, .zimab (and their capitals) be sufficient?

2. Done in fixup.

src/manager.cpp
```diff
@@ -257,18 +257,20 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
...
 void Manager::addBooksFromDirectory(const std::string& path,
-                                    const bool skipInvalid,
                                     const bool verboseFlag)
 {
   std::set<std::string> iteratedDirs;
   std::queue<std::string> dirQueue;
   dirQueue.push(fs::absolute(path).u8string());
+  int totalBooksAdded = 0;
+  if (verboseFlag)
+    std::cout << "Starting BFS traversal at root directory: " << dirQueue.front() << std::endl;
```
veloman-yunkan: I would prefer if the message reads "Adding books from the directory tree: /path/to/dir"

src/manager.cpp
veloman-yunkan: > Should checking the extensions for .zim, .zimaa, .zimab (and their capitals) be sufficient?

I think that you should go only after .zim and .zimaa files.

@kelson42 Please confirm

kelson42: 100% agree with @veloman-yunkan, this is already what I said in my first review, see https://github.com/kiwix/kiwix-tools/pull/743#issuecomment-3289809963

kelson42 (commented):
I wanted to compile and test the code but for me it does not compile:
```
skin/i18n/tr.json ['skin', 'i18n', 'tr_json']
skin/i18n/zh-hans.json ['skin', 'i18n', 'zh_hans_json']
skin/i18n/zh-hant.json ['skin', 'i18n', 'zh_hant_json']
[14/81] Compiling C++ object src/libkiwix.so.14.0.0.p/manager.cpp.o
FAILED: src/libkiwix.so.14.0.0.p/manager.cpp.o 
c++ -Isrc/libkiwix.so.14.0.0.p -Isrc -I../src -Iinclude -I../include -I/usr/include/kainjow -Istatic -I/usr/local/include -I/usr/include/x86_64-linux-gnu -I/usr/include/p11-kit-1 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=c++17 -O0 -g -fPIC -pthread -MD -MQ src/libkiwix.so.14.0.0.p/manager.cpp.o -MF src/libkiwix.so.14.0.0.p/manager.cpp.o.d -o src/libkiwix.so.14.0.0.p/manager.cpp.o -c ../src/manager.cpp
../src/manager.cpp: In member function ‘void kiwix::Manager::addBooksFromDirectory(const std::string&, bool)’:
../src/manager.cpp:263:8: error: ‘queue’ is not a member of ‘std’
  263 |   std::queue<std::string> dirQueue;
      |        ^~~~~
../src/manager.cpp:28:1: note: ‘std::queue’ is defined in header ‘<queue>’; this is probably fixable by adding ‘#include <queue>’
   27 | #include <iostream>
  +++ |+#include <queue>
   28 | #include <set>
../src/manager.cpp:263:25: error: expected primary-expression before ‘>’ token
  263 |   std::queue<std::string> dirQueue;
      |                         ^
../src/manager.cpp:263:27: error: ‘dirQueue’ was not declared in this scope
  263 |   std::queue<std::string> dirQueue;
      |                           ^~~~~~~~
[23/81] Compiling C++ object src/libkiwix.so.14.0.0.p/library.cpp.o
ninja: build stopped: subcommand failed.
```

kelson42:
@juuz0 Thx, I guess the compilation error is foxed? I will gove a new run on saturday.

kelson42:
> @kelson42 Yes, it should be. I didn't face that and probably veloman either, but from your error logs it showed missing the queue header. I added that explicitly now.

Yes, I use g++14 on Ubuntu 24.04.

kelson42:
> Just wanted to confirm no zim files are loaded into memory without request to do so...

@evrial Probably not the issue you wanted to comment on.

src/manager.cpp
juuz0: Done

src/manager.cpp
juuz0: Done

kelson42 (approved):
All my remarks have been taken in account and it LGTM.

I have rebased to ensure CI works well.

@veloman-yunkan Waiting your last iteration of technical review.

@juuz0 It seems to me the failing Windows CI finds its root cause in your PR considering that it works on `main` and that I just rebased. I would guess the same for the CI Codefactor.

src/manager.cpp
```diff
@@ -251,6 +259,58 @@ bool Manager::addBookFromPath(const std::string& pathToOpen,
...
+          std::cout << "Already iterated over " << pathString << ". Skipping..." << std::endl;
+      } else if (resolvedPathExtension == ".zim" || resolvedPathExtension == ".zimaa") {
+        if (!this->addBookFromPath(pathString, pathString, "", false)) {
+          std::cerr << "Could not add " << pathString << " into the library." << std::endl;
+        } else if (verboseFlag) {
+          std::cout << "Added " << pathString << " into the library." << std::endl;
+          totalBooksAdded++;
+        }
+      } else if (verboseFlag) {
+        std::cout << "Skipped " << pathString << " - unsupported file type or permission denied." << std::endl;
```
veloman-yunkan, line 304: 1. How can this happen because of permission problems?
2. Why should non-ZIM files be reported?

veloman-yunkan (commented):
LGTM, though I concur with CodeFactor's assessment regarding the complexity of the introduced method.

Shouldn't we also have a unit test for this new method? @kelson42

PR #1229, opened 2025-10-02
Add support for accessing books using a metadata-based ID 

### LINKED ISSUE #1057: Access books using a metadata-based ID
Although there are tickets that intersects with this discussion, such as #864; I'd like to express this one independently and clearly.

### Request

Being able to build a kiwix-serve URL with only in-ZIM (thus Metadata) information.

Ideally, I'd like to be able to call `/viewer#{bookId}` or `/content/{bookId}` and get to said book. `bookId` here being the ZIM UUID.

### Why would one need this?

- The `bookName` to a book is based on the filename on disk at time of addition to kiwix-serve. This is thus somewhat volatile.
- The `bookName` can be extracted from the OPDS catalog (link to `/content/`), of a running kiwix-serve. Needs live-querying of kiwix-serve.

Assuming I want to link to a specific content from outside of kiwix-serve ; if I can't rely on a running OPDS endpoint (as the libkiwix viewer does), I have to:

- known the filename passed to kiwix-serve
- [duplicate and maintain](https://github.com/offspot/offspot-config/blob/8a48b6f6f49f9930ec23fcbf24816aef5699a1f2/src/offspot_config/zim.py#L82) part of libkiwix code to generate the `bookName`.

---

Adding support for this seems quite simple.

If the _BookName part_ of URL looks like a UUID, it can be tested first and if it matches, a redirection to the current format happen.

This would ensures continuity of current behavior and allows UUID-based link to be presented as human in URL of browsers.

vighnesh-sawant:
Fixes #1057 

Changes:
* Added a filter to the catalog endpoint catalog/v2/entries?id="id", allows to filter based on uuid
* On the viewer endpoint added a check if supplied arg looks like a uuid using [regex](https://stackoverflow.com/questions/136505/searching-for-uuids-in-text-with-regex) and then fetched the bookName from the catalog endpoint and kept the rest same after that.
* `handle_location_hash_change()`made async (I dont think it will but comments will be helpful)
* On the `/content` endpoint, same check as viewer endpoint to see if it looks like a uuid, but this endpoint was handled in the server so just used a function to convert id to name and name to id.

Please do tell if i missed something! I would love to rectify it.

Also i just wanted to ask if there is a reason why /viewer is handled in client-side js while /content on server-side? I am just a little curious.

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -79,6 +79,7 @@ class Filter {
     uint64_t activeFilters;
     Tags _acceptTags;
     Tags _rejectTags;
+    std::string _id;
     std::string _category;
     std::string _lang;
     std::string _publisher;
@@ -141,7 +142,7 @@ class Filter {
      *  which case a book in any of those languages will match).
      */
     Filter& lang(std::string lang);
-
+    Filter& id(std::string id);
     Filter& publisher(std::string publisher);
     Filter& creator(std::string creator);
     Filter& maxSize(size_t size);
@@ -154,7 +155,8 @@ class Filter {
     bool hasQuery() const;
     const std::string& getQuery() const { return _query; }
     bool queryIsPartial() const { return _queryIsPartial; }
-
+    bool hasId() const;
+    const std::string& getId() const { return _id; }
     bool hasName() const;
     const std::string& getName() const { return _name; }
 
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -686,7 +686,10 @@ Xapian::Query langQuery(const std::string& commaSeparatedLanguageList)
 {
   return multipleParamQuery(commaSeparatedLanguageList, "L");
 }
-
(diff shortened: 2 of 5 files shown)


vighnesh-sawant:
The PR implements the redirect feature also, accessing by uuid redirects to the current format.
I'll wait for feedback and change the code as necessary!

benoit74:
I also don't see why kiwix-serve should prevent to serve multiple versions of the same ZIM Name (which would happen if we use ZIM Name as unique identifier instead of UUID).

This is not our usage as for now, but could be for some of our users, and this could even be handy for us in dev (https://dev.library.kiwix.org/) where one might want to be able to see all ZIMs we have on the filesystem to easily compare them.

The redirect from UUID to filename seems kinda mandatory to help users from my perspective.

PR #1231, opened 2025-10-08
Requires now libzim 9.4.0

kelson42:
Once https://github.com/openzim/libzim/pull/1008 merged


veloman-yunkan:
This will cause the kiwix-build CI/CD fail unless `main_project_versions` in [kiwixbuild/versions.py](https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/versions.py) is updated too.

PR #1234, opened 2025-10-26
Add functions which return displayable addresses

vighnesh-sawant:
Fixes https://github.com/kiwix/kiwix-tools/issues/764
Reference ->https://github.com/kiwix/kiwix-tools/pull/763

### DIFF
--- a/include/server.h
+++ b/include/server.h
@@ -69,6 +69,7 @@ namespace kiwix
        int getPort() const;
        IpAddress getAddress() const;
        IpMode getIpMode() const;
+       std::vector<std::string> getServerAccessUrls() const;
 
      protected:
        std::shared_ptr<Library> mp_library;
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -29,6 +29,22 @@
 
 namespace kiwix {
 
+namespace
+{
+
+std::string makeServerUrl(std::string host, int port, std::string root)
+{
+  const int httpDefaultPort = 80;
+
+  if (port == httpDefaultPort) {
+    return "http://" + host + root;
+  } else {
+    return "http://" + host + ":" + std::to_string(port) + root;
+  }
+}
+
+}  // unnamed namespace
+
 Server::Server(LibraryPtr library, std::shared_ptr<NameMapper> nameMapper) :
   mp_library(library),
   mp_nameMapper(nameMapper),
@@ -110,6 +126,18 @@ IpAddress Server::getAddress() const
 IpMode Server::getIpMode() const
 {
   return mp_server->getIpMode();
+}
+
+std::vector<std::string> Server::getServerAccessUrls() const
+{
+  std::vector<std::string> result;
+  if (!m_addr.addr.empty()) {
+    result.push_back(makeServerUrl(m_addr.addr, m_port, m_root));
+  }
+  if (!m_addr.addr6.empty()) {
+    result.push_back(makeServerUrl("[" + m_addr.addr6 + "]", m_port, m_root));
+  }
+  return result;
 }
 
 }


vighnesh-sawant:
@kelson42 Please do have a look at the two pull requests

vighnesh-sawant:
Is this fine? It works well on kiwix-serve and kiwix-desktop.
I'll apply this to the remaining platforms soon!

src/server.cpp
```diff
@@ -106,4 +106,43 @@ IpMode Server::getIpMode() const
   return mp_server->getIpMode();
 }
 
+std::string Server::getDisplayableAddress()
+{
+  kiwix::IpAddress addresses = Server::getAddress();   
+  if(addresses.addr.empty()) return getDisplayableAddress6(); //as a fallback return ipv6 address as one of them has to be set
```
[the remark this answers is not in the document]
vighnesh-sawant: The server would not start right if both addresses are missing? Do i need to handle this?

include/server.h
```diff
@@ -69,6 +69,8 @@ namespace kiwix
        int getPort() const;
        IpAddress getAddress() const;
        IpMode getIpMode() const;
+       std::string getDisplayableAddress();
+       std::string getDisplayableAddress6();
```
veloman-yunkan: 1. These methods can (and, therefore, should) be declared `const`
2. I would name these functions `getServerAccessUrl()`  and `getServerAccessIPv6Url()`. BTW, maybe it's better to have a single public function `std::vector<std::string> getServerAccessUrls() const`? @kelson42

src/server.cpp
```diff
@@ -106,4 +106,27 @@ IpMode Server::getIpMode() const
...
+  
+  std::string result = "http://";
```
veloman-yunkan: This variable is not justified. `return "http://" + addresses.addr + suffix;` is quite readable.

src/server.cpp
```diff
@@ -106,4 +106,27 @@ IpMode Server::getIpMode() const
   return mp_server->getIpMode();
 }
 
+std::string Server::getDisplayableAddress()
+{
+  kiwix::IpAddress addresses = Server::getAddress();
```
veloman-yunkan: I advise you to start practising defensive programming. A powerful technique is to declare `const` whatever can be declared `const`, which protects you against carelessly modifying things that are supposed to be immutable.

In particular, both `addresses` and `suffix` in this function can be declared `const`.

src/server/internalServer.cpp
```diff
@@ -272,6 +273,19 @@ struct NoDelete
...
+  std::string suffix;
+   if (m_port == HTTP_DEFAULT_PORT)
+  {
+    suffix = m_rootPrefixOfDecodedURL;
+  } else
+  {
+    suffix = ":" + std::to_string(m_port) + m_rootPrefixOfDecodedURL;
+  }
+  return suffix;
+}
```
veloman-yunkan: Please replace this member function of `InternalServer` with a free function in an unnamed namespace in `server.cpp` `std::string makeServerUrl(std::string host, int port, std::string root)`.

src/server.cpp
```diff
@@ -106,4 +108,31 @@ IpMode Server::getIpMode() const
...
+  }
+  return serverUrl;
+}
+
+std::vector<std::string> Server::getServerAccessUrls() const
+{
+  std::vector<std::string> serverAccessUrls;
+  const kiwix::IpAddress addresses = Server::getAddress();
+  const int port = Server::getPort();
+  const std::string root = mp_server->getRootPrefixOfDecodedURL();
```
veloman-yunkan: This is available in `Server` as `m_root`. So there is no need for the `getRootPrefixOfDecodedURL()` accessor in `InternalServer`.

src/server.cpp
```diff
@@ -106,4 +108,31 @@ IpMode Server::getIpMode() const
...
+    serverUrl = host + ":" + std::to_string(port) + root;
+  }
+  return serverUrl;
+}
+
+std::vector<std::string> Server::getServerAccessUrls() const
+{
+  std::vector<std::string> serverAccessUrls;
+  const kiwix::IpAddress addresses = Server::getAddress();
+  const int port = Server::getPort();
```
veloman-yunkan: Why don't you use `m_addr` and `m_port` directly?

src/server.cpp
```diff
@@ -106,4 +108,32 @@ IpMode Server::getIpMode() const
...
+  const std::string root = mp_server->getRootPrefixOfDecodedURL();
+  const std::string httpPrefix = "http://";
+  if (!addresses.addr.empty()) {
+    serverAccessUrls.push_back(
```
veloman-yunkan: By naming this variable `result` (and with the other proposed change) this expression should be short enough to be written on a single line.

vighnesh-sawant:
Incorporated all changes.
i have kept the httpPrefix variable since the other alternative will be to convert "[" to std::string
what is better?
httpPrefix + "[" + ....
"http://" + std::string("[") .....

src/server.cpp
```diff
@@ -106,4 +108,32 @@ IpMode Server::getIpMode() const
...
+        makeServerUrl(httpPrefix + addresses.addr, port, root));
```
veloman-yunkan: Let `makeServerUrl()` add the URL scheme prefix on its own.

src/server.cpp
```diff
@@ -106,4 +108,32 @@ IpMode Server::getIpMode() const
...
+std::string makeServerUrl(std::string host, int port, std::string root)
+{
+  std::string serverUrl;
+  if (port == HTTP_DEFAULT_PORT) {
+    serverUrl = host + root;
+  } else {
+    serverUrl = host + ":" + std::to_string(port) + root;
+  }
+  return serverUrl;
+}
```
veloman-yunkan: Hide this function in an unnamed namespace

src/server.cpp
```diff
@@ -27,6 +27,8 @@
 #include <zim/item.h>
 #include "server/internalServer.h"
 
+#define HTTP_DEFAULT_PORT 80
```
veloman-yunkan: `#define`s are the old C-style way of declaring constants. Better use a `const int` variable in an unnamed namespace.

src/server.cpp
```diff
@@ -106,4 +108,32 @@ IpMode Server::getIpMode() const
   return mp_server->getIpMode();
 }
 
+std::string makeServerUrl(std::string host, int port, std::string root)
+{
+  std::string serverUrl;
```
veloman-yunkan: This function is simple enough that you can avoid this non-const variable by returning the result of the expression directly.

src/server.cpp
veloman-yunkan: > in an unnamed namespace.

No, we don't need such a broad scope for this constant. Declare it right in `makeServerUrl()`.

src/server.cpp
vighnesh-sawant: m_root does not get normalized completely
m_rootPrefixOfDecodedURL of internal server is normalized and decoded 
m_root of internalserver gets encoded

src/server.cpp
vighnesh-sawant: m_addr of server class may be left empty so it wont work correctly
m_addr of internal server gets initialized correctly thats why
m_port works but i think its better they both are same

src/server.cpp
veloman-yunkan: I see. I was lost in the multitude of `root` and `m_root` variables.

Regarding normalization of root location - `Server::setRoot()` performs some steps in that direction but since it contains a loophole `InternalServer` has to correct. So I think that the code will be simpler if normalization is fully performed in `Server::setRoot()` and `InternalServer` receives a fully normalized value. Please make this change in a separate refactoring commit.

src/server.cpp
veloman-yunkan: You are right regarding `m_addr`. This is a pitfall (I tell you so after emerging from it with your help :smiley: ). It would be better if `Server::m_addr` and `InternalServer::m_addr` don't diverge. Please update `Server::m_addr` in `Server::start()` and make `Server::getAddress()` a simple accessor. Let this be another refactoring commit. Then you can use `m_addr` and `m_port` directly in this function.

src/server.cpp
```diff
@@ -112,4 +128,19 @@ IpMode Server::getIpMode() const
   return mp_server->getIpMode();
 }
 
+std::vector<std::string> Server::getServerAccessUrls() const
+{
+  std::vector<std::string> result;
+  const kiwix::IpAddress addresses = m_addr;
+  const int port = m_port;
+  const std::string root = m_root;
```
veloman-yunkan: Now these local variables are not justified. Use `m_addr`, `m_port` and `m_root` below directly.

src/server.cpp
```diff
@@ -112,4 +128,19 @@ IpMode Server::getIpMode() const
...
+  if (!addresses.addr.empty()) {
+    result.push_back(makeServerUrl(addresses.addr, port, root));
+  }
+  if (!addresses.addr6.empty()) {
+    result.push_back(makeServerUrl("[" + addresses.addr6 + "]", port, root));
+  }
+  return result;
 }
+
+}
```
veloman-yunkan: I am bringing to your attention that GitHub's diff visualization makes it conspicuous that the last line here doesn't end with a new-line symbol. Let's pay our tribute to all the man-hours invested in implementing that feature by addressing this otherwise invisible formatting issue.

veloman-yunkan (approved):

PR #1241, opened 2025-11-05
Avoid interpretation of content coming from zim by mustache

vighnesh-sawant:
Fixes https://github.com/kiwix/kiwix-desktop/issues/1365

While reproducing please use superuser.com_en_all_2024-10.zim(2024 version not 2025)(present on  web.archive.org)
This crash occurs because mustache is interpreting {{ as a tag which is present in a answer to a question.
mustache does not have a built in thing to stop it from doing that
The solution is to replace {{>snippet}} to {{{snippet}}}.

{{> }} interprets the content and renders it (does no escaping)
https://mustache.github.io/mustache.5.html ( Please look at the example given in the partials section)
{{{ }}} does no escaping and just returns the raw content.

So there should be no inconveniences by changing them!

### DIFF
--- a/static/templates/search_result.html
+++ b/static/templates/search_result.html
@@ -117,7 +117,7 @@
               {{title}}
             </a>
             {{#snippet}}
-              <cite>{{>snippet}}...</cite>
+              <cite>{{{snippet}}}...</cite>
             {{/snippet}}
             {{#bookInfo}}
               <div class="book-title">{{{bookInfo}}}</div>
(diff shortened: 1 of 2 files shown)


vighnesh-sawant:
@kelson42 The character instructs mustache to consider "snippet" as a partial template and interpret it and render it.
I dont understand what a partial in mustache is but in the code it seems to be checking if its a partial then using its partial value or else using the normal str value and then it goes on to interpret it whereas without the character it skips the interpretation part.

PR #1246, opened 2025-11-21
[WIP] Use nuspell for spelling correction

veloman-yunkan:
Fixes openzim/libzim#1014
Fixes https://github.com/openzim/libzim/issues/1012

This is the initial version of using nuspell for spelling correction, which yet has to be tuned.

Note that libnuspell must be available as a dependency.

Xapian-based code for spelling correction is not deleted.

### DIFF
--- a/src/spelling_correction.cpp
+++ b/src/spelling_correction.cpp
@@ -20,6 +20,7 @@
 #include "spelling_correction.h"
 #include "zim/archive.h"
 
+#include <fstream>
 #include <sstream>
 #include <stdexcept>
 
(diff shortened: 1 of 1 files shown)


gremid:
> @gremid I can agree with your opinion if it pertains solely to the DWDS application. I don't know much about it, but as far as I understand it targets learners of German. Mistakes made in short exercises by novices in a language can be quite different from those made by native speakers in longer texts in casual writing. In the latter case misspellings can be mostly typos (a result of fast typing), whereas in the former case mistakes originate from poor command of the language and can vary depending on the native language of the learner. For example, a (literate) native German speaker would hardly have ever spelled _Stuhl_ as _Schtuhl_, but such a misspelling can be expected from a new learner struggling with the phonetics of German digraphs _st_ & _sp_ at the beginning of syllables.

I did not express myself very clearly. My point was that _Stuhl_ vs. _Schtuhl_ has a Levenshtein distance of 2 and should therefore be matched by any suggestion technique covering that distance. It is independent of the origin of the misspelling, i.e., which user cohort is more prone to such misspellings. As documented in

https://github.com/gremid/xapian-spelling-suggestions/blob/main/test_spelling_suggestion.py#L12-L13

for some typical misspellings in our test dataset, covering a Levenshtein distance of 3 is needed for all tests to pass. But even with an acceptable distance of 2 we would cover a substantial subset in a language-independent way. This is why I would not want to tune Nuspell by manually adding rules, but would instead be happy with a technique that generally allows lookups for any match within a defined distance (2 or 3 are up for debate).

PR #1249, opened 2025-11-27
Fix for intermittent /content/blank.html errors

veloman-yunkan:
Fixes kiwix/kiwix-tools#786


benoit74:
Wouldn't this kind of change deserve a small explanation so that we do not fell into the same trap in few months / years?

I absolutely don't get why this fix the issue, and why the issue was only intermittent / on some OS.

benoit74:
My bad ... sorry about that noise

Issue #1193, opened 2025-05-04
Replace aria2c with libtorrent


kelson42:
We want to use more BitTorrent to allow Kiwix users to download and share efficiently ZIM files (and reduce our bandwidth costs).

Traditionally the plan has been to rely on Metalink/Mirrorbrain and `aria2c` to do that. But:
* It has been implemented only on Kiwix-Desktop/libkiwix because we had difficulties with `libaria2c` (`aria2c` binary instrumentation)
* BiTorrent Seeding has not been implemented
* Server side BitTorrent infrastructure (superseeder + Tracker) was not ready

Now we have the server side infrastructure and we clearly need to have a library to handle BitTorrent download/seeding. We could probably use libaria2c but:
* libaria2c is not as easy as libtorrent.
* libaria2c supports metalinks which is ultimatively maybe not something we need anymore because of Webseeds

This issue superseeds https://github.com/kiwix/libkiwix/issues/1172

aviralgarg05:
**Plan:** Replace aria2c with libtorrent for BitTorrent Downloads

Replace the current aria2c binary integration with libtorrent to enable efficient BitTorrent downloads and seeding of ZIM files, improving the download infrastructure while simplifying the library implementation.

**Steps:**

Remove aria2c dependencies - Delete [src/aria2.cpp], [src/aria2.h], update [src/meson.build] to remove aria2.cpp, and remove aria2 runtime dependency from documentation.

Create libtorrent wrapper - Implement new src/libtorrent.cpp and src/libtorrent.h with similar interface to Aria2 class (methods: addUri, tellStatus, tellActive, tellWaiting, pause, unpause, remove, saveSession, shutdown) plus new seeding functionality.

Update Downloader class - Modify [src/downloader.cpp]) and [include/downloader.h] to replace shared_ptr<Aria2> with shared_ptr<LibTorrent>, maintaining the same public API for Download and Downloader classes.

Add seeding support - Extend Download class with seeding-related methods (startSeeding, stopSeeding, getSeedingStatus) and update status tracking to include seeding states.

Update build system - Add libtorrent dependency to [meson.build], update [README.md] dependencies section, and ensure all platform-specific build configurations work.

Update tests - Modify download-related tests to work with libtorrent, add new tests for seeding functionality, and ensure all existing download workflows continue working.

**Open Questions:**

1. Session persistence format - Should we maintain aria2 session file format for backward compatibility, or migrate to libtorrent's format?
[...]

PR #1251, opened 2025-11-29
Add libtorrent dependency and minimal stub integration

### LINKED ISSUE #1238: Compiles against libtorrent
* Introduce hard dependency in the `meson.build`
* Create stub of libtorrent class to ensure we can include/link against
* Create minimal test to secure we can include/link against.

aviralgarg05:
Fixes #1238 (sub-issue of #1193) 

- Add libtorrent-rasterbar as a hard dependency in meson.build
- Create minimal LibTorrent stub class in src/libtorrent.h and src/libtorrent.cpp
- Add getVersion() method to verify libtorrent integration works
- Create test/libtorrent.cpp with tests for instantiation, version retrieval, and linking
- Add boost include path for libtorrent header dependencies
- Update meson.build to check for mustache header in source tree
- Add mustache.hpp header to include directory

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -63,13 +63,10 @@ zlib_dep = dependency('zlib', static:static_deps)
 xapian_dep = dependency('xapian-core', static:static_deps)
 libtorrent_dep = dependency('libtorrent-rasterbar', static:static_deps)
 
-
 if compiler.has_header('mustache.hpp')
   extra_include = []
 elif compiler.has_header('mustache.hpp', args: '-I/usr/include/kainjow')
   extra_include = ['/usr/include/kainjow']
-elif compiler.has_header('mustache.hpp', args: '-I/opt/homebrew/include')
-  extra_include = []
 else
   error('Cannot found header mustache.hpp')
 endif
@@ -103,10 +100,7 @@ all_deps = [thread_dep, libzim_dep, pugixml_dep, libcurl_dep, microhttpd_dep, zl
 # Dependencies as array
 all_deps += libicu_deps
 
-# Add boost include path for libtorrent
-boost_include = include_directories('/opt/homebrew/include')
-
-inc = include_directories('include', extra_include, '/opt/homebrew/include')
+inc = include_directories('include', extra_include)
 
 conf = configuration_data()
 conf.set('LIBKIWIX_VERSION', '"@0@"'.format(meson.project_version()))
--- a/src/libtorrent.cpp
+++ b/src/libtorrent.cpp
@@ -1,4 +1,21 @@
-
+/*
+ * Copyright 2025 Kiwix developers
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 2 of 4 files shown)


aviralgarg05:
> > > @aviralgarg05 Thank you for your first PR. My first remark, and this is about something very important, we have to rely on official libtorrent (and not on the rasterbar fork). Any very strong reason to go for this minor fork?
> > 
> > 
> > I think libtorrent-rasterbar IS the official libtorrent library from github.com/arvidn/libtorrent. The "rasterbar" name comes from the project's origins. This is the correct pkg-config name used on all platforms.
> 
> I´m not an expert of this library either but:
> 
> * If you look at https://github.com/airium/libtorrent-rasterbar, you see this is a fork of https://github.com/arvidn/libtorrent
> * https://github.com/arvidn/libtorrent has far more stars/contributions than `libtorrent-rasterbar`



Thank you for pointing this out! I understand the confusion now. The libtorrent-rasterbar package name does indeed refer to the official libtorrent library from [https://github.com/arvidn/libtorrent].

The "rasterbar" part comes from Rasterbar Software, which is Arvid Norberg's company that develops and maintains libtorrent. Looking at the repositories:

[https://github.com/arvidn/libtorrent]: The official repository with 5.8k stars, 1.1k forks, 137 contributors, and active development (latest release v2.0.11 on Jan 28)
[https://github.com/airium/libtorrent-rasterbar]: A minor personal fork with only 13 stars, 1 fork, and no releases published
The pkg-config name libtorrent-rasterbar is the standard name used across Linux distributions and package managers for the official library. This is confirmed by:

The official libtorrent website [https://libtorrent.org/] which links to the arvidn/libtorrent repository
Package listings on Repology [https://repology.org/project/libtorrent-rasterbar/versions] showing 206 packages across major distributions (Alpine, Arch, Debian, Fedora, Ubuntu, etc.) all using libtorrent-rasterbar for the official library
Our successful meson build that finds and links against libtorrent-rasterbar version 2.0.11

**So we are definitely using the official libtorrent library, not a fork. The naming convention just reflects the company behind the project.** 

The airium fork appears to be a personal project with minimal activity and no official status.

kelson42:
@veloman-yunkan After a first looks it looks good to me, at least it compiles and I don't see any obvious strange thing. Can you please have a look on your side?

We need as well to:
* Implement https://github.com/kiwix/kiwix-build/issues/890 to get the CI passing
* Decide if we want to merge this BitTorrent related thing to a dedicated branch or directly on `main`.

meson.build
```diff
@@ -94,12 +97,15 @@ endif
...
 # Dependencies as string
-all_deps = [thread_dep, libzim_dep, pugixml_dep, libcurl_dep, microhttpd_dep, zlib_dep, xapian_dep]
+all_deps = [thread_dep, libzim_dep, pugixml_dep, libcurl_dep, microhttpd_dep, zlib_dep, xapian_dep, libtorrent_dep]
 
 # Dependencies as array
 all_deps += libicu_deps
 
-inc = include_directories('include', extra_include)
+# Add boost include path for libtorrent
+boost_include = include_directories('/opt/homebrew/include')
```
veloman-yunkan, line 106: So we must add `boost` as a dependency. This is not the correct way of doing it (`libkiwix` has to compile in various environments, not just on MacOS with `boost` installed via `homebrew`).

meson.build
```diff
@@ -94,12 +97,15 @@ endif
...
+
+inc = include_directories('include', extra_include, '/opt/homebrew/include')
```
veloman-yunkan, line 108: Same here

include/mustache.hpp
veloman-yunkan, line 1: We already have `mustache.hpp` in our dependencies (Update: I see that the added file was removed in the next commit, but in a clean version of the PR it shouldn't have been added at all).

meson.build
veloman-yunkan: I see that this was reversed in the last commit.

meson.build
veloman-yunkan: I see that this was reversed in the last commit.

PR #1254, opened 2025-12-05
[Hack] Multiple spelling corrections are returned

veloman-yunkan:
This is likely a throwaway PR intended to enable the customer to test a solution to openzim/libzim#1012 that may be implemented in a cleaner way at more development cost.

### DIFF
--- a/include/spelling_correction.h
+++ b/include/spelling_correction.h
@@ -32,7 +32,7 @@ class Archive;
 
 namespace Xapian
 {
-class Database;
+class WritableDatabase;
 }
 
 namespace kiwix
@@ -50,7 +50,7 @@ public: // functions
   std::vector<std::string> getSpellingCorrections(const std::string& word, uint32_t maxCount) const;
 
 private: // data
-  std::unique_ptr<Xapian::Database> impl_;
+  std::unique_ptr<Xapian::WritableDatabase> impl_;
 };
 
 } // namespace kiwix
--- a/src/spelling_correction.cpp
+++ b/src/spelling_correction.cpp
@@ -43,15 +43,11 @@ std::vector<std::string> getAllTitles(const zim::Archive& a)
 void createXapianDB(std::string path, const zim::Archive& archive)
 {
   const int flags = Xapian::DB_BACKEND_GLASS|Xapian::DB_CREATE;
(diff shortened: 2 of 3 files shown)


kelson42:
I let @gremid give his feedback, but it seems to me that the results we get here, even by having only the results with the shortest distance, @re not the one expected.

veloman-yunkan:
> @veloman-yunkan Please can you put a copy of your `dwds_zim_titles.cleaned_up.spellingsdb` somewhere I can download it to poke at.

@kelson42 @gremid Can I publicly share the word list of the DWDS dictionary?

kelson42:
> > @veloman-yunkan Please can you put a copy of your `dwds_zim_titles.cleaned_up.spellingsdb` somewhere I can download it to poke at.
> 
> 
> 
> @kelson42 @gremid Can I publicly share the word list of the DWDS dictionary?

@veloman-yunkan I think this is OK. Go ahead.

ojwb:
This much shorter list is enough to trigger it:

```
Schiedsrichterstuhl
Stuhl
```

There's an n-gram filter which identifies candidate words to calculate the edit distance for.  What's happening is that `Schiedsrichterstuhl` matches 7 n-grams from `Schtuhl` (we have trigrams plus head, tail and bookend bigrams, and it matches `^Sc`, `Sch`, `cht`, `tuh`, `uhl`, `hl$` and `^S.*l$`) whereas `Stuhl` matches 4 (`tuh`, `uhl`, `hl$` and `^S.*l$`) and we consider `Schiedsrichterstuhl` first and record that we've seen a score of 7.  New candidates are required to have an n-gram score within 2 of the best, but that seems a flawed optimisation as a candidate which is actually much too long can get a high n-gram score.

I'm unsure of the logic behind this threshold - I don't see it documented in the code, but maybe it's in an old ticket somewhere.  Making the threshold 3 here fixes this case, but I think I could construct examples which fail with arbitrary thresholds so it seems this may just be flawed.  We may be able to just drop the check without too much impact - the edit distance calculation has been optimised and is now much faster than it used to be (I've done some experiments doing spelling correction without storing any special spelling data and just considering unstemmed terms as correctly spelled words and even that seems fast enough).

I'll investigate further and decide what the best way to address it is.

Thanks again for providing the data to reproduce.

PR #1261, opened 2025-12-27
Fix core stability, correctness, and memory issues

shbhmexe:
This PR addresses several critical stability and correctness issues in libkiwix. Due to reviewer feedback regarding ABI compatibility, the initial proposal to change the return type of `getBookById` was introduced and subsequently reverted to preserve the library's public API.

### Changes Merged:

1.  **Critical Bug Fix (Windows)**:
    - Fixed an infinite loop and memory leak in `getNetworkInterfacesWin`. The iteration counter is now correctly incremented and memory is freed before retrying on buffer overflow.
2.  **Exception Safety (String Tools)**:
    - Updated `removeAccents` in `stringTools.cpp` to use RAII (`std::unique_ptr`) for ICU resource management, ensuring objects are deleted even if an exception occurs.
3.  **Exception Correctness (Downloader)**:
    - Fixed exception slicing in `Downloader::getDownload`. The code now uses `throw;` to preserve the original exception type when re-throwing.
4.  **Code Quality (Server)**:
    - Replaced the magic number `416` with the official `MHD_HTTP_RANGE_NOT_SATISFIABLE` constant in `response.cpp`.

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -340,9 +340,9 @@ class Library: public std::enable_shared_from_this<Library>
   std::string getBestTargetBookId(const std::string& bookName, const std::string& preferedFlavour="", const std::string& minDate="") const;
 
   // XXX: This is a non-thread-safe operation
-  Book getBookById(const std::string& id) const;
+  const Book& getBookById(const std::string& id) const;
   // XXX: This is a non-thread-safe operation
-  Book getBookByPath(const std::string& path) const;
+  const Book& getBookByPath(const std::string& path) const;
 
   Book getBookByIdThreadSafe(const std::string& id) const;
 
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -203,8 +203,8 @@ std::string Library::getBestFromBookCollection(BookIdCollection books, const Boo
 
   sort(books, DATE, false);
   std::stable_sort(books.begin(), books.end(), [&](const std::string& bookId1, const std::string& bookId2) {
-    const auto book1 = getBookById(bookId1);
-    const auto book2 = getBookById(bookId2);
+    const auto& book1 = getBookById(bookId1);
+    const auto& book2 = getBookById(bookId2);
     bool same_flavour1 = book1.getFlavour() == bookmark.getBookFlavour();
     bool same_flavour2 = book2.getFlavour() == bookmark.getBookFlavour();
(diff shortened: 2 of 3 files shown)


shbhmexe:
> @shbhmexe Thank you for your various improvement proposals. When you are ready, we will assign a technical reviewer for your code. But before we come to that, please split your commit in smaller pieces with one commit per "logical" change. That way we can clearly understand and follow the changes, one by one.

@kelson42 Thank you for the feedback. I am ready to proceed. I will split the changes into separate logical commits as requested to make the review process easier. I'll adhere to this practice for future contributions as well.

shbhmexe:
These are bug fixes for existing code paths. They don't introduce new testable logic, so adding tests isn't applicable here. Could you override the codecov check?

include/library.h
```diff
@@ -340,9 +340,9 @@ class Library: public std::enable_shared_from_this<Library>
   std::string getBestTargetBookId(const std::string& bookName, const std::string& preferedFlavour="", const std::string& minDate="") const;
 
   // XXX: This is a non-thread-safe operation
-  const Book& getBookById(const std::string& id) const;
+  Book getBookById(const std::string& id) const;
   // XXX: This is a non-thread-safe operation
-  const Book& getBookByPath(const std::string& path) const;
+  Book getBookByPath(const std::string& path) const;
 
   Book getBookByIdThreadSafe(const std::string& id) const;
```
veloman-yunkan, line 347: As you can see `getBookById()` is stated to be a non-thread-safe operation, for which we have a thread-safe counterpart `getBookByIdThreadSafe()`. I guess I avoided changing the return type of `getBookById()` out of the concern for the backward compatibility of the ABI of `libkiwix`. Let's not include this change (and other changes triggered by it) in this PR. It deserves a ticket of its own where we should make a decision of how to proceed with that potential problem.

kelson42:
@shbhmexe Any news?

shbhmexe:
> @shbhmexe Any news?

@veloman-yunkan Thank you for the review. I understand the concern regarding ABI compatibility with getBookById and getBookByPath. I have reverted those changes in this PR as requested. I agree that the thread-safety issue warrants a separate discussion/ticket. This PR now only contains the other stability and correctness fixes.

PR #1264, opened 2026-01-05
Kiwix Server OPDS feed links should not be paginated (with all results)

kunalpatil-18:
Fixes kiwix/kiwix-tools#795.

This PR appends `&count=-1` to the unpaginated querystring in search_renderer.cpp. This ensures that Atom feeds return all items instead of the default page limit.

### DIFF
--- a/src/search_renderer.cpp
+++ b/src/search_renderer.cpp
@@ -121,9 +121,8 @@ kainjow::mustache::data buildQueryData
   query.set("pattern", kiwix::encodeDiples(pattern));
   std::ostringstream ss;
   ss << searchProtocolPrefix << "?pattern=" << urlEncode(pattern);
-  ss << "&" << bookQuery;
- // fixed the Issue #795: Force feed to return all items (no pagination)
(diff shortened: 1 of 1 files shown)


kelson42 (changes requested):
@kunalpatil-18 Please remove the comment (code is self-explaining) and commit message is there to explain. Better do the modification directly line 124.

kunalpatil-18:
@veloman-yunkan Thanks for the feedback.

I am currently developing on Windows and facing significant issues setting up the full build environment (Meson/Ninja dependencies), so I haven't been able to run `kiwix-serve` locally to verify the runtime behavior yet.

Regarding the fix:
1. I see the CI failed — I suspect I broke a test case in `test/` that expects the old URL format. I can fix that.
2. If `count=-1` doesn't force "all results" as I assumed, is there a standard parameter in `server.cpp` that triggers unpaginated results (e.g. `count=0`), or does this require a deeper change in the request handler itself?

Any guidance would be appreciated while I try to sort out my local build!

PR #1274, opened 2026-02-22
No pagination for Atom feed URL

### LINKED ISSUE #795: Kiwix Server feed (URL) does only deliver 10 items
I guess has been forgotten when we have introduced pagination. We do return only 10 results if not specified differently. This is actually correct, but if someone wants to subscribe to a feed, its feed reader won't deal with the pagination. Therefore we need to return all the results.

Solution seems trivial, we just need to add `count=-1` to the querystring and we will be good. This should be done for all places where the feed URL is given. At least:
* In the HTML header
* Behind the feed icon (top-right corner).

kelson42:
Fixes kiwix/kiwix-tools#795

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -36,7 +36,7 @@
                 filteredParams.set(key, value);
             }
         }
-        const feedLink = `${root}/catalog/v2/entries?${filteredParams.toString()}`;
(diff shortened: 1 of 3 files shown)


veloman-yunkan (approved):
Please merge the PR as a single commit.

Issue #1275, opened 2026-02-24
Apple OPDSParse benchmark, icu73 is the slowest part


BPerlakiH:
I did some benchmark on parsing the library, because it felt slow to me.

On average it's **about 5 seconds**, just the parsing bit of the current full catalog.

It seems that the "removeAccents()" call which is **creating icu73** instance **is the slowest part** taking up **2.9 sec**!

<img width="1217" height="496" alt="Image" src="https://github.com/user-attachments/assets/504a65f2-1208-4e08-9ab7-a798736572c8" />

I think (by accident) a related discussion to this was in: https://github.com/kiwix/libkiwix/issues/1101

This time my question is about the performance itself. Is there a way to improve on this ?

kelson42:
Thank you @BPerlakiH for reporting this!

@veloman-yunkan Do you confirm the measure? If confirmed, it sounds to me we could very easily optimise that part of code and get massive speed improvement for all our readers, right? We should try this for the next minor release `14.2.0` IMHO as the ration effort/improvement would be very very good!

veloman-yunkan:
@kelson42 I confirm that `removeAccents()` is a bottleneck in library loading and a large low hanging fruit. I will harvest it.

kelson42:
Fixed by #1276

PR #1276, opened 2026-02-27
Optimized removeAccents() for ASCII-only input

veloman-yunkan:
Quick fix for #1275

Since most of the strings passed to `removeAccents()` while loading the current Kiwix online library and populating the in-memory Xapian database of book metadata are ASCII-only, it makes sense to bypass expensive ICU transliteration on such input.

The optimization results in ~3X speedup (down from ~9.8 seconds to ~3.3 seconds) of starting `kiwix-serve` in catalog-only mode using https://download.kiwix.org/library/library_zim.xml (which as of this writing contains 3444 entries and weighs 19MiB). Such testing doesn't exactly reproduce the flow reported in the mentioned ticket where the input data for library initialization comes from an OPDS feed but the effect should be the same.

### DIFF
--- a/src/tools/stringTools.cpp
+++ b/src/tools/stringTools.cpp
@@ -35,6 +35,28 @@
 #include <iomanip>
 #include <regex>
 
+namespace
+{
+
+bool isAsciiOnly(const std::string& s)
+{
+  for ( const unsigned char c : s ) {
+    if ( c >= 128 )
+      return false;
+  }
+  return true;
+}
+
+std::string asciiToLower(std::string s)
+{
+    std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) {
+        return ('A' <= c && c <= 'Z') ? c - ('Z' - 'z') : c;
+        });
+    return s;
+}
+
+} // unnamed namespace
+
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan That was quick! Do you still foresee a PR to optimise the non-ascii part? For example by instanciating only once ICU transformer?

veloman-yunkan:
> @veloman-yunkan That was quick! Do you still foresee a PR to optimise the non-ascii part? For example by instanciating only once ICU transformer?

@kelson42 Not at this point. IMHO, the effort is not worth the effect.

veloman-yunkan:
BTW, the unit test of `removeAccents()` has revealed some facts about its behavior that ran counter my intuition. For example, it doesn't convert the polish letter [Ł](https://en.wikipedia.org/wiki/%C5%81) to **L**, doesn't decompose Æ to AE, Œ to OE, ß to ss, etc.

I think we should explicitly define all details of text normalization for indexing purposes.

veloman-yunkan:
Also, note that `removeAccents()` is duplicated between `libkiwix` and `libzim`. In the latter it has undergone an optimization of its own.

PR #1277, opened 2026-02-28
Update docs

### LINKED ISSUE #1210: Library header documentation to be fixed
While including the library.h in XCode, it warns that the following doc is not inline with the code:
https://github.com/kiwix/libkiwix/blob/699cbccf3864946b3b323243350483724cab13f0/include/library.h#L446-L453

Image

BPerlakiH:
Fixes: #1210

### DIFF
--- a/include/library.h
+++ b/include/library.h
@@ -444,11 +444,11 @@ class Library: public std::enable_shared_from_this<Library>
 
 
   /**
-   * Sort (in place) bookIds using the given comparator.
+   * Sort (in place) bookIds using the given arguments.
    *
    * @param bookIds the list of book Ids to sort
-   * @param comparator how to sort the books
-   * @return The sorted list of books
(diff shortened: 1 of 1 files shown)


include/library.h
```diff
@@ -447,8 +447,8 @@ class Library: public std::enable_shared_from_this<Library>
    * Sort (in place) bookIds using the given comparator.
    *
    * @param bookIds the list of book Ids to sort
-   * @param comparator how to sort the books
-   * @return The sorted list of books
+   * @param sortBy how to sort the books (UNSORTED, TITLE, SIZE, DATE, CREATOR, PUBLISHER)
```
veloman-yunkan, line 450: It would be nice to reformulate the description of the function too (reference to `comparator` no longer looks relevant there).

BPerlakiH: Please see it updated.

veloman-yunkan (approved):

PR #1169, opened 2024-12-24
Wait up to 1s to le aria2c to start

kelson42:
Fixes https://github.com/kiwix/kiwix-desktop/issues/1291

PR #1282, opened 2026-03-10
aria2: increase curl timeout

rmelotte:
The starting of aria2c is rather fragile as it depends on quite short fixed timeouts, and because kiwix-desktop cannot recover when it fails.

When this happens, it leads to "Cannot connect to aria2c rpc" when launching kiwix-desktop.

The waiting time has already been increased in the past to try to make it more robust:
https://github.com/kiwix/libkiwix/pull/1169

However, curl also has its own internal timeout that is currently shorter than MAX_WAITING_TIME_SECONDS. As a result when aria2 is a bit slow to start (on older systems for example), curl currently times out after 100ms, but we wait for 1 second regardless until we report the issue. The resulting curl error in that case is: "Connection timed out after 100 milliseconds"

As a quick fix to avoid having to change too much of the implementation, just increase curl's timeout to match MAX_WAITING_TIME_SECONDS.

### DIFF
--- a/src/aria2.cpp
+++ b/src/aria2.cpp
@@ -120,7 +120,7 @@ Aria2::Aria2(std::string sessionFileDir):
   curl_easy_setopt(p_curl, CURLOPT_PORT, m_port);
   curl_easy_setopt(p_curl, CURLOPT_POST, 1L);
   curl_easy_setopt(p_curl, CURLOPT_ERRORBUFFER, curlErrorBuffer);
-  curl_easy_setopt(p_curl, CURLOPT_TIMEOUT_MS, 100);
+  curl_easy_setopt(p_curl, CURLOPT_TIMEOUT_MS, 1000);
 
   typedef std::chrono::duration<double> Seconds;
 


kelson42:
I would strongly recommend to use one unique const for all places where the timeout should be set. Otherwise, next time we have to deal with that, we will forget to change half of the things.

veloman-yunkan (commented):
I think your understanding of the code is wrong. Curl requests (with a short timeout) for checking whether aria2c is running are repeated in a loop until the longer timeout is exceeded. The short timeout doesn't hurt as long as aria2c isn't slow to an extent that it, once up and running, fails to process the request within that duration.

rmelotte:
Thanks a lot for the review!

Looking at it now, I've no clue how I arrived at the conclusion that curl ran only once, but it's clearly completely wrong...

But then since the current change fixes the issue in my case (I'm running it on a 4-cores x86 Desktop from 2012 so it's old, but not _that_ slow either), it must mean even the last call times out after 100ms.

Would increasing both timeouts (for example the curl timeout to 1000ms and the global timeout to 5 seconds) be acceptable to you, to accommodate slower systems?

veloman-yunkan:
@rmelotte Your change can be equivalent (under certain conditions) to increasing the global timeout to 2seconds.

I am not in favor of discussing optimal values for hardcoded parameters. I'd rather make this setting user-controllable. @kelson42 What do you think?

kelson42:
Do we have a valid scenario for which current timeout is not enough? If "yes", what would be the right default values to fix it? What would be the other negative impact of such a value?

To my understanding, I see no real reason to not increase the timeout... if needed.

PR #1285, opened 2026-03-20
Add TOC feature to Kiwix Server

### LINKED ISSUE #754: Kiwix Server should provide a TOC
Table of content is the responsability of the reader (to oppose to the ZIM). Kiwix server should have a Chrome proposing one.

AzoteGwei:
Fixes [kiwix/kiwix-tools#754](https://github.com/kiwix/kiwix-tools/issues/754)

image

This PR implements Table of Contents for Reader.

For widely enough desktops or iPads, it will use dock just like wikipedia does. For smaller ones, it will use floating TOC. For phones, it will be a full-screen TOC.

For mini ZIMs, like `wikipedia_zh_physics_mini_2026-03.zim`, it will disable TOC because the content only have a h1.

Tested on Firefox & Chromium.

More i18n keys need to be added.

### DIFF
--- a/static/skin/i18n/en.json
+++ b/static/skin/i18n/en.json
@@ -95,4 +95,5 @@
 	, "book-category.wiktionary": "Wiktionary"
 	, "book-category.other": "Other"
 	, "text-loading-content": "Loading Content"
+	, "table-of-content-button-text": "View table of contents"
 }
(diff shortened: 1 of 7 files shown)


AzoteGwei:
I would sincerely apologize for my mistake by not adding new `cacheid` into the tests.

```
Ok:                21
Fail:              0
```

Now it should be able to pass all the tests.

image

GitHub Action ran on my own repo, passed test and packaging.

kelson42 (changes requested):
@AzoteGwei Thank you very much for your PR. I did a first pass and I realy like the over look and feel!

That been said, a few things should be fixed.

* If there is no TOC, then the button is disabled, which is good. But there is some kind of a grey visual artefact on the left of the button, see:
image

* We need to find a Unicode character symbolizing the list to replace the "TOC" string

* Might be good to have a keyboard shorcut to toggle the sidebar, would you have one to propose?

* Question: is the width of the sidebar fix? What will happen when the section titles are longer than the width of the sidebar?

Issue #915, opened 2023-03-15
Kiwix OPDS Feed Title


rgaudin:
Kiwix OPDS feed's title is `All Entries` which makes sense in the context of the `entries` endpoint within the API but when used as a Feed source (as advertise in the kiwix-serve viewer), it doesn't.

An `All Entries` subscription is not a good name.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<feed xmlns="http://www.w3.org/2005/Atom"
      xmlns:dc="http://purl.org/dc/terms/"
      xmlns:opds="https://specs.opds.io/opds-1.2"
      xmlns:opensearch="http://a9.com/-/spec/opensearch/1.1/">
  <id>8cad36a0-64e2-7996-0472-00c09bc7ec79</id>

  <link rel="self"
        href="/catalog/v2/entries"
        type="application/atom+xml;profile=opds-catalog;kind=acquisition"/>
  <link rel="start"
        href="/catalog/v2/root.xml"
        type="application/atom+xml;profile=opds-catalog;kind=navigation"/>
  <link rel="up"
        href="/catalog/v2/root.xml"
        type="application/atom+xml;profile=opds-catalog;kind=navigation"/>

  <title>All Entries</title>
  <updated>2023-03-15T11:56:05Z</updated>
```

Ii might be difficult to solve, especially given a meaningful name might not be generic, so maybe we assume that all syndication readers allow renaming and users would go through that extra step.

kelson42:
Thinking loud. The problem of kiwix-serve instance name/title is broader than just here. Welcome page title is a topic as well. I was thinking tha the problem shoykd be solved by a custom welcom page... but I wonder if we would still benefit of a --instanceTitle option?

rgaudin:
It could also be part of some library-wide metadata because in this context, the library is the instance.

I can't recall the content of the discussion we had around categories/tags in Paris but I have the feeling it was somewhat connected.

Are there already other *static* information you can think of that you'd see user wanting to customize?

kelson42:
@rgaudin Your last remark makes sense if we consider at some point to replace the custom library xml by a proper OPDS feed.

kelson42:
@rgaudin At this stage, my opinion on this issue has slightly evolved:
* A better name should be found like "Kiwix Server library entries"
* We could consider that customisation should be done via reverse proxy if wanted (this is machine readable content)

rgaudin:
> A better name should be found like "Kiwix Server library entries"

[...]

PR #1287, opened 2026-03-24
Set names in site.webmanifest for PWA

### LINKED ISSUE #799: Web app title

Roan-V:
Resolves https://github.com/kiwix/kiwix-tools/issues/799

### DIFF
--- a/static/skin/favicon/site.webmanifest
+++ b/static/skin/favicon/site.webmanifest
@@ -1,6 +1,6 @@
 {
-    "name": "",
-    "short_name": "",
+    "name": "Kiwix Server",
+    "short_name": "Kiwix",
     "icons": [
         {
             "src": "android-chrome-192x192.png",


rgaudin:
I don't think this is the right move. kiwix-serve is a tool to serve a content catalog. That catalog should have a name and that should not be “Kiwix” which is already overused and confusing. Add to that the fact that we have https://library.kiwix.org which is _The Kiwix Catalog_…

This should follow #915 IMO.

Roan-V:
Your point is certainly fair, but #915 looks pretty stale for now
I'd argue that (at least as a stopgap) "Kiwix" is preferable over letting the browser picking a default name. Firefox turns it into "Web site" for example

When #915 gets more progress this could quite easily be changed again
Don't let perfect be the enemy of ~~good~~ decent and all that :P

rgaudin:
I maintain that Kiwix is a terrible value. If we want to improve without fixing we should make it obvious that it's a generic name from a tool because those metadata are supposed to reference a content source.

"kiwix-serve instance" would work.

Given we recommend kiwix-serve to be behind a reverse proxy for public access cases, overriding it there would be a possibility

PR #1288, opened 2026-03-24
Fix white line at bottom of iframe kiwix/kiwix-tools#809

### LINKED ISSUE #809: White line at bottom of content iframe due to fixed height
Hi!  I think it is so cool that I can have a full (!!) copy of Wikipedia locally.

Something I am noticing when viewing a kiwix-serve instance of Wikipedia on macOS Safari is that the content iframe is not actually filling out the full viewport height.  This is evident with the dark theme, where there is a several pixel tall white line at the bottom of the page.

Image

If I set:
- `html` to `height: 100%`
- and then set `html`, `body`, and the `iframe` to
  - `display: flex`
  - `flex-direction: column`
  - `flex: 1`

 it works correctly with the content laid out as expected.

If I set the `html`, `body`, and `iframe` elements to `height: 100%`, at first it seems like it works, but after scrolling to the bottom of the content, the iframe is taller than the body as much as the height of the toolbar (which could be calculated out, I suppose), and the same line shows up at the bottom.  Removing the header element does not fix this.  Scroll to the bottom once, and then again (to catch the outer scroll), and you see the line.

kiwix-tools macos-arm64-3.8.2

Safari Version 26.3.1 (21623.2.7.11.7)
macOS Tahoe 26.3.1
MacBook Pro 14", Apple M3 Pro

Thank you so much for this very cool project!

Edit: related PR for libkiwix at https://github.com/kiwix/libkiwix/pull/1288

zeyad-elkholy:
Fixes kiwix/kiwix-tools#809

I removed the hardcoded Javascript math that was calculating the iframe height with an arbitrary - 4 pixel offset. I replaced it with the CSS Flexbox approach (display: flex, flex: 1) on the html, body, and iframe elements to allow it to scale naturally without leaving a gap at the bottom.

### DIFF
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -228,12 +228,6 @@ function updateToolbarVisibilityState() {
   previousScrollTop = st;
 }
 
-function handle_visual_viewport_change() {
- // const wh = window.visualViewport
-   //        ? window.visualViewport.height
-     //      : window.innerHeight;
-  //  contentIframe.height = wh - contentIframe.offsetTop - 4;
-}
 
 function setIframeUrl(path) {
   try {
@@ -583,9 +577,7 @@ function setupViewer() {
   // Defer the call of handle_visual_viewport_change() until after the
   // presence or absence of the taskbar as determined by this function
   // has been settled.
-  setTimeout(handle_visual_viewport_change, 0);
 
-  window.onresize = handle_visual_viewport_change;
 
   const kiwixToolBarWrapper = document.getElementById('kiwixtoolbarwrapper');
   if ( ! viewerSettings.toolbarEnabled ) {
--- a/static/skin/viewer.js
+++ b/static/skin/viewer.js
@@ -229,10 +229,10 @@ function updateToolbarVisibilityState() {
 }
 
 function handle_visual_viewport_change() {
-  const wh = window.visualViewport
-           ? window.visualViewport.height
-           : window.innerHeight;
-  contentIframe.height = wh - contentIframe.offsetTop - 4;
+ // const wh = window.visualViewport
+   //        ? window.visualViewport.height
+     //      : window.innerHeight;
+  //  contentIframe.height = wh - contentIframe.offsetTop - 4;
 }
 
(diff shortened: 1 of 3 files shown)


zeyad-elkholy:
Fixes kiwix/kiwix-tools#809

david-cako:
🎉 sweet!!

I hope this works on all dependents of libkiwix, but it should.  Looks like iOS and Android are webviews.  Still could be scaffolding that gets changed with this, and I'm sure someone can test it!

zeyad-elkholy:
i don't know to be honest but i don't see smth wrong in web, so check it on safari, it works in firefox

Yizhen-Zheng:
I've verified this works perfectly on Safari. 
Before:(notice the white line at the bottom
Screenshot 2026-03-25 at 12 01 01
After: 
Screenshot 2026-03-25 at 12 01 11

kelson42:
@zeyad-elkholy Thank you very much for your PR. I have assigned reviewers and will test things myself. Give us a few days to complete the review.

veloman-yunkan (changes requested):
I didn't test the fix (leaving it to @kelson42). If testing doesn't reveal any problems and the fix is going to be merged then we should simply drop the `handle_visual_viewport_change()` function (as suggested by Copilot).

Also, the cacheids in the server unit-test must be updated.

zeyad-elkholy:
@kelson42 @veloman-yunkan i removed `handle_visual_viewport_change()` and his calls and also updated the cache ids and SHA1 hash

veloman-yunkan:
@kelson42 This PR should be merged as a single (squashed) commit.

zeyad-elkholy:
@kelson42 could you review or merge the PR plz if u have time?

kelson42:
@zeyad-elkholy Please:
* Never merge within a feature branch. Use rebase instead
* Squash all your commits like requested by @veloman-yunkan

kelson42:
@zeyad-elkholy For next time, I made myself. Superseeded by #1295

PR #1289, opened 2026-03-26
Align download links order across JS modal and no-JS page

### LINKED ISSUE #775: Improve a bit kiwix-serve download widget UI

zeyad-elkholy:
image
image

Fixes kiwix/kiwix-tools#775

### DIFF
--- a/static/templates/no_js_download.html
+++ b/static/templates/no_js_download.html
@@ -7,6 +7,15 @@
     <title>{{translations.download-links-title}}</title>
 </head>
 <style>
+    .downloadRow {
+        display: flex;
+        align-items: center;
+        gap: 10px;
+    }
+    .downloadRow img {
+        width: 20px;
+        height: 20px;
+    }
     .downloadLinksTitle {
         text-align: center;
         font-size: 32px;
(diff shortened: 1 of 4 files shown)


zeyad-elkholy:
@rgaudin @kelson42 can someone review the PR if he have time?

kelson42 (changes requested):
@zeyad-elkholy Thank you for your PR.

Here are the requested changes:
* Magnet links should be visible in the PR description screenshot
* When the issue request a small vertical space or separator for the Hash, that means the logo should be also properly stayed aligned with the text. The idea is that we list first 3 ways to download the ZIM and then you have the file signature/hash (not a way to download the file, therefore the small gap with the 3 other solutions )
* cacheids should be updated
* unit-test should be run
* squash all commit in one commits
* rebase on `main` if necessary

If you are quick to fix, this will be merged in the next release to be made soon.

static/templates/no_js_download.html
```diff
@@ -12,6 +12,9 @@
         font-size: 32px;
         margin-bottom: 8px;
     }
+    .checksum-spacing {
+        margin-top: 8px; 
+    }
```
veloman-yunkan: This change doesn't make sense in this commit

PR #1293, opened 2026-04-30
"Fixed" the server unit-test for libmicrohttpd 1.x

### LINKED ISSUE #1291: [DEBIAN] libkiwix is marked for autoremoval from testing
```
libkiwix 14.0.0-1.1 is marked for autoremoval from testing on 2026-06-01

It is affected by these RC bugs:
1134293: libkiwix: FTBFS: 18/20 libkiwix:server           FAIL
 https://bugs.debian.org/1134293



For more information on the autoremoval process, including hints to prevent
autoremoval can be found on the wiki: https://wiki.debian.org/Autoremoval

This mail is generated by:
https://salsa.debian.org/release-team/release-tools/-/blob/master/mailer/mail_autoremovals.pl

Autoremoval data is generated by:
https://salsa.debian.org/qa/udd/-/blob/master/udd/testing_autoremovals_gatherer.pl
```

veloman-yunkan:
Part of the fix for #1291

libmicrohttpd 1.0.5 (the version used at this point under the debian testing distribution) seems to return an HTTP 400 (Bad Request) error if the URL contains an invalid URI-encoding sequence, thus breaking our unit-tests involving such URLs. Upgrading our libmicrohttpd dependency is not easy, so this change is an optimistic attempt to fix the build of the libkiwix package under debian testing.

For this change to fix #1291 the debian package of `libkiwix` must be upgraded to the most recent version or this patch must be applied to the 14.0 release.

--- 
**Update (2026-05-09):** This PR was followed-up by

- The first commit of #1294
- #1298

### DIFF
--- a/test/server.cpp
+++ b/test/server.cpp
@@ -439,6 +439,10 @@ TEST_F(ServerTest, CacheIdsOfStaticResourcesMatchTheSha1HashOfResourceContent)
 }
 
 const char* urls400[] = {
+#if MHD_VERSION >= 0x01000000
+  "/ROOT%23%",
+  "/ROOT%23%3",
+#endif // MHD_VERSION
   "/ROOT%23%3F/search",
   "/ROOT%23%3F/search?content=non-existing-book&pattern=asdfqwerty",
   "/ROOT%23%3F/search?content=non-existing-book&pattern=asd<qwerty",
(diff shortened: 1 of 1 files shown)


kelson42 (approved):
Thank you. LGTM

veloman-yunkan:
It looks like this patch works not like I intended it to. `MHD_VERSION` seems to remain undefined in `test/server.cpp` (because `microhttpd.h` is not `#include`d in that file) and the compiler doesn't complain about it but silently excludes both sections guarded by the compile-time condition upon its value.

kelson42:
@veloman-yunkan I guess we need a new PR then!?

veloman-yunkan:
> @veloman-yunkan I guess we need a new PR then!?

Done as a side-fix in #1294

PR #1297, opened 2026-05-09
Libkiwix 14.2.1

kelson42:
Related to #1296

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+libkiwix 14.2.1
+===============
+
+ * Documentation fix (@bperlakih #1277)
+ * Fix kiwix-serve unit-test for libmicrohttpd 1.x (@veloman-yunkan #1293)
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan Kind of look like we have built a regression here with #1294 ... or are we finally impacted by #1266?

veloman-yunkan:
> @veloman-yunkan Kind of look like we have built a regression here with #1294 ... or are we finally impacted by #1266?

This is due to the side-change in #1294 attempting to correct the noticed problem with #1293. But the latter PR turned out to contain two defects - the noticed one masking another one (described [here](https://github.com/kiwix/kiwix-desktop/issues/1488#issuecomment-4406645389)), and when the said mistake was fixed the other defect kicked in, resulting in failed CI. I guess we should pay more attention to CI when merging PRs.

veloman-yunkan:
Fixed by #1298

PR #1270, opened 2026-02-16
add-download-button-funtionality

rajvi0106:
fixes #1083

This PR adds a download button to the preview page on library.kiwix.org making it making it easier for users to directly download the ZIM file they are currently viewing without navigating away or searching for the download link.

**Changes Implemented**

- Added the download button to the UI.
Screenshot from 2026-02-17 22-31-29

- also added the modal box which appears when the user clicks on the download button, with different ways to download.
image

PR #1299, opened 2026-05-11
Topbar download button

### LINKED ISSUE #1083: Add download button in the preview of ZIMs in library.kiwix.org
Most users probably preview a ZIM before downloading, and thus it would make sense to have a download option somewhere in the preview; probably in the top-bar. I'll send a demo here shortly.

kelson42:
Fixes #1083

### DIFF
--- a/static/resources_list.txt
+++ b/static/resources_list.txt
@@ -24,6 +24,7 @@ skin/fonts/Poppins.ttf
 skin/fonts/Roboto.ttf
 skin/search_results.css
 skin/blank.html
+skin/download.js
 skin/polyfills.js
 skin/viewer.js
 skin/i18n.js
--- a/static/skin/download.js
+++ b/static/skin/download.js
@@ -1,0 +1,119 @@
+function downloadButtonText(zimSize) {
+    return $t("download") + (zimSize ? ` - ${zimSize}`: '');
+}
+
+/* hack for library.kiwix.org magnet links (created by MirrorBrain)
+   See https://github.com/kiwix/container-images/issues/242 */
+async function getFixedMirrorbrainMagnet(magnetLink) {
+    // parse as query parameters
+    const params = new URLSearchParams(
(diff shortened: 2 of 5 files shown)


soloturn:
Opened #1340 against this branch to finish the wiring. Summary of what it does:

- Adds `fetchBookDownloadInfo(book)` to `viewer.js`, following the same pattern already used to fetch the current book's *title* on every book switch (a small async call in `updateCurrentBook()`, since `viewer.html` is a static, book-agnostic shell reused across every book — it has no way to know the download link at page-load time). It queries the existing `GET /catalog/v2/entries?name={book}&count=1` and reads the acquisition link the same way `index.js`'s book-grid cards already do. `name` is matched as an exact boolean term server-side (`kiwix::nameQuery()`), not fuzzy, so it can't match the wrong book — and books with no download URL correctly get no link back at all (addresses the "can't fabricate a download URL" objection from #1270).
- Splits `insertDownloadZimModal` into a thin per-button wrapper (unchanged behavior for `index.js`'s fixed-per-card buttons) plus a reusable `showDownloadZimModal(downloadLink, downloadSize, root)` that `viewer.js` calls with fresh values at click time, since the taskbar button's target book changes as the user navigates.
- Shows/hides the taskbar button based on whether the current book actually has a download link.
- Moves `humanFriendlySize()`/`humanFriendlyNumStr()` from `index.js` into the shared `download.js` so `viewer.js` can format sizes the same way.
- Fixes two pre-existing gaps in this branch's own `CacheIdsOfStaticResources` test (found while updating it): `download.js` was never added to the resource/cacheid test list, and the root-page check still expected `index.js`'s now-moved image references.

Couldn't run `meson test` locally in the environment I had (full dependency chain would've taken longer than the fix), so I hand-verified every `cacheid` by recomputing `sha1sum(file)[:8]` and cross-checking against `test/server.cpp` line by line, plus `node --check` on the JS. CI on #1340 is currently sitting at `action_required` pending approval.

PR #1308, opened 2026-06-20
Rollback magnet link hack

### LINKED ISSUE #1307: Remove magnet hack
In #1001, we introduced code in kiwix-serve reader to fix the magnet output of mirrorbrain which had several issues:

- encoding was not supported by some clients
- there was no webseed
- no link to torrent

Without the fix, user needed a client accepting its extra encoding and entirely relied on the trackers to work.
With the fix, any webseed client would be happy. Even non-webseed could find the torrent in `xs` and work off that.

I've backported all those tweaks into (our) `mod_mirrorbrain` and now the ZIM download server outputs it directly. if this hack was only to support our own server, I think we can remove it.

Please note that the hack was written in a way that a fixed server would not disturb it ; it's just wasted resources. That said, it's making a web request and parsing XML for no reason now.

kelson42:
Fixes #1307

Rollback https://github.com/kiwix/libkiwix/pull/1001

### DIFF
--- a/static/skin/index.js
+++ b/static/skin/index.js
@@ -244,59 +244,6 @@
         }
     }
 
-    function makeURLSearchString(params, keysToURIEncode) {
-        let output = '';
-        for (const [key, value] of params.entries()) {
-            let finalValue = (keysToURIEncode.indexOf(key) >= 0) ? encodeURIComponent(value) : value;
-            output += `&${key}=${finalValue}`;
-        }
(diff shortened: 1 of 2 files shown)


rgaudin:
If it's just a rollback, I believe the `cacheid` needs to be updated

kelson42:
cacheid is updated!

rgaudin (approved):
LGTM

veloman-yunkan (approved):

PR #1311, opened 2026-07-23
Allow ingesting a library directly from an OPDS feed

### LINKED ISSUE #1309: Allow to indgest directly an OPDS feed
In place of the `library.xml`

Clear goal is to remove the support of `library.xml` to simplify evolution over time.

hamazaspavetisyan:
Fixes #1309

Summary                                                                                                                                                                                                        
  - Manager::readFile() now auto-detects whether the given file is a library.xml or an OPDS feed (Manager::detectFormat) and dispatches to the right parser, so callers no longer need to specify the format     
  explicitly.                                                                                                                                                                                                    
  - readFile() gained a urlHost parameter, used to resolve host-relative URLs (e.g. favicons) found in OPDS entries.                                                                                             
  - parseOpdsDom() now honors readOnly/trustLibrary like the XML path already did, and can re-read book metadata from the local ZIM when trustLibrary is false.                                                  
  - Book::updateFromOpds() now validates that a book's self path is actually readable (m_pathValid) before trusting it.                                                                                          
  - Added libopds_dumper and refactored library_dumper to share dumping logic, replacing the Jump to bottom (click)


include/library.h
```diff
@@ -224,7 +224,7 @@ using ConstLibraryPtr = std::shared_ptr<const Library>;
 class Library: public std::enable_shared_from_this<Library>
 {
  public:
-  typedef uint64_t Revision;
+  typedef uint64_t Revision; // xxxx todo
```
hamazaspavetisyan: similar situation with downloadId

kelson42:
@hamazaspavetisyan Thank you for the PR but I'm unsure about the status. It is in draft, so it seems you continue to work on it, but at the same time you have assigned me as reviewer... so maybe you expect some kind of feedback.

include/library.h
```diff
@@ -361,12 +361,12 @@ class Library: public std::enable_shared_from_this<Library>
   bool removeBookById(const std::string& id);
 
   /**
-   * Write the library to a file.
+   * Write the library to a file as a library.xml.
    *
    * @param path the path of the file to write to.
    * @return True if the library has been correctly saved.
    */
-  bool writeToFile(const std::string& path) const;
```
kelson42, line 369: We need to be careful here and don't break the ABI. We follow semantic versioning. As far as reasonably possible, I would like to avoid to have to do a major release. Maybe we can keep the older naming `writeToFile()` as a wrapup to `writeToXMLFile()` and keep so the backward compatibility?

include/manager.h
```diff
@@ -60,23 +60,35 @@ class LibraryManipulator
...
+   * Read a `library.xml` (or an OPDS feed) and add the books in the file to the
+   * library.
    *
-   * @param path The (utf8) path to the `library.xml`.
+   * @param path The (utf8) path to the file.
    * @param readOnly Set if the libray path could be overwritten latter with
    *                 updated content.
-   * @param trustLibrary use book metadata coming from XML.
+   * @param trustLibrary use book metadata coming from the file.
+   * @param urlHost Used (when format is FileFormat::OPDS) to resolve
```
kelson42, line 83: Here there is a confusion about the requirements. The OPDS content is available as a file, not an URL. The direct handling of a URL is an interesting topic, but please isolate this in a dedicated issue/PR (but not sure right now we need this).

PR #1316, opened 2026-07-31
Allow ingesting a library directly from an OPDS feed

hamazaspavetisyan:
Adds the ability for `Manager` to load a library directly from an OPDS feed,                                                                                         
  in addition to the existing `library.xml` format — as a step toward the                                                                                              
  issue's longer-term goal of simplifying library I/O around a single format.
 This PR doesn't remove `library.xml` support — it adds OPDS as a parallel                                                                                            
  input format, which is the enabling step before any future removal.

**Format auto-detection**: `Manager::readFile()`/`reload()` now detect                                                                                             
    whether a given path is `library.xml` or an OPDS feed                                                                                                              
    (`Manager::detectFormat`) and dispatch to the right parser transparently;                                                                                          
    callers don't need to know or specify the format. 

**Local path support in OPDS**: `Book::updateFromOpds()` now recognizes a                                                                                          
    `rel="self"` link as the book's local file path, so an OPDS document can                                                                                           
    describe on-disk books, not just remote/importable ones.

**`LibOPDSDumper`**: a new class to dump a `Library` to OPDS for                                                                                                   
    local/offline use (as opposed to the existing `OPDSDumper`, used for the                                                                                           
    live HTTP catalog, which never leaks local filesystem paths to remote                                                                                              
    clients). `library_dumper.{h,cpp}` was refactored into a shared base so                                                                                            
    `OPDSDumper` and `LibOPDSDumper` reuse the same entry-rendering logic                                                                                              
    (`fullEntryOpds`, renamed from `fullEntryXML` for clarity).

**Test coverage**: new dedicated suites (`libopds_dumper`, `libxml_dumper`,                                                                                        
    `opds_dumper`, `library_dumper`), new unit tests for `updateFromOpds`                                                                                              
    mirroring the existing `updateFromXml` ones, and two new OPDS test fixtures                                                                                        
    (`test/data/library.opds`, `test/data/lib_for_server_search_test.opds`)                                                                                            
    mirroring existing XML ones. Several existing suites (`manager`, `library`,                                                                                        
    `library_server`, `server_search`) were parameterized to run their existing                                                                                        
    test bodies against both the XML and OPDS forms of the same library data,                                                                                          
    rather than duplicating test logic.

Unrelated small fix: `ServerNegativeTest.UnusablePort` was flaky on macOS                                                                                          
    (address binding); now binds explicitly to `127.0.0.1`.


test/libxml_dumper.cpp
```diff
@@ -0,0 +1,304 @@
...
+{
+  pugi::xml_document doc;
+  const auto result = doc.load_string(xmlContent.c_str());
+  EXPECT_TRUE(result) << "XML content is not well-formed: "
+                      << result.description() << "\n"
+                      << xmlContent;
+  return doc;
+}
+
+class LibXMLDumperTest : public ::testing::Test
```
veloman-yunkan, line 61: Why don't you introduce this new unit-test in a commit of its own?

test/libxml_dumper.cpp
```diff
@@ -0,0 +1,304 @@
...
+  EXPECT_STREQ("2021-03-25", book.attribute("date").as_string());
+  EXPECT_STREQ("http://download.kiwix.org/zim/book1-id.zim",
+               book.attribute("url").as_string());
+  EXPECT_STREQ("42", book.attribute("articleCount").as_string());
+  EXPECT_STREQ("7", book.attribute("mediaCount").as_string());
+  // Size is stored on disk in KiB (bytes >> 10).
+  EXPECT_STREQ("120", book.attribute("size").as_string());
+  EXPECT_STREQ("download-book1-id", book.attribute("downloadId").as_string());
+  EXPECT_FALSE(book.attribute("origId"));
+}
```
veloman-yunkan, line 100: Instead of parsing the XML and checking the DOM elementwise, why don't you just check the produced XML as text? See how it's done for OPDS catalog in `test/library_server.cpp`.

This comment applies to all tests in this file as well as to its sibling unit-test for OPDS.

src/libopds_dumper.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * A tool to dump a `Library` into a basic OPDS feed.
+ *
+ */
+class LibOPDSDumper
+{
+ public:
+  LibOPDSDumper() = default;
+  LibOPDSDumper(const Library* library);
```
veloman-yunkan: Make `explicit`

src/libopds_dumper.h
```diff
@@ -0,0 +1,72 @@
...
+#include <string>
+#include <vector>
+
+#include "library.h"
+
+namespace kiwix
+{
+
+/**
+ * A tool to dump a `Library` into a basic OPDS feed.
```
veloman-yunkan: Let's not refer to OPDS as feed in contexts intended to be used as offline library.

src/libopds_dumper.h
```diff
@@ -0,0 +1,72 @@
...
+   * @param library The library to dump.
+   */
+  void setLibrary(const Library* library) { this->library = library; }
+
+ protected:
+  const kiwix::Library* library = nullptr;
+  std::string rootLocation;
+ private:
+  std::string handleBook(const Book& book) const;
+};
```
veloman-yunkan, line 71: Shouldn't this also have `baseDir` similar to `LibXMLDumper`?

src/libopds_dumper.h
```diff
@@ -0,0 +1,72 @@
...
+  LibOPDSDumper(const Library* library);
+  ~LibOPDSDumper();
+
+  /**
+   * Dump the OPDS feed.
+   *
+   * @param bookIds the ids of the books to include in the feed.
+   * @return The OPDS feed content.
+   */
+  std::string dumpOPDSContent(const std::vector<std::string>& bookIds);
```
veloman-yunkan: I think we should use the opportunity and start sending the output into a `std::ostream&` instead of returning a `std::string`.

src/libopds_dumper.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Set the root location used when generating urls.
+   *
+   * @param rootLocation the root location to use.
+   */
+  void setRootLocation(const std::string& rootLocation) { this->rootLocation = rootLocation; }
```
veloman-yunkan: I guess this was mimicked based on the online catalog but I don't think that it is justified for an offline library.

test/opds_dumper.cpp
```diff
@@ -0,0 +1,286 @@
...
+  return doc;
+}
+
+// NB: the macro parameter names must not collide with any token used in the
+// body (e.g. the pugixml member names "node"/"text"), since preprocessor
+// substitution is purely textual and would otherwise mangle those calls.
+#define EXPECT_XPATH_TEXT(xmlNode, xpathExpr, expectedText) \
+  EXPECT_EQ(std::string(expectedText), std::string((xmlNode).select_node(xpathExpr).node().text().as_string()))
+
+class OPDSDumperTest : public ::testing::Test {
```
veloman-yunkan, line 176: 1. This unit test too must be introduced in a separate commit.
2. Produced text output must be checked directly (without parsing and checking the DOM).

include/manager.h
```diff
@@ -63,20 +63,28 @@ class Manager
  public: // types
   typedef std::vector<std::string> Paths;
 
+  /**
+   * The format of the file passed to readFile().
+   */
+  enum class FileFormat { XML, OPDS };
```
veloman-yunkan, line 69: This is not passed into `readFile()` (I know that it is a leftover from earlier work).

include/manager.h
```diff
@@ -114,6 +122,7 @@ class Manager
    */
   bool readOpds(const std::string& content, const std::string& urlHost);
 
+  FileFormat detectFormat(const std::string& libraryPath);
```
veloman-yunkan: This commit (titled "OPDS file I/O integration into Manager and Library classes") contains two independent changes related to input & output operations. Let's split it into 2 commits.

src/libopds_dumper.cpp
```diff
@@ -45,7 +45,7 @@ std::string LibOPDSDumper::handleBook(const Book& book) const
   // Local/offline dump: the book's own id is used as the content id, there
   // is no content-access URL to link to (no server involved), and the
   // book's local path is safe to expose as the rel="self" link.
-  return fullEntryXML(book, rootLocation, /*contentAccessUrl=*/"", /*contentId=*/book.getId());
+  return fullEntryOpds(book, rootLocation, /*contentAccessUrl=*/"", /*contentId=*/book.getId());
```
veloman-yunkan: This change should come earlier in the history of the PR (maybe even before the function is made available for public use).

test/library_dumper.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+TEST(FullEntryOpdsTest, omitsAcquisitionLinkWhenUrlIsEmpty)
+{
+  const Book book = createBook();
+  const auto entryXml = fullEntryOpds(book, "http://root.location", "", "book-id");
+  const auto doc = parseEntry(entryXml);
+
+  const auto entry = doc.select_node("/entry").node();
+  EXPECT_FALSE(entry.select_node("link[@rel='http://opds-spec.org/acquisition/open-access']"));
+}
```
veloman-yunkan: 1. I think that the name of the unit-test file is misleading/confusing.
2. Again, I believe that the produced text must be checked directly.

test/book.cpp
```diff
@@ -66,6 +66,49 @@ TEST(BookTest, updateFromXMLTest)
     EXPECT_EQ(defaultIllustration->url, "http://who.org/zara.fav");
 }
 
+TEST(BookTest, updateFromOPDSTest)
```
veloman-yunkan, line 69: The commit with title "Modifying existing tests to work also with OPDS library files." can and should be split into smaller commits focusing on one unit-test at a time.

veloman-yunkan (changes requested):
Let's extract a portion of this PR related to refactoring/groundwork & adding missing test cases for existing code into a standalone PR that we can merge quickly. Don't close these PR. Instead,
1. Split some commits as requested in comments below
2. Move all preparatory changes to the beginning of the commit list
3. Create a new branch at the last commit of those preparatory changes
4. Submit a new PR using that new branch
5. Change the base branch of this PR to that new branch

test/libxml_dumper.cpp
hamazaspavetisyan: Done

test/libxml_dumper.cpp
hamazaspavetisyan: I think this approach is more readable, but changed that anyways, to check string match

test/opds_dumper.cpp
hamazaspavetisyan: Fixed

src/libopds_dumper.cpp
hamazaspavetisyan: Agree, done

test/library_dumper.cpp
hamazaspavetisyan: Fixed

src/libopds_dumper.h
hamazaspavetisyan: We have also "dumpLibXMLContent", for it too ?

src/libopds_dumper.h
hamazaspavetisyan: Added

src/libopds_dumper.h
hamazaspavetisyan: ok

include/manager.h
hamazaspavetisyan: This enum will be used in kiwix-tools with detectFormat method

src/libopds_dumper.h
hamazaspavetisyan: Removed

PR #1319, opened 2026-08-06
Reading OPDS library file

hamazaspavetisyan:
# Reading OPDS library file

## Why

Addresses issue [#1309](https://github.com/kiwix/libkiwix/issues/1309) — "Allow to ingest directly an OPDS feed." The long-term goal stated in the issue is to eventually drop `library.xml` support entirely in favor of OPDS, to simplify the format libkiwix has to maintain over time. This PR is the first step toward that: it makes `Manager` and `Server` able to read OPDS as a library source through the same entry point as XML, without yet removing XML support. It is a prerequisite for any follow-up PR that would deprecate or drop `library.xml`. The next PR will cover OPDS file writing functionality.

## What

`Manager::readFile()` now auto-detects whether the file it's given is a `library.xml` or an OPDS file, and parses it accordingly — previously OPDS content could only be loaded via the separate `readOpds()` string-based API. `parseOpdsDom()` was extended to accept the same `readOnly`/`trustLibrary` flags `readFile()` already exposes for XML, and `Book::updateFromOpds()` was fixed up to capture a book's own path (via the OPDS `rel="self"` link) and to not stomp on existing illustrations when re-parsing.

No breaking changes to the public API — `readFile(path, readOnly, trustLibrary)` keeps the same signature. Callers can now point it at either a `library.xml` or an OPDS feed and get consistent behavior (same read-only/trust-library semantics either way), instead of needing to know the format up front and call a different method. `trustLibrary=false` on an OPDS file now actually re-verifies book metadata from disk rather than blindly trusting the feed, matching existing XML behavior.

## Fundamental change

Format detection is content-based, not extension-based: `detectFormat()` sniffs for a `<feed` tag in the file content. This means `readFile()` is no longer XML-only — it's now a dual-format entry point, with OPDS parsing brought up to parity with XML parsing on the flags that control trust and mutability of loaded books.

## Important notes
kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.

### DIFF
--- a/include/book.h
+++ b/include/book.h
@@ -78,7 +78,7 @@ class Book
    * @param baseDir base directory used to resolve a relative `rel="self"`
    *                link into an absolute book path.
    */
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
   std::string getHumanReadableIdFromPath() const;
 
   bool readOnly() const { return m_readOnly; }
--- a/include/book.h
+++ b/include/book.h
@@ -70,7 +70,15 @@ class Book
   bool update(const Book& other);
   void update(const zim::Archive& archive);
   void updateFromXml(const pugi::xml_node& node, const std::string& baseDir);
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost);
+  /**
+   * Update the book's metadata from an OPDS entry XML node.
+   *
+   * @param node the `<entry>` node of an OPDS feed describing the book.
+   * @param urlHost host to prepend to relative illustration/thumbnail URLs.
+   * @param baseDir base directory used to resolve a relative `rel="self"`
+   *                link into an absolute book path.
+   */
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
   std::string getHumanReadableIdFromPath() const;
 
   bool readOnly() const { return m_readOnly; }
--- a/include/book.h
+++ b/include/book.h
@@ -71,14 +71,14 @@ class Book
   void update(const zim::Archive& archive);
   void updateFromXml(const pugi::xml_node& node, const std::string& baseDir);
   /**
- * Update the book's metadata from an OPDS entry XML node.
- *
- * @param node the `<entry>` node of an OPDS feed describing the book.
- * @param urlHost host to prepend to relative illustration/thumbnail URLs.
- * @param baseDir base directory used to resolve a relative `rel="self"`
- *                link into an absolute book path.
- */
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
+   * Update the book's metadata from an OPDS entry XML node.
+   *
+   * @param node the `<entry>` node of an OPDS feed describing the book.
+   * @param urlHost host to prepend to relative illustration/thumbnail URLs.
+   * @param baseDir base directory used to resolve a relative `rel="self"`
+   *                link into an absolute book path.
+   */
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
   std::string getHumanReadableIdFromPath() const;
 
   bool readOnly() const { return m_readOnly; }
--- a/include/book.h
+++ b/include/book.h
@@ -70,7 +70,7 @@ class Book
   bool update(const Book& other);
   void update(const zim::Archive& archive);
   void updateFromXml(const pugi::xml_node& node, const std::string& baseDir);
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost);
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
   std::string getHumanReadableIdFromPath() const;
 
   bool readOnly() const { return m_readOnly; }
--- a/include/book.h
+++ b/include/book.h
@@ -71,14 +71,14 @@ class Book
   void update(const zim::Archive& archive);
   void updateFromXml(const pugi::xml_node& node, const std::string& baseDir);
   /**
-   * Update the book's metadata from an OPDS entry XML node.
-   *
-   * @param node the `<entry>` node of an OPDS feed describing the book.
-   * @param urlHost host to prepend to relative illustration/thumbnail URLs.
-   * @param baseDir base directory used to resolve a relative `rel="self"`
-   *                link into an absolute book path.
-   */
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
+ * Update the book's metadata from an OPDS entry XML node.
+ *
+ * @param node the `<entry>` node of an OPDS feed describing the book.
+ * @param urlHost host to prepend to relative illustration/thumbnail URLs.
+ * @param baseDir base directory used to resolve a relative `rel="self"`
+ *                link into an absolute book path.
+ */
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
   std::string getHumanReadableIdFromPath() const;
 
   bool readOnly() const { return m_readOnly; }
--- a/include/manager.h
+++ b/include/manager.h
@@ -188,7 +188,7 @@ class Manager
                    bool trustLibrary);
   bool parseOpdsDom(const pugi::xml_document& doc,
                     const std::string& urlHost,
-                    const std::string& libraryPath,
+                    const std::string& baseDir,
                     bool readOnly);
 };
 }  // namespace kiwix
--- a/include/manager.h
+++ b/include/manager.h
@@ -182,6 +182,7 @@ class Manager
                    bool trustLibrary);
   bool parseOpdsDom(const pugi::xml_document& doc,
                     const std::string& urlHost,
+                    const std::string& libraryPath,
                     bool readOnly,
                     bool trustLibrary);
 };
--- a/include/manager.h
+++ b/include/manager.h
@@ -108,18 +108,18 @@ class Manager
    * Load a library content stored in a OPDS stream or a local library file.
    *
    * @param content The content of the OPDS stream.
-   * @param urlHost The host used to resolve relative acquisition/thumbnail
-   *                links.
+   * @param contentOriginUri Where content was read from: either the URL it
+   *        was fetched from (used to resolve relative acquisition/thumbnail
+   *        links) or the local filesystem path it was read from (used to
+   *        resolve a relative rel="self" link). See
+   *        kiwix::resolveContentOrigin() for the exact splitting rules.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
-   * @param libraryPath The library path (used to resolve a relative
-   *                     rel="self" link)
    * @return True if the content has been properly parsed.
    */
   bool readOpds(const std::string& content,
-                const std::string& urlHost,
-                bool readOnly = false,
-                const std::string& libraryPath = "");
+                const std::string& contentOriginUri,
+                bool readOnly = false);
 
 
   /**
--- a/include/manager.h
+++ b/include/manager.h
@@ -107,9 +107,8 @@ class Manager
    * Load a library content stored in a OPDS stream.
    *
    * @param content The content of the OPDS stream.
-   * @param readOnly Set if the library path could be overwritten later with
-   *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
+   * @param urlHost The host used to resolve relative acquisition/thumbnail
+   *                links.
    * @return True if the content has been properly parsed.
    */
   bool readOpds(const std::string& content, const std::string& urlHost);
--- a/include/manager.h
+++ b/include/manager.h
@@ -68,12 +68,13 @@ class Manager
   explicit Manager(LibraryPtr library);
 
   /**
-   * Read a `library.xml` and add book in the file to the library.
+   * Read a library XML or an OPDS file and add the books in the file to the
+   * library.
    *
-   * @param path The (utf8) path to the `library.xml`.
+   * @param path The (utf8) path to the library file.
    * @param readOnly Set if the libray path could be overwritten latter with
    *                 updated content.
-   * @param trustLibrary use book metadata coming from XML.
+   * @param trustLibrary use book metadata coming from the file.
    * @return True if file has been properly parsed.
    */
   bool readFile(const std::string& path, bool readOnly = true, bool trustLibrary = true);
@@ -104,7 +105,7 @@ class Manager
                bool trustLibrary = true);
 
   /**
-   * Load a library content stored in a OPDS stream.
+   * Load a library content stored in a OPDS stream/file.
    *
    * @param content The content of the OPDS stream.
    * @param readOnly Set if the library path could be overwritten later with
@@ -157,7 +158,7 @@ class Manager
 
   /**
    * Add all books from the directory tree into the library.
-   * 
+   *
    * @param path          The path of the directory to scan.
    * @param verboseFlag   Verbose logs flag.
    */
@@ -180,9 +181,10 @@ class Manager
                    const std::string& libraryPath,
                    bool trustLibrary);
   bool parseOpdsDom(const pugi::xml_document& doc,
-                    const std::string& urlHost);
-
+                    const std::string& urlHost,
+                    bool readOnly,
+                    bool trustLibrary);
 };
-}
+}  // namespace kiwix
 
 #endif
--- a/include/manager.h
+++ b/include/manager.h
@@ -68,9 +68,10 @@ class Manager
   explicit Manager(LibraryPtr library);
 
   /**
-   * Read a `library.xml` and add book in the file to the library.
+   * Read a library XML or an OPDS file and add the books in the file to the
+   * library.
    *
-   * @param path The (utf8) path to the `library.xml`.
+   * @param path The (utf8) path to the library file.
    * @param readOnly Set if the libray path could be overwritten latter with
    *                 updated content.
    * @param trustLibrary use book metadata coming from XML.
@@ -104,14 +105,18 @@ class Manager
                bool trustLibrary = true);
 
   /**
-   * Load a library content stored in a OPDS stream.
+   * Load a library content stored in a OPDS stream or a local library file.
    *
    * @param content The content of the OPDS stream.
    * @param urlHost The host used to resolve relative acquisition/thumbnail
    *                links.
+   * @param readOnly Set if the library path could be overwritten later with
+   *                 updated content.
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& urlHost,
+                bool readOnly = false);
 
 
   /**
@@ -156,7 +161,7 @@ class Manager
 
   /**
    * Add all books from the directory tree into the library.
-   * 
+   *
    * @param path          The path of the directory to scan.
    * @param verboseFlag   Verbose logs flag.
    */
@@ -179,9 +184,9 @@ class Manager
                    const std::string& libraryPath,
                    bool trustLibrary);
   bool parseOpdsDom(const pugi::xml_document& doc,
-                    const std::string& urlHost);
-
+                    const std::string& urlHost,
+                    bool readOnly);
 };
-}
+}  // namespace kiwix
 
 #endif
--- a/include/manager.h
+++ b/include/manager.h
@@ -112,11 +112,14 @@ class Manager
    *                links.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
+   * @param libraryPath The library path (used to resolve a relative
+   *                     rel="self" link)
    * @return True if the content has been properly parsed.
    */
   bool readOpds(const std::string& content,
                 const std::string& urlHost,
-                bool readOnly = false);
+                bool readOnly = false,
+                const std::string& libraryPath = "");
 
 
   /**
@@ -185,6 +188,7 @@ class Manager
                    bool trustLibrary);
   bool parseOpdsDom(const pugi::xml_document& doc,
                     const std::string& urlHost,
+                    const std::string& libraryPath,
                     bool readOnly);
 };
 }  // namespace kiwix
--- a/include/tools.h
+++ b/include/tools.h
@@ -38,6 +38,16 @@ struct IpAddress
     std::string addr6; /**< IPv6 address */
 };
 
+/**
+ * The pieces needed to resolve references found in OPDS content, split out
+ * of the location that content was read from. See resolveContentOrigin().
+ */
+struct ContentOrigin
+{
+    std::string urlHost; /**< scheme+host[:port], no trailing slash; "" if the origin is a local path */
+    std::string baseDir; /**< parent directory of a local path; "" if the origin is a URL */
+};
+
 typedef std::pair<std::string, std::string> LangNameCodePair;
 typedef std::vector<LangNameCodePair> FeedLanguages;
 typedef std::vector<std::string> FeedCategories;
@@ -125,6 +135,23 @@ std::string computeAbsolutePath(const std::string& path, const std::string& rela
  * @return a relative path (pointing to absolutePath, relative to path).
  */
 std::string computeRelativePath(const std::string& path, const std::string& absolutePath);
+
+/** Split a "content origin" locator into a urlHost/baseDir pair.
+ *
+ * contentOriginUri is either a URL that some content was fetched from (e.g.
+ * "https://library.kiwix.org/catalog/v2/entries") or a local filesystem path
+ * that content was read from (e.g. "/data/library.opds"). Classification is
+ * purely syntactic: a "://" substring marks the input as a URL; everything
+ * else (including a Windows drive-letter path) is treated as a local path.
+ * Exactly one of the two returned fields is non-empty.
+ *
+ * Known limitation: a "file://" URI is classified as a URL, not a local
+ * path (no current caller produces one).
+ *
+ * @param contentOriginUri the URL or local path some content was read from.
+ * @return the corresponding urlHost/baseDir pair.
+ */
+ContentOrigin resolveContentOrigin(const std::string& contentOriginUri);
 
 /** Sleep the current thread.
  *
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -145,13 +145,12 @@ static std::string fromOpdsDate(const std::string& date)
 
 
 #define VALUE(name) node.child(name).child_value()
-void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost)
+void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir)
 {
   m_id = VALUE("id");
   if (!m_id.compare(0, 9, "urn:uuid:")) {
     m_id.erase(0, 9);
   }
-  // No path on opds.
   m_title = VALUE("title");
   m_description = VALUE("summary");
   m_language = VALUE("language");
@@ -171,6 +170,12 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
   for(auto linkNode = node.child("link"); linkNode;
            linkNode = linkNode.next_sibling("link")) {
     std::string rel = linkNode.attribute("rel").value();
+
+    if (rel == "self") {
+      const std::string path = linkNode.attribute("href").value();
+      m_path = isRelativePath(path)? computeAbsolutePath(baseDir, path): path;
+      m_pathValid = fileReadable(m_path);
+    }
 
     if (rel == "http://opds-spec.org/acquisition/open-access") {
       m_url = linkNode.attribute("href").value();
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -251,7 +251,9 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       // XXX non-absolute URL is expected to be an absolute-path.
       favicon->url = isAbsoluteUrl(thumbnailUrl)? thumbnailUrl: joinUrl(urlHost, thumbnailUrl);
       favicon->mimeType = linkNode.attribute("type").value();
-      m_illustrations.push_back(favicon);
+      if ( !favicon->mimeType.empty() ) {
+        m_illustrations.push_back(favicon);
+      }
     }
   }
 
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -25,12 +25,47 @@
 #include "tools/networkTools.h"
 #include "tools/otherTools.h"
 #include "tools/stringTools.h"
-#include "tools/pathTools.h"
 #include "tools/archiveTools.h"
 
 #include <zim/archive.h>
 #include <zim/item.h>
 #include <pugixml.hpp>
+
+#include <sstream>
+
+namespace
+{
+/**
+ * Tells whether a URL string is already absolute (contains a scheme,
+ * e.g. "https://example.com/x.png") as opposed to being a relative path
+ * (e.g. "/x.png"). Only a relative URL should be prefixed with a base host
+ * or URL—doing so unconditionally would garble an already-absolute one.
+ */
+bool isAbsoluteUrl(const std::string& url)
+{
+  // Find the scheme separator
+  size_t pos = url.find("://");
+  if (pos == 0 || pos == std::string::npos) {
+    return false; // No scheme or empty scheme
+  }
+
+  // RFC 3986: Scheme must begin with a letter, followed by letters, digits, '+', '.', or '-'
+  if (!std::isalpha(static_cast<unsigned char>(url[0]))) {
+    return false;
+  }
+
+  // Validate that all remaining characters in the scheme comply with RFC 3986
+  for (size_t i = 1; i < pos; ++i) {
+    char c = url[i];
+    if (!std::isalnum(static_cast<unsigned char>(c)) && c != '+' && c != '-' && c != '.') {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+} // anonymous namespace
 
 namespace kiwix
 {
@@ -184,7 +219,9 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
     if (rel == "http://opds-spec.org/image/thumbnail") {
       const auto favicon = std::make_shared<Illustration>();
       favicon->data.clear();
-      favicon->url = urlHost + linkNode.attribute("href").value();
+      const std::string thumbnailUrl = linkNode.attribute("href").value();
+      // XXX non-absolute URL is expected to be an absolute-path.
+      favicon->url = isAbsoluteUrl(thumbnailUrl)? thumbnailUrl: urlHost + thumbnailUrl;
       favicon->mimeType = linkNode.attribute("type").value();
       m_illustrations.push_back(favicon);
     }
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -65,6 +65,34 @@ bool isAbsoluteUrl(const std::string& url)
   return true;
 }
 
+/**
+ * Joins a host and a reference into a single URL without producing a double slash,
+ * respecting the host's format.
+ */
+std::string joinUrl(const std::string& host, const std::string& ref)
+{
+  if (host.empty()) {
+    return ref;
+  }
+  if (ref.empty()) {
+    return host;
+  }
+
+  const bool hostEndsWithSlash = (host.back() == '/');
+  const bool refStartsWithSlash = (ref.front() == '/');
+
+  if (hostEndsWithSlash && refStartsWithSlash) {
+    // Both have a slash; omit one to avoid a double slash
+    return host + ref.substr(1);
+  } else if (!hostEndsWithSlash && !refStartsWithSlash) {
+    // Neither has a slash; insert one
+    return host + "/" + ref;
+  } else {
+    // Exactly one has a slash; simple concatenation is correct
+    return host + ref;
+  }
+}
+
 } // anonymous namespace
 
 namespace kiwix
@@ -221,7 +249,7 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       favicon->data.clear();
       const std::string thumbnailUrl = linkNode.attribute("href").value();
       // XXX non-absolute URL is expected to be an absolute-path.
-      favicon->url = isAbsoluteUrl(thumbnailUrl)? thumbnailUrl: urlHost + thumbnailUrl;
+      favicon->url = isAbsoluteUrl(thumbnailUrl)? thumbnailUrl: joinUrl(urlHost, thumbnailUrl);
       favicon->mimeType = linkNode.attribute("type").value();
       m_illustrations.push_back(favicon);
     }
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -145,13 +145,12 @@ static std::string fromOpdsDate(const std::string& date)
 
 
 #define VALUE(name) node.child(name).child_value()
-void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost)
+void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir)
 {
   m_id = VALUE("id");
   if (!m_id.compare(0, 9, "urn:uuid:")) {
     m_id.erase(0, 9);
   }
-  // No path on opds.
   m_title = VALUE("title");
   m_description = VALUE("summary");
   m_language = VALUE("language");
@@ -167,9 +166,18 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
   m_category = catnode.empty() ? getCategoryFromTags() : catnode.child_value();
   m_articleCount = strtoull(VALUE("articleCount"), 0, 0);
   m_mediaCount = strtoull(VALUE("mediaCount"), 0, 0);
-  for(auto linkNode = node.child("link"); linkNode;
+  for (auto linkNode = node.child("link"); linkNode;
            linkNode = linkNode.next_sibling("link")) {
     std::string rel = linkNode.attribute("rel").value();
+
+    if (rel == "self") {
+      std::string path = linkNode.attribute("href").value();
+      if (isRelativePath(path)) {
+        path = computeAbsolutePath(baseDir, path);
+      }
+      m_path = path;
+      m_pathValid = fileReadable(path);
+    }
 
     if (rel == "http://opds-spec.org/acquisition/open-access") {
       m_url = linkNode.attribute("href").value();
--- a/src/book.cpp
+++ b/src/book.cpp
@@ -225,7 +225,23 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       favicon->mimeType = linkNode.attribute("type").value();
       m_illustrations.push_back(favicon);
     }
- }
+  }
+
+  for (auto thumbnailNode = node.child("thumbnails").child("thumbnail"); thumbnailNode;
+           thumbnailNode = thumbnailNode.next_sibling("thumbnail")) {
+    const auto thumbnail = std::make_shared<Illustration>();
+    thumbnail->data = base64_decode(thumbnailNode.child_value());
+    thumbnail->mimeType = thumbnailNode.attribute("mimetype").value();
+    const std::string widthAttr = thumbnailNode.attribute("width").value();
+    if (!widthAttr.empty()) {
+      thumbnail->width = static_cast<uint16_t>(strtoul(widthAttr.c_str(), 0, 0));
+    }
+    const std::string heightAttr = thumbnailNode.attribute("height").value();
+    if (!heightAttr.empty()) {
+      thumbnail->height = static_cast<uint16_t>(strtoul(heightAttr.c_str(), 0, 0));
+    }
+    m_illustrations.push_back(thumbnail);
+  }
 }
 #undef VALUE
 
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -170,7 +170,7 @@ bool Manager::readXml(const std::string& xml,
 
 bool Manager::parseOpdsDom(const pugi::xml_document& doc,
                            const std::string& urlHost,
-                           const std::string& libraryPath,
+                           const std::string& baseDir,
                            bool readOnly)
 {
   pugi::xml_node libraryNode = doc.child("feed");
@@ -194,8 +194,6 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc,
     m_itemsPerPage = 0;
   }
 
-
-  const auto baseDir = removeLastPathElement(libraryPath);
   for (pugi::xml_node entryNode = libraryNode.child("entry"); entryNode;
        entryNode = entryNode.next_sibling("entry")) {
     kiwix::Book book;
@@ -222,7 +220,8 @@ bool Manager::readOpds(const std::string& content,
       = doc.load_buffer((void*)content.data(), content.size());
 
   if (result) {
-    return this->parseOpdsDom(doc, urlHost, libraryPath, readOnly);
+    const auto origin = resolveContentOrigin(libraryPath);
+    return this->parseOpdsDom(doc, urlHost, origin.baseDir, readOnly);
   }
 
   return false;
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -171,6 +171,7 @@ bool Manager::readXml(const std::string& xml,
 
 bool Manager::parseOpdsDom(const pugi::xml_document& doc,
                            const std::string& urlHost,
+                           const std::string& libraryPath,
                            bool readOnly,
                            bool trustLibrary)
 {
@@ -190,7 +191,7 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc,
     kiwix::Book book;
 
     book.setReadOnly(readOnly);
-    book.updateFromOpds(entryNode, urlHost);
+    book.updateFromOpds(entryNode, urlHost, removeLastPathElement(libraryPath));
 
     if (!trustLibrary && !book.getPath().empty()) {
       this->readBookFromPath(book.getPath(), &book);
@@ -212,7 +213,7 @@ bool Manager::readOpds(const std::string& content, const std::string& urlHost)
       = doc.load_buffer((void*)content.data(), content.size());
 
   if (result) {
-    this->parseOpdsDom(doc, urlHost, /*readOnly=*/false, /*trustLibrary=*/true);
+    this->parseOpdsDom(doc, urlHost, /*libraryPath=*/"", /*readOnly=*/false, /*trustLibrary=*/true);
     return true;
   }
 
@@ -236,7 +237,7 @@ bool Manager::readFile(
   if (result) {
     FileFormat format = detectFormat(path);
     if (format == FileFormat::OPDS) {
-      this->parseOpdsDom(doc, "", readOnly, trustLibrary);
+      this->parseOpdsDom(doc, "", path, readOnly, trustLibrary);
     } else {
       this->parseXmlDom(doc, readOnly, path, trustLibrary);
     }
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -32,6 +32,32 @@
 
 namespace fs = std::filesystem;
 
+namespace
+{
+
+/**
+ * The format of the file passed to readFile().
+ */
+enum class FileFormat { XML, OPDS };
+
+/**
+ * Detect the format of a library file based on its content.
+ *
+ * @param libraryPath The path of the file to inspect.
+ * @return FileFormat::OPDS if the file content looks like an OPDS feed,
+ *         FileFormat::XML otherwise.
+ */
+FileFormat detectFormat(const std::string& libraryPath)
+{
+  auto format
+      = (kiwix::getFileContent(libraryPath).find("<feed") != std::string::npos)
+            ? FileFormat::OPDS
+            : FileFormat::XML;
+  return format;
+}
+
+} // anonymous namespace
+
 namespace kiwix
 {
 
@@ -208,7 +234,12 @@ bool Manager::readFile(
 #endif
 
   if (result) {
-    this->parseXmlDom(doc, readOnly, path, trustLibrary);
+    FileFormat format = detectFormat(path);
+    if (format == FileFormat::OPDS) {
+      this->parseOpdsDom(doc, "", readOnly, trustLibrary);
+    } else {
+      this->parseXmlDom(doc, readOnly, path, trustLibrary);
+    }
   } else {
     retVal = false;
   }
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -142,7 +142,9 @@ bool Manager::readXml(const std::string& xml,
 
 
 
-bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& urlHost)
+bool Manager::parseOpdsDom(const pugi::xml_document& doc,
+                           const std::string& urlHost,
+                           bool readOnly)
 {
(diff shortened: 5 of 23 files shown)


veloman-yunkan:
> kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.

I think it is important to note that if `kiwix-manage` is given a library file in OPDS format, that file will be converted to XML by `add` or `remove` commands.

Regarding `kiwix-serve` does reloading of the library from an OPDS format work like it does with XML?

include/manager.h
```diff
@@ -63,17 +63,23 @@ class Manager
  public: // types
   typedef std::vector<std::string> Paths;
 
+  /**
+   * The format of the file passed to readFile().
+   */
+  enum class FileFormat { XML, OPDS };
+
```
veloman-yunkan: Can't this be hidden in an unnamed namespace in `manager.cpp`?

include/manager.h
```diff
@@ -104,16 +110,24 @@ class Manager
...
   bool readOpds(const std::string& content, const std::string& urlHost);
 
+  /**
+   * Detect the format of a library file based on its content.
+   *
+   * @param libraryPath The path of the file to inspect.
+   * @return FileFormat::OPDS if the file content looks like an OPDS feed,
+   *         FileFormat::XML otherwise.
+   */
+  FileFormat detectFormat(const std::string& libraryPath);
```
veloman-yunkan: Can't this be hidden in an unnamed namespace in `manager.cpp`?

src/manager.cpp
```diff
@@ -143,7 +143,10 @@ bool Manager::readXml(const std::string& xml,
 
 
 
-bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& urlHost)
+bool Manager::parseOpdsDom(const pugi::xml_document& doc,
+                           const std::string& urlHost,
+                           bool readOnly,
+                           bool trustLibrary)
```
veloman-yunkan: Please dedicate a separate commit to this change and enhance the unit-tests to cover it.

src/manager.cpp
```diff
@@ -201,7 +217,12 @@ bool Manager::readFile(
 #endif
 
   if (result) {
+    FileFormat format = detectFormat(path);
+    if (format == FileFormat::OPDS) {
+      this->parseOpdsDom(doc, "", readOnly, trustLibrary);
+    } else {
     this->parseXmlDom(doc, readOnly, path, trustLibrary);
+    }
```
veloman-yunkan: Indentation needs to be fixed (it's OK to do that in a separate reformatting commit)

src/book.cpp
```diff
@@ -171,6 +171,11 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
            linkNode = linkNode.next_sibling("link")) {
     std::string rel = linkNode.attribute("rel").value();
 
+    if (rel == "self") {
+      m_path = linkNode.attribute("href").value();
+      m_pathValid = fileReadable(m_path);
+    }
+
```
veloman-yunkan, line 242: 1. Let's do this in a separate commit (with detailed enough description) and update the unit-tests to cover this enhancement.

2. Make sure that relative paths are supported

veloman-yunkan (changes requested):

hamazaspavetisyan:
> > kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.
> 
> I think it is important to note that if `kiwix-manage` is given a library file in OPDS format, that file will be converted to XML by `add` or `remove` commands.
> 
> Regarding `kiwix-serve` does reloading of the library from an OPDS format work like it does with XML?

Yes, in this branch we have only read from OPDS, so remove and add will dump XML.
Yes. we have auto-reload when specifying `--monitorLibrary` flag. Note, that UI is not refreshed automatically, because we do not have WebSockets.

include/manager.h
hamazaspavetisyan: Soon I will use it in kiwix-tools

include/manager.h
hamazaspavetisyan: Same as below, this will be us3d soon in kiwix-tools(my next PR).

src/manager.cpp
hamazaspavetisyan: What if I do that in parseOpdsDom params related commit that you suggested ?

veloman-yunkan: That's also fine. I just wanted to make it clear that it is OK to make code changes in such a way that the diff is minimal (without having to explicitly suppress changes in whitespace) and then fixing formatting separately.

kelson42:
I will review this PR together with the one of kiwix-manage. Link it here please when ready.

hamazaspavetisyan:
> I will review this PR together with the one of kiwix-manage. Link it here please when ready.

@kelson42 

Soon I'll open another PR in **libkiwix** that changes the OPDS dumper output. This change will also affect **kiwix-tools**, so I'll open another PR there as well.

You can either review this PR now, or wait until I submit the other two PRs and review all three together.

test/data/library.opds
```diff
@@ -0,0 +1,112 @@
...
+    <articleCount>284</articleCount>
+    <mediaCount>2</mediaCount>
+    <link rel="self" href="./zimfile_raycharles.zim" />
+    <link rel="http://opds-spec.org/acquisition/open-access"
+          type="application/x-zim"
+          href="https://github.com/kiwix/libkiwix/raw/master/test/data/zimfile_raycharles.zim"
+          length="569344" />
+    <link rel="http://opds-spec.org/image/thumbnail"
+          type="image/png"
+          href="/favicon/raycharles.png" />
```
kelson42: I'm surprised to see here a path, the two clear scenarios are:
* a HTTPS URL
* a base64 embbeded value

I don't remember that a local path support was requested!

test/book.cpp
```diff
@@ -137,11 +156,11 @@ kiwix::Book makeBook(const std::string& attr, const std::string& baseDir="")
     return book;
 }
 
-kiwix::Book makeBookFromOpds(const std::string& entryContent, const std::string& urlHost="")
+kiwix::Book makeBookFromOpds(const std::string& entryContent, const std::string& urlHost="", const std::string& baseDir="")
 {
     const XMLDoc opds("<entry>" + entryContent + "</entry>");
     kiwix::Book book;
-    book.updateFromOpds(opds.child("entry"), urlHost);
+    book.updateFromOpds(opds.child("entry"), urlHost, baseDir);
```
kelson42, line 163: I'm surprised of this new parameter. I don't understand to which feature this is bound. What is the story behind this? Why do we need suddently this new parameter?

As a reminder, library.xml had relative links, but this was resolve at runtime (based on library.xml own path)... so I still don't understand why we would need this parameter there.

Logically we should save nay local library (also in the readers) using OPDS then. You confirm?

hamazaspavetisyan: `baseDir` isn't solving a new problem — it's applying the pre-existing XML-library path-resolution rule to OPDS now that OPDS can also be an on-disk library file.  
The remote-feed call site (readOpds) is unaffected: it still passes libraryPath="", so baseDir is empty and relative self-links there resolve to nothing meaningful (matching the old "no path on opds" behavior for that path). 
So that parameter isn't a new feature bolted onto OPDS parsing in general — it's the direct consequence of turning .opds files into an on-disk library format in this PR series, on par with .xml.

kelson42:
> kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.

kiwix-manage has to:
* Been able to read old library.xml files
* But should always write OPDS files

kiwix-serve should loose his ability to deal with old library.xml.

@hamazaspavetisyan Do you confirm it will work that way?

hamazaspavetisyan:
> > kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.
> 
> kiwix-manage has to:
> 
> * Been able to read old library.xml files
> * But should always write OPDS files
> 
> kiwix-serve should loose his ability to deal with old library.xml.
> 
> @hamazaspavetisyan Do you confirm it will work that way?

Yes, that should come with kiwix-tools PR

kelson42:
> on an OPDS file now actually re-verifies book metadata from disk rather than blindly trusting the feed, matching existing XML behavior.

I'm not sure of all of this, do you confirm that:
* There is no way anymore to specify `trustLibrary` in an OPDS feed?
* `trustLibrary` was a way to tell the reader to trust the data of the ZIM OR the data of the XML?

hamazaspavetisyan:
> > on an OPDS file now actually re-verifies book metadata from disk rather than blindly trusting the feed, matching existing XML behavior.
> 
> I'm not sure of all of this, do you confirm that:
> 
> * There is no way anymore to specify `trustLibrary` in an OPDS feed?
> * `trustLibrary` was a way to tell the reader to trust the data of the ZIM OR the data of the XML?

`trustLibrary` still settable for OPDS, for the file-based path. `Manager::readFile(path, readOnly, trustLibrary)` (manager.h:80) auto-detects .xml vs .opds and forwards trustLibrary straight through to whichever parser applies (manager.cpp:240 for OPDS, :242 for XML) — that plumbing is unchanged by this PR.

Direction of trustLibrary: it means "trust the metadata written in the XML/OPDS file itself" as opposed to "re-derive metadata from the actual ZIM archive on disk." 
Concretely, in `Manager::parseXmlDom/parseOpdsDom` (manager.cpp:145-147 and :196-198):                                                                                 
                                                                                                                                                                       
  if (!trustLibrary && !book.getPath().empty()) {                                                                                                                      
    this->readBookFromPath(book.getPath(), &book);                                                                                                                     
  }                                                                                                                                                                    

`readBookFromPath` (manager.cpp:353-369) opens the ZIM at that path with zim::Archive and calls `book->update(archive)`, which overwrites whatever the XML/OPDS entry said (title, description, counts, etc.) with what's actually inside the ZIM. 
So:                                                                                     
- `trustLibrary=true` -> trust the file's own metadata, skip touching the ZIM.                                                                                          
- `trustLibrary=false` -> distrust the file, re-derive metadata from the real ZIM (only possible if the book resolved to a local path).

test/data/library.opds
veloman-yunkan: I concur on this. `href` must be a full URI.

veloman-yunkan: And we have to support embedded base64-encoded thumbnails too (and have it tested).

veloman-yunkan:
> > kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.
> 
> kiwix-manage has to:
> 
>     * Been able to read old library.xml files
> 
>     * But should always write OPDS files
> 
> 
> kiwix-serve should loose his ability to deal with old library.xml.
> 
> @hamazaspavetisyan Do you confirm it will work that way?

Eventually it should definitely work that way. But shouldn't the next release of `kiwix-manage` by default preserve the format of the input library file? When being passed an old style XML file it may output a warning that this format is being phased out but it shouldn't silently convert it to OPDS.

test/data/library.opds
hamazaspavetisyan: @kelson42 @veloman-yunkan Could we get a final decision on how to implement thumbnails? The last discussion left the approach open between:

1. `<link rel="[http://opds-spec.org/image/thumbnail](http://opds-spec.org/image/thumbnail)">base64EmbeddingHere</link>`

2. `<thumbnails><thumbnail>base64EmbeddingHere</thumbnail></thumbnails>`

veloman-yunkan: I reiterate my point of view that providing image data inside a `<link>` tag is wrong.

kelson42:
> > > kiwix-manage and kiwix-serve automatically work with these new changes, so we will not introduce a PR for kiwix-tools.
> > 
> > 
> > kiwix-manage has to:
> > ```
> > * Been able to read old library.xml files
> > 
> > * But should always write OPDS files
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > kiwix-serve should loose his ability to deal with old library.xml.
> > @hamazaspavetisyan Do you confirm it will work that way?
> 
> Eventually it should definitely work that way. But shouldn't the next release of `kiwix-manage` by default preserve the format of the input library file? When being passed an old style XML file it may output a warning that this format is being phased out but it shouldn't silently convert it to OPDS.

OK for me.

kelson42:
> > > on an OPDS file now actually re-verifies book metadata from disk rather than blindly trusting the feed, matching existing XML behavior.
> > 
> > 
> > I'm not sure of all of this, do you confirm that:
> > 
> > * There is no way anymore to specify `trustLibrary` in an OPDS feed?
> > * `trustLibrary` was a way to tell the reader to trust the data of the ZIM OR the data of the XML?
> 
> `trustLibrary` still settable for OPDS, for the file-based path. `Manager::readFile(path, readOnly, trustLibrary)` (manager.h:80) auto-detects .xml vs .opds and forwards trustLibrary straight through to whichever parser applies (manager.cpp:240 for OPDS, :242 for XML) — that plumbing is unchanged by this PR.
> 
> Direction of trustLibrary: it means "trust the metadata written in the XML/OPDS file itself" as opposed to "re-derive metadata from the actual ZIM archive on disk." Concretely, in `Manager::parseXmlDom/parseOpdsDom` (manager.cpp:145-147 and :196-198):
> 
> if (!trustLibrary && !book.getPath().empty()) { this->readBookFromPath(book.getPath(), &book); }
> 
> `readBookFromPath` (manager.cpp:353-369) opens the ZIM at that path with zim::Archive and calls `book->update(archive)`, which overwrites whatever the XML/OPDS entry said (title, description, counts, etc.) with what's actually inside the ZIM. So:
> 
>     * `trustLibrary=true` -> trust the file's own metadata, skip touching the ZIM.
> 
>     * `trustLibrary=false` -> distrust the file, re-derive metadata from the real ZIM (only possible if the book resolved to a local path).

@hamazaspavetisyan Thank you very much for the clarification. I wanted to be 100% before sharing here my instructions.

This `trustLibrary` is a very old thing and I'm convinced we don't need it anymore. If you have an XML feed (whatever it is old or new format), trust it over the ZIM file. At least this is not a property of the feed to decide if we want to take the metadata from the ZIM file or from the feed.

What is sure is that you should not have code reparsing and rechecking the data of the ZIM against the data of the feed.

test/data/library.opds
kelson42: I care only about the respecting the Atom/OPDS norm as much as possible. Here it seems we should use `link` for online HTTPS illustrations and `thumbnail` for embeded ones.

@veloman-yunkan I'm glad if you can put here a link to the comment about base64 and link... I don't know where it is.

veloman-yunkan: @kelson42 Looks like it was expressed only in live discussion with @hamazaspavetisyan 

But I want to clarify that I am fine with using `<link>` for thumbnails provided via a URL. My [comment above](https://github.com/kiwix/libkiwix/pull/1319#discussion_r3734987248) was aimed at (base64-encoded) image data directly embedded in (enhanced) OPDS.

hamazaspavetisyan: Very good, i will go with that path

kelson42: @veloman-yunkan Out of curiosity and for the record, can you elaborate here why `link` is not appropriate  to have a base64 content... and why `thumbnails` is better?

veloman-yunkan: @kelson42 For the same reason that `<link>` is not appropriate for presenting in our OPDS output extra info about our "books" via non-OPDS tags `<mediaCount>`, `<articleCount>`, `<flavour>`, etc. It could have been done, for example, as follows:

```XML
<link rel="http://kiwix.org/zim/flavour" flavour="maxi" />
<link rel="http://kiwix.org/zim/count" type="media" value="123" />
<link rel="http://kiwix.org/zim/count" type="article" value="99" />
```

I am sure that no one even thought of abusing `<link>` in such a way. I wonder why this option is at the table for thumbnail data.

test/manager.cpp
```diff
@@ -80,6 +104,91 @@ TEST(ManagerTest, readXml)
...
+    )");
+
+    {
+      auto lib = kiwix::Library::create();
+      TestManager manager(lib);
+      EXPECT_TRUE(manager.parseXmlDom(doc, /*readOnly=*/false, LIB_ABS_PATH, /*trustLibrary=*/true));
+
+      EXPECT_EQ(lib->getBooksIds(), (kiwix::Library::BookIdCollection{"book1", "book2"}));
+      EXPECT_FALSE(lib->getBookById("book1").readOnly());
+      EXPECT_FALSE(lib->getBookById("book2").readOnly());
```
veloman-yunkan: Please check a little more fields of the book objects. In particular the resolved path of the book files is of special interest.

test/manager.cpp
```diff
@@ -80,6 +104,91 @@ TEST(ManagerTest, readXml)
...
+    )");
+
+    EXPECT_TRUE(manager.parseXmlDom(doc, true, "./test/lib.xml", /*trustLibrary=*/false));
+
+    const auto ids = lib->getBooksIds();
+    ASSERT_EQ(ids.size(), 1U);
+    kiwix::Book book = lib->getBookById(ids[0]);
+    EXPECT_TRUE(book.isPathValid());
+    EXPECT_NE(book.getTitle(), "Stale title from XML");
+}
```
veloman-yunkan: `truslLibrary` is going to be dropped. Let's not test it.

test/manager.cpp
```diff
@@ -188,6 +189,91 @@ TEST(ManagerTest, parseXmlDomWithTrustLibraryFalseAndValidPathAdoptsZimMetadata)
...
+    TestManager manager(lib);
+
+    const XMLDoc doc(sampleOpdsFeed);
+
+    EXPECT_TRUE(manager.parseOpdsDom(doc, "http://example.com"));
+
+    EXPECT_TRUE(manager.m_hasSearchResult);
+    EXPECT_EQ(manager.m_totalBooks, 2U);
+    EXPECT_EQ(manager.m_startIndex, 0U);
+    EXPECT_EQ(manager.m_itemsPerPage, 2U);
```
veloman-yunkan: I don't know how important these fields of `kiwix::Manager` are, but if you decided to test that they are initialized correctly by `parseOpdsDom()` then you better use somewhat different values for them (`totalBooks` ≠ `itemsPerPage`≠ count of books in the OPDS, `startIndex` ≠ 0)

test/manager.cpp
```diff
@@ -3,9 +3,33 @@
...
+
+// XMLDoc helper needs to build a pugi::xml_document from a string before
+// handing it to parseXmlDom
+struct XMLDoc : pugi::xml_document
+{
+  explicit XMLDoc(const std::string& xml)
+  {
+    load_buffer(xml.c_str(), xml.size());
+  }
+};
```
veloman-yunkan: Please test the public method `readXml()` instead. Then you won't need these. Don't forget to update the commit message.

test/manager.cpp
```diff
@@ -188,6 +189,91 @@ TEST(ManagerTest, parseXmlDomWithTrustLibraryFalseAndValidPathAdoptsZimMetadata)
...
+)";
+
+TEST(ManagerTest, parseOpdsDomAddsEntriesAndParsesSearchMetadata)
+{
+    auto lib = kiwix::Library::create();
+    TestManager manager(lib);
+
+    const XMLDoc doc(sampleOpdsFeed);
+
+    EXPECT_TRUE(manager.parseOpdsDom(doc, "http://example.com"));
```
veloman-yunkan: Please test the public method `readOpds()` instead. Don't forget to update the test-case names and the commit message.

veloman-yunkan (commented):
I guess you have been somewhat disoriented by the fact that `Manager::readFile()` directly calls `Manager::parseOpdsDom()` instead of `Manager::readXml()`. Let's first reimplement  `Manager::readFile()` via `Manager::readXml()` (the latter has to be fixed to stop always returning `true`). Then continue from such a starting point

src/manager.cpp
```diff
@@ -138,7 +137,7 @@ bool Manager::readXml(const std::string& xml,
     this->parseXmlDom(doc, readOnly, libraryPath, trustLibrary);
   }
 
-  return true;
+  return !!result;
```
veloman-yunkan: This may be confusing (one may miss one of the `!`). 

Besides the return value of `parseXmlDom()` is ignored (the fact that the latter always returns `true` doesn't matter in the scope of this function). So better rewrite as

```
if ( result ) {
  return parseXmlDom(...);
} else {
  return false;
}

File: src/manager.cpp [hunk 14]
```diff
@@ -191,20 +190,8 @@ bool Manager::readFile(
...
-#else
-  pugi::xml_parse_result result = doc.load_file(path.c_str());
-#endif
-
-  if (result) {
-    this->parseXmlDom(doc, readOnly, path, trustLibrary);
-  } else {
-    retVal = false;
-  }
+  const std::string xml = getFileContent(path);
```
veloman-yunkan: What happens if the file doesn't exist?

test/manager.cpp
```diff
@@ -87,6 +87,95 @@ TEST(ManagerTest, readXml)
     EXPECT_EQ(678U*1024, book.getSize());
 }
 
+const char sampleOpdsFeed[] = R"(
```
veloman-yunkan, line 242: 1. The description of this commit (one titled "Enhance readOpds() tests") reads like a mystery to anyone unaware how it came to be that way :)

2. Currently `sampleOpdsFeed` is used in a single test. The other tests define the OPDS string inside the test function. Why is `sampleOpdsFeed` defined globally then?

test/manager.cpp
```diff
@@ -87,6 +87,95 @@ TEST(ManagerTest, readXml)
...
+<feed xmlns="http://www.w3.org/2005/Atom"
+      xmlns:opds="https://specs.opds.io/opds-1.2">
+  <totalResults>2</totalResults>
+  <startIndex>7</startIndex>
+  <itemsPerPage>100</itemsPerPage>
```
veloman-yunkan: These numbers are removed from reality. If there are total 2 results, a query requesting results starting from 7th would return nothing. Also please don't use the number 2 here (so that it cannot be confused with the count of entries below).

include/manager.h
```diff
@@ -104,15 +105,18 @@ class Manager
                bool trustLibrary = true);
 
   /**
-   * Load a library content stored in a OPDS stream.
+   * Load a library content stored in a OPDS stream/file.
```
veloman-yunkan: 1. This change may be confusing. `readOpds()` doesn't read from a file. Please make it clear that you mean OPDS content representing a local library file.

2. Regarding the commit message - please refer to `readOpds()` instead of `parseOpdsDom()` and also elaborate (in the long description) what you mean by "library file"

include/manager.h
```diff
@@ -104,15 +105,18 @@ class Manager
...
    *
    * @param content The content of the OPDS stream.
+   * @param urlHost The host used to resolve relative acquisition/thumbnail
+   *                links.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
```
veloman-yunkan, line 112: The description of `readOnly` doesn't make sense without library path. It's obvious that the documentation of (the old version of) `readOpds()` has been copied&pasted&not-updated from `readXml()`. Let's correct the documentation in the commit introducing the unit-tests for `readOpds()` and then update it as new parameters are added.

test/manager.cpp
```diff
@@ -87,6 +87,95 @@ TEST(ManagerTest, readXml)
...
+    <title>Book Two</title>
+    <link rel="http://opds-spec.org/acquisition/open-access"
+          type="application/x-zim"
+          href="https://example.com/book2.zim"
+          length="222" />
+  </entry>
+</feed>
+)";
+
+TEST(ManagerTest, readOpdsWithNoEntriesReturnsTrue)
```
veloman-yunkan, line 178: Please also test `readOpds()` on malformed input (check that it doesn't create any book objects even if the input starts as valid OPDS).

src/manager.cpp
```diff
@@ -31,6 +31,32 @@
...
+ *         FileFormat::XML otherwise.
+ */
+FileFormat detectFormat(const std::string& libraryPath)
+{
+  auto format
+      = (kiwix::getFileContent(libraryPath).find("<feed") != std::string::npos)
+            ? FileFormat::OPDS
+            : FileFormat::XML;
+  return format;
+}
```
veloman-yunkan: Running format detection on a file (rather than on its content) results on the file being loaded twice in `readFile()`

test/data/library.opds
```diff
@@ -0,0 +1,106 @@
...
+                                 data, so favicon's placeholder text becomes
+                                 a placeholder href instead)
+-->
+<feed xmlns="http://www.w3.org/2005/Atom"
+      xmlns:opds="https://specs.opds.io/opds-1.2">
+  <id>7f8a2b1e-0000-4000-8000-000000000000</id>
+  <title>Test OPDS library</title>
+  <totalResults>4</totalResults>
+  <startIndex>0</startIndex>
+  <itemsPerPage>4</itemsPerPage>
```
veloman-yunkan: I don't think that these should be present in library OPDS files (especially that they are not standard OPDS).

test/manager.cpp
```diff
@@ -190,6 +190,101 @@ TEST(ManagerTest, readOpdsAddsEntriesAndParsesSearchMetadata)
...
+TEST(ManagerTest, readFileDetectsOpdsFormat)
+{
+    // library.opds contains a "<feed" element, so it's routed through
+    // parseOpdsDom() instead. Note readFile() always passes an empty
+    // urlHost to parseOpdsDom(), unlike readOpds() which forwards the
+    // caller-supplied one - so any relative link in the feed is left as-is.
+    auto lib = kiwix::Library::create();
+    kiwix::Manager manager(lib);
+
+    EXPECT_TRUE(manager.readFile("./test/library.opds", /*readOnly=*/false, /*trustLibrary=*/true));
```
veloman-yunkan: Any reason why the `trustLibrary` parameter is provided in this test but not in `readFileDetectsXmlFormat()`?

test/manager.cpp
```diff
@@ -190,6 +190,101 @@ TEST(ManagerTest, readOpdsAddsEntriesAndParsesSearchMetadata)
     EXPECT_FALSE(book1.readOnly());
 }
 
+TEST(ManagerTest, readFileDetectsXmlFormat)
+{
+    // readFile() sniffs the file content for a "<feed" substring to tell
+    // OPDS files apart from XML library files (see detectFormat() in
+    // manager.cpp). library.xml contains no such substring, so it's routed
+    // through parseXmlDom(), which - unlike the OPDS branch - resolves
```
veloman-yunkan: Please don't refer to `parse*Dom()` methods unless really necessary.

src/book.cpp
veloman-yunkan: This works under the assumption that the URL of the illustration is an absolute path.

src/manager.cpp
```diff
@@ -201,14 +202,15 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc,
 
 bool Manager::readOpds(const std::string& content,
                        const std::string& urlHost,
-                       bool readOnly)
+                       bool readOnly,
+                       const std::string& libraryPath)
```
veloman-yunkan: I wonder if we can safely combine `urlHost` and `baseDir` into a single  (`std::string`) parameter `contentOriginUri` that will work for all scenarios.

When dealing with an OPDS feed generated by `https://library.kiwix.org/catalog/v2/entries`, `readOpds(opdsContent, /*contentOriginUri=*/ "https://library.kiwix.org/catalog/v2/entries")` should automatically deduce `urlHost="https://library.kiwix.org/"` and set `baseDir` to empty.

When dealing with a local OPDS library file "/data/library.opds" `readOpds(opdsContent, /*contentOriginUri=*/ "/data/library.opds")` should set `urlHost` to empty and `baseDir` to `"/data"`. This is based on the assumption that all URLs in a local library file should be full URLs (and therefore no need to be amended with `urlHost`).

src/book.cpp
```diff
@@ -227,10 +228,16 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
...
+      m_illustrations.push_back(favicon);
     }
- }
+  }
 
+  for (auto thumbnailNode = node.child("thumbnails").child("thumbnail"); thumbnailNode;
+           thumbnailNode = thumbnailNode.next_sibling("thumbnail")) {
+    const auto thumbnail = std::make_shared<Illustration>();
+    thumbnail->data = base64_decode(thumbnailNode.child_value());
+    m_illustrations.push_back(thumbnail);
```
veloman-yunkan, line 273: This leaves the illustration object's dimensions (width & height) at their default values of 48. The same bug is present in `Book::updateFromXml()`

veloman-yunkan (changes requested):

src/manager.cpp
hamazaspavetisyan: function just returns an empty

veloman-yunkan: What's the end result? Did the old implementation return `false` in that case and the new one will return `true`?

src/tools/pathTools.cpp
```diff
@@ -386,7 +399,11 @@ std::string kiwix::getExecutablePath(bool realPathOnly)
   return std::string(binRootPath);
 #else
   char binRootPath[PATH_MAX];
-  ssize_t size = readlink("/proc/self/exe", binRootPath, PATH_MAX);
+  // /proc/self/exe is a kernel-maintained symlink private to this process
+  // (not an attacker-controllable filename to race), and the resulting
+  // string below is bounded by the `size` readlink() returns, not by
+  // (missing) NUL-termination.
+  ssize_t size = readlink("/proc/self/exe", binRootPath, PATH_MAX); // Flawfinder: ignore
```
hamazaspavetisyan: I added this because automatic tool was complaining

src/manager.cpp
hamazaspavetisyan: I have created a commit, the last one in the PR list, please look at it. It covers what you proposed. We can keep it as a separate commit or I can fix it up.

test/manager.cpp
```diff
@@ -87,6 +87,120 @@ TEST(ManagerTest, readXml)
...
+    EXPECT_TRUE(manager.readOpds(R"(<feed xmlns="http://www.w3.org/2005/Atom"></feed>)", "http://example.com"));
+    EXPECT_TRUE(lib->getBooksIds().empty());
+}
+
+TEST(ManagerTest, readOpdsWithMalformedInputAddsNoBooks)
+{
+    // The first <entry> is well-formed, but the second's <entry> tag is
+    // never closed, making the feed as a whole malformed - doc.load_buffer()
+    // fails, parseOpdsDom() is never invoked, and no books get added at all
+    // (not even the first, valid entry).
```
veloman-yunkan: Don't explain the implementation.

test/manager.cpp
```diff
@@ -87,6 +87,120 @@ TEST(ManagerTest, readXml)
...
+    EXPECT_FALSE(manager.readOpds(feed, "http://example.com"));
+    EXPECT_TRUE(lib->getBooksIds().empty());
+}
+
+TEST(ManagerTest, readOpdsWithoutSearchMetadata)
+{
+    // strtoull() on the empty string returned for missing <totalResults> /
+    // <startIndex> / <itemsPerPage> elements yields 0 without throwing, so
+    // the try/catch in parseOpdsDom() never actually observes an error here
+    // - m_hasSearchResult ends up true, with all counters at 0.
```
veloman-yunkan: Move this down (just above the related checks).

test/manager.cpp
```diff
@@ -87,6 +87,120 @@ TEST(ManagerTest, readXml)
...
+{
+    auto lib = kiwix::Library::create();
+    kiwix::Manager manager(lib);
+
+    const std::string feed = R"(
+<feed xmlns="http://www.w3.org/2005/Atom"
+      xmlns:opds="https://specs.opds.io/opds-1.2">
+  <totalResults>2</totalResults>
+  <startIndex>7</startIndex>
+  <itemsPerPage>100</itemsPerPage>
```
veloman-yunkan: https://github.com/kiwix/libkiwix/pull/1319#discussion_r3764786043 has not been addressed

include/manager.h
```diff
@@ -104,15 +105,18 @@ class Manager
...
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& urlHost,
+                bool readOnly = false);
```
veloman-yunkan, line 122: https://github.com/kiwix/libkiwix/pull/1319#discussion_r3764867934 has not been addressed

src/manager.cpp
```diff
@@ -171,14 +173,16 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
...
+                       bool readOnly)
 {
   pugi::xml_document doc;
   pugi::xml_parse_result result
       = doc.load_buffer((void*)content.data(), content.size());
 
   if (result) {
-    this->parseOpdsDom(doc, urlHost);
+    this->parseOpdsDom(doc, urlHost, readOnly);
     return true;
```
veloman-yunkan: Return the value of `parseOpdsDom()`

src/manager.cpp
```diff
@@ -198,8 +224,15 @@ bool Manager::readFile(
...
-  const bool retVal = this->readXml(xml, readOnly, path, trustLibrary);
+  const std::string content = getFileContent(path);
+  bool retVal;
+
+  FileFormat format = detectFormat(content);
+  if (format == FileFormat::OPDS) {
+    retVal = this->readOpds(content, "", readOnly);
+  } else {
+    retVal = this->readXml(content, readOnly, path, trustLibrary);
+  }
```
veloman-yunkan: This can be rewritten to comply with the immutable local variables philosophy of defensive programming by using the conditional operator instead:

```c++
const bool retVal = detectFormat(content) == FileFormat::OPDS
                  ? readOpds(...)
                  : readXml(...);
```

It is also more concise.

src/manager.cpp
```diff
@@ -188,7 +189,7 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc,
     kiwix::Book book;
 
     book.setReadOnly(readOnly);
-    book.updateFromOpds(entryNode, urlHost);
+    book.updateFromOpds(entryNode, urlHost, removeLastPathElement(libraryPath));
```
veloman-yunkan: Compute the base dir outside the loop

test/book.cpp
```diff
@@ -326,6 +326,27 @@ TEST(BookTest, updateFromOPDSThumbnailWithAbsoluteHrefIgnoresUrlHostTest)
...
+              href="/zara-96.png" />
+    )");
+
+    const auto& illustrations = book.getIllustrations();
+    ASSERT_EQ(illustrations.size(), 2U);
+    EXPECT_EQ(illustrations[0]->url, "/zara-48.png");
+    EXPECT_EQ(illustrations[0]->width, 48);
+    EXPECT_EQ(illustrations[1]->url, "/zara-96.png");
+    EXPECT_EQ(illustrations[1]->width, 96);
+}
```
veloman-yunkan: I think that we can extract this change (with all of the enhancements on which it depends) into a small bugfix PR that can be merged independently.

src/tools/imageTools.cpp
```diff
@@ -0,0 +1,127 @@
...
+        return false;
+      }
+      height = readUint16BE(data, pos + 3);
+      width = readUint16BE(data, pos + 5);
+      return true;
+    }
+    pos += segmentLength;
+  }
+  return false;
+}
```
veloman-yunkan: I am not sure that we support GIF and JPEG thumbnails. Need to discuss.

src/book.cpp
```diff
@@ -246,6 +247,14 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       m_illustrations.push_back(favicon);
     }
   }
+
+  for (auto thumbnailNode = node.child("thumbnails").child("thumbnail"); thumbnailNode;
+           thumbnailNode = thumbnailNode.next_sibling("thumbnail")) {
+    const auto thumbnail = std::make_shared<Illustration>();
+    thumbnail->data = base64_decode(thumbnailNode.child_value());
+    getImageDimensions(thumbnail->data, thumbnail->width, thumbnail->height);
```
veloman-yunkan: Sorry to bring this up late. I just recalled https://github.com/openzim/libzim/issues/937 and figured out that there may be more to a thumbnail than just its width and height. Maybe we should provide that data via attributes of the "<thumbnail>" tag: `<thumbnail width="48" height="48" scale="2">...</thumbnail>`. Need to discuss.

test/manager.cpp
hamazaspavetisyan: not sure what you mean

test/book.cpp
hamazaspavetisyan: Potentially we can, but it will take another day to merge the PR

src/book.cpp
hamazaspavetisyan: Why to provide them if we can understand that runtime ? Plus that is nice to have feature that was discussed in 2024, we can open a new issue on this and start implementation. By the way why we do not consider vector graphics like SVG ?

test/manager.cpp
veloman-yunkan: This comment explains why `m_hasSearchResult` happens to be set to `true`. So it must appear just above the check of `m_hasSearchResult`

test/book.cpp
veloman-yunkan: But it will be merged much earlier than this one and will make this PR smaller. Moving in small steps is preferable to making big strides.

src/book.cpp
veloman-yunkan: > Why to provide them if we can understand that runtime ?

The target scale (`devicePixelRatio`) of an illustration cannot be derived from the image data (unless it is contained in the metadata).

> By the way why we do not consider vector graphics like SVG ?

PNG seems to be the standard format for thumbnails. SVG may result in rendering artifacts for small sizes, PNG provides a guaranteed result.

test/manager.cpp
```diff
@@ -87,6 +87,112 @@ TEST(ManagerTest, readXml)
...
+        <entry>
+          <id>urn:uuid:book1</id>
+          <title>Book One</title>
+        </entry>
+      </feed>
+    )";
+
+    EXPECT_TRUE(manager.readOpds(feed, "http://example.com"));
+
+    EXPECT_TRUE(manager.m_hasSearchResult);
```
veloman-yunkan: You lost the comment explaining this behaviour. I personally am fine with that (tests shouldn't care about the implementation). But a note that this looks like a bug is due.

veloman-yunkan (commented):
Please extract the first three commits into a separate PR which we can merge immediately.

While doing so, please remove the long description from the commit titled "Enhance readOpds() tests"

test/manager.cpp
hamazaspavetisyan: I just open a new PR with that bugfix. I can move that fix also into current PR.

New PR is here: https://github.com/kiwix/libkiwix/pull/1326

include/manager.h
```diff
@@ -104,15 +105,16 @@ class Manager
...
-   *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
+   * @param urlHost The host used to resolve relative acquisition/thumbnail
+   *                links.
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& urlHost,
+                bool readOnly = false);
```
veloman-yunkan, line 122: Please fix/update the documentation of (the current/old version of) `readOpds()` in a separate commit (as the first commit of this PR).

test/manager.cpp
```diff
@@ -190,54 +190,70 @@ TEST(Manager, reload)
...
+  <entry>
+    <id>urn:uuid:book2</id>
+    <title>Book Two</title>
+    <link rel="http://opds-spec.org/acquisition/open-access"
+          type="application/x-zim"
+          href="https://example.com/book2.zim"
+          length="222" />
+  </entry>
+</feed>
+)";
```
veloman-yunkan, line 265: Make this change (introducing `sampleOpdsFeed`) a separate commit.

test/manager.cpp
```diff
@@ -122,30 +122,30 @@ TEST(ManagerTest, readXmlInvalid)
 
 TEST(ManagerTest, readXmlNotXml)
 {
-    auto lib = kiwix::Library::create();
-    kiwix::Manager manager = kiwix::Manager(lib);
+  auto lib = kiwix::Library::create();
+  kiwix::Manager manager = kiwix::Manager(lib);
```
veloman-yunkan: Avoid unnecessary whitespace changes. If a significant amount of reformatting has to be done, please do it in a separate commit.

test/manager.cpp
```diff
@@ -280,4 +280,123 @@ TEST(ManagerTest, readOpdsWithoutSearchMetadata)
...
+          "charlesray",
+          "inaccessiblezim",
+          "raycharles",
+          "raycharles_uncategorized"
+    }));
+
+    kiwix::Book book = lib->getBookById("raycharles");
+    EXPECT_EQ(book.getPath(),
+              kiwix::computeAbsolutePath(kiwix::removeLastPathElement("./test/library.xml"),
+                                          "./zimfile_raycharles.zim"));
```
veloman-yunkan: This is too clumsy. Please introduce here (in the unit test file) a helper function `resolveAbsPath(basePath, relPath)`. Do it in a separate commit, replacing an existing usage of `kiwix::computeAbsolutePath(kiwix::removeLastPathElement())`. Then reuse it in this commit.

test/manager.cpp
```diff
@@ -280,4 +280,123 @@ TEST(ManagerTest, readOpdsWithoutSearchMetadata)
...
+        "raycharles_uncategorized"
+  }));
+
+  manager.reload({ "./test/library.xml" });
+  EXPECT_EQ(lib->getBooksIds(), kiwix::Library::BookIdCollection({
+        "charlesray",
+        "inaccessiblezim",
+        "raycharles",
+        "raycharles_uncategorized"
+  }));
```
veloman-yunkan: 1. This test case looks out-of-place in this commit. Since it tests `Manager::reload()` only on XML input it can be introduced early (before any changes are made to implementation). Then a sibling test case for OPDS version must be added.

include/book.h
```diff
@@ -70,7 +70,7 @@ class Book
   bool update(const Book& other);
   void update(const zim::Archive& archive);
   void updateFromXml(const pugi::xml_node& node, const std::string& baseDir);
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost);
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
```
veloman-yunkan, line 81: Please enhance `Book::updateFromOpds()` in a separate commit (decouple if from changes to `Manager`).

include/manager.h
```diff
@@ -110,11 +110,16 @@ class Manager
...
+   *                 updated content.
+   * @param libraryPath The library path (used to resolve a relative
+   *                     rel="self" link)
    * @return True if the content has been properly parsed.
    */
   bool readOpds(const std::string& content,
                 const std::string& urlHost,
-                bool readOnly = false);
+                bool readOnly = false,
+                const std::string& libraryPath = "");
```
veloman-yunkan: `readOnly` should have been documented earlier.

src/book.cpp
```diff
@@ -172,6 +171,15 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
            linkNode = linkNode.next_sibling("link")) {
     std::string rel = linkNode.attribute("rel").value();
 
+    if (rel == "self") {
+      std::string path = linkNode.attribute("href").value();
+      if (isRelativePath(path)) {
+        path = computeAbsolutePath(baseDir, path);
+      }
+      m_path = path;
```
veloman-yunkan: make `path` `const` & assign to `m_path` using a conditional expression

test/manager.cpp
```diff
@@ -31,10 +31,12 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
 # define UNITTEST_ZIM_PATH "zimfiles\\unittest.zim"
 # define LIB_ABS_PATH  "C:\\data\\lib.xml"
 # define ZIM_ABS_PATH  "C:\\data\\zimfiles\\unittest.zim"
+# define TEST_LIBRARY_OPDS_PATH ".\\test\\library.opds"
 #else
 # define UNITTEST_ZIM_PATH "zimfiles/unittest.zim"
 # define LIB_ABS_PATH "/data/lib.xml"
 # define ZIM_ABS_PATH "/data/zimfiles/unittest.zim"
+# define TEST_LIBRARY_OPDS_PATH "./test/library.opds"
```
veloman-yunkan: Inconsistency: for testing the XML mode an imaginary absolute path is used, while OPDS uses a relative path.

test/manager.cpp
```diff
@@ -256,6 +258,32 @@ TEST(ManagerTest, readOpdsAddsEntriesAndParsesSearchMetadata)
...
+      <feed xmlns="http://www.w3.org/2005/Atom">
+        <entry>
+          <id>urn:uuid:book1</id>
+          <title>Book From OPDS</title>
+          <link rel="self" href="does-not-exist.zim" />
+        </entry>
+      </feed>
+    )";
+
+  EXPECT_TRUE(manager.readOpds(feed, "http://example.com", /*readOnly=*/false, /*libraryPath=*/"./test/lib.xml"));
```
veloman-yunkan: `readOpds()` is invoked with the path of an XML file!

test/manager.cpp
```diff
@@ -323,13 +349,13 @@ TEST(ManagerTest, readFileDetectsOpdsFormat)
...
-    // library.opds is the OPDS analogue of library.xml - same books, with
-    // every XML attribute mapped to its corresponding OPDS element/link
-    // (see the comment at the top of library.opds). Check the "raycharles"
-    // entry field-by-field, the same way ManagerTest.readXml does for its
-    // XML-sourced equivalent.
     kiwix::Book book = lib->getBookById("raycharles");
-    EXPECT_EQ(book.getPath(), ""); // no path on OPDS
+
+    EXPECT_EQ(book.getPath(),
+            kiwix::computeAbsolutePath(kiwix::removeLastPathElement(TEST_LIBRARY_OPDS_PATH),
```
veloman-yunkan: Don't forget to reuse `resolveAbsPath()` here

src/book.cpp
```diff
@@ -187,7 +204,11 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
     if (rel == "http://opds-spec.org/image/thumbnail") {
       const auto favicon = std::make_shared<Illustration>();
       favicon->data.clear();
-      favicon->url = urlHost + linkNode.attribute("href").value();
+      std::string thumbnailUrl = linkNode.attribute("href").value();
+      if (!isAbsoluteUrl(thumbnailUrl)) {
+        thumbnailUrl = urlHost + thumbnailUrl;
+      }
+      favicon->url = thumbnailUrl;
```
veloman-yunkan: make `thumbnailUrl` `const` and assign to `favicon->url` using a condition expression.

src/book.cpp
```diff
@@ -32,6 +32,23 @@
...
+{
+/**
+ * Tells whether a URL string is already absolute (contains a scheme,
+ * e.g. "https://example.com/x.png") as opposed to being a relative path
+ * (e.g. "/x.png"). Only a relative URL should be prefixed with a base host
+ * or URL—doing so unconditionally would garble an already-absolute one.
+ */
+bool isAbsoluteUrl(const std::string& url)
+{
+  return url.find("://") != std::string::npos;
```
veloman-yunkan: This is not a reliable check. Add a TODO or XXX comment

test/book.cpp
```diff
@@ -259,6 +259,30 @@ TEST(BookTest, updateFromOPDSThumbnailWithAbsoluteHrefIgnoresUrlHostTest)
     EXPECT_EQ(illustration->url, "https://example.com/favicon/zara.png");
 }
 
+TEST(BookTest, updateFromOPDSMultipleBase64ThumbnailsTest)
+{
+    const kiwix::Book book = makeBookFromOpds(R"(
+        <thumbnails>
+          <thumbnail>Zmlyc3QtdGh1bWJuYWls</thumbnail>
+          <thumbnail>c2Vjb25kLXRodW1ibmFpbA==</thumbnail>
```
veloman-yunkan: Let's also support the following attributes for the `<thumbnail>` element:
- `width`
- `height`
- `mimetype`

Don't forget to update the commit description

src/book.cpp
```diff
@@ -47,6 +47,24 @@ bool isAbsoluteUrl(const std::string& url)
...
+ */
+std::string joinUrl(const std::string& host, const std::string& ref)
+{
+  if (host.empty()) {
+    return ref;
+  }
+  std::string h = host;
+  while (!h.empty() && h.back() == '/') {
+    h.pop_back();
+  }
```
veloman-yunkan: Discarding trailing slashes is in general wrong. Ideally, there should be no need to discard any slashes at all.

src/tools/pathTools.cpp
veloman-yunkan: This change - if it has to be preserved - must be in a separate commit.

include/tools.h
```diff
@@ -38,6 +38,16 @@ struct IpAddress
...
 
+/**
+ * The pieces needed to resolve references found in OPDS content, split out
+ * of the location that content was read from. See resolveContentOrigin().
+ */
+struct ContentOrigin
+{
+    std::string urlHost; /**< scheme+host[:port], no trailing slash; "" if the origin is a local path */
+    std::string baseDir; /**< parent directory of a local path; "" if the origin is a URL */
+};
```
veloman-yunkan, line 49: I don't think we need to preserve separate `urlHost` and `baseDir`. The seeming need for it comes from the current implementation, which tacitly assumes (and misbehaves otherwise) that `<link>` elements are allowed to contain only absolute path URLs. Suppose that the implementation is enhanced to support relative URLs too. Then you will have to replace `urlHost` with `baseUrl`. After that the next step will be supporting the `file://` scheme in `baseUrl`. Such an implementation path will require supporting resolution of relative URLs which is a serious enhancement on its own. Need to discuss how to proceed with this.

veloman-yunkan (changes requested):

include/manager.h
hamazaspavetisyan: I do not got why. I touched the method here so also updated its doc, why not ?

test/manager.cpp
hamazaspavetisyan: Not sure why, I changed the argument list of the method, then decided to add a test and reuse some parameter(sampleOpdsFeed). Why to move it to another commit ?

include/manager.h
veloman-yunkan: Because the documentation was wrong (as a result of a copy&paste&forget-to-update error). So it needs to be updated first so that updates to it caused by this change are easier to understand. Note that you have been misled by the wrong documentation too. When adding the `readOnly` parameter, you instead deleted its description in the documentation.

test/manager.cpp
veloman-yunkan: Small changes are easier to understand, review & validate. The combined change of moving a variable outside a function and introducing a new test between the old test and the new variable result in a bigger textual change. Another way to minimize is by adding the new test *after* the existing one.

hamazaspavetisyan: But in one of your previous comments, you requested not to introduce `sampleOpdsFeed` if it has a single usage. Plus if we move this to another commit, then we will have a commit with changes in arg list of `readOpds` and with no test cases to cover that arg introduction .

veloman-yunkan: I remember that and I still think the same way. Code should be of optimal shape at every state of the current implementation. When transitioning between stable states, intermediate states deviating from that requirement may be introduced in order to enable a smooth transition via a series of small enough changes. The commit message of such interim commits should explain the purpose of the change (that becomes fully justified only by the subsequent commit(s)).

include/manager.h
```diff
@@ -104,14 +105,16 @@ class Manager
...
    *
    * @param content The content of the OPDS stream.
    * @param urlHost The host used to resolve relative acquisition/thumbnail
    *                links.
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& urlHost,
+                bool readOnly = false);
```
veloman-yunkan, line 122: The new parameter must be mentioned in the documentation.

src/manager.cpp
```diff
@@ -181,15 +183,16 @@ bool Manager::parseOpdsDom(const pugi::xml_document& doc, const std::string& url
...
+                       bool readOnly)
 {
   pugi::xml_document doc;
   pugi::xml_parse_result result
       = doc.load_buffer((void*)content.data(), content.size());
 
   if (result) {
-    this->parseOpdsDom(doc, urlHost);
-    return true;
+    return this->parseOpdsDom(doc, urlHost, readOnly);
```
veloman-yunkan: I thought that this change was included in #1327

include/manager.h
```diff
@@ -110,11 +110,16 @@ class Manager
    * @param content The content of the OPDS stream.
    * @param urlHost The host used to resolve relative acquisition/thumbnail
    *                links.
+   * @param readOnly Set if the library path could be overwritten later with
+   *                 updated content.
```
veloman-yunkan, line 117: This should have been added in an earlier commit

test/manager.cpp
```diff
@@ -256,16 +269,35 @@ TEST(ManagerTest, readOpdsAddsEntriesAndParsesSearchMetadata)
...
     EXPECT_EQ(book1.getUrl(), "https://example.com/book1.zim");
-    // OPDS entries carry no local path at this point (unlike XML's "path"
-    // attribute - see ManagerTest.readXml above): this feed has no
-    // rel="self" link, so the resolved path stays empty and invalid.
+
     EXPECT_EQ(book1.getPath(), "");
     EXPECT_FALSE(book1.isPathValid());
-    // Unlike readXml() (readOnly defaults to true), OPDS-sourced books are
-    // always mutable.
+
```
veloman-yunkan: This comment should have been removed in the commit introducing the `readOnly` parameter of `readOpds()`

src/book.cpp
```diff
@@ -25,13 +25,47 @@
...
+
+  // Validate that all remaining characters in the scheme comply with RFC 3986
+  for (size_t i = 1; i < pos; ++i) {
+    char c = url[i];
+    if (!std::isalnum(static_cast<unsigned char>(c)) && c != '+' && c != '-' && c != '.') {
+      return false;
+    }
+  }
+
+  return true;}
```
veloman-yunkan: Misplaced closing brace

src/book.cpp
```diff
@@ -184,7 +218,8 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
     if (rel == "http://opds-spec.org/image/thumbnail") {
       const auto favicon = std::make_shared<Illustration>();
       favicon->data.clear();
-      favicon->url = urlHost + linkNode.attribute("href").value();
+      const std::string thumbnailUrl = linkNode.attribute("href").value();
+      favicon->url = isAbsoluteUrl(thumbnailUrl)? thumbnailUrl: urlHost + thumbnailUrl;
```
veloman-yunkan: Please add a `// XXX` comment documenting the assumption that a non-absolute URL is expected to be an absolute-path.

test/book.cpp
```diff
@@ -259,6 +259,55 @@ TEST(BookTest, updateFromOPDSThumbnailWithAbsoluteHrefIgnoresUrlHostTest)
     EXPECT_EQ(illustration->url, "https://example.com/favicon/zara.png");
 }
 
+TEST(BookTest, updateFromOPDSMultipleBase64ThumbnailsTest)
+{
+    const kiwix::Book book = makeBookFromOpds(R"(
+        <thumbnails>
+          <thumbnail width="96" height="96" mimetype="image/jpeg">Zmlyc3QtdGh1bWJuYWls</thumbnail>
```
veloman-yunkan: Use different values for width and height

veloman-yunkan (commented):
Didn't review the last two commits.

src/manager.cpp
hamazaspavetisyan: Than was for XML

test/manager.cpp
hamazaspavetisyan: I cleaned up all comments from test cases as I touched that file

veloman-yunkan:
Dropping this PR since both reading and writing of OPDS library files will be merged via #1321

Issue #1302, opened 2026-05-26
OPDS to provide multiple links per book


rgaudin:
Due to https://github.com/openzim/cms/issues/300 and the reliance of both ZIM and Metalink links by readers, we'd like to be able to get both an `application/x-zim` and an `application/metalink4+xml` link for our books

This is blocked by https://github.com/kiwix/libkiwix/issues/1309

rgaudin:
FYI here's the current readers usage of this information.

| reader | Using URL from | `.meta4` suffix handling | Multiple links safe? |
| --- | --- | --- | --- |
| `libkiwix` | Sets `downloadUrl` from link with `rel="http://opds-spec.org/acquisition/open-access"` | - | ❌ will use last `open-access` |
| PWA | Uses Link with both `rel="http://opds-spec.org/acquisition/open-access" type="application/x-zim"` | Treat `.meta4` as Metalink | ✅ will use first matching |
| Apple | `downloadUrl` from `libkiwix` | removes suffix if it exists | ❌ `libkiwix` |
| Android  | `downloadUrl` from `libkiwix` | removes suffix if it exists | ❌ `libkiwix` |
| Desktop | `downloadUrl` from `libkiwix` | passes it directly to aria2 or libtorrent | ⚠️ may not benefit from metalink/torrent |
| Cardshop | Uses Link with `type="application/x-zim"` | removes suffix if it exists | ✅ will use last `x-zim` |
| Seeder | Uses Link with `type="application/x-zim"` | removes suffix if it exists | ✅ will use last `x-zim` |

`libkiwix` will need to decide how to expose the other links and what goes into the `downloadUrl` (or rename this prop). Kiwix-desktop will have to adapt as it can download differently.

---

```xml
 <entry>
    <id>urn:uuid:00072115-bf0f-1e6e-e1cd-9d8c0751f7ed</id>
    <title>Klexikon – das Kinderlexikon</title>
    <updated>2026-05-09T00:00:00Z</updated>
    <summary>Klexikon ist die Wikipedia für Kinder.</summary>
    <language>deu</language>
    <name>klexikon_de_all</name>
    <flavour>nopic</flavour>
    <category></category>
    <tags>zum;_pictures:no;_videos:no;_details:yes;_ftindex:yes</tags>
    <articleCount>5659</articleCount>
    <mediaCount>73</mediaCount>
    <link rel="http://opds-spec.org/image/thumbnail"
[...]

PR #1321, opened 2026-08-06
Writing and Reading an OPDS library file

hamazaspavetisyan:
# Writing and Reading an OPDS library file

## Why

Addresses #1309 — "Allow to ingest directly an OPDS feed." The issue's stated
long-term goal is to eventually drop `library.xml` support entirely in favor
of OPDS, to simplify the format libkiwix has to maintain over time. This PR
gets a `Library` all the way to that goal for the local/offline case: it can
now read *and* write OPDS, so a library round-trips through OPDS alone with
no `library.xml` involved.

## What (implementation details)

- `Manager::readFile()` auto-detects `library.xml` vs. OPDS by content
  (sniffs for a `<feed` tag) and parses accordingly.
- `readOpds()` gained a `readOnly` parameter, and its origin argument now
  accepts a local filesystem path as well as a URL.
- `Book::updateFromOpds()` gained a `baseDir` parameter and now resolves an
  entry's `rel="self"` link into `Book::getPath()`, mirroring `updateFromXml()`.
- Added `Library::writeToOPDSFile()` / `dumpOpds()`. `writeToFile()` is now a
  deprecated alias for the renamed `writeToXMLFile()`.
- `fullEntryOpds()` gained `selfPath`/`isLiveCatalog` params: file dumps get a
  `rel="self"` link to the book's local path; the live catalog never does, to
  avoid leaking server filesystem paths.
- Thumbnail rendering was unified and made round-trip-capable: an
  illustration with an external URL renders a normal link; an embedded-only
  illustration renders as a `data:` URI (OPDS 1.2 §5.2.2) in file dumps and
  is decoded back on read; width/height/scale now round-trip through the
  link's `type` attribute.
- Moved `setOpenSearchInfo()` from `LibraryDumper` down to `OPDSDumper`
  (only meaningful for the live, paginated catalog).
- Done some cleanups and refactoring, dead code elimination.
- Tests parameterized across `.xml`/`.opds` for `Library`, `Manager`,
  `Server`, and search.

No source-breaking API changes. Concretely:

- `Library::writeToFile()` keeps working exactly as before (XML); switch to
  `writeToXMLFile()` to silence the new deprecation, or to
  `writeToOPDSFile()` to write OPDS instead.
- `Manager::readFile(path, readOnly, trustLibrary)` keeps its signature and
  now transparently accepts an OPDS file in addition to `library.xml` — no
  caller changes needed to gain OPDS support.
- A library round-tripped through `writeToOPDSFile()` → `readFile()` comes
  back with the same book paths (via `rel="self"`) and the same thumbnails
  (via direct URL or embedded `data:` URI), so OPDS is now a viable
  standalone replacement for `library.xml` for local use.
- **Known gap:** `downloadId` is still not persisted through OPDS
  read/write (flagged in review, not yet addressed) — a book's download
  bookkeeping doesn't survive an OPDS round-trip today.

## Fundamental change

`Library` (and the OPDS entry renderer it shares with the live catalog) is no
longer XML-only for serialization, and OPDS is no longer read-only. The same
`fullEntryOpds()` rendering path now serves both the live HTTP catalog and
local/offline file dumps.

## Notes

The known `downloadId` gap above is intentionally left as follow-up
rather than blocking this PR.


kelson42:
@hamazaspavetisyan AFAIK witting library.xml capacity is not fully dropped, right?

hamazaspavetisyan:
> @hamazaspavetisyan AFAIK witting library.xml capacity is not fully dropped, right?

Yes, as we discussed under previous PR, first we will have symmetric behavoiur for XML and ODPS, then will drop the XML.

src/library.cpp
```diff
@@ -447,6 +453,21 @@ bool Library::writeToFile(const std::string& path) const
...
+
+  LibOPDSDumper dumper(this);
+  std::string opds;
+  {
+    std::lock_guard<std::recursive_mutex> lock(m_mutex);
+    std::ostringstream oss;
+    dumper.dumpOPDSContent(allBookIds, oss);
+    opds = oss.str();
+  };
+  return writeTextFile(path, opds);
```
veloman-yunkan: Why don't you write directly to a file via `std::ofstream`?

test/book.cpp
```diff
@@ -329,6 +329,35 @@ TEST(BookTest, getHumanReadableIdFromPath)
...
+  EXPECT_EQ("abc",     opdsPath2HumanReadableId("âbç.zim"));
+  EXPECT_EQ("ancient", opdsPath2HumanReadableId("ancient.zimbabwe"));
+  EXPECT_EQ("ab_cd",   opdsPath2HumanReadableId("ab cd.zim"));
+#ifdef _WIN32
+  EXPECT_EQ("abc",     opdsPath2HumanReadableId("C:\\Data\\ZIM\\abc.zim"));
+#else
+  EXPECT_EQ("abc",     opdsPath2HumanReadableId("/Data/ZIM/abc.zim"));
+#endif
+  EXPECT_EQ("3plus2",  opdsPath2HumanReadableId("3+2.zim"));
+}
```
veloman-yunkan: I don't think that we need this unit test (at least in this form). The only thing that we need to test that the self path of a book is stored correctly in the book object when initializing it from OPDS input.

src/libopds_dumper.h
```diff
@@ -0,0 +1,74 @@
...
+  explicit LibOPDSDumper(const Library* library);
+  ~LibOPDSDumper();
+
+  /**
+   * Dump the library's OPDS content into the provided stream.
+   *
+   * @param bookIds the ids of the books to include.
+   * @param os the output stream to write the OPDS content into.
+   */
+  void dumpOPDSContent(const std::vector<std::string>& bookIds, std::ostream& os);
```
veloman-yunkan: I don't think that we need to filter the library when saving it.

src/libopds_dumper.h
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Set the library to dump.
+   *
+   * @param library The library to dump.
+   */
+  void setLibrary(const Library* library) { this->library = library; }
```
veloman-yunkan: I don't think we need this method. In fact I don't even think that we need a user-visible class for dumping a library. It looks like a free-function `void dumpLibraryOpds(const Library& lib, const std::string& outputPath)` is all we need as user facing API (where `baseDir` is derived from `outputPath`). Wait! `Library::writeToOPDSFile()` *is* that function. Why can't the OPDS export be implemented there directly? I guess, because we want that functionality to be unit testable. Then we can have `void dumpLibraryOpds(const Library& lib, const std::string& baseDir, std::ostream& out)` as a testable workhorse and the two-parameter version as a simple wrapper around it.

src/library.cpp
```diff
@@ -447,6 +453,21 @@ bool Library::writeToFile(const std::string& path) const
...
+bool Library::writeToOPDSFile(const std::string& path) const
+{
+  const auto allBookIds = getBooksIds();
+
+  LibOPDSDumper dumper(this);
+  std::string opds;
+  {
+    std::lock_guard<std::recursive_mutex> lock(m_mutex);
+    std::ostringstream oss;
+    dumper.dumpOPDSContent(allBookIds, oss);
```
veloman-yunkan: base dir is not set

veloman-yunkan (commented):
This was a quick superficial review. Take it with a grain of salt.

src/library.cpp
hamazaspavetisyan: @veloman-yunkan 
writeTextFile() isn't a plain ofstream wrapper — it opens the file via _wopen(Utf8ToWide(path), ...) on Windows so that UTF-8 library paths with non-ASCII characters work correctly (see pathTools.cpp:397). A raw std::ofstream(path) would go through the narrow-string/local-codepage path on Windows and break on those paths.

src/libopds_dumper.h
hamazaspavetisyan: @veloman-yunkan I mirrored this from dumpLibXMLContent, if we need symmetric behaviour, then having such argument is good idea

src/libopds_dumper.h
hamazaspavetisyan: @veloman-yunkan I agree with you, we can remove LibOPDSDumper entirely and keep only writer method that will be called from Library::writeToOPDSFile. But again I prefer to have symmetric structure for XML and OPDS, then clean up codes in scope of separate issue. What you think ?

test/library.cpp
```diff
@@ -1345,4 +1347,13 @@ TEST_P(LibraryTest, removeBooksNotUpdatedSince)
   );
 };
 
+TEST(LibraryOpdsExportTest, allInOne)
+{
+  auto lib = kiwix::Library::create();
+  kiwix::Manager manager(lib);
+  manager.readOpds(sampleLibraryOpds, OPDS_LIBRARY_PATH);
+
+  EXPECT_EQ(lib->dumpOpds(OPDS_LIBRARY_PATH), sampleLibraryOpds);
```
veloman-yunkan, line 1364: Please add a note (both here and above the definition of `sampleLibraryOpds`) that the test OPDS content is intentionally formatted in a way to stay invariant after being read and dumped back.

static/templates/catalog_v2_entry.xml
```diff
@@ -14,8 +14,7 @@
           href="{{root}}/catalog/v2/illustration/{{{id}}}/?size={{icon_size}}"
           type="{{icon_mimetype}}"/>
     {{/icons}}{{#contentAccessUrl}}<link type="text/html" href="{{contentAccessUrl}}/{{{content_id}}}" />
-    {{/contentAccessUrl}}
-    <author>
+    {{/contentAccessUrl}}<author>
```
veloman-yunkan, line 18: Let's do this in a separate commit.

test/opds_entry_rendering.cpp
```diff
@@ -213,6 +213,45 @@ TEST(FullEntryOpdsTest, rendersThumbnailLinkForBookIllustration)
...
+    "    </author>\n"
+    "    <publisher>\n"
+    "      <name>Some Publisher</name>\n"
+    "    </publisher>\n"
+    "    <dc:issued>2021-03-25T00:00:00Z</dc:issued>\n"
+    "    <link rel=\"self\" href=\"/local/path/book.zim\" type=\"application/x-zim\"/>\n"
+    "  </entry>\n"
+  );
+}
+
```
veloman-yunkan: Please merge this commit into the one introducing support for outputting self-links.

static/templates/catalog_entries.xml
```diff
@@ -10,5 +10,7 @@
   <itemsPerPage>{{itemsPerPage}}</itemsPerPage>
 {{/filter}}
   <link rel="self" href="" type="application/atom+xml" />
+{{#include_search_link}}
   <link rel="search" type="application/opensearchdescription+xml" href="{{root}}/catalog/searchdescription.xml" />
+{{/include_search_link}}
```
veloman-yunkan: Do we still need this change/commit?

src/library_dumper.cpp
```diff
@@ -31,28 +31,40 @@ std::string getIllustrationMimeTypeStr(const Book::Illustration& illustration)
...
+  for ( const auto& illustration : book.getIllustrations() ) {
+    // Live catalog: keep every icon (its own endpoint can serve embedded or
+    // remote data). File dump: keep only icons with an external url, since
+    // there is no server there to serve embedded data from.
+    if (!(isLiveCatalog || !illustration->url.empty())) {
+      continue;
     }
-    return illustrations;
+    const std::string iconSizeStr = to_string(illustration->width);
+        const std::string thumbnailUrl = isLiveCatalog
```
veloman-yunkan: Indentation is off

src/library_dumper.cpp
```diff
@@ -31,28 +31,40 @@ std::string getIllustrationMimeTypeStr(const Book::Illustration& illustration)
...
+        ? rootLocation + "/catalog/v2/illustration/" + kiwix::urlEncode(book.getId()) + "/?size=" + iconSizeStr
+        : illustration->url;
```
veloman-yunkan: Because of this, the commit doesn't qualify as a refactoring only change.

Please don't mention "refactoring" in the commit message and elaborate in its description the functional enhancement enabled by this change.

Also, previously, URL-encoding was not applied to book ids in catalog illustration URLs. For real book id values (as well as for those used in the unit tests) this didn't make any difference because they have been free of special characters. In any case, this also should be noted in the description of the commit.

test/opds_entry_rendering.cpp
```diff
@@ -261,6 +261,57 @@ TEST(FullEntryOpdsTest, omitsThumbnailLinkForEmbeddedOnlyIllustrationInFileDump)
...
+  Manager manager(lib);
+  const char sampleXML[] = R"(
+<library version="1.0">
+  <book
+        id="book-with-icon-url"
+        path="/local/path/book.zim"
+        title="Book With Icon Url"
+        favicon="AAAA"
+        faviconMimeType="image/png"
+        faviconUrl="/favicon.png"
```
veloman-yunkan, line 281: Use a full URL here.

src/library_dumper.cpp
```diff
@@ -31,28 +31,40 @@ std::string getIllustrationMimeTypeStr(const Book::Illustration& illustration)
...
 
-} // namespace
-
-kainjow::mustache::list getBookIllustrationInfo(const Book& book, bool isLiveCatalog)
+/**
+ * Get the illustration data (size/mimetype) of a book, for use in OPDS entry
+ * rendering.
+ */
+kainjow::mustache::list
+getBookIllustrationInfo(const Book& book, const std::string& rootLocation, bool isLiveCatalog = true)
```
veloman-yunkan: Looks like this function should now be hidden in the unnamed namespace.

static/templates/catalog_v2_entry.xml
```diff
@@ -10,10 +10,17 @@
...
-    {{#icons}}<link rel="http://opds-spec.org/image/thumbnail"
+    {{#linkIcons}}<link rel="http://opds-spec.org/image/thumbnail"
           href="{{icon_url}}"
           type="{{icon_mimetype}}"/>
-    {{/icons}}{{#contentAccessUrl}}<link type="text/html" href="{{contentAccessUrl}}/{{{content_id}}}" />
+    {{/linkIcons}}{{#base64IconsNotEmpty}}<thumbnails>
+      {{#base64Icons}}<thumbnail rel="http://opds-spec.org/image/thumbnail"
+                                 width="{{icon_width}}"
+                                 height="{{icon_height}}"
+                                 mimetype="{{icon_mimetype}}">{{icon_base64_data}}</thumbnail>
```
veloman-yunkan: Please calibrate the indention to be nice in the output. If you want it to be nice in this template as well, you can add a line break before `<thumbnail` - it is fine to have empty lines between `<thumbnail>` tags.

src/library_dumper.cpp
```diff
@@ -35,31 +36,41 @@ std::string getIllustrationMimeTypeStr(const Book::Illustration& illustration)
...
+      {"icon_url", thumbnailUrl},
+      {"icon_width", iconSizeWidth},
+      {"icon_height", to_string(illustration->height)},
+      {"icon_base64_data", thumbnailData.empty()? "" : base64_encode(thumbnailData)}
     });
   }
 
-  return illustrations;
+  return {linkBasedThumbnails, base64DataBasedThumbnails};
 }
```
veloman-yunkan: This function has become too convoluted. Let's have two separate functions for dataful and dataless illustrations.

test/opds_entry_rendering.cpp
```diff
@@ -312,6 +318,172 @@ TEST(FullEntryOpdsTest, rendersIllustrationUrlDirectlyInFileDump)
...
+)";
+  ASSERT_TRUE(manager.readOpds(sampleOpds, "http://root.location"));
+  const Book& book = lib->getBookById("book-with-remote-icon");
+  ASSERT_EQ(book.getIllustrations().size(), 1U);
+
+  kiwix::testing::CapturedStderr stderror;
+  fullEntryOpds(book, "http://root.location", "", "book-with-remote-icon");
+
+  EXPECT_EQ(std::string(stderror), "");
+  EXPECT_TRUE(book.getIllustrations().at(0)->getData().empty());
```
veloman-yunkan, line 347: This is a fragile way of testing the sought behaviour relying on the fact that `Book::Illustration::getData()` prints to `stderr` if it tries to download the thumbnail data. A more robust test would start an HTTP server and check that no request has been made to it. At least put a TODO or XXX comment about it here.

veloman-yunkan (changes requested):
We also need to persist `downloadId` in OPDS

src/library_dumper.cpp
hamazaspavetisyan: it is there

test/book.cpp
```diff
@@ -287,6 +287,75 @@ TEST(BookTest, updateFromOPDSThumbnailLinkTypeWithPartialSizeSuffixTest)
...
+              href="https://example.com/favicon.png" />
+        <link rel="http://opds-spec.org/image/thumbnail"
+              type="image/jpeg"
+              href="data:image/jpeg;base64,Zmlyc3QtdGh1bWJuYWls" />
+    )");
+
+    const auto& illustrations = book.getIllustrations();
+    ASSERT_EQ(illustrations.size(), 2U);
+    EXPECT_EQ(illustrations[0]->url, "https://example.com/favicon.png");
+    EXPECT_EQ(illustrations[0]->getData(), "");
```
veloman-yunkan: The getData() check passes because the GET request to https://example.com/favicon.png fails with 404 Not Found error. This unit test may start to fail for reasons out of our control.

src/book.cpp
```diff
@@ -302,8 +302,19 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       const auto favicon = std::make_shared<Illustration>();
       favicon->data.clear();
```
veloman-yunkan: I wonder why data is cleared for a newborn Illustration object.

veloman-yunkan (commented):
@hamazaspavetisyan Please update the description of this PR assuming that we will drop #1319 and will merge this one directly into main. While doing so provide elaborate information about user-observable effects/format changes/etc.

veloman-yunkan:
@kelson42 Please test this PR together with kiwix/kiwix-tools#836. Once you confirm that kiwix-manage and kiwix-serve work as expected, I will go through final review.

src/book.cpp
```diff
@@ -145,13 +262,12 @@ static std::string fromOpdsDate(const std::string& date)
 
 
 #define VALUE(name) node.child(name).child_value()
-void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost)
+void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir)
```
kelson42, line 273: Please document this method, this is not clear to me what it does and why we need now a new paramater.

hamazaspavetisyan: Done, as describe din the commit message, we are mirroring updateFromXml. baseDr is used for full path resolution.

include/manager.h
```diff
@@ -107,9 +107,8 @@ class Manager
    * Load a library content stored in a OPDS stream.
    *
    * @param content The content of the OPDS stream.
-   * @param readOnly Set if the library path could be overwritten later with
-   *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
+   * @param urlHost The host used to resolve relative acquisition/thumbnail
+   *                links.
```
veloman-yunkan: Please add the following sentence:

> It is assumed that those links come in the form of an absolute path component, so the full URL is obtained by prepending the `urlHost` string to the href value of the link.

include/manager.h
```diff
@@ -104,14 +105,18 @@ class Manager
...
    * @param urlHost The host used to resolve relative acquisition/thumbnail
    *                links.
+   * @param readOnly Set if the library path could be overwritten later with
+   *                 updated content.
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& urlHost,
+                bool readOnly = false);
```
veloman-yunkan: This is an ABI change. @kelson42 Are we allowed to break the libkiwix ABI with this PR? If not, then there is a simple workaround (introducing an overload of `readOpds()` with the new signature).

include/book.h
```diff
@@ -71,14 +71,14 @@ class Book
...
-   */
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
+ * Update the book's metadata from an OPDS entry XML node.
+ *
+ * @param node the `<entry>` node of an OPDS feed describing the book.
+ * @param urlHost host to prepend to relative illustration/thumbnail URLs.
+ * @param baseDir base directory used to resolve a relative `rel="self"`
+ *                link into an absolute book path.
+ */
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir);
```
veloman-yunkan, line 83: This is a public API change (on top of a previous public ABI change).

test/manager.cpp
```diff
@@ -312,7 +312,7 @@ TEST(ManagerTest, readOpdsWithInvalidSelfPath)
       </feed>
     )";
 
-  EXPECT_TRUE(manager.readOpds(feed, "http://example.com", /*readOnly=*/false, /*libraryPath=*/"./test/library.opds"));
+  EXPECT_TRUE(manager.readOpds(feed, "http://example.com", /*readOnly=*/false));
```
veloman-yunkan: The previous version had both `urlHost` and `libraryPath`. The purpose of this test was to test based on `libraryPath`, but it has been dropped. I guess you should have replaced the value of `urlHost` with the value of `libraryPath`.

test/library_server.cpp
```diff
@@ -198,7 +203,28 @@ TEST_F(LibraryServerTest, catalog_root_xml)
...
+  // of the library format
+  EXPECT_TRUE(kiwix::contains(r->body,
+    "<link rel=\"http://opds-spec.org/image/thumbnail\"\n"
+    "          href=\"/ROOT%23%3F/catalog/v2/illustration/raycharles/?size=48\"\n"
+    "          type=\"image/png;width=48;height=48;scale=1\"/>"
+  ));
+
+  EXPECT_FALSE(kiwix::contains(r->body, "illustration/charlesray/"));
+  EXPECT_FALSE(kiwix::contains(r->body, "illustration/inaccessiblezim/"));
+}
```
veloman-yunkan, line 220: Why was this test introduced in this commit?

src/library_dumper.h
```diff
@@ -40,11 +40,25 @@ kainjow::mustache::list getBookIllustrationInfo(const Book& book);
...
+ * @param book The book to render the OPDS entry for.
+ * @param rootLocation The root URL/path the catalog is served from, used to
+ *                 build absolute links within the entry.
+ * @param contentAccessUrl The URL (or URL prefix) used to access the book's
+ *                 content itself (as opposed to the catalog entry).
+ * @param contentId The identifier of the book's content, used when building
+ *                 content access links.
+ * @param selfPath If non-empty, rendered as the entry's rel="self" link (its
+ *                 local file path). Only meant for local/offline dumps
+ *                 (LibOPDSDumper) - leave empty for the live HTTP catalog
```
veloman-yunkan: Didn't we get rid of `LibOPDSDumper`?

test/opds_entry_rendering.cpp
```diff
@@ -217,6 +217,45 @@ TEST(FullEntryOpdsTest, rendersThumbnailLinkForBookIllustration)
...
+    "      <name>Some Creator</name>\n"
+    "    </author>\n"
+    "    <publisher>\n"
+    "      <name>Some Publisher</name>\n"
+    "    </publisher>\n"
+    "    <dc:issued>2021-03-25T00:00:00Z</dc:issued>\n"
+    "    <link rel=\"self\" href=\"/local/path/book.zim\" type=\"application/x-zim\"/>\n"
+    "  </entry>\n"
+  );
+}
```
veloman-yunkan: In these new (as well as in a number of old) test case the value of the `rootLocation` doesn't make a difference (because the used book object lacks the attributes that result in the generation of respective `<link>` tags). 

To the best of my understanding `rootLocation` should be empty when OPDS is saved for an offline/local library file. If so please update the `rendersSelfLinkWhenSelfPathIsSet` test case which seems to be the first one corresponding to the scenario of producing an OPDS library file.

include/library.h
```diff
@@ -470,6 +470,8 @@ class Library: public std::enable_shared_from_this<Library>
    */
   uint32_t removeBooksNotUpdatedSince(Revision rev);
 
+  std::string dumpOpds(const std::string& outputPath) const;
```
veloman-yunkan, line 509: Add doc comment

test/opds_entry_rendering.cpp
```diff
@@ -74,7 +74,6 @@ TEST(FullEntryOpdsTest, rendersCoreBookFields)
   EXPECT_EQ(fullEntryOpds(book, "http://root.location", /*contentAccessUrl=*/"", /*contentId=*/"book-id"),
     "  <entry>\n"
     CORE_ENTRY_BODY
-    "    \n"
```
veloman-yunkan, line 77: Changes like this one in this commit are an immediate consequence of the previous commit and must belong there.

include/library.h
```diff
@@ -361,13 +361,33 @@ class Library: public std::enable_shared_from_this<Library>
...
+   */
+  bool writeToXMLFile(const std::string& path) const;
+
+  /**
+   * Write the library to a file as an OPDS document.
+   *
+   * @param path the path of the file to write to.
+   * @return True if the library has been correctly saved.
+   */
+  bool writeToOPDSFile(const std::string& path) const;
```
veloman-yunkan: Please rename these to `writeAsXML()` and `writeAsOPDS()`

src/library_dumper.h
```diff
@@ -49,10 +49,9 @@ kainjow::mustache::list getBookIllustrationInfo(const Book& book);
  * @param contentId The identifier of the book's content, used when building
  *                 content access links.
  * @param selfPath If non-empty, rendered as the entry's rel="self" link (its
- *                 local file path). Only meant for local/offline dumps
- *                 (LibOPDSDumper) - leave empty for the live HTTP catalog
- *                 (OPDSDumper), which must not leak server-side filesystem
- *                 paths to remote clients.
+ *                 local file path). Only meant for local/offline dumps, leave
+ *                 empty for the live HTTP catalog (OPDSDumper), which must not
+ *                 leak server-side filesystem paths to remote clients.
```
veloman-yunkan: We shouldn't see this change here. The `selfPath` parameter should have been documented in this way in the commit where it was introduced.

test/libxml_dumper.cpp
```diff
@@ -184,9 +184,6 @@ TEST_F(LibXMLDumperTest, aliasBookWithOrigIdOnlyEmitsIdAndOrigId)
 
 TEST_F(LibXMLDumperTest, dumpLibXMLContentThrowsForUnknownBookId)
 {
-  // Unlike LibOPDSDumper::dumpOPDSContent (which silently skips book ids no
-  // longer present in the library), LibXMLDumper::dumpLibXMLContent uses
-  // Library::getBookById() directly and lets std::out_of_range propagate.
```
veloman-yunkan, line 189: We shouldn't see this change here. This comment shouldn't have been written at all (I mean, in the cleaned-up version of the PR).

src/library_dumper.cpp
```diff
@@ -91,13 +91,6 @@ LibraryDumper::~LibraryDumper()
 {
 }
 
-void LibraryDumper::setOpenSearchInfo(int totalResults, int startIndex, int count)
-{
-  m_totalResults = totalResults;
-  m_startIndex = startIndex,
-  m_count = count;
-}
-
```
veloman-yunkan, line 76: Please move this commit to the beginning of the PR

test/book.cpp
```diff
@@ -287,6 +287,74 @@ TEST(BookTest, updateFromOPDSThumbnailLinkTypeWithPartialSizeSuffixTest)
...
+TEST(BookTest, updateFromOPDSDataUriThumbnailWithoutCommaTest)
+{
+    const kiwix::Book book = makeBookFromOpds(R"(
+        <link rel="http://opds-spec.org/image/thumbnail"
+              type="image/png"
+              href="data:image/png;base64" />
+    )");
+
+    const auto illustration = book.getIllustrations().at(0);
+    EXPECT_EQ(illustration->getData(), "");
```
veloman-yunkan: I guess that (at least for testing purposes) we need a direct data accessor in `Book::Illustration` that doesn't attempt to download the data.

include/manager.h
```diff
@@ -104,15 +105,21 @@ class Manager
...
+   *        kiwix::resolveContentOrigin() for the exact splitting rules.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& contentOriginUri,
+                bool readOnly = false);
```
veloman-yunkan: This is an ABI change. @kelson42 Are we allowed to break the libkiwix ABI with this PR? If not, then there is a simple workaround (introducing an overload of `readOpds()` with the new signature).

include/tools.h
```diff
@@ -38,6 +38,16 @@ struct IpAddress
...
+/**
+ * The pieces needed to resolve references found in OPDS content, split out
+ * of the location that content was read from. See resolveContentOrigin().
+ */
+struct ContentOrigin
+{
+    std::string urlHost; /**< scheme+host[:port], no trailing slash; "" if the origin is a local path */
+    std::string baseDir; /**< parent directory of a local path; "" if the origin is a URL */
+};
+
```
veloman-yunkan: This must not be exposed in the public API of `libkiwix`. Move this and `resolveContentOrigin()` into src/tools/otherTools.{h,cpp}

src/tools/pathTools.cpp
```diff
@@ -228,6 +228,21 @@ std::string kiwix::removeLastPathElement(const std::string& path)
...
+    return { /*urlHost=*/"", removeLastPathElement(contentOriginUri) };
+  }
+
+  const auto hostEnd = contentOriginUri.find_first_of("/?#", schemeEnd + 3);
+  const auto urlHost = hostEnd == std::string::npos
+                      ? contentOriginUri
+                      : contentOriginUri.substr(0, hostEnd);
+
+  return { urlHost, /*baseDir=*/"" };
+}
```
veloman-yunkan: Move to `src/tools/otherTools.cpp`

src/library_dumper.cpp
```diff
@@ -1,38 +1,124 @@
...
+    const std::string dataUri = thumbnailData.empty() ? ""
+      : "data:" + illustration->mimeType + ";base64," + base64_encode(thumbnailData);
+    thumbnailLinks.push_back(kainjow::mustache::object{
+      {"icon_mimetype", getIllustrationMimeTypeStr(*illustration)},
+      {"icon_url", dataUri}
+    });
+  }
+
+  return thumbnailLinks;
 }
```
veloman-yunkan, line 101: Now that there is no `<thumbnail>` tag, it seems that these two functions no longer need to be separate - you can produce all the illustration "link"s with a single function.

test/opds_entry_rendering.cpp
```diff
@@ -217,6 +214,313 @@ TEST(FullEntryOpdsTest, rendersThumbnailLinkForBookIllustration)
...
+
+  const std::string rendered = fullEntryOpds(book, "http://root.location", "", "book-with-embedded-icon");
+
+  EXPECT_EQ(rendered.find("data:"), std::string::npos) << rendered;
+  EXPECT_NE(rendered.find(
+    "    <link rel=\"http://opds-spec.org/image/thumbnail\"\n"
+    "          href=\"http://root.location/catalog/v2/illustration/book-with-embedded-icon/?size=48\"\n"
+    "          type=\"image/png;width=48;height=48;scale=1\"/>\n"),
+    std::string::npos) << rendered;
+}
```
veloman-yunkan, line 538: Why do you check the output via `rendered.find()` in these tests? Why not simply comparing against the expected string?

test/pathTools.cpp
```diff
@@ -178,6 +178,50 @@ TEST(pathTools, removeLastPathElement)
...
+    ASSERT_EQ(origin.baseDir, "C:\\data");
+  }
+#endif
+  {
+    const auto origin = kiwix::resolveContentOrigin("");
+    ASSERT_EQ(origin.urlHost, "");
+    ASSERT_EQ(origin.baseDir, "");
+  }
+}
+
```
veloman-yunkan: Move to test/otherTools.cpp

veloman-yunkan (changes requested):
This was the last full review. If nothing pops up in the way the requested changes are implemented we will be good to merge.

The most important issue remaining is that of backward compatibility of the public API/ABI.

test/pathTools.cpp
hamazaspavetisyan: Not logical to have implementation in pathTools, test in otherTools. So I moved implementation to otherTools.cpp too.

src/library_dumper.cpp
hamazaspavetisyan: I don’t think so. Each function has its own distinct responsibility, even though they eventually push into the same list. Keeping them separate also makes the function bodies shorter and easier to read.

include/book.h
hamazaspavetisyan: Not sure what you mean

veloman-yunkan: `baseDir` was added in an earlier commit with a default value. That change preserved the public API but changed the ABI. Removing the default value breaks the public API too.

veloman-yunkan:
Please also rebase on recent `main` (your branch is out-of-date).

veloman-yunkan (commented):
LGTM, except that this PR contains backward incompatible API/ABI changes (which should be easy to avoid).

kelson42:
> LGTM, except that this PR contains backward incompatible API/ABI changes (which should be easy to avoid).

We should break ABI only when really necessary.

hamazaspavetisyan:
> LGTM, except that this PR contains backward incompatible API/ABI changes (which should be easy to avoid).

@veloman-yunkan , @kelson42 

backward compatibility issues are fixed.

include/book.h
```diff
@@ -70,7 +70,15 @@ class Book
...
-  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost);
+  /**
+   * Update the book's metadata from an OPDS entry XML node.
+   *
+   * @param node the `<entry>` node of an OPDS feed describing the book.
+   * @param urlHost host to prepend to relative illustration/thumbnail URLs.
+   * @param baseDir base directory used to resolve a relative `rel="self"`
+   *                link into an absolute book path.
+   */
+  void updateFromOpds(const pugi::xml_node& node, const std::string& urlHost, const std::string& baseDir = "");
```
veloman-yunkan: This one must be fixed too (because it breaks the ABI).

include/manager.h
```diff
@@ -104,15 +105,41 @@ class Manager
...
+  *        links) or the local filesystem path it was read from (used to
+  *        resolve a relative rel="self" link). See
+  *        kiwix::resolveContentOrigin() for the exact splitting rules. It is
+  *        assumed that URL links come in the form of an absolute path
+  *        component, so the full URL is obtained by prepending the urlHost
+  *        string to the href value of the link.
+  * @return True if the content has been properly parsed.
+  */
+  bool readOpds(const std::string& content,
+                const std::string& contentOriginUri);
```
veloman-yunkan: Please match the formatting of this piece of code (indentation of the doc comment,  function declaration) with that of `main` so that the diff shows that a new function with a third parameter was added while for this one only the doc comment was updated. Also explain that this function is a simple wrapper around the three parameter version preserved for backward compatibility reasons.

kelson42 (changes requested):
It seems we have still a problem, the writer newly writes something like:
```
    <link rel="http://opds-spec.org/acquisition/open-access" type="application/x-zim" href="dd" length="32846848" />
    <link rel="self" href="php.net_en_all_2024-08.zim" type="application/x-zim"/>
```

We have many problems here:
* First, for the moment, it is either local content or online content. Not both (see https://github.com/kiwix/libkiwix/issues/1302 for the support of multiple links). Looks like a regression against `library.xml`.
* Then the local version use `rel="self"` which is different than online `http://opds-spec.org/acquisition/open-access`, it should not because it will make the parsing of sources difficult.
* Finally the local version has not `length` attribute, which makes AFAIK impossible to know the size of the file, so a regression

kelson42:
@hamazaspavetisyan Once this PR merged, the code will be included in the readers... before having a kiwix-serve OPDS stream generated with it. Therefore it is really important the reader is backward compatible (still able to parser properly the current version of the OPDS stream). Please confirm, this is the case.

include/manager.h
```diff
@@ -104,15 +105,42 @@ class Manager
...
+  *        component, so the full URL is obtained by prepending the urlHost
+  *        string to the href value of the link.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
+  * @return True if the content has been properly parsed.
+  */
+  bool readOpds(const std::string& content,
+                const std::string& contentOriginUri,
+                bool readOnly);
```
veloman-yunkan: Reference to `urlHost` doesn't make sense in this context. Let's rather preserve that name for the second parameter of the other (legacy) overload and move the sentence about `urlHost` there.

include/manager.h
```diff
@@ -104,15 +105,42 @@ class Manager
...
+   * @param contentOriginUri Where content was read from: either the URL it
+   *        was fetched from (used to resolve relative acquisition/thumbnail
+   *        links) or the local filesystem path it was read from (used to
+   *        resolve a relative rel="self" link). See
+   *        kiwix::resolveContentOrigin() for the exact splitting rules.
    * @return True if the content has been properly parsed.
    */
-  bool readOpds(const std::string& content, const std::string& urlHost);
+  bool readOpds(const std::string& content,
+                const std::string& contentOriginUri);
```
veloman-yunkan: 1. Rename `contentOriginUri` to `urlHost` and don't mention `resolveContentOrigin()` in the doc comment (restore the original description of `urlHost`.
2. Write the function declaration on a single line (so that there is no diff with the old version).

include/manager.h
```diff
@@ -104,15 +105,42 @@ class Manager
...
+  *        resolve a relative rel="self" link). See
+  *        kiwix::resolveContentOrigin() for the exact splitting rules. It is
+  *        assumed that URL links come in the form of an absolute path
+  *        component, so the full URL is obtained by prepending the urlHost
+  *        string to the href value of the link.
    * @param readOnly Set if the library path could be overwritten later with
    *                 updated content.
-   * @param libraryPath The library path (used to resolve relative path)
+  * @return True if the content has been properly parsed.
+  */
```
veloman-yunkan: Fix the indentation of the comment margin (to match that of all other doc comments).

veloman-yunkan:
> It seems we have still a problem, the writer newly writes something like:
> 
> ```
>     <link rel="http://opds-spec.org/acquisition/open-access" type="application/x-zim" href="dd" length="32846848" />
>     <link rel="self" href="php.net_en_all_2024-08.zim" type="application/x-zim"/>
> ```

@kelson42 

I assume that you redacted the path of the local file and `href="dd"` doesn't mean that we have a bug there.


> 
> We have many problems here:
> 
> * First, for the moment, it is either local content or online content. Not both (see #1302 for the support of multiple links). Looks like a regression against `library.xml`.
> 

Consider the following command (using v3.7.0 of `kiwix-tools`):

```
$ kiwix-manage library.xml add test/data/zimfile.zim --url http://example.com/zimfile.zim
```

It creates a `library.xml` file with the following content which has both `path` and `url` attributes (I omitted all other attributes for readability):


```xml
<library version="20110515">
  <book path="test/data/zimfile.zim" url="http://example.com/zimfile.zim" .../>
</library>
```

The OPDS version corresponds to that case.


---

> * Then the local version use `rel="self"` which is different than online `http://opds-spec.org/acquisition/open-access`, it should not because it will make the parsing of sources difficult.

It's not difficult. It amounts to the following small piece of code:

```c++
    if (rel == "self") {
      const std::string path = linkNode.attribute("href").value();
      m_path = isRelativePath(path)? computeAbsolutePath(baseDir, path): path;
      m_pathValid = fileReadable(m_path);
    }
```
Usage of `rel="self"` instead of `rel="http://opds-spec.org/acquisition/open-access"` facilitates recognizing that the value of `href` must be used to set the path of book's ZIM file rather than its URL.



---

> 
> * Finally the local version has not `length` attribute, which makes AFAIK impossible to know the size of the file, so a regression

That is a valid concern.

hamazaspavetisyan:
@kelson42

1. `library.xml` allows us to specify both a URL and a path. Should we provide the same functionality for OPDS, or should the two options be mutually exclusive?
2. If we decide to support both, there is an additional question: which link should we use to get the size, since theoretically, both links may have a `size` attribute?

kelson42:
@veloman-yunkan @hamazaspavetisyan Thank you for clarifying a few points.

>     1. `library.xml` allows us to specify both a URL and a path. Should we provide the same functionality for OPDS, or should the two options be mutually exclusive?

Lets keep it like it is. I'm wrong here. They are not exclusive.

Still, please remove the `self` value. It should be `http://opds-spec.org/acquisition/open-access`. It should be trivial to list all sources. The differentiation of the protocol to use to access the content should be done based on the `href`, nothing else.
 
>     2. If we decide to support both, there is an additional question: which link should we use to get the size, since theoretically, both links may have a `size` attribute?

I guesse you mean the `lengh` attribute in OPDS? [Here is my opinion](https://github.com/kiwix/libkiwix/issues/1302#issuecomment-5467728594). `lenght` attribute should be everywhere. But I would be glad if - ultimately - we could save this is in a dedicated OPDS node. We do have nothing to do that?

hamazaspavetisyan:
@kelson42, @veloman-yunkan to satisfy all needs that we have with these urls, paths, I have an idea.

OPDS/Atom entries already support `xml:base`(RFC 4287 §2 lists it as a common attribute, legal on any element
including individual `<link>`s). That gives us a standards-based way to satisfy "differentiate by href, nothing else" without a dedicated`rel="self"`:

```xml
<entry>
  <link xml:base="file:///home/user/library/"
        rel="http://opds-spec.org/acquisition/open-access"
        type="application/x-zim" href="book.zim" length="123456" />
  <link xml:base="https://example.com/books/"
        rel="http://opds-spec.org/acquisition/open-access"
        type="application/x-zim" href="book.zim" length="123456" />
</entry>
```

Both links can share the identical relative href; the reader resolves each against its own xml:base and branches on the resulting scheme (file: -> local path, http(s): -> remote URL). Same `rel` everywhere, length on every link, and it extends cleanly to other source types (e.g. BitTorrent from #1302) without more special-cased `rel` values.
By the way, the `entity` tag supports the `xml:base` attribute too. This can be used for download URLs. In cases where both exist, we just need to define a clear precedence rule—such as prioritizing links over `entity`s when resolving their respective `xml:base` attributes.

kelson42:
@hamazaspavetisyan redundant and useless complexification. Just do the minimal parser to identify local/remote hrefs.

src/tools/archiveTools.cpp
```diff
@@ -105,10 +105,6 @@ bool getArchiveFavicon(const zim::Archive& archive, unsigned size,
   return false;
 }
 
-unsigned int getArchiveFileSize(const zim::Archive& archive) {
-  return archive.getFilesize() / 1024;
-}
-
```
veloman-yunkan, line 111: Make this change in a separate commit (and include the `BookTest.updateFromArchiveSetsByteExactSize` unit-test in it)

test/library.cpp
```diff
@@ -342,7 +342,7 @@ const char sampleLibraryOpds[] = R"(<feed xmlns="http://www.w3.org/2005/Atom"
       <name>Kiwix &amp; Some Enthusiasts</name>
     </publisher>
     <dc:issued>2021-04-11T00:00:00Z</dc:issued>
-    <link rel="self" href=")" EXAMPLE_PATH R"(" type="application/x-zim"/>
+    <link rel="self" href=")" EXAMPLE_PATH R"(" type="application/x-zim" length="0"/>
```
veloman-yunkan: Why 0???

src/book.cpp
```diff
@@ -309,7 +310,13 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
       }
       const std::string length = linkNode.attribute("length").value();
       if (!length.empty()) {
+        if (!lengthSetByHref.empty()) {
+          std::cerr << "Book '" << m_id << "': both acquisition links '"
+                    << lengthSetByHref << "' and '" << href
+                    << "' specify a length (" << length << ")." << std::endl;
+        }
         m_size = strtoull(length.c_str(), 0, 0);
+        lengthSetByHref = href;
```
veloman-yunkan: I think that the warning should be issued only if the two length values don't match.

test/book.cpp
```diff
@@ -164,6 +165,33 @@ TEST(BookTest, updateFromOPDSTwoAcquisitionLinksTest)
...
+    )");
+
+    kiwix::Book book;
+    kiwix::testing::CapturedStderr stderror;
+    book.updateFromOpds(opds.child("entry"), "http://who.org", DATA_ABS_PATH);
+
+    const std::string warning = stderror;
+    EXPECT_NE(warning.find("zara"), std::string::npos);
+    EXPECT_NE(warning.find("zara.zim"), std::string::npos);
+    EXPECT_NE(warning.find("https://who.org/zara.zim"), std::string::npos);
```
veloman-yunkan: Check for the full warning text.

src/library_dumper.cpp
```diff
@@ -135,10 +137,9 @@ std::string fullEntryOpds(const Book& book,
       {"media_count", to_string(book.getMediaCount())},
       {"author_name", book.getCreator()},
       {"publisher_name", book.getPublisher()},
-      {"url", onlyAsNonEmptyMustacheValue(book.getUrl())},
+      {"acquisition_href", onlyAsNonEmptyMustacheValue(acquisitionHref)},
       {"size", to_string(book.getSize())},
       {"thumbnailLinks", thumbnailLinks},
-      {"local_path", onlyAsNonEmptyMustacheValue(localPath)},
```
veloman-yunkan: I don't see why we need the commit titled "Emit a single acquisition link chosen from dump context". A library OPDS file may contain a download URL for a ZIM file (just like in the case of library.xml).

Update: I see that it was reverted in the next commit.

include/manager.h
```diff
@@ -104,16 +105,39 @@ class Manager
...
+  bool readOpds(const std::string& content,
+                const std::string& contentOriginUri,
+                bool readOnly);
 
+  /**
+   * Load a library content stored in a OPDS stream or a local library file.
+   *
+   * A simple wrapper around the three-parameter readOpds() above, kept for
+   * backward compatibility. Equivalent to calling
+   * readOpds(content, contentOriginUri, false).
```
veloman-yunkan: contentOriginUri → urlHost

veloman-yunkan (changes requested):

src/library_dumper.cpp
hamazaspavetisyan: I can fixup these both commits if you want

src/book.cpp
```diff
@@ -199,7 +199,7 @@ void Book::update(const zim::Archive& archive) {
   m_category = getCategoryFromTags();
   m_articleCount = archive.getArticleCount();
   m_mediaCount = archive.getMediaCount();
-  m_size = static_cast<uint64_t>(getArchiveFileSize(archive)) << 10;
+  m_size = archive.getFilesize();
```
veloman-yunkan, line 202: This should have gone in the commit that gets rid of `getArchiveFileSize()`. That commit must come before this one and its title must be "Got rid of getArchiveFileSize()".

test/book.cpp
```diff
@@ -164,6 +165,34 @@ TEST(BookTest, updateFromOPDSTwoAcquisitionLinksTest)
...
+              href="https://who.org/zara.zim"
+              length="222" />
+      </entry>
+    )");
+
+    kiwix::Book book;
+    kiwix::testing::CapturedStderr stderror;
+    book.updateFromOpds(opds.child("entry"), "http://who.org", DATA_ABS_PATH);
+
+    const std::string warning = stderror;
```
veloman-yunkan: Why do you need an extra variable?

test/book.cpp
```diff
@@ -164,6 +165,34 @@ TEST(BookTest, updateFromOPDSTwoAcquisitionLinksTest)
...
+    EXPECT_EQ(
+      "Book 'zara': both acquisition links 'zara.zim' and "
+      "'https://who.org/zara.zim' specify a length (222).\n",
+      warning);
```
veloman-yunkan: The warning text and the commit description need to be changed (since the warning is only issued when lengths don't match).

veloman-yunkan (changes requested):

src/library_dumper.cpp
veloman-yunkan: Yes, please drop those two commits.

veloman-yunkan (approved):

PR #1327, opened 2026-08-15
Reimplemented Manager::readFile() via Manager::readXml()

hamazaspavetisyan:
Refactor readFile() to use readXML() instead of parseXmlDom(). This is a prerequisite for upcoming OPDS format library enhancements.

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -20,7 +20,6 @@
 #include "manager.h"
 
 #include "tools.h"
-#include "tools/pathTools.h"
 
 #include <pugixml.hpp>
 #include <filesystem>
@@ -201,20 +200,12 @@ bool Manager::readFile(
   bool readOnly,
   bool trustLibrary)
 {
-  bool retVal = true;
-  pugi::xml_document doc;
-
-#ifdef _WIN32
-  pugi::xml_parse_result result = doc.load_file(Utf8ToWide(path).c_str());
-#else
-  pugi::xml_parse_result result = doc.load_file(path.c_str());
-#endif
-
-  if (result) {
-    this->parseXmlDom(doc, readOnly, path, trustLibrary);
-  } else {
-    retVal = false;
+  if (!kiwix::fileExists(path)) {
+    return false;
   }
+
+  const std::string xml = getFileContent(path);
+  const bool retVal = this->readXml(xml, readOnly, path, trustLibrary);
 
   /* This has to be set (although if the file does not exists) to be
    * able to know where to save the library if new content are
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -135,10 +135,10 @@ bool Manager::readXml(const std::string& xml,
       = doc.load_buffer((void*)xml.data(), xml.size());
 
   if (result) {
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
Please fix the CI first.

hamazaspavetisyan:
> Please fix the CI first.

Done

src/manager.cpp
```diff
@@ -135,10 +134,10 @@ bool Manager::readXml(const std::string& xml,
       = doc.load_buffer((void*)xml.data(), xml.size());
 
   if (result) {
-    this->parseXmlDom(doc, readOnly, libraryPath, trustLibrary);
+    return this->parseXmlDom(doc, readOnly, libraryPath, trustLibrary);
   }
 
-  return true;
+  return false;
```
veloman-yunkan, line 140: This is a change of behaviour of `readXml()`. Let's not combine this change with the re-implementation of readFile(). Please move this change into the commit enhancing the unit-test of `readXml()` (and add a test case checking how invalid input XML is handled).

veloman-yunkan (changes requested):
Please reorder the commits so that unit-test enhancements come before the re-implementation of `readFile()` (the logic of such sequence is that we first make sure that readXml() can be trusted and then re-use it in `readFile()`).

test/manager.cpp
```diff
@@ -80,6 +87,31 @@ TEST(ManagerTest, readXml)
     EXPECT_EQ(678U*1024, book.getSize());
 }
 
+TEST(ManagerTest, readXmlInvalid)
+{
+    auto lib = kiwix::Library::create();
+    kiwix::Manager manager = kiwix::Manager(lib);
+
+    // Malformed XML (unclosed tag) - load_buffer() should fail to parse it.
+    const std::string invalidXML = "<library version=\"1.0\"><book></library>";
```
veloman-yunkan: 1. Please use instead a truncated library.xml file content (with one or more complete book entries) so that we test that books are added only when the entire file is valid XML (similar to `readOpds()` coming in the next commit).

2. This is a blackbox test. Don't explain the implementation of the functionality being tested.

src/manager.cpp
```diff
@@ -135,10 +135,10 @@ bool Manager::readXml(const std::string& xml,
...
```
veloman-yunkan, line 140: The commit message must be updated to mention this bug-fix

src/manager.cpp
```diff
@@ -201,21 +200,13 @@ bool Manager::readFile(
...
-  if (result) {
-    this->parseXmlDom(doc, readOnly, path, trustLibrary);
-  } else {
-    retVal = false;
+  if (!kiwix::fileExists(path)) {
+    return false;
   }
 
+  const std::string xml = getFileContent(path);
+  const bool retVal = this->readXml(xml, readOnly, path, trustLibrary);
```
veloman-yunkan, line 208: Please mention in the commit description the change of behavior of `readFile()` on input files that are not valid XML.

veloman-yunkan: Oops! I got it wrong here. Unlike `readXml()`, `readFile()` didn't return `true` if the input was not valid XML.

veloman-yunkan (approved):
LGTM

PR #1328, opened 2026-08-15
Rewrite beautifyFileSize to use SI (base-1000) units

lakshyajn:
To address the issue under `kiwix-tools` , modified the function to display human-readable size formats.
It can now be used in `kiwix-tools` to display sizes nicely.

Also added new unit tests as per the changes

### DIFF
--- a/src/tools/stringTools.cpp
+++ b/src/tools/stringTools.cpp
@@ -31,6 +31,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cmath>
 #include <iostream>
 #include <iomanip>
 #include <regex>
@@ -53,6 +54,11 @@ std::string asciiToLower(std::string s)
         return ('A' <= c && c <= 'Z') ? c - ('Z' - 'z') : c;
         });
     return s;
+}
+
+double roundToTwoDecimalPlaces(double value)
+{
+    return std::round(value * 100.0) / 100.0;
 }
 
 } // unnamed namespace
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan I would propose to rename to `beaitifyMemorySize()`, keep `beautifyFileSize` as a wrapper (with deprecation warning and remove next major release). What do you think?

src/tools/stringTools.cpp
```diff
@@ -126,16 +126,39 @@ std::string kiwix::beautifyInteger(uint64_t number)
...
+    999995000000000ULL,      // PB
+    999995000000000000ULL    // EB
+  };
+  static_assert(sizeof(thresholds) / sizeof(thresholds[0]) == unitCount,
+                "thresholds and units arrays must have the same length");
+
+  size_t unit = 0;
+  while (unit + 1 < unitCount && number >= thresholds[unit + 1]) {
+    ++unit;
+  }
```
veloman-yunkan: The problem that this code is trying to address is too unimportant (this function is not used to control a space telescope; rounding that is slightly off for values falling right on the rounding boundary for the case of exact arithmetic can be easily tolerated). Therefore I consider this solution as overengineered and too cumbersome. I'd rather stick to the original simpler solution and instead adapt the unit tests to allow for either result for an edge case value. With a properly defined helper macro `EXPECT_ONE_OF()` the check can be written as:

```c++
EXPECT_ONE_OF(beautifyFileSize(999995000ULL), ("999.99 GB", "1.00 GB")  );
```

veloman-yunkan:
> @veloman-yunkan I would propose to rename to `beaitifyMemorySize()`, keep `beautifyFileSize` as a wrapper (with deprecation warning and remove next major release). What do you think?

Not worth the effort. In case you choose to do it, at least don't let the typo to sneak in the public API :)

lakshyajn:
@veloman-yunkan  kindly have a look, I added the macro and replaced that "rocket engineering optimisation" 😅 code with almost as the previous one.

veloman-yunkan:
@lakshyajn For some reason, the diff is currently unavailable, but I looked at your changes via [the diff with the previous version](https://github.com/kiwix/libkiwix/compare/ac47877874d6046f24f3955fca844bf8c9e9d980..16124ae39277d943cbb026d46f271eda393fbf71).

Can't you incorporate the correction step into the main unit detection loop by changing the condition to test the rounded value? Note that we don't mind about the performance of this function.

BTW, in order to make the code more readable/clear I would introduce a small helper function `double roundToTwoDecimalPlaces(double)` in an unnamed namespace above.

lakshyajn:
> Can't you incorporate the correction step into the main unit detection loop by changing the condition to test the rounded value?

Yeah sorry about that, it made the function more simpler for checking.
Thanks :)

veloman-yunkan (approved):
LGTM

PR #1329, opened 2026-08-17
Added Flawfinder error suppression comment.

hamazaspavetisyan:
I was facing error in github since Flawfinder was detecting potential issue with the readlink call. The fix is just writing comment that will be parsed by Flawfinder to ignore the issue. Actually we do not have issue in sources.


kelson42:
@veloman-yunkan Anything stopping us to quickly review and merge this trivial PR?

veloman-yunkan (approved):

PR #1332, opened 2026-08-20
Fix Manager::readFile() losing writableLibraryPath for new libraries

hamazaspavetisyan:
## Why

`Manager::readFile()` is how a library file gets loaded — including the still-nonexistent path of a brand-new library about to be created. The `fileExists()` early-return skipped setting `writableLibraryPath` in that case, so a caller who then added a book with a relative `--zimPathToSave` had it resolved against the process's CWD instead of the new library's own directory — silently wrong output paths for anyone creating a library from scratch.

## What

- In `Manager::readFile()`, moved the `writableLibraryPath` assignment ahead of the `fileExists()` early return, so it's set whether or not the file exists yet.
- Added `resolveAbsPath()` test helper (wraps the recurring `computeAbsolutePath(removeLastPathElement(...), ...)` pattern) and used it in the existing `addBookFromPathAndGetIdTest`.
- Added `readFileSetsWritableLibraryPathEvenIfFileDoesNotExist` regression test, using the new helper.

No API changes. Fixes incorrect path resolution for any caller (e.g. `kiwix-manage`) that creates a new library and saves a book with a relative path before the library file itself exists on disk.

## Fundamental change

None — this is an ordering fix inside `readFile()`; behavior for existing/readable libraries is unchanged.

### DIFF
--- a/test/manager.cpp
+++ b/test/manager.cpp
@@ -6,6 +6,16 @@
 #include <iostream>
 #include <fstream>
 
+namespace
+{
+
+std::string resolveAbsPath(const std::string& basePath, const std::string& relPath)
+{
+    return kiwix::computeAbsolutePath(kiwix::removeLastPathElement(basePath), relPath);
+}
+
+} // unnamed namespace
+
 TEST(ManagerTest, addBookFromPathAndGetIdTest)
 {
     auto lib = kiwix::Library::create();
@@ -20,7 +30,7 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
     const std::string url = "url";
     bookId = manager.addBookFromPathAndGetId("./test/example.zim", pathToSave, url, true);
     book = lib->getBookById(bookId);
-    auto savedPath = kiwix::computeAbsolutePath(kiwix::removeLastPathElement(manager.writableLibraryPath), pathToSave);
+    auto savedPath = resolveAbsPath(manager.writableLibraryPath, pathToSave);
     EXPECT_EQ(book.getPath(), savedPath);
     EXPECT_EQ(book.getUrl(), url);
 }
--- a/test/manager.cpp
+++ b/test/manager.cpp
@@ -35,6 +35,27 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
     EXPECT_EQ(book.getUrl(), url);
 }
 
+TEST(ManagerTest, readFileSetsWritableLibraryPathEvenIfFileDoesNotExist)
+{
+    auto lib = kiwix::Library::create();
+    kiwix::Manager manager(lib);
+
+    const std::string nonExistentPath
+        = kiwix::computeAbsolutePath(
+              kiwix::computeAbsolutePath(kiwix::getCurrentDirectory(), "test"),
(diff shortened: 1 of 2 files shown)


test/manager.cpp
```diff
@@ -25,6 +25,32 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
...
+
+    EXPECT_FALSE(manager.readFile(nonExistentPath, /*readOnly=*/false));
+    EXPECT_EQ(manager.writableLibraryPath, nonExistentPath);
+
+    const std::string pathToSave = "./relative.zim";
+    auto bookId = manager.addBookFromPathAndGetId("./test/example.zim", pathToSave);
+    ASSERT_NE(bookId, "");
+    kiwix::Book book = lib->getBookById(bookId);
+    auto savedPath = kiwix::computeAbsolutePath(
+        kiwix::removeLastPathElement(nonExistentPath), pathToSave);
```
veloman-yunkan: Why don't you move commit a9332572c1ebb72f6e645a498c2003f5eda0eff6 from #1319 into this PR and take advantage of it here?

veloman-yunkan (changes requested):
Please rewrite this PR as two commits:

1. Commit introducing the helper function
2. Bugfix (immediately taking advantage of the new helper in its unit test).

hamazaspavetisyan:
> Please rewrite this PR as two commits:
> 
> 1. Commit introducing the helper function
> 2. Bugfix (immediately taking advantage of the new helper in its unit test).

I think it would be more natural if, in the first commit you suggested, we also refactored the existing `addBookFromPathAndGetIdTest` to use it. Then, in the second commit, we could implement the bug fix along with the new test case.

Please check it out.

veloman-yunkan:
> > Please rewrite this PR as two commits:
> > 
> > 1. Commit introducing the helper function
> > 2. Bugfix (immediately taking advantage of the new helper in its unit test).
> 
> I think it would be more natural if, in the first commit you suggested, we also refactored the existing `addBookFromPathAndGetIdTest` to use it. Then, in the second commit, we could implement the bug fix along with the new test case.
> 
> Please check it out.

That's what I meant.

test/manager.cpp
```diff
@@ -35,6 +35,27 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
...
+    const std::string nonExistentPath
+        = kiwix::computeAbsolutePath(
+              kiwix::computeAbsolutePath(kiwix::getCurrentDirectory(), "test"),
+              "does_not_exist.xml");
+
+    EXPECT_FALSE(manager.readFile(nonExistentPath, /*readOnly=*/false));
+    EXPECT_EQ(manager.writableLibraryPath, nonExistentPath);
+
+    const std::string pathToSave = "./relative.zim";
+    auto bookId = manager.addBookFromPathAndGetId("./test/example.zim", pathToSave);
```
veloman-yunkan, line 52: This is the first time that I am exposed to the `pathToSave` parameter of `Manager::addBookFromPathAndGetId()` and it looks fragile to me. We will have to revisit it.

veloman-yunkan (approved):

PR #1333, opened 2026-08-25
fix: stop decoding the whole filters cookie

### LINKED ISSUE #1310: Encoding/Interpretation of kiwix-serve parameter seems wrong
Reported at https://github.com/kiwix/kiwix-android/issues/4954#issuecomment-4756736368

Here we search for `D&D`, and the server returns this result.

Image

Here I have searched for the encoded one `D%26D` and it is showing the D&D files.

Image

cestercian:
Fixes #1310

## Summary

Catalog search for `D&D` broke because `getCookie()` ran `decodeURIComponent` over the *entire* cookie header. Filters are stored from `URLSearchParams.toString()` (so `q=D%26D`). Decoding the whole cookie turned `%26` into `&` before parse, so `q` became `D`.

## Changes

- `setCookie` / `getCookie` own cookie-value encoding: encode on set, decode **only the extracted value** on get (never the whole `document.cookie` header).
- `queryUrlBuilder` appends `nonEmptyParams(params).toString()` like the feed link (keeps `q=D%26D`, drops empty `tag=` / `lang=`).
- Tests assert OPDS titles/self-links keep single-encoded `q=D%26D`.
- `index.js` cacheid updated.

## Related

- https://github.com/kiwix/libkiwix/issues/1310
- Origin: https://github.com/kiwix/kiwix-android/issues/4954

## Testing

1. `meson . build && ninja -C build`
2. `meson test -C build`

Local `meson test`: 24/24 passed.


kelson42:
@cestercian Thank you for you PR. Please just solve the encoding issue. Just ensure that the `value` is properly URI encoded. You should not try to redo URL querystring parsing like a browser or MHD does.

cestercian:
Thanks @kelson42 — will drop the querystring re-join logic and just make sure `value` is properly URI-encoded (no browser/MHD-style reparsing).

kelson42:
@cestercian Looking at your PR, something has to be wrong, we should not have the `%` character encoded twice (and get `D%2526D`). Therefore `D%26D` seems to be the correct URI value. So then the question is why the `q` value retrieves `D` in place of `D&D` and I suspect we have a double decoding problem or the URI is parsed after URI decoding or something like that.

cestercian:
@kelson42 Agreed — the correct wire form for searching `D&D` is `q=D%26D`, not `q=D%2526D`.

What this PR changes is only the library UI (`queryUrlBuilder`): it was concatenating the already-decoded `URLSearchParams` value, so a search for `D&D` left the server as `q=D&D`. MHD then splits on `&` and `get_argument("q")` becomes `D`. With `encodeURIComponent` once, the request is `q=D%26D` and MHD delivers `D&D` (encode/decode once on the C++ side was already correct in `add_argument` / `get_query`).

The `D%2526D` bits in the tests are only documenting decode-once vs encode-once for a *double-encoded input*; they are not a client path we want. I’ll trim those so the PR is clearly about producing `D%26D`, and add a clearer assertion that `q=D%26D` is the intended encoding.

If you still see `q` → `D` when the request line truly contains `q=D%26D` (not an unencoded `&`), point me at that trace and I’ll dig the double-decode path further.

kelson42:
> What this PR changes is only the library UI (queryUrlBuilder): it was concatenating the already-decoded URLSearchParams value, so a search for D&D left the server as q=D&D

The server should get `q=D%26D` and not parse the `&` in the middle. This is why I say there is a decoding which happens at the wrong place.

With your patch now you add a double encoding of the input, so of course it workarounds the problem.

cestercian:
@kelson42 Fair point — I was chasing the wrong layer.

Looking again: when filters are restored from the cookie we run `decodeURIComponent` over the *whole* cookie string. That turns a stored `q=D%26D` into `q=D&D` before `URLSearchParams` ever sees it, so the `&` gets treated as a separator. That’s decode in the wrong place.

I’ll drop the “just encode harder in queryUrlBuilder” framing and fix that. Catalog fetch should end up with `q=D%26D` on the wire the same way the feed link already does via `params.toString()`.

cestercian:
Pushed `e9b3bd9` with that fix:

- `getCookie` no longer runs `decodeURIComponent` over the whole cookie header
- `queryUrlBuilder` uses `params.toString()` the same way as the feed link
- dropped the `encodeURIComponent` bandage

Local `meson test` is green. Title/body updated to match.

static/skin/index.js
```diff
@@ -53,8 +53,11 @@
     function queryUrlBuilder() {
         let url = `${root}/catalog/v2/entries?`;
         url += Object.keys(incrementalLoadingParams).map(key => `${key}=${incrementalLoadingParams[key]}`).join("&");
-        params.forEach((value, key) => {url+= value ? `&${key}=${value}` : ''});
-        return (url);
+        const filterQuery = params.toString();
+        if (filterQuery) {
+            url += `&${filterQuery}`;
+        }
```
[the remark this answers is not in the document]
cestercian: Good catch — pushed `35f4cb3`.

`queryUrlBuilder` now goes through a shared `nonEmptyParams()` helper (same empty-skip as `updateFeedLink`) before `toString()`, so cleared filters no longer emit `&tag=` / `&lang=`. Encoded values like `q=D%26D` are unchanged.

kelson42:
@cestercian @veloman-yunkan From a user perspective it works fine now.

But I still have a doubt about the fix. Reasons are:
* Cookies should be encoded
* `setCookie()` and `getCookie()` should deal with unencoded variables

Therefore I suspect that this is a bad idea to remove the decoding from `getCookie()` (but why there is no encoding in `setCookie()`?!) and actually we have an encoding/decoding problem somewhere else (probably deeper in the code where we use the free text `q` value).

@veloman-yunkan I leave it to you for the last part of the review.

cestercian:
@kelson42 Agreed on the layering — thanks for spelling it out.

Pushed `0138a71`:
- `setCookie` now `encodeURIComponent`s the value
- `getCookie` splits the raw cookie header, then `decodeURIComponent`s **only** the extracted value (still no whole-header decode)
- callers keep passing logical values (`params.toString()` etc.)

So `setCookie`/`getCookie` deal in unencoded variables again; the cookie attribute itself is encoded. Left the empty-filter skip + `params.toString()` catalog URL path as-is.

@veloman-yunkan ready whenever you take a look.

kelson42:
@cestercian Please rework the PR to not merge `main` in the feature branch. This is something we don't do at Kiwix to keep the revision log clear. Please to a rebase instead.

cestercian:
@kelson42 Rebased this branch onto current `main` (no merge commit). The three fix commits are now replayed linearly on top of upstream main.

kelson42:
@veloman-yunkan Any feedback here? In particular to https://github.com/kiwix/libkiwix/pull/1333#issuecomment-5523321230

veloman-yunkan (commented):
Why isn't this PR limited just to the changes in the `setCookie()` and `getCookie()` functions?

cestercian:
@veloman-yunkan Fair ask.

The cookie helpers are the core of the bug: decoding the *whole* `document.cookie` header turned a stored `q=D%26D` into `q=D&D` before `URLSearchParams` parsed it, so `q` became `D`. `setCookie`/`getCookie` now encode on set and decode only the extracted value, which matches the layering kelson described.

I’m keeping one sibling change in `queryUrlBuilder`. On main it still builds the catalog fetch with `params.forEach((value, key) => … ${value})`, which emits a raw `q=D&D`. MHD then splits on `&`, so a live `D&D` search breaks even with correct cookies. The feed link already uses `toString()`; this makes the catalog fetch do the same (and still skips empty filter params). Without that, cookie-only would fix restore-from-cookie but not a fresh search.

I dropped the C++ encode/OPDS assertions that weren’t exercising the filters-cookie path. Happy to split the `queryUrlBuilder` bit into a follow-up if you’d rather this PR stay cookie-only.

static/skin/index.js
```diff
@@ -64,18 +71,18 @@
...
-        document.cookie = `${cookieName}=${cookieValue};${exp}sameSite=Strict`;
+        document.cookie = `${cookieName}=${encodeURIComponent(cookieValue)};${exp}sameSite=Strict`;
     }
 
     function getCookie(cookieName) {
         const name = cookieName + "=";
         let result;
-        decodeURIComponent(document.cookie).split('; ').forEach(val => {
+        document.cookie.split('; ').forEach(val => {
             if (val.indexOf(name) === 0) {
```
veloman-yunkan: Please change to `val.startsWith(name)`

static/skin/index.js
```diff
@@ -53,8 +57,11 @@
     function queryUrlBuilder() {
         let url = `${root}/catalog/v2/entries?`;
         url += Object.keys(incrementalLoadingParams).map(key => `${key}=${incrementalLoadingParams[key]}`).join("&");
-        params.forEach((value, key) => {url+= value ? `&${key}=${value}` : ''});
-        return (url);
+        const filterQuery = nonEmptyParams(params).toString();
+        if (filterQuery) {
+            url += `&${filterQuery}`;
+        }
+        return url;
```
veloman-yunkan: To the best of my understanding, the only problem related to the issue being addressed is that the values present in `params` were not URI-encoded when building the URL. Dropping of params with empty values has nothing to do with that. Let's undo the change introducing the `nonEmptyParams()` function, and just restrict this change to the correct way of appending the filtering params to the URL:

``` javascript
if ( parmas.size ) {
  url += `&${params}`;
}
```

veloman-yunkan (changes requested):
Let's keep this PR as to the point as possible, with minimal changes (all in one commit).

veloman-yunkan:
Please rebase your branch on recent main and make sure that this PR consists of a single commit.

cestercian:
@veloman-yunkan Addressed:

- Rebased onto current `main` and squashed to a **single commit** (`e6015da`)
- `getCookie` now uses `val.startsWith(name)`
- Dropped `nonEmptyParams()`; `queryUrlBuilder` only does `if (params.size) url += \`&${params}\`;`
- Left `updateFeedLink`’s existing empty-filter loop alone
- Diff is just `static/skin/index.js` + the `index.js` cacheid bump in `test/server.cpp`

veloman-yunkan (approved):
LGTM. Thank you!

PR #1341, opened 2026-09-04
1302 multiple links per book

### LINKED ISSUE #1302: OPDS to provide multiple links per book

hamazaspavetisyan:
Fixes #1302

## Why

An OPDS entry today exposes a single acquisition link, conflating "the ZIM" with whatever URL happens to be set - often actually a `.meta4` file labeled `type="application/x-zim"`. Per the design agreed on that issue, direct-ZIM, Metalink, and BitTorrent need to be exposed as separate, distinguishable download options so readers (PWA, Kiwix Desktop, Cardshop, Seeder, etc.) can pick the one they actually support instead of guessing from a `.meta4` suffix.

## What

- `Book` now stores a list of typed acquisition links (`AcquisitionLink{mimeType, url}`) instead of a single URL string, via a new `setUrl(mimeType, url)` / `getAcquisitionLinks()` pair.
- `updateFromOpds()` reads each link's `type` attribute and keeps every distinct mime type it finds (previously it silently kept only the last remote link, discarding the rest).
- The OPDS entry renderer now emits one `rel="open-access"` `<link>` per stored acquisition link, each with its own `type` and the shared `length` (the target ZIM's size, per the issue's agreed convention) - covering the live catalog, the legacy v1 feed, and the offline library dump, since they all share the same renderer.
- `getUrl()` is deprecated in favor of `getUrl(mimeType)`; every internal caller (search filtering, the plain-HTML dumper, the legacy XML dumper, and the server's no-JS download page) now asks for `application/x-zim` explicitly, and the no-JS download page prefers an explicit `application/metalink4+xml` link when a catalog provides one.

- `getUrl()` still returns the same value it always did (the book's `application/x-zim` link) - no behavior change for existing callers, just a deprecation notice pointing them at the new API.
- `setUrl(url)` still works exactly as before for anyone only ever setting one link.
- New: `setUrl(mimeType, url)` to record more than one link, `getUrl(mimeType)` to read a specific one, and `getAcquisitionLinks()` to get the full list.
- A single-link book's OPDS output is byte-identical to before; a book with several typed links now gets all of them rendered, everywhere the library is serialized to OPDS.

## Fundamental change

A book's downloadable content is no longer "one URL" - it's a set of interchangeable acquisition links (ZIM, Metalink, BitTorrent), and every OPDS-producing code path in the library now speaks that model uniformly.

## Notes

Implements #1302. `kiwix-serve`'s no-JS download page only partially takes advantage of the new links so far (it falls back to the old `.meta4`-suffix-stripping trick when no distinct link is available); fully wiring up an explicit `application/x-bittorrent` link there is a reasonable follow-up but isn't required by this issue.


kelson42:
@veloman-yunkan Looks good on the principle to me, can you please review?

include/book.h
```diff
@@ -63,6 +63,16 @@ class Book
 
   typedef std::vector<std::shared_ptr<const Illustration>> Illustrations;
 
+  struct AcquisitionLink
+  {
+    std::string mimeType;
```
veloman-yunkan, line 68: MIME-type doesn't look right for discriminating between link/access types. It suggests a wide range of possibilities like `book.getUrl("text/plain")` or `book.getUrl("audio/ogg")`.

I think that an enumeration like `enum LinkKind { DIRECT, BITTORRENT, MAGNET, META4 }` would be better, however I realize that then we will have to deal with the problem of mapping between link info found in the OPDS and this `enum`.

include/book.h
```diff
@@ -163,7 +182,7 @@ class Book
   std::string m_creator;
   std::string m_publisher;
   std::string m_date;
-  std::string m_url;
+  std::vector<AcquisitionLink> m_urls;
```
veloman-yunkan, line 185: This is an ABI change (`sizeof(kiwix::Book)` may be affected).

veloman-yunkan (commented):
This is a preliminary review, pointing out the biggest potential problems with the approach and implementation.

include/book.h
hamazaspavetisyan: I initially considered using an enum, but went with const strings to avoid writing mapping functions. On the other hand, this makes it easier to support other types in the future. If you'd like, I can add a check that throws an error for unsupported types.

include/book.h
hamazaspavetisyan: Since we will be storing more URLs, I think a change to the book size is acceptable. Another option—though a bit of an antipattern—would be to serialize the AcquisitionLink URLs into a single string separated by a delimiter.

kelson42: We had a discussion about the `length` value in the issue. We are aware about the redundancy of the `length` attribute. For the moment, return the value of the links with the following order:
* Path
* (If no path) Direct http link
* (If no http link) BitTorrent link
* (Still if no values) Any other entry in the order they appear.

hamazaspavetisyan:
@kelson42 We need a better understanding of the scope of the two issues that @veloman-yunkan raised.

1. **Changing the `url` field type:** If we change the `url` field from a simple string to a vector of strings, the memory layout/size of the object may change. This is not related to the `length` attribute of the URL itself, but to the internal representation of the `Book` object. This means that clients using the `Book` class would need to be rebuilt, and we may need to bump the `libkiwix` version as well. Having compatibility will require an ugly fix.

2. **Number of links per book:** We currently have direct HTTP, Meta4, and BitTorrent links. How many links of each type can exist within a single book entity? Are we restricted to one link per type? Note that Meta4 now also covers mirroring for direct HTTP download links.

3. **The `length` attribute:** If we are going to release a new version of `libkiwix` anyway, we could consider moving the `length` attribute to the `entity` tag itself.

kelson42:
>     1. **Changing the `url` field type:** If we change the `url` field from a simple string to a vector of strings, the memory layout/size of the object may change. This is not related to the `length` attribute of the URL itself, but to the internal representation of the `Book` object. This means that clients using the `Book` class would need to be rebuilt, and we may need to bump the `libkiwix` version as well. Having compatibility will require an ugly fix.

Yes, libkiwix will have anyway a new minor version. What is at stack is only the major version.

If you don't break the API, we don't have to release a new major. I request here you don't break the API. Keep the old API working like usual (returning the direct HTTP link). Still make them `DEPRECATED`

>     2. **Number of links per book:** We currently have direct HTTP, Meta4, and BitTorrent links. How many links of each type can exist within a single book entity? Are we restricted to one link per type? Note that Meta4 now also covers mirroring for direct HTTP download links.

Only one of each type.

>     3. **The `length` attribute:** If we are going to release a new version of `libkiwix` anyway, we could consider moving the `length` attribute to the `entity` tag itself.

Yes, how would that look?

include/book.h
veloman-yunkan: @kelson42 Your feedback seems to pertain to some other comment.

veloman-yunkan:
> > ```
> > 1. **Changing the `url` field type:** If we change the `url` field from a simple string to a vector of strings, the memory layout/size of the object may change. This is not related to the `length` attribute of the URL itself, but to the internal representation of the `Book` object. This means that clients using the `Book` class would need to be rebuilt, and we may need to bump the `libkiwix` version as well. Having compatibility will require an ugly fix.
> > ```
> 
> Yes, libkiwix will have anyway a new minor version. What is at stack is only the major version.
> 
> If you don't break the API, we don't have to release a new major. I request here you don't break the API. Keep the old API working like usual (returning the direct HTTP link). Still make them `DEPRECATED`

@kelson42 The change in the internals of `kiwix::Book` preserves the API but breaks the ABI.

include/book.h
kelson42: @veloman-yunkan Then we don't understand each other. Looks like anyway not relevant if we go to have a dedicated OPDS/Atom node to store the file size.

veloman-yunkan: @kelson42

> Then we don't understand each other.

Indeed. I see now why you have been confused.

> I think a change to the book size is acceptable

This was intended to refer to the size of the `kiwix::Book` C++ object rather than the length of the ZIM file associated with a book. The current state of this PR increases the size of the `kiwix::Book` object in memory and shifts its data members coming after the (old/replaced) data member `m_url`. That's why the ABI is broken.

kelson42:
What is the status here?

hamazaspavetisyan:
> What is the status here?

It is ready to be merged! But note, we must release this together with https://github.com/kiwix/libkiwix/pull/1349

kelson42:
@hamazaspavetisyan If ready, PR should be approved. if depending of an other PR, then it should be based on its feature branch, and clearly written that this is blocked by an other PR.

hamazaspavetisyan:
> @hamazaspavetisyan If ready, PR should be approved. if depending of an other PR, then it should be based on its feature branch, and clearly written that this is blocked by an other PR.

Not blocked, but as we have ABI change in scope of this PR, to release this change to users requires https://github.com/kiwix/libkiwix/pull/1349 also to be merged. Technically 1349 can be merged after this one.

What about approval, @veloman-yunkan please take a look at recent changes that we discussed today.

include/book.h
```diff
@@ -137,7 +143,7 @@ class Book
   void setCreator(const std::string& creator) { m_creator = creator; }
   void setPublisher(const std::string& publisher) { m_publisher = publisher; }
   void setDate(const std::string& date) { m_date = date; }
-  void setUrl(const std::string& url) { m_url = url; }
+  void setUrl(const std::string& url) { m_urls.push_back(url); }
```
veloman-yunkan: `m_urls.assign(1, url)` would be better

include/book.h
```diff
@@ -63,6 +63,15 @@ class Book
 
   typedef std::vector<std::shared_ptr<const Illustration>> Illustrations;
 
+  struct AcquisitionLink
+  {
+    enum LinkKind { DIRECT, BITTORRENT, MAGNET, META4 };
+
+    LinkKind mimeType;
```
veloman-yunkan: The name of this field had to be changed too (and the commit message must be updated)

include/book.h
```diff
@@ -143,7 +147,8 @@ class Book
   void setCreator(const std::string& creator) { m_creator = creator; }
   void setPublisher(const std::string& publisher) { m_publisher = publisher; }
   void setDate(const std::string& date) { m_date = date; }
-  void setUrl(const std::string& url) { m_urls.push_back(url); }
+  void setUrl(const std::string& url) { addUrl(AcquisitionLink::DIRECT, url); }
+  void addUrl(AcquisitionLink::LinkKind mimeType, const std::string& url) { m_urls.push_back({mimeType, url}); }
```
veloman-yunkan: Such an API allows for multiple links of the same kind:

```
book.addUrl(AcquisitionLink::DIRECT, "https://download.kiwix.org/zims/fairytales.zim")
book.addUrl(AcquisitionLink::DIRECT, "https://archive.org/kiwixmirror/zims/fairytales.zim")
```

At least the implementation has to be fixed. What about the function name, `setUrl()` is a better fit for the [current stated constraints](https://github.com/kiwix/libkiwix/pull/1341#issuecomment-5615543517).

include/book.h
```diff
@@ -63,6 +63,15 @@ class Book
 
   typedef std::vector<std::shared_ptr<const Illustration>> Illustrations;
 
+  struct AcquisitionLink
+  {
+    enum LinkKind { DIRECT, BITTORRENT, MAGNET, META4 };
+
+    LinkKind mimeType;
+    std::string url;
+    static LinkKind fromMimeType(const std::string& mimeType);
```
veloman-yunkan: We don't need this method in our public API (at least at this point).

test/book.cpp
```diff
@@ -192,6 +192,37 @@ TEST(BookTest, updateFromOPDSDuplicateLengthWarnsTest)
     EXPECT_EQ(book.getSize(), 222U);
 }
 
+TEST(BookTest, addUrlWithMimeTypeAddsToAcquisitionLinks)
+{
+    kiwix::Book book;
+    book.setUrl("http://who.org/zara.zim");
+    book.addUrl(kiwix::Book::AcquisitionLink::fromMimeType("application/metalink4+xml"), "http://who.org/zara.zim.meta4");
+    book.addUrl(kiwix::Book::AcquisitionLink::fromMimeType("application/x-bittorrent"), "http://who.org/zara.zim.torrent");
```
veloman-yunkan: Why don't you use `LinkKind` enum values directly?

test/book.cpp
```diff
@@ -192,6 +192,37 @@ TEST(BookTest, updateFromOPDSDuplicateLengthWarnsTest)
...
+    kiwix::Book book;
+    book.addUrl(kiwix::Book::AcquisitionLink::fromMimeType("application/metalink4+xml"), "http://who.org/zara.zim.meta4");
+    book.addUrl(kiwix::Book::AcquisitionLink::fromMimeType("application/x-bittorrent"), "http://who.org/zara.zim.torrent");
+
+    EXPECT_EQ(book.getUrl(), "");
+
+    book.setUrl("http://who.org/zara.zim");
+    book.addUrl(kiwix::Book::AcquisitionLink::fromMimeType("application/x-bittorrent"), "http://who.org/zara.zim.torrent");
+
+    EXPECT_EQ(book.getUrl(), "http://who.org/zara.zim");
```
veloman-yunkan: This piece of code illustrates the [problem mentioned above](https://github.com/kiwix/libkiwix/pull/1341/changes/30adcc69f48b67195d6dcae0fc998b15f9c31729#r4053088442). By looking only at these three lines, you cannot expect it to work - the code will fail if preceded by `book.setUrl("something other than http://who.org/zara.zim");`.

src/book.cpp
```diff
@@ -326,15 +326,17 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
...
-      // book (an absolute URL) or a local one (a filesystem path, absolute
-      // or relative to baseDir) - a single entry may carry one of each.
       const std::string href = linkNode.attribute("href").value();
-      if (isAbsoluteUrl(href)) {
-        m_urls = { {AcquisitionLink::DIRECT, href} };
-      } else {
+      std::string type = linkNode.attribute("type").value();
+      if (type.empty()) {
+        type = AcquisitionLink::DIRECT;
+      }
```
veloman-yunkan: This code won't compile (obviously, it is fixed in a later commit).

src/book.cpp
```diff
@@ -166,11 +167,36 @@ Book::~Book()
...
+    return META4;
+  }
+  if (mimeType == "application/x-bittorrent") {
+    return BITTORRENT;
+  }
+  if (mimeType == "application/x-magnet") {
+    return MAGNET;
+  }
+  return DIRECT;
+}
```
veloman-yunkan: All MIME-type values different from "application/metalink4+xml", "application/x-bittorrent", and "application/x-magnet" are converted to `DIRECT`. That's somewhat dubious.

src/book.cpp
```diff
@@ -326,15 +326,17 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
...
-      // or relative to baseDir) - a single entry may carry one of each.
       const std::string href = linkNode.attribute("href").value();
-      if (isAbsoluteUrl(href)) {
-        m_urls = { {AcquisitionLink::DIRECT, href} };
-      } else {
+      std::string type = linkNode.attribute("type").value();
+      if (type.empty()) {
+        type = AcquisitionLink::DIRECT;
+      }
+      auto linkType = AcquisitionLink::fromMimeType(type);
```
veloman-yunkan: I think that the helper utility for recognizing the link kind from the MIME-type should appear in this commit.

src/library.cpp
```diff
@@ -59,6 +59,17 @@ bool booksReferToTheSameArchive(const Book& book1, const Book& book2)
       && book1.getPath() == book2.getPath();
 }
 
+// updateFromXml() always stores a DIRECT AcquisitionLink, even when the xml
+// carries no url attribute, so getAcquisitionLinks() alone (unlike
+// getUrl(), which only returns non-empty urls) can be non-empty for a
+// purely local book. Filter those empty-url entries out here.
```
veloman-yunkan: Maybe we should define that setting an empty URL is equivalent to unsetting that link entry?

include/manager.h
```diff
@@ -163,6 +175,28 @@ class Manager
...
+   * @param checkMetaData Tell if we check metadata before adding book to the
+   *                      library.
+   * @return The id of the book if the book has been added to the library.
+   *         Else, an empty string.
+   */
+  std::string addBookFromPathAndGetId(const std::string& pathToOpen,
+                                 const std::string& pathToSave,
+                                 const BookAcquisitionUrls& urls,
+                                 const bool checkMetaData = false);
+
```
veloman-yunkan, line 190: Is there any urgency for adding this new method?

src/library_dumper.cpp
```diff
@@ -103,14 +103,26 @@ getEmbeddedThumbnailLinks(const Book& book, bool isLiveCatalog = true)
...
-      continue;
+      return;
     }
     acquisitionLinks.push_back(kainjow::mustache::object{
       {"mimetype", Book::AcquisitionLink::toMimeType(link.mimeType)},
       {"href", link.url}
     });
+  };
+  // DIRECT is the one mime type every OPDS client is guaranteed to
+  // understand, so emit it last as the reliable fallback.
```
veloman-yunkan: This (in a slightly elaborated form) should be also put in the commit message as the justification for this change.

src/library_dumper.cpp
```diff
@@ -103,14 +103,26 @@ getEmbeddedThumbnailLinks(const Book& book, bool isLiveCatalog = true)
...
+  for (const auto& link : book.getAcquisitionLinks()) {
+    if (link.mimeType != Book::AcquisitionLink::DIRECT) {
+      addLink(link);
+    }
+  }
+  for (const auto& link : book.getAcquisitionLinks()) {
+    if (link.mimeType == Book::AcquisitionLink::DIRECT) {
+      addLink(link);
+    }
```
veloman-yunkan: If the constraint for one link per kind is respected, then the links can be stored in sorted order, and you can take advantage of that order here.

veloman-yunkan (changes requested):
The API for setting and getting links must be well thought through.

src/library.cpp
hamazaspavetisyan: @veloman-yunkan 
ok

include/manager.h
hamazaspavetisyan: Because I am going to use this imediatelly after merging the PR.

include/book.h
```diff
@@ -63,6 +65,16 @@ class Book
 
   typedef std::vector<std::shared_ptr<const Illustration>> Illustrations;
 
+  enum AcquisitionLinkKind { BITTORRENT = 0, MAGNET, META4, DIRECT };
```
veloman-yunkan: 1. Make `enum class`
2. I see that `DIRECT` comes last so that links of that kind appear in the end when generating OPDS. However this makes ABI vulnerable if we ever need to add another link kind. Better keep `DIRECT` at the beginning and adjust the OPDS code accordingly.

include/book.h
```diff
@@ -63,6 +65,16 @@ class Book
...
+
+  static constexpr size_t ACQUISITION_LINK_KIND_COUNT = 4;
```
veloman-yunkan: Make this the last member named `COUNT` of `AcquisitionLinkKind`

src/book.cpp
```diff
@@ -303,11 +335,17 @@ void Book::updateFromOpds(const pugi::xml_node& node, const std::string& urlHost
...
       const std::string href = linkNode.attribute("href").value();
-      if (isAbsoluteUrl(href)) {
-        m_url = href;
-      } else {
+      std::string type = linkNode.attribute("type").value();
+      if (type.empty()) {
+        type = fromLinkKindToMimeType(DIRECT);
+      }
+
+      auto linkType = fromMimeTypeToLinkKind(type);
```
veloman-yunkan: This looks silly. Change to

```c++
const auto linkType = type.empty()
                    ? AcquisitionLinkKind::DIRECT
                    : fromMimeTypeToLinkKind(type);
```

src/server/internalServer.cpp
```diff
@@ -871,7 +871,13 @@ std::unique_ptr<Response> InternalServer::handle_viewer_settings(const RequestCo
...
 {
   const auto book = mp_library->getBookById(bookId);
-  auto bookUrl = kiwix::stripSuffix(book.getUrl(), ".meta4");
+  // Today's real remote catalogs (e.g. download.kiwix.org) don't emit a
+  // distinct metalink4+xml link; they tag the .meta4 URL as AcquisitionLinkKind::DIRECT
+  // instead, so fall back to that when there's no dedicated metalink link.
+  auto metalinkUrl = book.getUrl(kiwix::Book::AcquisitionLinkKind::META4);
+  auto bookUrl = kiwix::stripSuffix(
+    metalinkUrl.empty() ? book.getUrl(kiwix::Book::AcquisitionLinkKind::DIRECT) : metalinkUrl,
+    ".meta4");
```
veloman-yunkan: This is dubious.

Please properly update the `no_js_download_html` template and this method to use actual links of different kinds instead of generating them from a single (allegedly, direct) URL.

src/server/internalServer.cpp
veloman-yunkan, line 880: Should we also have a `HASH` (or `SHA256HASH`) link kind?

include/book.h
```diff
@@ -107,7 +119,16 @@ class Book
...
-  const std::string& getUrl() const { return m_url; }
+  /**
+      * Get the book's URL.
+      *
+      * @deprecated A book may have several acquisition links (one per mime
+      * type). Use getAcquisitionLinks() instead.
+      */
+  DEPRECATED const std::string& getUrl() const;
+  const std::string& getUrl(AcquisitionLinkKind kind) const;
+  AcquisitionLinkMap getAcquisitionLinks() const;
```
veloman-yunkan: Something in me speaks against introducing this method (and its return type) now. Public API must not be enhanced lightheartedly. `hasAcquisitionLink()` and `getAcquisitionLinkData()` can be implemented without it and in unit-tests a free helper function can be used.

include/manager.h
```diff
@@ -63,6 +63,18 @@ class Manager
...
+   * A set of acquisition URLs for a book, keyed by acquisition mime type.
+   *
+   * A field left empty means no acquisition link of that type is recorded
+   * for the book.
+   */
+  struct BookAcquisitionUrls {
+    std::string zim;
+    std::string meta4;
+    std::string torrent;
+  };
```
veloman-yunkan: You can use `Book::AcquisitionLinks` instead of this. But either way a new point of binary backward compatibility appears for situations when we will need to add a new link kind. This can probably be countered by changing the container underlying `Book::AcquisitionLinks` to `std::vector`. Need a deeper analysis.

include/book.h
```diff
@@ -63,6 +65,16 @@ class Book
...
 
+  enum AcquisitionLinkKind { BITTORRENT = 0, MAGNET, META4, DIRECT };
+
+  static constexpr size_t ACQUISITION_LINK_KIND_COUNT = 4;
+  // Acquisition URLs indexed by AcquisitionLinkKind (empty = not set).
+  typedef std::array<std::string, ACQUISITION_LINK_KIND_COUNT> AcquisitionLinks;
+  // Only the acquisition URLs that are set (non-empty), keyed by kind.
+  typedef std::map<AcquisitionLinkKind, std::string> AcquisitionLinkMap;
+
+  static std::string fromLinkKindToMimeType(AcquisitionLinkKind kind);
```
veloman-yunkan: This should be a free helper function in an unnamed namespace in `src/library_dumper.cpp`. Its usage in `Book::updateFromXml()` is absolutely unjustified.

veloman-yunkan (changes requested):

src/server/internalServer.cpp
hamazaspavetisyan: @kelson42

include/book.h
hamazaspavetisyan: Then we need getUrls() method right that will return all urls even empty ones ?

include/manager.h
hamazaspavetisyan: Maybe std::list is better ? If we are going to do iterations to find required kind, then index based access has zero advantage. What you think ?

include/book.h
veloman-yunkan: The current use for such a method would be then to iterate over its output:

```c++
for ( const auto& url : book.getUrls() ) { ... }
```

The same can be achieved with:

```c++
for ( int linkKind = 0; linkKind < AcquisitionLinkKind::COUNT; ++linkKind ) {
  const auto& url = book.getUrl(AcquisitionLinkKind(linkKind));
 ...
}
```

include/manager.h
veloman-yunkan: Index based access enables straightforward access to the elements (`acquisitionUrls.zim` becomes `acquisitionLinks[AcquisitionLinkKind::DIRECT]` and so on).

hamazaspavetisyan: Good catch. Since this is a general ABI-stability concern rather than something specific to this PR, I'd rather handle it as part of [libkiwix#1349](https://github.com/kiwix/libkiwix/pull/1349), which is already tackling exactly this class of issue. Let's pick it up there.

src/book.cpp
```diff
@@ -150,10 +154,33 @@ ParsedIllustrationType parseIllustrationType(const std::string& type)
...
+  }
+  if (mimeType == "application/x-magnet") {
+    return kiwix::Book::AcquisitionLinkKind::MAGNET;
+  }
+  if (mimeType == "application/x-zim") {
+    return kiwix::Book::AcquisitionLinkKind::DIRECT;
+  }
+  if (mimeType == "application/x-sha256") {
+    return kiwix::Book::AcquisitionLinkKind::SHA256;
+  }
```
veloman-yunkan: We'd rather get input from @kelson42 before introducing this link kind

src/html_dumper.cpp
```diff
@@ -64,6 +64,19 @@ kainjow::mustache::list getTagList(std::string tags)
...
+  size_t count = 0;
+  const auto urls = book.getUrls();
+  for (const auto& url : urls)
+  {
+    if (!url.empty()) {
+      ++count;
+    }
+  }
+  return count;
+}
```
veloman-yunkan: Duplication with `src/library.cpp`. Should be shared via `src/tools/otherTools.{h,cpp}`.

src/html_dumper.cpp
```diff
@@ -98,7 +111,9 @@ std::string HTMLDumper::dumpPlainHTML(kiwix::Filter filter) const
     const auto bookDescription = bookObj.getDescription();
     const auto bookIconUrl = rootLocation + "/catalog/v2/illustration/" + bookId +  "/?size=48";
     const auto tags = bookObj.getTags();
-    const auto downloadAvailable = (bookObj.getUrl() != "");
+    // The no-JS download page this flag's link leads to lists whichever
+    // acquisition links the book has, so it is available if there is any.
+    const bool downloadAvailable = nonEmptyLinksCount(bookObj) > 0;
```
veloman-yunkan: Having `AcquisitionLinkKind::SHA256` creates a problem with this logic.

include/book.h
```diff
@@ -107,7 +114,15 @@ class Book
...
-  const std::string& getUrl() const { return m_url; }
+  /**
+       * Get the book's URL.
+       *
+       * @deprecated A book may have several acquisition links (one per mime
+       * type). Use getAcquisitionLinks() instead.
+       */
+  DEPRECATED const std::string& getUrl() const;
+  const std::string& getUrl(AcquisitionLinkKind kind) const;
+  const AcquisitionLinks& getUrls() const;
```
veloman-yunkan, line 125: Why do we need this? It looks like so far we can get along with `getUrl()` alone.

src/book.cpp
```diff
@@ -230,7 +267,7 @@ void Book::updateFromXml(const pugi::xml_node& node, const std::string& baseDir)
   m_creator = ATTR("creator");
   m_publisher = ATTR("publisher");
   m_date = ATTR("date");
-  m_url = ATTR("url");
+  m_urls[linkIndex(AcquisitionLinkKind::DIRECT)] = ATTR("url");
```
veloman-yunkan: Use `setUrl()`

src/library_dumper.cpp
```diff
@@ -100,6 +100,57 @@ getEmbeddedThumbnailLinks(const Book& book, bool isLiveCatalog = true)
...
+      {"href", link}
+    });
+  }
+
+  if (!directLinkUrl.empty()) {
+    acquisitionLinks.push_back(kainjow::mustache::object{
+      {"mimetype", fromLinkKindToMimeType(Book::AcquisitionLinkKind::DIRECT)},
+      {"href", directLinkUrl}
+    });
+  }
```
veloman-yunkan, line 150: An easier way of putting the direct link in the end is to iterate in the reverse order.

src/manager.cpp
```diff
@@ -276,7 +289,11 @@ std::string Manager::addBookFromPathAndGetId(const std::string& pathToOpen,
     if (!checkMetaData
         || (!book.getTitle().empty() && !book.getLanguages().empty()
             && !book.getDate().empty())) {
-      book.setUrl(url);
+      for (size_t i = 0; i < urls.size(); ++i) {
+        if (!urls[i].empty()) {
```
veloman-yunkan, line 293: This filtering is redundant

static/templates/catalog_v2_entry.xml
```diff
@@ -22,7 +22,7 @@
       <name>{{publisher_name}}</name>
     </publisher>
     <dc:issued>{{book_date}}</dc:issued>
-    {{#url}}
-    <link rel="http://opds-spec.org/acquisition/open-access" type="application/x-zim" href="{{{url}}}" length="{{{size}}}" />
-    {{/url}}{{#local_path}}<link rel="http://opds-spec.org/acquisition/open-access" href="{{local_path}}" type="application/x-zim" length="{{{size}}}"/>{{/local_path}}
+    {{#acquisitionLinks}}
+    <link rel="http://opds-spec.org/acquisition/open-access" type="{{mimetype}}" href="{{{href}}}" length="{{{size}}}" />
```
veloman-yunkan, line 26: Another problem with `SHA256` is the value of `length`

static/templates/no_js_download.html
```diff
@@ -16,18 +16,21 @@
...
+    </a>{{/sha256_url}}{{#magnet_url}}
+    <a href="{{magnet_url}}" target="_blank">
         <div>{{translations.magnet-link-text}}</div>
-    </a>
-    <a href="{{url}}.torrent" download>
+    </a>{{/magnet_url}}{{#torrent_url}}
+    <a href="{{torrent_url}}" download>
         <div>{{translations.torrent-download-link-text}}</div>
-    </a>
+    </a>{{/torrent_url}}
```
veloman-yunkan, line 34: We must be careful here. Before this change, additional URLs were being generated from the value of the direct URL. Now this has to be somehow done when importing the library.

test/book.cpp
```diff
@@ -15,6 +15,24 @@ struct XMLDoc : pugi::xml_document
...
+  size_t count = 0;
+  const auto urls = book.getUrls();
+  for (const auto& url : urls)
+  {
+    if (!url.empty()) {
+      ++count;
+    }
+  }
+  return count;
+}
```
veloman-yunkan, line 34: Get rid of this code duplication

test/book.cpp
```diff
@@ -192,6 +210,68 @@ TEST(BookTest, updateFromOPDSDuplicateLengthWarnsTest)
...
+    EXPECT_EQ(book.getSize(), 345678U);
+
+    const auto urls = book.getUrls();
+    EXPECT_EQ(nonEmptyLinksCount(book), 3U);
+    EXPECT_EQ(urls[linkIndex(kiwix::Book::AcquisitionLinkKind::DIRECT)], "https://who.org/zara.zim");
+    EXPECT_EQ(urls[linkIndex(kiwix::Book::AcquisitionLinkKind::META4)], "https://who.org/zara.zim.meta4");
+    EXPECT_EQ(urls[linkIndex(kiwix::Book::AcquisitionLinkKind::BITTORRENT)], "https://who.org/zara.zim.torrent");
+}
+
+TEST(BookTest, setUrlWithMimeTypeAddsToAcquisitionLinks)
```
veloman-yunkan: The name of this unit test must be changed.

test/library_server.cpp
```diff
@@ -1542,15 +1545,6 @@ TEST_P(LibraryServerTest, no_name_mapper_catalog_v2_individual_entry_access)
...
   "    </a>\n" \
-  "    <a href=\"https://github.com/kiwix/libkiwix/raw/master/test/data/zimfile_raycharles_uncategorized.zim.sha256\" download>\n" \
-  "        <div>SHA-256 checksum</div>\n" \
-  "    </a>\n" \
-  "    <a href=\"https://github.com/kiwix/libkiwix/raw/master/test/data/zimfile_raycharles_uncategorized.zim.magnet\" target=\"_blank\">\n" \
-  "        <div>Magnet link</div>\n" \
-  "    </a>\n" \
-  "    <a href=\"https://github.com/kiwix/libkiwix/raw/master/test/data/zimfile_raycharles_uncategorized.zim.torrent\" download>\n" \
-  "        <div>BitTorrent</div>\n" \
-  "    </a>\n" \
```
veloman-yunkan, line 1553: Here it is. Non-direct links have been lost.

test/library_server.cpp
```diff
@@ -1593,6 +1587,82 @@ TEST_P(LibraryServerTest, noJS) {
...
+std::string writeNoJSDownloadLinksLibrary()
+{
+  const auto cwd = kiwix::getCurrentDirectory();
+  const auto libPath = cwd + "/test/nojs_download_links.opds";
+  const auto zimPath = cwd + "/test/zimfile_raycharles_uncategorized.zim";
+  const auto zimPath2 = cwd + "/test/zimfile_raycharles.zim";
+  std::ofstream(libPath) << replace(replace(NOJS_DOWNLOAD_LINKS_OPDS, "@ZIMPATH@", zimPath),
+                                    "@ZIMPATH2@", zimPath2);
+  return libPath;
+}
```
veloman-yunkan: Why don't you just add a new file `nojs_download_links.opds` under `test/data` instead of generating a library file at runtime?

test/library_server.cpp
```diff
@@ -1593,6 +1587,82 @@ TEST_P(LibraryServerTest, noJS) {
...
+  const auto libPath = writeNoJSDownloadLinksLibrary();
+  ZimFileServer server(8003, ZimFileServer::DEFAULT_OPTIONS, libPath);
+
+  auto r = server.GET("/ROOT%23%3F/nojs/download/zimfile_raycharles_uncategorized");
+  EXPECT_EQ(r->status, 200);
+  EXPECT_NE(r->body.find("<a href=\"https://example.com/all.zim\" download>"), std::string::npos);
+  EXPECT_NE(r->body.find("<a href=\"https://example.com/all.zim.meta4\" download>"), std::string::npos);
+  EXPECT_NE(r->body.find("<a href=\"https://example.com/all.zim.checksum\" download>"), std::string::npos);
+  EXPECT_NE(r->body.find("<a href=\"magnet:?xt=urn:btih:allzim\" target=\"_blank\">"), std::string::npos);
+  EXPECT_NE(r->body.find("<a href=\"https://example.com/all.zim.tor\" download>"), std::string::npos);
```
veloman-yunkan: I don't like this kind of testing. Let's check the full output.

test/library_server.cpp
```diff
@@ -1593,6 +1587,82 @@ TEST_P(LibraryServerTest, noJS) {
...
+  ZimFileServer server(8003, ZimFileServer::DEFAULT_OPTIONS, libPath);
+
+  auto r = server.GET("/ROOT%23%3F/nojs/download/zimfile_raycharles");
+  EXPECT_EQ(r->status, 200);
+  EXPECT_NE(r->body.find("<a href=\"https://example.com/only.zim.meta4\" download>"), std::string::npos);
+  EXPECT_EQ(r->body.find("<div>Direct</div>"), std::string::npos);
+  EXPECT_EQ(r->body.find("SHA-256"), std::string::npos);
+  EXPECT_EQ(r->body.find("Magnet"), std::string::npos);
+  EXPECT_EQ(r->body.find("BitTorrent"), std::string::npos);
+  EXPECT_EQ(r->body.find(".sha256"), std::string::npos);
```
veloman-yunkan: Same here

test/manager.cpp
```diff
@@ -14,6 +14,19 @@ std::string resolveAbsPath(const std::string& basePath, const std::string& relPa
...
+  size_t count = 0;
+  const auto urls = book.getUrls();
+  for (const auto& url : urls)
+  {
+    if (!url.empty()) {
+      ++count;
+    }
+  }
+  return count;
+}
```
veloman-yunkan, line 28: Get rid of this code duplication

test/manager.cpp
```diff
@@ -45,7 +58,37 @@ TEST(ManagerTest, addBookFromPathAndGetIdTest)
...
+    urls[static_cast<size_t>(kiwix::Book::AcquisitionLinkKind::META4)] = meta4Url;
+    urls[static_cast<size_t>(kiwix::Book::AcquisitionLinkKind::BITTORRENT)] = torrentUrl;
+    auto bookId = manager.addBookFromPathAndGetId("./test/example.zim", "", urls);
+    ASSERT_NE(bookId, "");
+    kiwix::Book book = lib->getBookById(bookId);
+    EXPECT_EQ(book.getUrl(kiwix::Book::AcquisitionLinkKind::DIRECT), zimUrl);
+    EXPECT_EQ(book.getUrl(kiwix::Book::AcquisitionLinkKind::META4), meta4Url);
+    EXPECT_EQ(book.getUrl(kiwix::Book::AcquisitionLinkKind::BITTORRENT), torrentUrl);
+
+    // Empty fields are not recorded as acquisition links.
```
veloman-yunkan: This is a dubious formulation. `getUrl(kiwix::Book::AcquisitionLinkKind::A_NON_DIRECT_LINK_KIND)` should still succeed and return an empty string.

veloman-yunkan (changes requested):

PR #1347, opened 2026-09-12
Escape angle brackets in KIWIX_RESPONSE_DATA regardless of case

kah-ja:
### What

`ContentResponseBlueprint::Data::asJSON()` (`src/server/response.cpp:358`, pre-patch) neutralizes a `</script` substring in the JSON blob that becomes `window.KIWIX_RESPONSE_DATA` with `std::regex_replace(oss.str(), std::regex("</script"), "</scr\\ipt")`. The regex has no `std::regex::icase` flag, so `</SCRIPT>` or `</ScRiPt>` inside a string value pass through unchanged.

### Why

`KIWIX_RESPONSE_DATA` reaches the page through a raw, triple-mustache substitution (`{{{KIWIX_RESPONSE_DATA}}}`), which applies no HTML escaping of its own. The case-sensitive substring match is the only thing between a string value and the surrounding `<script>` tag, and it covers only one of several equivalent spellings of the closing tag.

`escapeForJSON()` (`src/tools/otherTools.cpp:336`), applied per string value inside `dumpJSON()`, does not escape `<` or `>` either. `KIWIX_RESPONSE_TEMPLATE`, the sibling field set a few lines below in `generateResponseObject()` (`src/server/response.cpp:380`), goes through a double-brace mustache substitution, which does HTML-escape its value; that field is unaffected by this issue.

### How

`asJSON()` now escapes every `<` and every `>` in the JSON text as a `\u`-escape, replacing the single case-sensitive substring match. Object keys and structural punctuation in this code path come from static field names, not from data, so `<`/`>` only ever occur inside an already JSON-string-escaped value; escaping them there removes any `</script`-like sequence regardless of case, and the browser's parser turns the escapes back into the original characters once it evaluates the object literal.

`escapeForJSON()` and its other call sites (`KIWIX_RESPONSE_TEMPLATE`, the suggestions JSON endpoint) are left untouched: changing that shared helper would also change the already-safe `KIWIX_RESPONSE_TEMPLATE` output and the suggestions API's byte content.

### Testing

Local build via meson + ninja, dependencies from apt plus a locally built `libzim` 9.4.1 (distro package is 9.2.3):

- `ninja -C builddir -j1`: full build succeeds.
- `./builddir/test/response` before the patch: new test `shouldEscapeAngleBracketsInResponseDataRegardlessOfCase` fails, `</SCRIPT>` and `<script>` from a `PATTERN` parameter reach `window.KIWIX_RESPONSE_DATA` unescaped. After the patch all 3 pass, the 2 pre-existing ones unchanged.
- `./builddir/test/otherTools`: unaffected, all 8 pass.
- `meson test --num-processes 1` on unpatched `origin/main`: all 55 `test/server` cases and all 24 test binaries pass, no pre-existing failures.
- `test/server.cpp`'s own `escapeJsString()` helper duplicates this escaping to build the 404/500 error page fixtures. It still matched only the case-sensitive `</script>` literal, so `ServerTest.HttpSexy404HtmlError` failed once `asJSON()` started escaping every `<`/`>`. This PR updates `escapeJsString()` to the same scheme.
- `meson test --num-processes 1` on the patched branch: all 55 `test/server` cases and all 24 test binaries pass.

### References

GHSA-6pfr-c959-wh6q ([found during penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by [turingpoint](https://www.turingpoint.de), reported privately, unpublished at time of writing)


src/server/response.cpp
```diff
@@ -349,13 +349,18 @@ std::string ContentResponseBlueprint::Data::asJSON() const
...
+  // any HTML escaping. A JSON string value may contain "<" as part of a
+  // "</script" (or "</SCRIPT", "</ScRiPt", ...) substring, which would let it
+  // break out of that tag regardless of letter case. Escaping every "<" (and,
+  // symmetrically, every ">") in the JSON text as a \u-escape removes the
+  // byte from the output while leaving the decoded JSON value unchanged; this
+  // is stricter than only matching the literal, case-sensitive "</script".
+  const std::string json = oss.str();
+  return std::regex_replace(
+      std::regex_replace(json, std::regex("<"), "\\u003c"),
+      std::regex(">"), "\\u003e");
```
veloman-yunkan: Let's not encode `>` for symmetry. We gain nothing, but penalize performance.

test/response.cpp
```diff
@@ -99,3 +99,39 @@ R"(<!DOCTYPE html>
...
+  const auto dataLineEnd = content.find('\n', dataPos);
+  const std::string dataLine = content.substr(dataPos, dataLineEnd - dataPos);
+
+  // window.KIWIX_RESPONSE_DATA is inserted into the page unescaped (raw
+  // mustache substitution) inside an inline <script> tag. No "<" (and, for
+  // symmetry, no ">") coming from a data value may survive into that line,
+  // no matter its letter case, or it can be used to break out of the tag.
+  EXPECT_EQ(dataLine.find('<'), std::string::npos) << dataLine;
+  EXPECT_EQ(dataLine.find('>'), std::string::npos) << dataLine;
+}
```
veloman-yunkan: We only care about absence of `</script>` inside a `<script>` tag. Can't this test be renamed and simplified accordingly?

veloman-yunkan (changes requested):

src/server/response.cpp
kah-ja: Agreed, dropped. `<` is the only way into a `</script` sequence, so escaping it alone is
enough, back to a single regex_replace, same cost as the original code.

test/response.cpp
kah-ja: Renamed to `shouldNotAllowBreakingOutOfTheScriptTag` and cut to one assertion: no
`</script`, in any case, in the rendered `window.KIWIX_RESPONSE_DATA` line; `rfind`
replaces the manual skip past the template line.

Built in the CI image: 24/24 pass. Negative control with the old case-sensitive regex
restored, the test fails on `"</script><script>alert(1)</scr\ipt>"`, uppercase survives,
lowercase is neutralised.

test/response.cpp
```diff
@@ -99,3 +101,28 @@ R"(<!DOCTYPE html>
...
+
+  // The rendered assignment is the last one in the page (the earlier
+  // occurrence is literal text inside the HTML-escaped template string).
+  const std::string content = getResponseContent(errResp);
+  const auto dataPos = content.rfind("window.KIWIX_RESPONSE_DATA = ");
+  ASSERT_NE(dataPos, std::string::npos);
+  std::string dataLine = content.substr(dataPos, content.find('\n', dataPos) - dataPos);
+
+  std::transform(dataLine.begin(), dataLine.end(), dataLine.begin(), ::tolower);
+  EXPECT_EQ(dataLine.find("</script"), std::string::npos) << dataLine;
```
veloman-yunkan: This is still too complex for the current state of affairs. I believe that ensuring that the entire response contains only a single occurrence of `</script>` is enough.

veloman-yunkan (commented):
I just merged another PR, necessitating this one to resolve a conflict with updated cacheids. Please simplify the test further, and we will merge.

test/response.cpp
kah-ja: Done. The test now only checks that the whole response contains a single `</script>`
(lowercased first, since the HTML parser matches that sequence case-insensitively).

Rebased on main, cacheids updated. 24/24 test binaries pass; with the old case-sensitive
regex restored the test fails with two occurrences.

veloman-yunkan (approved):
LGTM. Thank you!

PR #1348, opened 2026-09-12
Escape untrusted book metadata before inserting it as HTML

kah-ja:
### What

Two places in `static/skin` build markup from ZIM book metadata by interpolating it into a string later assigned to `innerHTML`, without escaping it first:

- `viewer.js:124` builds the reader's home button as `` `<button>${title}</button>` ``, `title` being the raw text of `./raw/<book>/meta/Title` (`viewer.js:147-149`).
- `index.js:216` interpolates a book's `Description` into both a `title="${description}"` attribute and the following element text.

### Why

`viewer.js:111-118` defines `setTitle()`, which assigns the same `title` value to `element.title`/`ariaLabel` as a DOM property, not as markup. `setCurrentBook()` (`viewer.js:120`) calls `setTitle()` one line above the `innerHTML` line, so the safe pattern for this value already exists right next to the unsafe one.

`index.js:116-118` defines `htmlEncode()`, used at `:102`, `:138` and `:513` for tag/title values, encoding `<`, `>` and `&` but not the quote character. Since `description` sits inside a `title="..."` attribute as well as in text content, a value containing a double quote closes the attribute early and adds arbitrary attributes, independent of `<`/`>` handling.

### How

`viewer.js`: `setCurrentBook()` now creates the button element and sets its label with `textContent`, mirroring `setTitle()`, instead of building a `<button>...</button>` string.

`index.js`: `htmlEncode()` now also encodes `"` and `'`, so it is safe inside a quoted attribute and not only inside text content; its three existing call sites are unaffected in practice. `description` is passed through `htmlEncode()` at both places on `:216`.

Left unchanged, as follow-up candidates: the book `title` (`:214`) and the favicon `style` attribute built from `iconUrl` (`:207`) interpolate the same kind of untrusted value into `innerHTML` without escaping, structurally identical to the `description` call site fixed here. The `language` attribute next to them (`:208`) is not equally exposed: it is already rendered through `humanFriendlyTitle()` and `htmlEncode()` before use.

### Testing

A test suite for these files exists: `test/server.cpp` requests every static resource over HTTP and checks its byte content indirectly, through a `cacheid` query parameter that is the first 8 hex characters of the SHA1 of the file's content. Since this patch changes the content of `index.js` and `viewer.js`, their cache ids change too, and `test/server.cpp` hardcoded the old ones in several places.

- `meson test --num-processes 1` on unpatched `origin/main`: all 24 test binaries pass.
- With only the skin files patched, before updating the fixtures: 7 failures (`ServerTest.200`, `CacheIdsOfStaticResources`, `CacheIdsOfStaticResourcesMatchTheSha1HashOfResourceContent`, `CompressibleContentIsCompressedIfAcceptable`, `HeadMethodIsSupported`, `CacheControlOfStaticContent`, `IfNoneMatchRequestsWithMismatchingETagResultIn200Responses`), each requesting one of the two changed files by its old cache id and getting a 404.
- This PR updates the cache ids in `test/server.cpp` in the same commit. All 24 test binaries then pass again.
- No JavaScript unit test exists for `htmlEncode()` itself. Ran the extracted function in Node against the reported payload (`X" tabindex="0" autofocus onfocus="..." z="`): unpatched it still breaks out of `title="..."`; patched, every quote becomes a numeric character reference and stays inert text.

### References

GHSA-f65j-hj4g-jf27 ([found during penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by [turingpoint](https://www.turingpoint.de), reported privately, unpublished at time of writing)
GHSA-8m2p-2jch-x3p2 ([found during penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by [turingpoint](https://www.turingpoint.de), reported privately, unpublished at time of writing)


static/skin/index.js
```diff
@@ -213,7 +215,7 @@
             <div class="book__header">
                 <div id="book__title">${title}</div>
```
veloman-yunkan, line 226: Shouldn't title be HTML-encoded here as well?

kelson42: We should be careful here.... Pretty sure I had already fix this. The encoding might be done in the Cpp using a moustache primitive.

kah-ja: @kelson42 The C++ side does escape ({{title}}/{{description}} go through kainjow's
html_escape(), & < > " '), but it does not protect this sink: index.js parses the feed with
DOMParser(..., 'application/xml'), which decodes those entities again, and getInnerHtml()
re-serializes with XML rules that escape & < > but not " / '. The quote is raw again in the
template literal.

Stock kiwix-serve 3.8.2 / libkiwix 14.2.0, welcome page, no interaction, ZIM description
`X" tabindex="0" autofocus onfocus="fetch('/BEACON')" z="` :

| index.js            | /BEACON           | `Questions & Answers about <science>` |
|---------------------|-------------------|---------------------------------------|
| `main`              | fires, zero-click | ok                                    |
| this PR, 1st commit | no                | `&amp;` / `&lt;` double-encoded       |
| this PR, now        | no                | ok                                    |

4dfcfbe and 2fdd206 fixed the viewer / search-result path; this is the welcome-page tile,
untouched since 063bb8c (10.0.0).

@veloman-yunkan Yes, but not by wrapping the current value, that is the double-encoding row
above. Same situation as your 846404e ("Tags in the OPDS feed are HTML encoded and must be
decoded [...] must be HTML encoded when injected into the DOM"); title and description never
got that. The new commit reads both via a getTextContent() helper and encodes at every sink.

veloman-yunkan: @kah-ja Which "new commit" do you refer to? Did you forget to push (or missed the `--force` option to it)?

kah-ja: @veloman-yunkan Sorry, the push was late; `85aed84` is the PR head now.

static/skin/index.js
```diff
@@ -175,8 +185,8 @@
                 iconUrl = link.getAttribute('href');
             }
         });
-        const title =  getInnerHtml(book, 'title');
-        const description = getInnerHtml(book, 'summary');
+        const title = getTextContent(book, 'title');
+        const description = getTextContent(book, 'summary');
```
veloman-yunkan: HTML-encoding can be performed right here (rather than in the interpolation).

veloman-yunkan (changes requested):
Please rebase your PR once more and squash all commits into one.

static/skin/index.js
kah-ja: Moved the encoding to the assignment, rebased on main and squashed into one commit.

veloman-yunkan (approved):
LGTM. Thanks!

kelson42:
@kah-ja Thank you very much for your security fix

PR #1349, opened 2026-09-14
Stabilize libkiwix's public ABI via PIMPL

### LINKED ISSUE #1346: Book's internal state via PIMPL to stabilize ABI across releases
## Problem

`kiwix::Book` (`book.h`) stores its state as protected data members directly in the class layout:

```cpp
protected:
  std::string m_id;
  std::string m_path;
  ...
  Illustrations m_illustrations;
```

Since libkiwix is consumed via static *and* dynamic linking, any change to these members (add/remove/reorder/retype a field) changes `Book`'s memory layout — an **ABI break**. Clients using the shared library must recompile and relink even though the public API (method signatures) hasn't changed. This makes routine maintenance disproportionately costly and discourages incremental evolution of `Book`, since layout changes can only ship in infrequent, coordinated major releases.

## Proposed solution

Apply the **PIMPL idiom**: move all data members into a private, opaque `Book::Impl`, and have `Book` hold only a pointer to it.

```cpp
class Book
{
 public:
  Book();
  ~Book();
  // public API (getters/setters) unchanged

 private:
  class Impl;
  std::unique_ptr<Impl> mp_impl;
};
```

With `sizeof(Book)` fixed, future changes to tracked fields become implementation-only and can ship in minor/patch releases without forcing client rebuilds.

### Scope of work

- Move today's members (`m_id`, `m_path`, `m_title`, `m_urls`, `m_illustrations`, etc.) into `Book::Impl`.
- Route all internal access through `mp_impl->...`.
- Define copy/move ctors, assignment operators, and the destructor in the `.cpp` (required since `Impl` is incomplete where `unique_ptr<Impl>` is declared).
- Check other libkiwix classes with public data members (e.g. `Illustration`) for the same treatment — can be split into follow-up tickets.
- Update anything relying on `Book`'s layout/`sizeof` (serialization, comparisons).

### Consequences

- This change is itself ABI-breaking (layout → single pointer), so it needs a **major version bump** — but it's a one-time break that buys lasting ABI stability afterward.
- Public method signatures stay the same, so most client code using the API (not the raw layout) should stay source-compatible.
- Document clearly in release notes/migration guide: why the break happened, and that `sizeof(Book)` changes (relevant to anyone embedding `Book` by value or doing manual serialization).

hamazaspavetisyan:
Fixes #1346

## Why

PIMPL `Book` so its internal layout can evolve without forcing every static/dynamic consumer to recompile. Since the next libkiwix release is already a planned major version bump, the maintainers decided to spend that one-time break on the whole public API instead of just `Book`, so this doesn't have to happen again.

## What

- `Book`, `Bookmark`, `Manager` (+ `LibraryManipulator`), and `Server` now hold their state behind a private `Impl`, reached only through methods — their size is now fixed regardless of future internal changes.
- `Book::Illustration`'s `width`/`height`/`mimeType`/`url` are private now, reached through new `getWidth()`/`getHeight()`/`getMimeType()`/`getUrl()`.
- `Download` and `SearchRenderer` can no longer be constructed directly — `Download` is `Downloader`-only (friendship), `SearchRenderer` gets a `create()` factory — matching how both were already actually used.
- `Library::create()`'s body moved out of the header into the `.cpp`. `Library` was already fully encapsulated, but its inline factory body still baked the caller's `sizeof(Library)` into client code.
- `Manager`'s five public fields (`writableLibraryPath`, `m_hasSearchResult`, `m_totalBooks`, `m_startIndex`, `m_itemsPerPage`) are now reached through getters/setters; its copy and move operations are explicitly deleted (never used anywhere).

This is an ABI break requiring a recompile, plus a couple of source changes:
- Direct reads of `Book::Illustration`'s fields must switch to the new getters — one confirmed spot in kiwix-desktop (`urlschemehandler.cpp`: `illustration->mimeType` → `illustration->getMimeType()`).
- Direct reads/writes of `Manager`'s five former public fields must switch to the new getters/setters — none found in kiwix-tools or kiwix-desktop, only in this repo's own tests.
- `Book`, `Bookmark`, `Manager`, `Server`, `Library` keep their existing construction/usage patterns (stack, by value, `Library::create()`, `Server`'s public constructor) — no other call-site changes needed.
- `Download` and `SearchRenderer` are never constructed directly by any downstream code found — no expected fallout there.

## Fundamental change

Nothing about the public API's behavior changes — only its resilience to future internal changes. From here on, adding, removing, or reordering an internal field on any of these classes only needs a recompile of libkiwix itself, not of every consumer.

## Notes

Each class was handled in its own commit for easier review. Two related items are intentionally out of scope: `IpAddress`/`DownloadedFile` stay plain structs (too trivial to warrant it), and adding `LIBKIWIX_API` visibility macros plus a symbol-version script (every symbol is currently exported by default) is left for a follow-up ticket.


include/book.h
```diff
@@ -49,14 +49,19 @@ class Book
   {
     friend class Book;
    public:
+    uint16_t getWidth() const;
+    uint16_t getHeight() const;
+    const std::string& getMimeType() const;
+    const std::string& getUrl() const;
+    const std::string& getData() const;
```
veloman-yunkan, line 56: Let's not hurry with the API of Book::Illustration.

Returning these by (`const`) reference can still be a serious restraining factor in any future changes to the implementation of these methods (`getData()` is of the most concern).

include/book.h
```diff
@@ -100,84 +104,65 @@ class Book
...
-  const std::string& getFlavour() const { return m_flavour; }
-  const std::string& getOrigId() const { return m_origId; }
-  const uint64_t& getArticleCount() const { return m_articleCount; }
-  const uint64_t& getMediaCount() const { return m_mediaCount; }
-  const uint64_t& getSize() const { return m_size; }
+  const std::string& getFlavour() const;
+  const std::string& getOrigId() const;
+  const uint64_t& getArticleCount() const;
+  const uint64_t& getMediaCount() const;
+  const uint64_t& getSize() const;
```
veloman-yunkan, line 129: Let's consider removing the reference from the return types of those accessors where absolutely not justified (like for these `uin64_t` attributes) or where they still may prove to be a restraining factor in the future.

include/book.h
```diff
@@ -100,84 +104,65 @@ class Book
...
   DEPRECATED const std::string& getFavicon() const;
   DEPRECATED const std::string& getFaviconUrl() const;
   DEPRECATED const std::string& getFaviconMimeType() const;
```
veloman-yunkan, line 132: Shouldn't we also delete deprecated methods?

include/manager.h
```diff
@@ -188,16 +201,15 @@ class Manager
...
-  uint64_t m_totalBooks = 0;
-  uint64_t m_startIndex = 0;
-  uint64_t m_itemsPerPage = 0;
-
- protected:
-  kiwix::LibraryManipulator manipulator;
+  bool hasSearchResult() const;
+  uint64_t getTotalBooks() const;
+  uint64_t getStartIndex() const;
+  uint64_t getItemsPerPage() const;
```
veloman-yunkan, line 210: We should use the opportunity to fix this quirk with `Manager`'s API rather than perpetuate it.

include/search_renderer.h
```diff
@@ -41,7 +42,7 @@ class SearchRenderer
    * @param start The start offset used for the srs.
    * @param estimatedResultCount The estimatedResultCount of the whole search
    */
-  SearchRenderer(zim::SearchResultSet srs, unsigned int start, unsigned int estimatedResultCount);
+  static std::unique_ptr<SearchRenderer> create(zim::SearchResultSet srs, unsigned int start, unsigned int estimatedResultCount);
```
veloman-yunkan, line 45: You're mistaken about 

> the only production construction site anywhere in the codebase is a single stack variable inside
InternalServer's own search-handling code - no external client construction or copying exists.

`SearchRenderer` is used in `kiwix-desktop`.

veloman-yunkan (changes requested):
Need to discuss how to move forward.

kelson42:
What is the status here?

include/manager.h
hamazaspavetisyan: @kelson42 We would like to get rid of this API in kiwix::Manager. 

```
bool m_hasSearchResult = false;
uint64_t m_totalBooks = 0;
uint64_t m_startIndex = 0;
uint64_t m_itemsPerPage = 0;
```

It was intended to get pagination information after importing an OPDS. There is no usage of it across all kiwix projects. Can we do it?

include/book.h
hamazaspavetisyan: @kelson42 Is it OK to drop all deprecated API from libkiwix in 15.0?

PR #1351, opened 2026-09-19
fix: treat dated letter-suffix ZIM names as permalink aliases

### LINKED ISSUE #1318: Some ZIM library permalink redirects are not working
https://browse.library.kiwix.org/viewer#terraria.wiki.gg_en_all is not found, while it should (currently) redirect to https://browse.library.kiwix.org/viewer#terraria.wiki.gg_en_all_2026-07a 

I suspect the `a` suffix is not supported.

cestercian:
## Summary

`HumanReadableNameMapper` only stripped `_YYYY-MM`. Extra same-month releases (`_2026-07a`, `_2026-07aa`) never got a date-less alias, so permalinks like `/viewer#terraria.wiki.gg_en_all` 404ed.

Strip `_YYYY-MM` plus an optional letter suffix.

Fixes #1318

## Test plan

- [x] `name_mapper` tests including letter-suffix cases


benoit74:
This change matches the new convention we have at https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention#ZIM_filename ; it is a bit more permissive (allow any number of alpha characters as suffix) but I feel like it is the proper implementation, no need to be rigid here.

Examples used in tests are adequate and match what we currently expect to see in the wild.

test/name_mapper.cpp
```diff
@@ -202,3 +202,68 @@ TEST_F(NameMapperTest, UpdatableNameMapperWithAliases)
...
+  kiwix::HumanReadableNameMapper nm(*lib, true);
+
+  EXPECT_EQ("wiki-2026-07a",  nm.getIdForName("terraria.wiki.gg_en_all"));
+  EXPECT_EQ("wiki-2026-07aa", nm.getIdForName("other.wiki_en_all"));
+  EXPECT_EQ("wiki-2021-10",   nm.getIdForName("plain.wiki_en_all"));
+
+  EXPECT_EQ("wiki-2026-07a",  nm.getIdForName("terraria.wiki.gg_en_all_2026-07a"));
+  EXPECT_EQ("wiki-2026-07aa", nm.getIdForName("other.wiki_en_all_2026-07aa"));
+  EXPECT_EQ("wiki-2021-10",   nm.getIdForName("plain.wiki_en_all_2021-10"));
+}
```
veloman-yunkan: Why don't you enhance the `HumanReadableNameMapperWithoutAliases`/`HumanReadableNameMapperWithAliases` test cases instead?

cestercian: fair point, folded them into the existing HumanReadableNameMapperWithoutAliases / WithAliases cases and dropped the extra fixture

src/name_mapper.cpp
```diff
@@ -34,7 +34,8 @@ HumanReadableNameMapper::HumanReadableNameMapper(const kiwix::Library& library,
     if (!withAlias)
       continue;
 
-    auto aliasName = replaceRegex(bookName, "", "_[[:digit:]]{4}-[[:digit:]]{2}$");
+    // Date suffix is _YYYY-MM, or _YYYY-MMll for extra releases in the same month.
+    auto aliasName = replaceRegex(bookName, "", "_[[:digit:]]{4}-[[:digit:]]{2}[[:alpha:]]*$");
```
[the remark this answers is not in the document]
cestercian: good catch, suffix is {0,2} now so three-plus letters no longer get a date-less alias

test/name_mapper.cpp
```diff
@@ -24,6 +24,9 @@ const char libraryXML[] = R"(
   <book id="06plus"     path="C:\data\zero_plus_six.zim"></book>
   <book id="07-super"   path="C:\data\zero_seven.zim"></book>
   <book id="07-sub"     path="C:\data\subdir\zero_seven.zim"></book>
+  <book id="08-2026-07a"    path="C:\data\zero_eight_2026-07a.zim"></book>
+  <book id="09-2026-07aa"   path="C:\data\zero_nine_2026-07aa.zim"></book>
+  <book id="10-2026-07abc"  path="C:\data\zero_ten_2026-07abc.zim"></book>
```
veloman-yunkan: Having stepped into double digit range, the `zero_` prefix in the filename is no longer needed :)

veloman-yunkan (changes requested):
Please squash all your changes into a single commit.

test/name_mapper.cpp
```diff
@@ -24,6 +24,9 @@ const char libraryXML[] = R"(
   <book id="06plus"     path="C:\data\zero_plus_six.zim"></book>
   <book id="07-super"   path="C:\data\zero_seven.zim"></book>
   <book id="07-sub"     path="C:\data\subdir\zero_seven.zim"></book>
+  <book id="08-2026-07a"    path="C:\data\eight_2026-07a.zim"></book>
+  <book id="09-2026-07aa"   path="C:\data\nine_2026-07aa.zim"></book>
```
veloman-yunkan: eight and nine are still single digit numbers and therefore must be prefixed with `zero_`  (the text of the filename should match the id string)

test/name_mapper.cpp
cestercian: yep, dropped the zero_ prefix now that they are double-digit

cestercian:
squashed to a single commit on latest main

test/name_mapper.cpp
cestercian: ah yep, good catch — eight/nine are still single-digit so they need the zero_ prefix back. fixing that now (ten stays without it).

cestercian: yep, zero_eight / zero_nine restored; ten stays bare

veloman-yunkan (approved):

